PR #33, opened 2019-10-28
RFC: spreading props

sergonius:
In this RFC, we propose introducing the spread syntax for React component props.

[View formatted RFC](https://github.com/transferwise/neptune/blob/rfc-spread-props/packages/docs/pages/rfc/0001-spread-props.md)


packages/docs/pages/rfc/0001-spread-props.md
```diff
@@ -0,0 +1,74 @@
...
+
+[Spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) allows to conveniently and flexibly pass props to a chosen element or method in a React component. This comes with pros and cons and should be used responsibly.
+
+Something to keep in mind when talking about flexibility is that the TransferWise component library is a public one, so in most cases hindering built-in functionality of HTML elements should be avoided.
+
+# Basic example
+
+```javascript
+const Button = ({ children, variant = 'default', isLoading, ...rest }) => (
+	<button {...rest} className={classnames('btn', `btn-${variant}`, isLoading && 'is-loading')}>
```
arikanmstf-tw: problem about spreading is, when you do something like this `<button {...rest}` you pass some unrelated props to the button. for example lets say `isVisible` and in the browser you get warning `isVisible is not a valid prop for element button`, and after some time there are hundreds of warnings in one page. and you can't see real important warnings.

sergonius: So you get a warning. Nothing still breaks or changes, browsers handle this really well.
And we assume people will do this a lot for some reason and that and it will pass CR-s?
The question is, does this benign negative outcome outweigh the benefits?

twrdjpalmer: The warning only gets thrown if somewhere you have the equivilent of `<Button isVisible="someVal" />`. As a mentioned in slack, it's a non-issue as you shouldn't be placing unsupported  props on a component (i.e., per the component's prop-types OR the standardised props that will work when spreaded). Only time I can imagine you'll get this otherwise is if you're blindly spreading props many levels deep, which is a smell

helendurranttw:
I'm closing this PR: We have agreed that this RFC describes a sensible approach to prop spreading that would be useful to adopt in our components. As the RFC describes, developers should use their best judgement to use it where appropriate.

I raised [another PR](https://github.com/transferwise/neptune-web/pull/365/files) to merge this in to Neptune: This one is a long time out of date against master. Thanks @sergonius and everyone else who took part in the discussion.

PR #100, opened 2019-11-25
feat(tabs): handle resizing

sergonius:
## ❓ Context
`<Tabs />` currently break when you resize the window.

## 🚀 Changes
This PR handles the resizing by setting a new container width and translate values.

- added a new function to handle resizing
- adding/removing an event listener that runs the function
- created a new function to set the container width
- added the current tab index to state to use in the new width calculation when resizing
- changed the old ref function to store the container for later use and call the new container width setting function

## ✅ Checklist


sergonius:
Both `retranslate` and `react-intl` can give you a string instead of a node as well via hoc/hook.

oliverviljamaa:
> Both `retranslate` and `react-intl` can give you a string instead of a node as well via hoc/hook.

I'm aware, but we'd be using a workaround that's not consistent with how we mostly use these libraries (with components), and is different/harder to test.

andreapiras:
image
 Maybe this would solve the issue

oliverviljamaa:
> image
> Maybe this would solve the issue.
> So we can specify both string and node of type Message

Potentially, I'm worried that we'd need to come back to change the type for new versions of translation libraries, though.

andreapiras:
> > image
> > Maybe this would solve the issue.
> > So we can specify both string and node of type Message
> 
> Potentially, I'm worried that we'd need to come back to change the type for new versions of translation libraries, though.

I think if that's the case we'll think about it, as for now I d defo prefer this to have node which is way too open. Another solution could be create a custom proptype check function so we can reuse that for this kind of scenarios and if we need to modify in future we can do from one place

```customProp: function(props, propName, componentName) {
    if (!/matchme/.test(props[propName])) {
      return new Error(
        'Invalid prop `' + propName + '` supplied to' +
        ' `' + componentName + '`. Validation failed.'
      );
    }
  },```

Tankenstein:
Guys, It's not worth it. I don't want to be an asshole, but you're talking about proptype checking with regexes atm. I really hope we all have better things to do here. Let's just use node.

andreapiras:
@Tankenstein 
> Guys, It's not worth it. I don't want to be an asshole, but you're talking about proptype checking with regexes atm. I really hope we all have better things to do here. Let's just use node or force using the hooks.

I think the conversation is indeed worth it and useful, but maybe we can address this change into another PR and go ahead with this.

Tankenstein:
Fair compromise

PR #107, opened 2019-12-03
[WIP] Footer: Docs, accessible text and links

danlaush:
## 📎 Jira ticket

n/a

## ❓ Context  

A regulator in the US state of Alaska has required us to fix the low colour contrast in the footer. [#design-system conversation](https://transferwise.slack.com/archives/C2XG7R3EZ/p1573142193205900)

## 🚀 Changes 

1. Copy the bootstrap footer styles into the Neptune styles (incl. updating to custom properties
1. Change the footer's links to be underlined and the grey text at the bottom to be white
1. Add a section to the Neptune docs for non-React components so they can be documented and worked on

image

image

## 💬 Considerations 

The last change is worth discussing - I think non-React components should be have documentation which is kept up-to-date if it's still in use. For some purposes we can link people to the [Bootstrap documentation](https://bootstrap.transferwise.com/components/#footer) but that will only get further out of date as time goes on. Not all of the components/styles will end up as React components, so it's worth having documentation for those style-only components.

All of the wording ("CSS Components") was the first thing to come to mind, so if there are suggestions I'm keen to discuss. 

## ✅ Checklist


anttio:
Can you confirm that all of the links in the footer needs the underlined style? What I understood from the original discussion was that only the link in the legal (tiny) text needed the underline.

danlaush:
@anttio I took the styles from how it was changed in `public-navigation`, but I think all underlined would actually be better for accessibility.

PR #109, opened 2019-12-03
fix: docgen issues

andreapiras:
## ❓ Context  
New documentation generates props via react-docegen. Some refactors are needed in order to go around reactdocgen issues.

## 🚀 Changes 
- Move Chevron out of common (currently excluded from docgen)
- Refactor popOver placement export and included into popOver file
- Refactor flex and box proptype exports
- Updated test
- Unified props and centralised prop exports

## ✅ Checklist


packages/components/src/common/placement.js
```diff
@@ -0,0 +1,10 @@
+export const Placement = {
+  TOP: 'top',
```
oliverviljamaa: Any reason for not using the exact same strings? So: `TOP: 'TOP`?

packages/components/src/common/flexProperties.js
```diff
@@ -1,3 +1,3 @@
 export const JustifyContent = ['start', 'end', 'center', 'between', 'around'];
 export const AlignItems = ['start', 'end', 'center', 'stretch', 'baseline'];
-export const flexDirection = ['row', 'column', 'row-reverse', 'column', 'column-reverse'];
+export const FlexDirection = ['row', 'column', 'row-reverse', 'column', 'column-reverse'];
```
oliverviljamaa: Would be cool if these were enum-like as well!

packages/components/src/index.js
```diff
@@ -26,7 +26,7 @@ export { default as RadioGroup } from './radioGroup';
 export { default as Alert } from './alert';
 export { default as InputWithDisplayFormat } from './inputWithDisplayFormat';
 export { default as TextareaWithDisplayFormat } from './textareaWithDisplayFormat';
-export { default as Chevron } from './common/Chevron';
+export { default as Chevron } from './chevron';
```
oliverviljamaa: 👏

packages/components/src/common/placement.js
andreapiras: Tbh I don't know, I ll create a ticket to go through all these proptypes const and align them. I guess in this case is consistent with
image

packages/components/src/common/flexProperties.js
andreapiras: The reason why those are no enum is because they ll never going to change as they are mapped from the css flex specification

packages/components/src/common/breakpoints.js
```diff
@@ -0,0 +1 @@
+export const Breakpoints = [0, 576, 768, 992, 1200];
```
dozortsev: could you please at least leave a comment about what this breakpoints for

dozortsev: I guess at some point we would need add documentation for them

packages/components/src/common/placement.js
dozortsev: @andreapiras JFYI we already have ticket for this https://transferwise.atlassian.net/browse/DS-81 🙂

PR #110, opened 2019-12-04
feat: new InstructionList for instructions / requirements

sbrandwoo:
Selection_051

## ❓ Context  

When opening bank details and ordering cards there is now a new requirements list, much like the existing instructions list on mobile.

## 🚀 Changes 

Adds `InstructionList`.

## 💬 Considerations 

This PR should probably also remove `InstructionsList`, which could be better named as `DosAndDontsList`. I'll await feedback on the new component before doing this.

The TypeScript definitions do not make it into the output bundle :(
I'm not aware how to do this in the current build flow.

## ✅ Checklist

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -752,7 +752,7 @@ Fix failing test and make code more clear
 ## Disable autocomplete for MoneyInput
 
 # v4.15.0
-## Add InstructionList component
+## Add InstructionsList component
 
 # v4.14.1
 ## Improve `MoneyInput` search
--- a/packages/components/src/index.js
+++ b/packages/components/src/index.js
@@ -16,6 +16,7 @@ export { default as DateLookup } from './dateLookup';
 export { default as PhoneNumberInput } from './phoneNumberInput';
 export { default as DateInput } from './dateInput';
 export { default as InstructionsList } from './instructionsList';
+export { default as InstructionList } from './instructionList';
 export { default as ProcessIndicator } from './processIndicator';
 export { default as Upload } from './upload';
 export { default as DefinitionList } from './definitionList';
--- a/packages/components/src/instructionList/InstructionList.docs.js
+++ b/packages/components/src/instructionList/InstructionList.docs.js
@@ -9,7 +9,7 @@ const InstructionListDocs = () => {
   const instructions = [
     {
       icon: <MoneyIcon size="md" />,
-      text: 'Add money to your account',
+      text: <strong>Add money to your account</strong>,
       help: {
         title: 'Add money to your account',
         content: (
@@ -24,11 +24,11 @@ const InstructionListDocs = () => {
     },
     {
       icon: <IdIcon size="md" />,
-      text: 'You uploaded you identity documents',
+      text: <strong>You uploaded you identity documents</strong>,
     },
     {
       icon: <HomeIcon size="md" />,
-      text: 'You entered your card delivery address',
+      text: <strong>You entered your card delivery address</strong>,
       body: (
         <>
           Flat 3 Chaplin House, Shepperton Road
@@ -57,7 +57,7 @@ const InstructionListDocs = () => {
               <pre className="tw-docs-code">{`<InstructionList instructions={[
   {
     icon: <HomeIcon size="md" />,
-    text: 'You entered your card delivery address',
+    text: <strong>You entered your card delivery address</strong>,
     body: (
       <>
         Flat 3 Chaplin House, Shepperton Road
--- a/packages/components/src/instructionList/InstructionList.docs.js
+++ b/packages/components/src/instructionList/InstructionList.docs.js
@@ -31,13 +31,11 @@ const InstructionListDocs = () => {
       text: 'You entered your card delivery address',
       body: (
         <>
-          <p>
-            Flat 3 Chaplin House, Shepperton Road
-            <br />
-            London N1 2NE
-            <br />
-            <a href="#foo">Edit address</a>
-          </p>
+          Flat 3 Chaplin House, Shepperton Road
+          <br />
+          London N1 2NE
+          <br />
+          <a href="#foo">Edit address</a>
         </>
       ),
       help: {
@@ -62,13 +60,11 @@ const InstructionListDocs = () => {
     text: 'You entered your card delivery address',
     body: (
       <>
-        <p>
-          Flat 3 Chaplin House, Shepperton Road
-          <br />
-          London N1 2NE
-          <br />
-          <a href="#foo">Edit address</a>
-        </p>
+        Flat 3 Chaplin House, Shepperton Road
+        <br />
+        London N1 2NE
+        <br />
+        <a href="#foo">Edit address</a>
       </>
     ),
     help: {
--- a/packages/components/src/instructionList/InstructionList.docs.js
+++ b/packages/components/src/instructionList/InstructionList.docs.js
@@ -1,0 +1,93 @@
+import React from 'react';
+import MoneyIcon from '@transferwise/icons/react/money';
+import IdIcon from '@transferwise/icons/react/id';
+import HomeIcon from '@transferwise/icons/react/home';
+
+import InstructionList from './InstructionList';
+
+const InstructionListDocs = () => {
+  const instructions = [
+    {
+      icon: <MoneyIcon size="md" />,
+      text: 'Add money to your account',
+      help: {
+        title: 'Add money to your account',
+        content: (
+          <>
+            To open your account, you’ll need to add money to it at least once.
+            <br />
+            This isn’t a fee — it’s still your money. It just means that you’re ready to go once
+            you’re verified.
+          </>
+        ),
+      },
+    },
+    {
+      icon: <IdIcon size="md" />,
+      text: 'You uploaded you identity documents',
+    },
+    {
+      icon: <HomeIcon size="md" />,
+      text: 'You entered your card delivery address',
+      body: (
+        <>
+          <p>
+            Flat 3 Chaplin House, Shepperton Road
+            <br />
+            London N1 2NE
+            <br />
+            <a href="#foo">Edit address</a>
+          </p>
+        </>
+      ),
+      help: {
+        title: 'Enter your card address',
+        content: 'When your card is ready we’ll post it to your home address.',
+      },
+    },
+  ];
+
+  return (
+    <div className="container">
+      <section className="section">
+        <div className="row m-t-4">
+          <div className="col-md-6">
+            <h2>Instruction list / Requirement list</h2>
+            <p>Things the customer needs to do or has done.</p>
+            <div className="m-t-2">
+              {/* eslint-disable react/jsx-indent */}
+              <pre className="tw-docs-code">{`<InstructionList instructions={[
+  {
+    icon: <HomeIcon size="md" />,
+    text: 'You entered your card delivery address',
+    body: (
+      <>
+        <p>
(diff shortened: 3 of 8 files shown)


packages/components/CHANGELOG.md
```diff
@@ -752,7 +752,7 @@ Fix failing test and make code more clear
 ## Disable autocomplete for MoneyInput
 
 # v4.15.0
-## Add InstructionList component
+## Add InstructionsList component
```
oliverviljamaa: Oops.

packages/components/src/instructionList/InstructionList.less
```diff
@@ -0,0 +1,34 @@
+@import 'bootstrap/less/variables.less';
```
oliverviljamaa: @nabaroa Should we import from `neptune-css` here already? If yes, do we have an example?

packages/components/src/instructionList/index.d.ts
```diff
@@ -0,0 +1,17 @@
+import React, { ReactNode } from 'react';
+
+export interface Instruction {
```
oliverviljamaa: 👏

packages/components/src/instructionList/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+    content: ReactNode;
+  } | null;
+  body?: ReactNode | null;
+}
+
+export interface InstructionListProps {
+  instructions: Instruction[];
+}
+
+export const InstructionList: React.FunctionComponent<InstructionListProps>;
```
oliverviljamaa: 😍

packages/components/src/instructionList/index.js
```diff
@@ -0,0 +1,3 @@
+import InstructionList from './InstructionList';
```
oliverviljamaa: This is often used across the company, as it allows us to omit the intermediate variable:
```suggestion
export { default } from './InstructionList';
```

packages/components/src/instructionList/index.d.ts
```diff
@@ -0,0 +1,17 @@
+import React, { ReactNode } from 'react';
+
+export interface Instruction {
+  icon: ReactNode;
+  text: ReactNode;
+  help?: {
+    title?: ReactNode | null;
```
oliverviljamaa: `null` is a valid `ReactNode`.
```suggestion
    title?: ReactNode | null;
```

packages/components/src/instructionList/InstructionList.less
```diff
@@ -0,0 +1,34 @@
...
+      [data-toggle='popover']:not(.btn) {
+        // Bootstrap popover has !important dotted border bottom
+        // This doesn't make sense on a help icon
+        border-bottom: none !important;
+      }
+    }
+
+    &__body {
+      margin-left: 40px;
+      font-size: 14px;
```
oliverviljamaa: Can we use variables for these as well?

packages/components/src/instructionList/InstructionList.js
```diff
@@ -0,0 +1,51 @@
...
+        </div>
+        {body && <div className="instruction__body m-t-1">{body}</div>}
+      </div>
+    );
+  });
+
+  return <div className="instruction-list">{items}</div>;
+};
+
+InstructionList.propTypes = {
```
oliverviljamaa: Would be cool if we could somehow ensure `propTypes` and the types will always be aligned 🤔

packages/components/src/instructionList/InstructionList.js
```diff
@@ -0,0 +1,51 @@
...
+              </Popover>
+            </div>
+          )}
+        </div>
+        {body && <div className="instruction__body m-t-1">{body}</div>}
+      </div>
+    );
+  });
+
+  return <div className="instruction-list">{items}</div>;
```
oliverviljamaa: Do you think the `items` variable is necessary?

packages/components/src/instructionList/InstructionList.js
```diff
@@ -0,0 +1,51 @@
...
+
+const InstructionList = ({ instructions }) => {
+  const items = instructions.map(({ icon, text, help, body }, i) => {
+    return (
+      // eslint-disable-next-line react/no-array-index-key
+      <div key={i} className="instruction m-t-3">
+        <div className="instruction__main">
+          <div className="instruction__icon m-r-2">{icon}</div>
+          <div className="instruction__text">
+            <div>
```
oliverviljamaa: Do we need that `div`?

packages/components/src/instructionList/index.d.ts
gusso: omg yes

packages/components/src/instructionList/InstructionList.docs.js
```diff
@@ -0,0 +1,93 @@
...
+      body: (
+        <>
+          <p>
+            Flat 3 Chaplin House, Shepperton Road
+            <br />
+            London N1 2NE
+            <br />
+            <a href="#foo">Edit address</a>
+          </p>
+        </>
```
gusso: This example doesn't need a fragment, right? Are you including it to encourage every body property to have one?

packages/components/src/index.js
```diff
@@ -16,6 +16,7 @@ export { default as DateLookup } from './dateLookup';
 export { default as PhoneNumberInput } from './phoneNumberInput';
 export { default as DateInput } from './dateInput';
 export { default as InstructionsList } from './instructionsList';
+export { default as InstructionList } from './instructionList';
```
andreapiras: How come we are not modifying the existing InstructionsList. I find this InstructionsList vs InstructionList etxremely confusing

packages/components/src/instructionList/InstructionList.js
```diff
@@ -0,0 +1,51 @@
...
+              <strong>{text}</strong>
+            </div>
+          </div>
+          {help && (
+            <div className="instruction__help m-l-2">
+              <Popover preferredPlacement="top" title={help.title} content={help.content}>
```
andreapiras: maybe deconstruct help and content?

packages/components/src/instructionList/InstructionList.js
andreapiras: div can wrap directly the map

packages/components/src/instructionList/InstructionList.js
```diff
@@ -0,0 +1,51 @@
...
+                <HelpCircleIcon size="sm" />
```
andreapiras: Do we always need this? can we add a propr to control whether the helpcircle is there or not?

packages/components/src/instructionList/InstructionList.docs.js
sbrandwoo: Just bad copy-pasting

packages/components/src/instructionList/InstructionList.js
sbrandwoo: This is a very mild case, but I'm not a huge fan of nesting absolutely everything inside a single statement. So if there's a `.map` or other complex structure, I'll split it out for readability.

packages/components/src/instructionList/InstructionList.less
sbrandwoo: What would I use?

packages/components/src/instructionList/index.d.ts
sbrandwoo: Hmm, I don't want that. `icon` and `text` shouldn't allow null :thinking:

packages/components/src/instructionList/index.d.ts
sbrandwoo: I discovered an unfortunate side-effect. If I copy this file into `/node_modules/@transferwise/components` then TypeScript is like "OMG I know `InstructionList`!"  - which is great. But then it thinks that all other components don't exist because there's no types of them. It'd usually just apply `Button` as `any`, but with this in place it'll be an error.

I don't know if there's a way to tell TypeScript that this is just some of the types, and to import anything else as `any`?

packages/components/src/instructionList/InstructionList.less
oliverviljamaa: `@font-size-small` for `14px` font and `@spacer * x` for spacing (`@spacer` is `8px`).

packages/components/src/instructionList/InstructionList.less
dozortsev: I guess these styles also should be in neptune/css package

dozortsev: re. variables

yes, I think you can import them but not all variables are there yet https://github.com/transferwise/neptune/blob/master/packages/css/src/less/variables/_variables.less

I would try import it and check current neptune `_variables.less` satisfy needs

dozortsev: here is the example of importing color variables from neptune https://github.com/transferwise/neptune/blob/91fdd6a3b0124c6127e33e03df6f3b52e562a716/packages/components/src/tabs/Tabs.less#L2

packages/components/src/instructionList/InstructionList.js
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import Types from 'prop-types';
+import HelpCircleIcon from '@transferwise/icons/react/help-circle';
+
+import Popover from '../popover';
+
+import './InstructionList.css';
+
+const InstructionList = ({ instructions }) => (
```
dozortsev: according to this [rfc](https://github.com/transferwise/neptune/blob/master/packages/docs/pages/rfc/0001-always-pass-classname.md) please expose `className` prop 🙂

dozortsev: in most outer DOM element

packages/components/src/instructionList/InstructionList.less
sbrandwoo: Really? Why?

packages/components/src/instructionList/InstructionList.less
sbrandwoo: thanks!

sbrandwoo:
Outstanding questions for the component interface:
- Should the text be by default in `<strong>` elements, or do we let the parent decide?
- Should the left icon be forced to be navy, or do we let the parent decide?
Both are difficult to overwrite if you don't want them. If we want to support red and green icons, then we'll need to do _something_...

sbrandwoo:
Source design, for reference:

image

sbrandwoo:
@dozortsev I've addressed the popover bug in https://github.com/transferwise/neptune/pull/122 and changed it to open bottom-left so that it's more likely to fit on the screen.

packages/components/src/instructionList/InstructionList.js
oliverviljamaa: The simplest way of doing it would probably be to make this a TypeScript file and use `FC` with defined prop types, but adding TypeScript to this repo is a choice we're probably not ready to make yet?

packages/components/src/instructionList/InstructionList.js
```diff
@@ -0,0 +1,53 @@
+import classNames from 'classnames';
+import React from 'react';
+import Types from 'prop-types';
+import HelpCircleIcon from '@transferwise/icons/react/help-circle';
+
+import Popover from '../popover';
+
+import './InstructionList.css';
+
+const InstructionList = ({ className, instructions }) => (
```
oliverviljamaa: Is it always going to be an _instruction_ list? It's icons, texts, and statuses, so it could be used for more than instructions?

gusso:
Replaced with #319

PR #111, opened 2019-12-04
BREAKING: change dismiss handling in Alert component

dozortsev:
## ❓ Context  

tl;dr component doesn't show up on page after user dismisses it, and this is because it's set to [dismissed state for "forever"](https://github.com/transferwise/neptune/blob/master/packages/components/src/alert/Alert.js#L48-L54) (only page refresh helps).

a bit more here: https://transferwise.slack.com/archives/CJNBX9CP6/p1575462336041100

## 🚀 Changes 

Make `Alert` dismissible only by parent component. For that I exposed one new prop `onDismiss` which is required if `dismissible` is true and remove dismissed value from state.

Breaking change here is that now if your component is dismissible (`dismissible={true}`) then you have to pass `onDismiss` handler and render component under condition, e.g:

```jsx
// inside render() method

{showAlert &&
   <Alert
     type="info"
     onDismiss={() => setState(({ showAlert }) => ({ showAlert: !showAlert }))}
     dismissible>
       ...
   </Alert>
}
```

## 💬 Considerations 

[Make sure to check deployment](https://transferwise.github.io/neptune/branch/change-alerts-dismissing-props/index.html).

## ✅ Checklist


packages/components/src/alert/Alert.js
```diff
@@ -80,7 +71,7 @@ export default class Alert extends Component {
             type="button"
             className="close"
             data-dismiss="alert"
-            onClick={() => this.dismiss()}
```
andreapiras: Can we not keep both? I mean if its passed used the one passed from outside otherwise use the built in?

PR #112, opened 2019-12-04
feat: rename InstructionsList as DosAndDonts, part of Field

sbrandwoo:
## ❓ Context  

The old `InstructionsList` was only used in `Field` and is really a list of dos and don'ts.

## 🚀 Changes 

Rename `InstructionsList` as `DosAndDonts` and move it inside `Field`.

## 💬 Considerations 

Selection_052

## ✅ Checklist


sbrandwoo:
The build fail is the new documentation.

sbrandwoo:
Selection_053

For reference, if we used the new `InstructionList` for this, it'd look like the above, and you barely save any code in `DoAndDonts` because you have to map to the instruction objects.

I do not think we should do this, there is barely any similarity between the components, better to have two components that do their own purpose well.

gusso:
I disagree, they're both different types of instructions, but instructions nonetheless. Why would we teach designers and engineers 2 components that are so similar? Here's how things are organised in Figma.
Screenshot 2019-12-04 at 3 35 56 pm
https://www.figma.com/file/KiieyaL6uedMA6lePWyJWTat/Neptune-Mobile?node-id=3325%3A958

andreapiras:
> Selection_053
> 
> For reference, if we used the new `InstructionList` for this, it'd look like the above, and you barely save any code in `DoAndDonts` because you have to map to the instruction objects.
> 
> I do not think we should do this, there is barely any similarity between the components, better to have two components that do their own purpose well.

Yeah in this case I may to disagree too. I think the components are basically the same component. and there should be only a configurable one that meets all design. In fact the pattern is icon + text + helpIcon. What if in future we'll want a instructionsList with help tooltip? that won't be achievable with dos and donts component. So I think we should create a component like the one we discussed with Gusso too that can meet all requirements

sbrandwoo:
So in that design file those happy and sad faces match more closely to the new `InstructionList`. Perhaps I should remove the icon colour from the `InstructionsList` component and allow the parent to decide a colour, so they can choose navy, red or green.

We don't currently have those happy/sad instructions on web. The current use case of Dos and Don'ts is much simpler and exists only as a tooltip on Field. In this case I think the similarity is only that there is an icon next to some text - both things that are already part of the design system.

sbrandwoo:
> What if in future we'll want a instructionsList with help tooltip?

If you want dos and don'ts with help tooltips, and with extra information, then it shouldn't exist inside `Field` and you should use the new `InstructionList`. There isn't room for extra content inside field and it's impossible to interact with help tooltips due to the behaviour of `Field`.

This PR addresses an existing use case, not any future ones.

oliverviljamaa:
Could we reconsider the naming? We're considering using this component for features supported or not supported by bank details, for example, and I'm not sure whether `DosAndDonts` is correct for this use case. Of course, neither is `InstructionsList`.

packages/components/src/field/dosAndDonts/DosAndDonts.js
```diff
@@ -3,35 +3,35 @@ import Types from 'prop-types';
...
         <CloseIcon className="dont" />
         <p className="m-l-2 m-b-0 text-primary">{dont}</p>
       </div>
     ))}
   </div>
 );
 
-InstructionsList.defaultProps = {
+DosAndDonts.defaultProps = {
   dos: [],
```
oliverviljamaa: Could we accept just one array to allow not grouping the items?

sbrandwoo: How do you know which are red and which are green?

anttio:
Where is this design coming from? AFAIK the instruction list was never meant to be used inside alert.

packages/components/src/field/dosAndDonts/DosAndDonts.js
oliverviljamaa: We'd need to make it an array of objects indeed — seems to be not worth it indeed for an use case we're not sure of.

gusso:
Replaced with #319

PR #114, opened 2019-12-04
fix: remove the extra border in select components

anttio:
## 📎 Jira ticket
DS-141

## ❓ Context  
Neptune CSS was missing `.tw-select-filter` from BT which removed the extra border in the select dropdown search.

**Before**
image

**After**
image

## ✅ Checklist


packages/css/src/less/forms.less
```diff
@@ -11,3 +11,8 @@
 @import (reference) 'bootstrap/less/mixins/tab-focus.less';
 
 @import 'bootstrap/less/forms.less';
+
+// from transferwise.less on bootstrap
+.tw-select-filter {
```
andreapiras: should this be here . or inside the select.css in components?

nabaroa: I'd add it here, otherwise the selector when using only bootstrap would be broken. That fix has been on transferwise.less as a fix for three years.

anttio: Ideally yes, but there is also Angular version that needs this.

PR #121, opened 2019-12-06
Fix Upload postData

czebe:
## ❓ Context  
Fixing minor issues on the `<Upload>` component and `<FlowNavigation>`

## 🚀 Changes 
- add `tw-flow-navigation--inverse` class to `<FlowNavigation>` when the DARK theme is applied, so that consumers can set it's background colour
- set the default `POST` method in `postData` 
- throw an `Error` when upload fails and include the response (so you can access data returned from the API)
- don't queue success/error callbacks if the upload was cancelled 

## ✅ Checklist


packages/components/src/upload/utils/postData/postData.js
```diff
@@ -9,8 +10,10 @@ export const postData = (httpOptions, data = {}) =>
   })
     .then(response => {
       if (!response.ok) {
-        const errorMessage = { status: response.status, statusText: response.statusText };
-        throw errorMessage;
+        const error = new Error(response.statusText);
+        error.status = response.status;
+        error.response = response;
+        throw error;
```
dozortsev: could you please improve tests according to this changes (`postData.js`)?

PR #122, opened 2019-12-06
Popover fixes

sbrandwoo:
We use Popover in the new InstructionList and I ran into some issues with the `bottom-left` positioning.

### feat: add more Popover repositioning when it'd go off-screen

bottom-left and bottom-right will now reposition if they go off-screen

### feat: correctly position popover on first render
When using bottom-left, bottom-right, left-top or right-top the popover would be incorrectly positioned when it first rendered.

This is because the positioning relied on checking for the current position of the popover arrow, and this was not stable as the popover had not yet been rendered.

The offset of the arrow in relation to the bubble is now hard-coded to match the CSS coming from Bootstrap.


packages/components/src/popover/positioning.js
```diff
@@ -85,18 +95,11 @@ function getPositionRelativeToViewport(popover, placement) {
   );
 }
 
+const ARROW_OFFSET = 30;
```
andreapiras: where is this value coming from?

sbrandwoo: It's the value the CSS is designed to use:
image
The red line is 30px.

The original solution tried to calculate this value by reading actual computed values from the DOM, but I found this doesn't work on the first render and thus most of the use cases.

It isn't exactly future proof, but it was the only way I could get it to work.

sbrandwoo:
If you want to see the bug, click on a ? on the Instruction List component here: https://transferwise.github.io/neptune/branch/instruction-list/index.html

PR #123, opened 2019-12-09
fix: group propTypes and distribute as static props

andreapiras:
## ❓ Context  
This change is another steps towards unifying the way we define and distribute prop types. Also it groups the propDefinition which will be eventually unified

## 🚀 Changes 
1) Updates static propTypes
2) Group propType constants

## ✅ Checklist


packages/components/docs/utils/generateInput/generateInput.js
```diff
@@ -15,20 +14,20 @@ const generateInput = (knobs, componentThis) => {
   let newPlaceholder = placeholder;
 
   switch (type) {
-    case FormControlType.SELECT:
+    case FormControl.Types.SELECT:
```
oliverviljamaa: Do we keep it plural to not create a breaking version?

andreapiras: I introduce this static so I guess it won't be a breaking change either plural or singular no?

andreapiras: But actually Button is Type so I guess makes sense to keep it consistent

packages/components/src/common/propsValues/sizes.js
```diff
@@ -12,6 +12,20 @@ export const Sizes = {
...
 };
 
+export const SizesXSML = {
+  ...extraSmall,
+  ...small,
+  ...medium,
+  ...large,
+};
+
+export const SizesSMLX = {
```
oliverviljamaa: I feel having separate exportables for all these subsets is a bit excessive and it'd be more explicit to just use `Size`  keys and values instead.

andreapiras: What about this assignments? Button.Size = SizesXSML; how would u write this?
Button.Size = {..Sizes.EXTRASMALL,...Sizes.SMALL ...}

packages/components/src/common/index.js
```diff
@@ -1,9 +1,18 @@
...
-export { Sizes, SizesSML, SizesSL } from './propsValues/sizes';
-export { DateMode } from './propsValues/dateMode';
-export { MonthFormat } from './propsValues/monthFormat';
-export { Breakpoints } from './breakpoints';
-export { Placement } from './placement';
 export { addModalOpenBodyClass, removeModalOpenBodyClass } from './DOMOperations';
-export { JustifyContent, AlignItems, FlexDirection } from './flexProperties';
 export { Theme } from './theme';
+// Props
+export { Breakpoints } from './propsValues/breakpoints';
```
dozortsev: let call them consistently

either plurally versus singularly

packages/components/src/common/propsValues/type.js
```diff
@@ -0,0 +1,6 @@
+export const Type = {
+  PRIMARY: 'primary',
+  PAY: 'pay',
+  SECONDARY: 'secondary',
+  DANGER: 'danger',
```
dozortsev: those ones only needed for Button component no?

here I would probably use `MessageType` (which should be `Type`):

```js
PAY: Type.SUCCESS,
PRIMARY: Type.INFO,
..
```

dozortsev: actually 'pay' and 'secondary' it's something new and custom, we should names like here https://bootstrap.transferwise.com/css/#buttons-options I think

packages/components/src/alert/Alert.js
```diff
@@ -28,22 +17,22 @@ const ArrowPosition = {
...
-  static Type = Type;
+  static Type = MessageType;
   static ArrowPosition = ArrowPosition;
 
   static propTypes = {
-    type: Types.string,
+    type: Types.oneOf([Alert.Type.INFO, Alert.Type.WARNING, Alert.Type.ERROR, Alert.Type.SUCCESS]),
     children: Types.node.isRequired,
-    size: Types.oneOf(Object.values(Size)),
+    size: Types.oneOf([Alert.Size.SMALL, Alert.Size.LARGE]),
```
oliverviljamaa: Why expose `Size` with values we don't allow? 🤔

packages/components/src/box/Box.js
```diff
@@ -3,9 +3,9 @@ import { jsx } from '@emotion/core';
 import classNames from 'classnames';
 import Types from 'prop-types';
 
-import { Sizes, JustifyContent, AlignItems, Breakpoints } from '../common';
+import { Size, JustifyContent, AlignItems, Breakpoints } from '../common';
 
-const mediaQueries = Breakpoints.map(bp => `@media (min-width: ${bp}px)`);
+const mediaQueries = Object.values(Breakpoints).map(bp => `@media (min-width: ${bp}px)`);
```
oliverviljamaa: Do we polyfill `Object.values`?

packages/components/src/common/propsValues/breakpoints.js
```diff
@@ -0,0 +1,7 @@
+export const Breakpoints = {
```
oliverviljamaa:
```suggestion
export const Breakpoint = {
```

packages/components/src/common/propsValues/breakpoints.js
```diff
@@ -0,0 +1,7 @@
...
+  EXTRA_SMALL: 0,
```
oliverviljamaa: Where would we have an use for this?

packages/components/src/common/index.js
oliverviljamaa: Why are we not using default exports here? No strong opinion, but not sure what the benefit is in this scenario 😊

packages/components/src/common/propsValues/messageType.js
```diff
@@ -0,0 +1,6 @@
+export const MessageType = {
```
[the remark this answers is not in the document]
andreapiras: We have Type already ```export const Type = {
  PRIMARY: 'primary',
  PAY: 'pay',
  SECONDARY: 'secondary',
  DANGER: 'danger',
};```

packages/components/src/common/propsValues/messageType.js
```diff
@@ -0,0 +1,6 @@
...
+  INFO: 'info',
```
oliverviljamaa: If we're already doing a breaking change, might as well change to capitalised? WDYT?

packages/components/src/common/propsValues/position.js
```diff
@@ -1,4 +1,4 @@
-const Placement = {
+export const Position = {
   TOP: 'top',
```
oliverviljamaa: Same here, would it be a good time to change to capitalised values?

packages/components/src/typeahead/Typeahead.docs.js
```diff
@@ -1,10 +1,9 @@
 import React, { Component } from 'react';
 import { Typeahead } from '..';
 import { generateCodeBlock, generateState, generateInput } from '../../docs/utils';
-import { Sizes } from '../common';
 import Alert from '../alert';
 
-const SIZES = [Sizes.MEDIUM, Sizes.LARGE];
+const SIZES = [Typeahead.Size.MEDIUM, Typeahead.Size.LARGE];
```
oliverviljamaa: 👏

packages/components/src/processIndicator/ProcessIndicator.docs.js
```diff
@@ -32,8 +32,8 @@ const KNOBS = {
         label: value,
       })),
       defaultState: {
-        value: ProcessIndicator.Size.ExtraLarge,
-        label: ProcessIndicator.Size.ExtraLarge,
+        value: ProcessIndicator.Size.EXTRALARGE,
```
oliverviljamaa: Is this intentional? We seem to be using `EXTRA_LARGE` elsewhere.

packages/components/src/alert/Alert.js
andreapiras: Size is used in several places. And all of them need different subset. Initially I exposed different subset SizeXSM, SizeSML, etc which u asked me to remove 😅

packages/components/src/box/Box.js
andreapiras: Actually good point. I checked the docs in IE11 and it was working ok...but... Docs is still generated using webpack that includes polyfill while lib and umd are generated using rollup and babel so they are not polyfilled. Do we want to serve polyfill with this library or should we leave to the consumer to take care of that to avoid duplicate polyfills?

packages/components/src/common/propsValues/breakpoints.js
andreapiras: I m using this is in FLex and Box css-in-js bit https://github.com/transferwise/neptune-web/pull/123/files#diff-46a308455f6e0f21332eb2ed8ebea780R39

packages/components/src/common/propsValues/position.js
andreapiras: uhm not sure about this. All the other prop apart from Layouts are not capitalised. I'd do it but not in this PR as I m a bit concerned of introducing too many Breaking changes at once

packages/components/src/common/propsValues/messageType.js
andreapiras: https://transferwise.atlassian.net/browse/DS-253

PR #132, opened 2019-12-19
fix: force to clean content type header

arikanmstf-tw:
force to clean content type header on upload component at http options

## ❓ Context  
with current implementation, it is not possible to remove Content-Type  header.

In httpOptions file:

```js
headers: {
  'Content-Type': undefined,
},
```

* but didn’t work. client’s header turns out like in the image

`Content-Type: undefined` (like a string)

## 💬 Considerations 
Screenshot 2019-12-18 at 17 14 20

## ✅ Checklist


czebe:
Closing this PR, we'll remove the default `Content-type` as discussed with @arikanmstf-tw

PR #133, opened 2019-12-19
docs: fix mobile scroll

andreapiras:
## ❓ Context  
Currently mobile width of docs is overflowing and doesn't allow scroll

## 🚀 Changes 

1) Fix width for Box container
2) Prevent docs changes to bump css and component version
3) Reorganized utils files

## ✅ Checklist


packages/docs/utils/packageVersion/packageVersion.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import Types from 'prop-types';
+
+const PackageVersion = ({ version }) => <small>v{version}</small>;
```
dozortsev: I would pass version value in this component as `children` 

```
<PackageVersion>2.4.3</PackageVersion>
```

dozortsev: maybe let's use `<code />` tag here instead of `<small />` as we discussed yesterday? 🙂

Screen Shot 2019-12-18 at 4 08 42 PM

andreapiras: NOOOOO it looks so ugly!! 😅

PR #136, opened 2020-01-02
Don't specify default Content-type

czebe:
## ❓ Context  
Upload works better when we let the browser to define the `Content-type` with the `boundary` hash

## 🚀 Changes 
Remove the default `Content-type`, essentially reverting back to the original implementation.

## ✅ Checklist


packages/components/src/upload/Upload.docs.js
```diff
@@ -342,7 +342,7 @@ httpOptions: {
   fileInputName: '', // Optional. Defaults to the name of the file being uploaded
   data: { ... }, // Optional. Additional key/value pairs to be included in the form submission
   method: '', // Optional. Defaults to 'POST'
-  headers: {},  // Optional. Additional headers for fetch. 'Content-type' defaults to: 'multipart/form-data'
+  headers: {},  // Optional. Additional headers for fetch. Eg. define custom 'Content-type': 'multipart/form-data' or add access tokens
```
andreapiras: Can we move this documentation here? https://github.com/transferwise/neptune-web/blob/master/packages/docs/pages/components/Upload.mdx
So it will appear in the new docs 
https://transferwise.github.io/neptune-web/components/Upload

packages/components/src/upload/utils/postData/postData.js
```diff
@@ -1,12 +1,8 @@
 export const postData = (httpOptions, data = {}) =>
   fetch(`${httpOptions.url}`, {
-    ...httpOptions,
     method: 'POST',
     body: data,
-    headers: {
-      'Content-Type': 'multipart/form-data',
-      ...httpOptions.headers,
-    },
+    ...httpOptions,
```
andreapiras: Is this a breaking change?

packages/components/src/upload/Upload.docs.js
czebe: I added more specific `propType` definition, let's see if this is clear enough.

packages/components/src/upload/utils/postData/postData.js
czebe: No, it's not, the original implementation did not define the `Content-type` (correctly).

PR #137, opened 2020-01-03
fix(button): explicitly expose className prop

dozortsev:
## ❓ Context 

https://transferwise.slack.com/archives/CJNBX9CP6/p1578072335003900?thread_ts=1578065577.002700&cid=CJNBX9CP6

## ✅ Checklist


packages/components/src/button/Button.spec.js
```diff
@@ -63,4 +63,11 @@ describe('Button', () => {
     wrapper.simulate('click');
     expect(props.onClick).toBeCalled();
   });
+
+  it('tests work of custom CSS classes passed via className prop', () => {
```
oliverviljamaa: How about simplifying it a bit?
```suggestion
  it('adds passed class name to existing class names', () => {
```

PR #140, opened 2020-01-08
fix: align common constants

dozortsev:
## ✅ Checklist


packages/components/src/common/index.js
```diff
@@ -15,3 +15,14 @@ export { FormControlType } from './propsValues/formControlType';
...
+
+export function pickProps(originalProps, values) {
+  const extractedProps = {};
+  Object.keys(originalProps).forEach(key => {
+    const value = originalProps[key];
+    if (values.indexOf(value) > -1) {
+      extractedProps[key] = value;
+    }
+  });
+  return extractedProps;
```
dozortsev: perhaps instead of custom function we could use [`.pickBy`](https://lodash.com/docs/4.17.15#pickBy)

PR #142, opened 2020-01-09
fix: validate max length for empty string correctly

farah:
❓ Context
When  `value` prop is an empty string and `maxLength` validation is present, you get a validation error. 
`isValidMaxLength` function will always return `false` when `value` is an empty string which is incorrect.

🚀 Changes
Add a manual check if `value` is an empty string.


andreapiras:
Merged with https://github.com/transferwise/neptune-web/pull/144

farah:
hey, why don't I get credit for that commit? ...... not fair man...

farah:
Can I be credited for future PRs or is this the way you do it for this project? This was my first Github PR and was looking for a project to contribute to (that I also use). thanks

PR #147, opened 2020-01-13
Alerts docs to CSS components

nabaroa:
## 📎 Jira ticket
https://transferwise.atlassian.net/browse/DS-258

## 🚀 Changes 
Alerts docs to CSS components

imagen

Little extra: Adds the CSS file size to the layout.
imagen

## ✅ Checklist


packages/docs/pages/css/Alerts.mdx
```diff
@@ -0,0 +1,99 @@
...
+  <a href="#" class="alert-link">...</a>
+</div>
+<div class="alert alert-danger" role="alert">
+  <a href="#" class="alert-link">...</a>
+</div>
+```
+
+export const meta = {
+  abstractURL: '',
+  name: 'Alerts',
```
anttio: We started to rename the components to be in singular, see React components which has Alert.

packages/docs/pages/css/Alerts.mdx
```diff
@@ -0,0 +1,99 @@
...
+  size: '4 KB',
```
anttio: Is this really needed? If we could get this automatically from the file that would be nice, otherwise I think this will be outdated quite easily.

nabaroa: Agree, removed by now.

PR #149, opened 2020-01-17
Fix tabs overflow issues and focus issues

fergusjordantw:
## ❓ Context  

There are currently a few issues with the tabs component:
1. When any prop of tabs changes, any input you were focused in loses focus
1. It doesn't take into account overflow (see image below)

image

### To fix the focus issue

We were updating the tab focus on a `useEffect` hook so that the selected tab label had focus (instead of other tab labels). I've updated this so it now only does that when the `selected` attribute on the `<Tab />` changes. This will let us keep the nice accessibility consideration, while removing the problematic behaviour since updating the focus will be linked to the explicit action of selecting a new tab. 

As part of this, I also updated the `:hover` styles so that we don't apply them on mobile clients (which effectively amounts to `onClick` sometimes). This fixes an edge case where when you clicked a tab to select it, then swiped back to the other tab on mobile, both would be highlighted because one was `:focus`ed and the other was `:hover`ed.

### To fix the overflow issue

I've changed the way we deal with overflowed tab content. It used to be that we were setting the `tab-slider-container` width to `100% * tabsLength` with `overflow: hidden`. The other tabs were always there, just off screen and out of harms way due to overflow. 

This is still the case when you click to change tab or start swiping (`isAnimating || isSwiping`) so that you can still see the new tab coming in. Only now, when the tab is "selected" (i.e. the content is rested in one place and you're interacting with the content, not the tab components itself), the slider container is brought back to `100%` width and the other tabs are set to `display: none`. This lets us set `overflow: visible` on the panel container so that any select lists will show properly.

Note: this also means I've had to set `position: relative` on the `.tabs` and introduce a `zIndex` property so the consumer can set that if needed (see example of issue without it below).

Screenshot 2020-01-17 at 15 01 30

Note: this will still require us to update the Select component to be attached to the body via a react portal on mobile (easy because its fixed position). This is because Tabs use `transform: translateX` for switching between tabs and any child of an element using a `transform: translate` that has `position: fixed` is positioned according to the parent instead of the viewport. See https://stackoverflow.com/a/24996421

## 💬 Considerations 

Now that the non-selected tab panels have `display: none`, those tabs don't have a height the way they used to. This means that if the content passed as children to `<Tabs />` have different heights, the `<Tabs />` height itself will be the height of whichever tab is selected (or the highest one during tab switch). Visually, this means that if you have content below where the tabs are, it could potentially jump up/down as you switch tabs depending on the height of the tab content. 

## ✅ Checklist


packages/components/src/tabs/Tabs.spec.js
```diff
@@ -13,6 +13,9 @@ jest.mock('react-spring/renderprops.cjs', () => ({
   Spring: jest.fn(props => {
     const { children, to, onRest, config } = props;
 
+    // Queue this behind a timer so we can controll when it runs
```
Tankenstein:
```suggestion
    // Queue this behind a timer so we can control when it runs
```

packages/components/src/tabs/Tabs.js
```diff
@@ -304,7 +304,7 @@ class Tabs extends React.Component {
   };
 
   render() {
-    const { tabs: allTabs, changeTabOnSwipe, name, selected, className } = this.props;
+    const { tabs: allTabs, changeTabOnSwipe, name, selected, className, zIndex } = this.props;
```
Tankenstein: Why shouldn't consumers control `zIndex` through classes/styles?

Tankenstein:
Super nice solution, was worried it would be worse (portals on select etc)

fergusjordantw:
> Super nice solution, was worried it would be worse (portals on select etc), love playing with the tabs.

To be clear, we will still need to use portals on select for mobile, but that's nice and easy because it already has a fixed position to the bottom of the screen. So we avoid the pitfalls of excessive positioning calculations.

packages/components/src/tabs/Tabs.js
```diff
@@ -360,7 +365,13 @@ class Tabs extends React.Component {
             }}
           />
         </TabList>
-        <div className="tabs__panel-container" ref={this.setContainerRefAndWidth}>
+        <div
+          className="tabs__panel-container"
+          ref={this.setContainerRefAndWidth}
+          style={{
+            overflow: hidePanelOverflow ? 'hidden' : 'visible',
```
andreapiras: Since we are already using [emotion](https://github.com/transferwise/neptune-web/blob/master/packages/components/src/box/Box.js) I'd have give it a go so we write css in js in a consistent way. Plus there are also some method to generate [snapshots](https://github.com/transferwise/neptune-web/blob/master/packages/components/src/box/__snapshots__/Box.spec.js.snap) of the css and I've found it pretty good

fergusjordantw: Don’t really think it’s necessary here, it’s only a couple switch statements, don’t think it would provide any great value

packages/components/src/tabs/Tabs.js
fergusjordantw: Thought about it a bit more and got some input from @sergonius - removed the zIndex. Will leave it up to the consumer to set via CSS

PR #152, opened 2020-01-21
feat(button): add LINK type

oliverviljamaa:
## 📎 Jira ticket
[BORAT-3500](https://transferwise.atlassian.net/browse/BORAT-3500)

## ❓ Context  
We don't yet have support for buttons styled as links and our designs often use them.

## 🚀 Changes 
`LINK` type is added to `Button`.

## 💬 Considerations 
N/A

## ✅ Checklist


andreapiras:
Hey Oliver don't forget to update the docs!

oliverviljamaa:
@dozortsev Great catch! I thought the same, and thought it was OK indeed, as it's still a `Button`, just styled as an anchor.

oliverviljamaa:
@dozortsev Do you think the additional complexity is worth it? How would you specify that the `disabled` prop has no effect only when the type is `LINK`?

oliverviljamaa:
Thanks for the approval!

Although slightly controversial, I'll merge this, as link-type buttons still exist in our designs, even new ones, and if we don't provide this type, consumers would resort to HTML buttons, making them even harder to unify/remove one day.

PR #153, opened 2020-01-22
Move mobile select menu into a portal

fergusjordantw:
## ❓ Context  
For some usage of tabs, we have a select list inside of a TabPanel (see https://github.com/transferwise/neptune-web/pull/149). On mobile, the Select component tried to fix the menu to the bottom of the screen. However, when you set something to `position: fixed` inside of a parent who has a `transform: translate()` applied to it, it positions it relative to the parent, not the body. 

We want to fix this behaviour, but not mess around with the Select menu being rendered on the body via a portal every time it's used, since there are a lot of positioning headaches around that.

## 🚀 Changes 

This now renders the menu in a portal **only on mobile** where we have the fixed position at the bottom of the screen. This fixes the issue of it being a child of a transformed parent while not encountering any of the headaches of positioning.

## ✅ Checklist


packages/components/src/select/Select.js
```diff
@@ -296,6 +297,35 @@ export default class Select extends Component {
...
+    const list = (
+      <ul className={dropdownClass} role="menu">
+        {!required && !canSearch ? this.renderPlaceHolderOption() : ''}
+        {canSearch ? this.renderSearchBox() : ''}
+        {this.renderOptions()}
+      </ul>
+    );
+
+    // @media (max-width: @screen-xs-max)
+    if (window.innerWidth <= 575 && typeof document !== 'undefined') {
```
andreapiras: Please use breakpoint costant https://github.com/transferwise/neptune-web/blob/master/packages/components/src/common/propsValues/breakpoint.js#L3

andreapiras: Maybe it s also good to add a check on whether the window object exists (for SSR)

packages/components/src/select/Select.less
```diff
@@ -45,16 +46,22 @@ html.ios-click {
...
+  .dropdown-menu {
+    &--open {
       position: fixed;
       bottom: 0;
       right: 0;
       left: 0;
       top: auto;
       max-width: 100%;
-      width: 100%;
+      width: 100% !important;
```
andreapiras: How come we need important here? Can we achieve this incrementing the specificity?

packages/components/src/select/Select.spec.js
```diff
@@ -532,4 +546,18 @@ describe('Select', () => {
     expect(component.find('.dropdown-toggle').exists()).toBe(false);
     expect(component.find('.dropdown-toggle_TWISAWESOME125').exists()).toBe(true);
   });
+
+  it('creates a portal when rendering on mobile', () => {
+    expect(createPortal).not.toHaveBeenCalled();
+
+    window.innerWidth = 576;
```
andreapiras: Please do use constant again here

packages/components/src/select/Select.spec.js
```diff
@@ -532,4 +546,18 @@ describe('Select', () => {
     expect(component.find('.dropdown-toggle').exists()).toBe(false);
     expect(component.find('.dropdown-toggle_TWISAWESOME125').exists()).toBe(true);
   });
+
+  it('creates a portal when rendering on mobile', () => {
```
andreapiras: Nice tests 👍

packages/components/src/select/Select.js
fergusjordantw: Perfect, thought there would be a JS list of it somewhere but couldn't find it. Figured I'd just commit this and it could get flagged 👍

packages/components/src/select/Select.less
fergusjordantw: Can increase specificity by changing the selector to `.dropdown-menu.dropdown-menu--open`, but thought it'd be better to use the `!important` since `width` was the only property with conflicting styles. Let me know which you prefer

packages/components/src/select/Select.js
```diff
@@ -1,10 +1,12 @@
 import React, { Component, createRef } from 'react';
+import { createPortal } from 'react-dom';
 import Types from 'prop-types';
 import classNames from 'classnames';
 import Transition from 'react-transition-group/Transition';
 
 import Option from './option';
 import KeyCodes from '../common/keyCodes';
+import { Breakpoint } from '../common';
```
Tankenstein: Where is this used?

packages/components/src/select/Select.js
```diff
@@ -296,6 +298,34 @@ export default class Select extends Component {
...
+
+    const list = (
+      <ul className={dropdownClass} role="menu">
+        {!required && !canSearch ? this.renderPlaceHolderOption() : ''}
+        {canSearch ? this.renderSearchBox() : ''}
+        {this.renderOptions()}
+      </ul>
+    );
+
+    if (window && window.innerWidth < Breakpoint.SMALL && typeof document !== 'undefined') {
```
Tankenstein: could you check `typeof window !== 'undefined'` instead of just `window`? Checking `window` when it's undefined will just throw.

Tankenstein: Also, this doesn't handle resizing. is that alright?

fergusjordantw: I figured that would be okay, since all the styling is still done via CSS, the portal only really applies when the Select is a child of something that's been transformed. I'm guessing the amount of times somebody resizes from desktop to mobile and interacts with a select that's nested in something like Tabs is going to be relatively small. Can definitely add a resize listener though and move the `typeof window !== 'undefined' && window.innerWidth < Breakpoint.SMALL && typeof document !== 'undefined'` check to some setState to trigger the re-render

Tankenstein: I think either is fine. Just wanted to make sure it was an explicit decision.

packages/components/src/select/Select.less
fergusjordantw: Actually, adding that level of specificity is still getting overwritten by some media queries. So I think the `!important` is just the better way of handling it this time around

packages/components/src/select/Select.js
fergusjordantw: Have added the event listener, figured it doesn't hurt to have it

fergusjordantw:
> Breakpoints pls sir

Turns out nothing needs to change here, there's just a bit of confusion around sizes.

image

The confusion came from the multiple and inconsistent variables we're using for media queries. In the `_custom-media.less` that's in the monorepo, we have `screen-xs-max` set as `480px`. Over in bootstrap https://github.com/transferwise/bootstrap/blob/master/less/variables.less#L608 (which is where we're getting the actual value from), it's set to `575px` and surrounded by a lot of supposed deprecation. 

Nothing needs to change in this PR now, but this confusion should be looked at. Furthermore, because of bootstraps styling, it's only setting `dropdown-menu-right` to actually go to the right `>= 768px` which leads to a bit of inconsistency. 

Desktop:

Screenshot 2020-01-29 at 14 58 31

Weird `~150px` where we aren't yet in mobile but bootstrap hasn't appleid the `right` styles:

Screenshot 2020-01-29 at 14 58 28

Mobile (now in a portal):

Screenshot 2020-01-29 at 14 58 26

Not necessarily for today. But this stuff should get looked at at some stage

PR #156, opened 2020-01-27
Adding onOpen callback to popover component

gregoirelinot:
## ❓ Context  
We need an easy way to know when the tooltip is opened by the user.

## 🚀 Changes 
Adding a callback to Popover which is called when the popover opens

## ✅ Checklist

- [X] All changes are covered by tests
- [X] The changes are covered in docs
- [X] The commits follow conventional commits standards


packages/components/src/popover/Popover.js
```diff
@@ -30,12 +30,14 @@ export default class Popover extends Component {
       Popover.Placement.BOTTOM_LEFT,
     ]),
     classNames: Types.objectOf(Types.string),
+    onOpen: Types.func,
```
dozortsev: I would do this a bit differently, create a prop `open` and let parent decide when to open `Popover`, just like this done in [`Modal`](https://github.com/transferwise/neptune-web/blob/master/packages/components/src/modal/Modal.js#L28) component.

dozortsev: though I'm not sure how to handle [this](https://github.com/transferwise/neptune-web/blob/master/packages/components/src/popover/Popover.js#L57-L61) with suggested approach 🤔

oliverviljamaa: Do you think that's necessary? I find the fact that it handles its own state quite nifty — in fact, I've missed that functionality with `Tabs`, for example.

phildobsontw: I agree with @oliverviljamaa personally that encapsulating the behaviour within this component is quite handy ☝️. If we were to change this to be a dumb/presentational component, it'll be costly (change of contract as Greg mentioned) and would force each consumer to implement relatively similar patterns across the TW code base.

I can see an argument for this being split into both a presentational and container component (and exporting both) to give the consumer an option of which to use. But may be it will be better to do that when the requirement arises, rather than doing it proactively. Our current requirement is simple - when the popover opens, we want it to call an optional callback. The use case of this isn't to maintain state in our own application - only to enable us to fire a tracking event.

dozortsev: > Do you think that's necessary?

@oliverviljamaa I think yes 🙂because it makes interface of component more flexible, it will be possible to show component on diff actions/events (click, mouse over etc.) which make sense for Popover, right? it's not a Tooltip (which should only appear on hover).

I did think long about this but I struggle to find other use cases when suggested `onOpen` prop will be useful in this component? I don't think adding a prop only for firing MX & GA events is reasonable.

I still would add `open` prop, have follow similar approach in other components (Alert, Modal, Dimmer (maybe a few more)).

---

re. your use case ("to fire a tracking event")

@andreapiras brought seems nice idea, fire MX event on `onClick` of children element, wdyt?

```javascript
<Popover
  title="I am the Popover title"
  content="Lorem ipsum dolor sit amet"
>
  <a
    onClick={() => { mixpanle.event(...) }}
  >
    Click here to Open Popover!
  </a>
</Popover>;
```

gregoirelinot:
Found a way to not modify the component

packages/components/src/popover/Popover.js
oliverviljamaa: Seems a solution was found!

I would allow an outside force to open the popover by default as well indeed, but wouldn't force all consumers to manage their own state.

Regarding the `a` idea, I don't find it nice at all — it's semantically not an anchor at all.

PR #157, opened 2020-01-28
feat: add breakpoint 480

andreapiras:
## ❓ Context  
This PR updates the breakpoints and add EXTRA_SMALL => 480px while before EXTRA_SMALL was 0

## 🚀 Changes 
1) EXTRA_SMALL => 480px while before EXTRA_SMALL was 0
2) Add Login to handle extra breakpoint to flex
2) Add Logincto handle extra breakpoint to box

## ✅ Checklist


packages/components/src/box/Box.js
```diff
@@ -67,6 +69,8 @@ const Box = props => {
   ) : null;
 };
 
+Box.Size = { ...Size, SMALLEST: 'default' };
```
fergusjordantw: Looking at this again, is there a reason this needs to be attached to the `Box` component? Can we not move it above the component declaration? i.e.

```javascript
const boxSizes = { ...Size, SMALLEST: 'default' };

const Box = () => {
    // component
};

Box.propTypes = {
    size: Types.shape({
        [boxSizes.EXTRA_LARGE]: Types.number,
        ...
    })
};
```

andreapiras: Some of the other components are already using this pattern, the reason is because this can be used as sort of extra prop checking. So for example when people use the component they can write this `size={{
      [Box.EXTRA_SMALL]: 0.25,
      [Box.SMALL]: 0.25,
      [Box.MEDIUM]: 0.3333333333333333,
    }}`

And this will be available as soon as u import Box without any extra import/export. You can check another example here where probably the same pattern is more clear
https://transferwise.github.io/neptune-web/components/Alert

PR #158, opened 2020-01-29
fix: remove text/csv type from uploads accepted formats

arikanmstf-tw:
BREAKING CHANGE: Remove text/csv from upload component's accepted formats

## ❓ Context  
At Win7 Chrome/Firefox/IE11 mime type looks empty, and so upload is blocked. Developer can forget about it, so it is better to remove `text/csv` support at all.

Edit: Skip type checking if mime type is empty string `""` and operation system is windows. 

https://transferwise.slack.com/archives/CJNBX9CP6/p1580208301003500

## 💬 Considerations 
Other workaround is, skipping type checking if it is empty string `""` , but this can cause unexpected results. There is no clear fix for this issue. 
Here are some discussions for the issue:

https://github.com/react-dropzone/react-dropzone/issues/391
https://github.com/danialfarid/ng-file-upload/issues/1211

## ✅ Checklist

### DIFF
--- a/packages/components/src/upload/Upload.js
+++ b/packages/components/src/upload/Upload.js
@@ -8,17 +8,7 @@ import './Upload.css';
 import ProcessIndicator from '../processIndicator';
 
 const PROCESS_STATE = ['error', 'success'];
-/*
- * Do not add `text/csv` to the accepted formats.
- * This mime type is not supported by win7 browsers.
- * At Win7 Chrome/Firefox/IE11 mime type looks empty, and so upload is blocked
- * Use `*` for `text/csv` and make type check in backend.
- * Reference:
- * https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/ms775147(v=vs.85)?redirectedfrom=MSDN
- * https://cs.chromium.org/chromium/src/net/base/mime_util.cc?q=mime&sq=package:chromium&g=0&l=1
- * https://transferwise.slack.com/archives/CJNBX9CP6/p1580208301003500
- */
-const ACCEPTED_FORMAT = ['*', 'image/*', 'application/*'];
+const ACCEPTED_FORMAT = ['*', 'image/*', 'application/*', 'text/csv'];
 
 /*
  * This delay is required for the isError/isSuccess to be fired after isProcessing so the processIndicator, will be
--- a/packages/components/src/upload/Upload.js
+++ b/packages/components/src/upload/Upload.js
@@ -8,7 +8,17 @@ import './Upload.css';
 import ProcessIndicator from '../processIndicator';
 
 const PROCESS_STATE = ['error', 'success'];
-const ACCEPTED_FORMAT = ['*', 'image/*', 'application/*', 'text/csv'];
+/*
+ * Do not add `text/csv` to the accepted formats.
+ * This mime type is not supported by win7 browsers.
+ * At Win7 Chrome/Firefox/IE11 mime type looks empty, and so upload is blocked
+ * Use `*` for `text/csv` and make type check in backend.
+ * Reference:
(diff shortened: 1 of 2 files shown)


packages/components/src/upload/Upload.js
```diff
@@ -8,7 +8,17 @@ import './Upload.css';
...
+ * Do not add `text/csv` to the accepted formats.
+ * This mime type is not supported by win7 browsers.
+ * At Win7 Chrome/Firefox/IE11 mime type looks empty, and so upload is blocked
+ * Use `*` for `text/csv` and make type check in backend.
+ * Reference:
+ * https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/ms775147(v=vs.85)?redirectedfrom=MSDN
+ * https://cs.chromium.org/chromium/src/net/base/mime_util.cc?q=mime&sq=package:chromium&g=0&l=1
+ * https://transferwise.slack.com/archives/CJNBX9CP6/p1580208301003500
+ */
+const ACCEPTED_FORMAT = ['*', 'image/*', 'application/*'];
```
andreapiras: Hey I'm not sure about this solution. This removed the check also for system that actually support it. I prefer validation in some cases vs never validate. Maybe we can check if the type is empty (like you suggested) and then put some logic around. I would not leave only backend validation

arikanmstf-tw: > Hey I'm not sure about this solution. This removed the check also for system that actually support it. I prefer validation in some cases vs never validate. Maybe we can check if the type is empty (like you suggested) and then put some logic around. I would not leave only backend validation

after your comments. I came up with a idea: skipping type checks for windows os

packages/components/src/upload/utils/isTypeValid/isTypeValid.js
```diff
@@ -1,4 +1,21 @@
 export const isTypeValid = (file, rule) => {
+  if (file.type === '') {
```
andreapiras: https://stackoverflow.com/questions/52696000/wrong-mime-type-in-windows-browser-for-text-csv-files

I would at least add a check on the file extension ( I know it doesn't guarantee the type) but still better than nothing?

arikanmstf-tw: but what if the file is valid csv file without extension ?

andreapiras: I guess that would be a problem also for OS as if the file doesn't have extension probably it will appear as disabled for the used in the upload window

arikanmstf-tw: no, in windows all type of files are enabled even if you set usAccept to "text/csv" , because for windows, it is unknown.
in mac/linux. there is no problem at all.

sbrandwoo: I would agree on checking the file extension. I would expect the code to be:

```
if (file.type) {
  validateType
} else {
  validateExtension
}
```

And then the server side needs to revalidate the file using a more reliable method.

andreapiras: Maybe we can improve the validation message in case the extension is missing?

arikanmstf-tw: > I would agree on checking the file extension. I would expect the code to be:
> 
> ```
> if (file.type) {
>   validateType
> } else {
>   validateExtension
> }
> ```
> 
> And then the server side needs to revalidate the file using a more reliable method.

if statement should be like this
`if (file.type !== '') { ... }`

because windows gives empty string for file type.

helendurranttw:
Closing this for now. I've raised a new issue to improve validation on upload components against Neptune Web.

PR #163, opened 2020-02-04
fix(select): rm focus after open for touch devices

sergonius:
## 📎 Jira ticket
https://transferwise.atlassian.net/browse/SEN-437

## ❓ Context
The select has buggy behaviour on mobile (or iOS, at least) and works differently from what we have on the app (we don't focus the search when you open the currency picker).

## 🚀 Changes
Added touch device detection (used this method: https://stackoverflow.com/a/52855084).
Added a check that you're not using a touch device before focusing.

## 💬 Considerations
Another "fix" would have been to add a timeout to the focus, but it's not very pretty as you can see below.

### Before: iOS
___before

### After: iOS
___after--nofocus

### After (alternative test with timeout): iOS
___after--focus-timeout

### After: desktop (remains the same)
___after--desktop


danlaush:
How does this behave on dual-mode devices, like laptops with touchscreens or ipads with pencils?

sergonius:
> How does this behave on dual-mode devices, like laptops with touchscreens or ipads with pencils?

From the stackoverflow answer I linked:
> The above solution is to detect if a "coarse pointer" - usually a touch screen - is the primary input device. In case you want to dectect if a device with e.g. a mouse also has a touch screen you may use any-pointer: coarse instead.

danlaush:
Is it possible to detect the pointer from the event itself, instead of the window? So then it can feel natural when I'm sausage fingers or penciling.

PR #165, opened 2020-02-04
fix: increase react classnames specificity

andreapiras:
## 📎 Jira ticket

https://transferwise.atlassian.net/browse/DS-262
## ❓ Context  
We identified a problem when using components inside nextJS. When the bundler runs the order of the css changes giving also a warning
Conflicting order between:
@transferwise/neptune-docs:  * css /Users/andrea.piras/Projects/design-system/neptune/node_modules/css-loader??ref--8-1!/Users/andrea.piras/Projects/design-system/neptune/node_modules/@transferwise/components/lib/tabs/Tabs.css
The problem we are facing is due to this structure
import 'bootstrap/dist/css/bootstrap.css';
....
<Button .../> 
<Select />
where Button and Select imports its own css automatically. The problem is that bootstrap still overrides the component's styles.
This cause some css to not be applied correctly. We didn't identify a solution yet but for now you can import
import 'bootstrap/dist/css/bootstrap.css';
in this order. The downside is that you'll get more css (~55Kb + every component you use css)

## 🚀 Changes 
Increase specificity of React components overrides. This chages should potentially be merged into neptune-css if they apply to also html components

## ✅ Checklist


packages/docs/pages/_app.js
```diff
@@ -5,6 +5,8 @@ import Layout from '../components/Layout';
 import 'bootstrap/dist/css/bootstrap.css';
 import 'currency-flags/dist/currency-flags.min.css';
 import '@transferwise/icons/dist/icons.min.css';
+// @TODO to be fixed as https://transferwise.atlassian.net/browse/DS-262
+import '@transferwise/components/build/main.css';
```
mikknarusk: does this not only fix neptune docs? Sorry, not very familiar with the structure of this repo

andreapiras: Yes this fix only neptune docs for now until we find a solution.

packages/components/src/modal/Modal.less
```diff
@@ -1,13 +1,13 @@
...
 
-.modal-np {
+.tw-modal {
   height: 100%;
   width: 100vw;
   display: flex;
   justify-content: center;
   align-items: center;
-  .modal-np-dialog {
+  .tw-modal-dialog {
```
nabaroa: Could we use BEM here and in all the siblings?
`.tw-modal__dialog`

packages/components/src/tooltip/Tooltip.js
```diff
@@ -75,7 +75,7 @@ class Tooltip extends Component {
           this.elementReference = elementReference;
         }}
         aria-describedby={tooltipId}
-        className="tooltip-container"
+        className="tw-tooltip-container"
```
nabaroa: BEM?

packages/components/src/upload/Upload.less
```diff
@@ -1,19 +1,19 @@
 // This css is a quick workaround on the height of the upload container collapsing if UploadImageStep is
 // not rendered.
-.droppable-lg {
+.tw-droppable-lg {
```
nabaroa: `.tw-droppable--lg`

packages/components/src/button/Button.js
```diff
@@ -20,6 +20,7 @@ const Button = ({
 }) => {
   const classes = classNames(
     `btn btn-${size}`,
+    `tw-btn tw-btn-${size}`,
```
nabaroa: `.tw-btn--${size}`

packages/components/src/tooltip/Tooltip.js
andreapiras: For now I just prefixed the existing one. In a next step we should ideally move all these classes to neptune and maybe at the point we can rename them?

PR #166, opened 2020-02-05
feat: add alignment property to instructions list component

istvanmakary:
## Description:
To re-implement [choose template](https://transferwise.com/template-instructions/templates/choose-a-template) card components, I had to create a right aligned variation of the instructionsList component.

## 📎 Jira ticket
No tickets involved.

## ❓ Context  
See description

## 🚀 Changes 
- InstructionsList.js
- InstructionsList.less
- InstructionList.story.js
- instructionList.code.js

## ✅ Checklist


packages/components/src/instructionsList/InstructionList.story.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
-import InstructionsList from './InstructionsList';
-import { array } from '@storybook/addon-knobs';
+import InstructionsList, { ALIGNMENTS } from './InstructionsList';
```
andreapiras: Hey Istvan, we have this values defined [here](https://github.com/transferwise/neptune-web/blob/master/packages/components/src/common/propsValues/placement.js
) can we try and reuse them please

packages/components/src/instructionsList/InstructionsList.js
```diff
@@ -5,20 +5,25 @@ import TickCircleIcon from '@transferwise/icons/react/tick-circle';
 
 import './InstructionsList.css';
 
-const InstructionsList = ({ dos, donts }) => (
-  <div className="instructions">
+export const ALIGNMENTS = {
```
andreapiras: Maybe we can expose this values as part of the component something like [this](https://github.com/transferwise/neptune-web/blob/master/packages/components/src/button/Button.js#L44)  
So we can access it later without having to import separately but just using InstructionsList.Alignments

packages/components/src/instructionsList/InstructionsList.less
```diff
@@ -4,9 +4,24 @@
...
+        margin-right: auto;
+      }
+    }
+  }
+
   .instruction {
     display: flex;
 
+    .instructions-text {
+      margin-left: @spacer * 2;
```
andreapiras: Can we not switch those classes m-l-2 and the others based on the value of props. That would help to 

1. Reusing existing css
2. Can be tested

oliverviljamaa:
Do our designs **really** need to have instructions with icons aligned both left and right? Could you challenge the designers on that?

gusso:
Had a chat with @istvanmakary, he explained his intention was to maintain the source design of the current page, which is why he wanted to update the component style.

We agreed that in this case there is no reason to have both layout options, as scanning icons on the right makes it harder to associate with the text content. The page can accept the visual change and he'll use the component as it is.

PR #168, opened 2020-02-05
fix(select): tab will select item under focus in select

jonstieglitz:
## 📎 Jira ticket

N/A

## ❓ Context  

I don't like hitting the enter key on forms for fear of submitting the form. Selects often appear in forms. Also, tab is usually used to navigate through a form. It would be great if in addition to hitting ENTER, you could hit TAB and the item that is already highlighted would be selected in the Select

## 🚀 Changes 

This PR calls the selectKeyboardFocusedOption when tab is hit instead of the current behavior of blocking the tab. Also added a test to cover this.

## 💬 Considerations 

If this functionality was already considered and rejected then please reject this PR

## ✅ Checklist


twrdjpalmer:
I'm not sure about this. I agree we shouldn't block the user from tabbing, but to match the user's expectations it should tab _out_ of the select and onto the next focusable element. From a quick play, the current behaviour (aside from the blocking of the tab) is expected when compared to the browser-native experience

jonstieglitz:
> I'm not sure about this. I agree we shouldn't block the user from tabbing, but to match the user's expectations it should tab _out_ of the select and onto the next focusable element. From a quick play, the current behaviour (aside from the blocking of the tab) is expected when compared to the browser-native experience

I thought someone might suggest that. Hard to argue with it. You're saying ENTER is for selecting and TAB is for leaving. I would argue that ESC is for leaving and that TAB is in a middle ground. 

The way I see it, the user already entered some text and has an item under focus and then presses TAB. If I did that, I would be annoyed that my work was wasted and deleted.

fergusjordantw:
> I agree we shouldn't block the user from tabbing, but to match the user's expectations it should tab out of the select and onto the next focusable element.

Or, it tabs through each item in the select list until it reaches the end, at which point if they hit tab again it closes the select and goes to the next focusable element

jonstieglitz:
> > I agree we shouldn't block the user from tabbing, but to match the user's expectations it should tab out of the select and onto the next focusable element.
> 
> Or, it tabs through each item in the select list until it reaches the end, at which point if they hit tab again it closes the select and goes to the next focusable element

The arrow keys work so well for navigating through the options and if TAB went through each item in the list it would get very annoying for long lists (like the 50 States in the State selector). Why duplicate the arrow functionality with TAB when it could add new functionality?

twrdjpalmer:
>  If I did that, I would be annoyed that my work was wasted and deleted.

Comparing to the native select box: Using arrow keys to move to another item & tabbing selects the currently highlighted value, closes the select and moves the user on to the next focusable item.

I can see now that this is actually what your PR achieves! My apologies, I understood it to be what @fergusjordantw was suggesting 😛 

> Or, it tabs through each item in the select list until it reaches the end, at which point if they hit tab again it closes the select and goes to the next focusable element or cycles back to the beginning of the available options.

@fergusjordantw trouble with this is that it's totally different behaviour to the native select. IMO we should avoid this where we can and go for as close to a native experience as possible

fergusjordantw:
> @fergusjordantw trouble with this is that it's totally different behaviour to the native select.

Fair, it's been a long time since I've had to interact with a native select 😂

If that's the behaviour, then keep it inline with native behaviour. Only suggested what I did because to me, conceptually, when I open a select list I'm interacting with a new layer of content which I would hope is tab navigate-able the same way the rest of a page would be, going from element to element 🤷‍♂

jonstieglitz:
> Can we make it so it has the rest of the behaviour as well? E.g., we enable tab when the select is close, we close, select the item and move onto the next select when the select is open, space opens the select/selects an item, `esc` closes it etc. etc.

I think this PR already has all of that behavior but correct me if I'm wrong.

https://drive.google.com/file/d/1SzKQOKwcBRBgv5293dBAnA25MDeiEGdB/view?usp=sharing

This is a video of the behavior of this PR. Hard to tell exactly what is going on because you can't see what keys i'm pressing but i'm tabbing through the fields. The first time I open the select I use SPACE to open it, ARROWS to get to an item, and then TAB to leave. The second time I open it I use ESC to leave without selecting the item under focus.

twrdjpalmer:
Awesome stuff in that case!

oliverviljamaa:
> > Can we make it so it has the rest of the behaviour as well? E.g., we enable tab when the select is close, we close, select the item and move onto the next select when the select is open, space opens the select/selects an item, `esc` closes it etc. etc.
> 
> I think this PR already has all of that behavior but correct me if I'm wrong.
> 
> https://drive.google.com/file/d/1SzKQOKwcBRBgv5293dBAnA25MDeiEGdB/view?usp=sharing
> 
> This is a video of the behavior of this PR. Hard to tell exactly what is going on because you can't see what keys i'm pressing but i'm tabbing through the fields. The first time I open the select I use SPACE to open it, ARROWS to get to an item, and then TAB to leave. The second time I open it I use ESC to leave without selecting the item under focus.

For the future — every branch creates a docs deployment so anyone reviewing can play with the live docs for that branch 😊

PR #175, opened 2020-02-10
refactor!: update bundles format

andreapiras:
## ❓ Context  
This PR aims to unify the transpilation and polyfilling between the two different bundles: UMD,ES.
It also introduces the polyfilled and unpolyfilled version for each type. To get a better understanding please check this [page](https://transferwise.github.io/neptune-web/branch/polyfill-umd-bundle/components/About)

## 🚀 Changes 
1) Remove component's  babelrc and replace with babel.config.js
2) Add polyfilled and unpolyfilled version for both UMD and ES
3) Bump version for rollup

## ✅ Checklist


packages/components/.babelrc
```diff
@@ -3,7 +3,7 @@
     [
```
andreapiras: Do NOT CR this file as it has been removed in favour of babel.config.js in next commit

sergonius:
Is there an easy way to know which polyfills are being included with each component?

andreapiras:
> Is there an easy way to know which polyfills are being included with each component?

Polyfills are added automatically using  useBuiltIns: 'usage' preset-env option. I guess one way to discover would be to check the es polyfilled version and see what files gets included

> Is there an easy way to know which polyfills are being included with each component?

Hey the polyfills are included automatically using preset-env usage and babel-plugin-transform-runtime. If you open the es polyfilled version you'll see which polyfill gets included at the top of the file. I'm not sure there's an easier way

packages/components/.browserslistrc
```diff
@@ -0,0 +1 @@
+last 1 version
```
oliverviljamaa: Are we sure that's the browser list we want to go with? For example, we have Safari 11–13 in the 95% of front page visitors:
https://mixpanel.com/report/260146/jql-console/#edit/3277505

packages/components/babel.config.js
```diff
@@ -0,0 +1,100 @@
+const umdConfig = {
+  presets: [
+    [
+      // @babel/preset-env is a smart preset that allows you to use the latest JavaScript without needing to micromanage which
+      // syntax transforms (and optionally, browser polyfills) are needed by your target environment(s).
+      '@babel/preset-env',
+      {
+        // This option configures how @babel/preset-env handles polyfills.
+        // useBuiltIns: 'usage' Adds specific imports for polyfills when they are used in each file.
+        // We take advantage of the fact that a bundler will load the same polyfill only once.
```
oliverviljamaa: Cool.

packages/docs/utils/bundleSize/getBundleSize.js
```diff
@@ -0,0 +1,48 @@
...
+  return returnArray;
+};
+
+const getTotalSize = directoryPath => {
+  const arrayOfFiles = getAllFiles(directoryPath);
+
+  let totalSize = 0;
+
+  arrayOfFiles.forEach(filePath => {
+    totalSize += fs.statSync(filePath).size;
```
oliverviljamaa: Is this gzipped? If not, we should provide gzipped size, as that's the size that actually gets served to the customers.

packages/components/.browserslistrc
andreapiras: This query https://browserl.ist/?q=last+2+version%2Cnot+IE+10 shoudl cover better then. Even If I don't think that we have any particular feature that will break Safari 12 and not 13. But until we have a proper multibrowser test better to stay on the safe side

packages/docs/utils/bundleSize/getBundleSize.js
andreapiras: Do you actually gzip the file on you side or do you need me to createa gzip version of the bundles?

oliverviljamaa: The server does it, so nothing to do from components' side, but we should gzip in this file to show gzipped size in the docs 😊

andreapiras: Okie, I ll add this to docs then :)

packages/components/.browserslistrc
oliverviljamaa: I think I'd make a custom list according to our needs, as we don't need to support the last 2 versions of _every_ browser — we don't have any Blackberry visits, for example. Shouldn't be too hard.

oliverviljamaa: I've updated the doc with the recommended `browserslist` string from the query above:
https://transferwise.atlassian.net/wiki/spaces/FE/pages/223674484/Supported+browsers+and+resolutions

andreapiras: Ok but if we use a custome query like Chrome >=71, Firefox >=71, iOS >=11, Safari >=11.1, IE 11, Opera >=65, Edge >=18 it means that periodically we have to check the browser. Does mixpanle has an API that we can use to update the string automatically? Maybe we can launch with the string above and I can check if we can get the json from mixpanel

PR #178, opened 2020-02-12
fix(tabs): animate to tab on mount

sergonius:
There were a few bugs caused by not calling this method.

1. `selectedTabIndex` was not set when you started resizing, breaking the tabs
2. When the `selected` prop was initially anything but `0`, the UI didn't really reflect that

This one-liner solves both of the issues.


sergonius:
Sorry, the branch name doesn't work with our CI. Closing in favour of https://github.com/transferwise/neptune-web/pull/179

PR #180, opened 2020-02-12
fix(select): put the overlay in a portal

sergonius:
## ❓ Context
The previous solution of having the overlay in the dropdown as a pseudo element doesn't play well with `transform`s on parents, causing stuff like this to happen:
image
(a bit of a poor image, but the overlay should cover the tabs and the barely visible side-bits as well)

## 🚀 Changes
Having a portal for the overlay same way as for the list itself solves the problem.
Also changed the use of `window.innerWidth` to `window.matchMedia`.

## ✅ Checklist


packages/components/src/select/Select.js
```diff
@@ -36,7 +36,8 @@ function getShouldRenderWithPortal() {
   return (
     typeof document !== 'undefined' &&
     typeof window !== 'undefined' &&
-    window.innerWidth < Breakpoint.SMALL
+    window.matchMedia &&
+    window.matchMedia(`(max-width: ${Breakpoint.SMALL}px)`).matches
```
fergusjordantw: 😍

packages/components/src/select/Select.js
```diff
@@ -458,6 +459,16 @@ export default class Select extends Component {
     return <span className={this.style('form-control-placeholder')}>{placeholder}</span>;
   }
 
+  renderOverlay() {
+    const { open, shouldRenderWithPortal } = this.state;
+
+    if (open && shouldRenderWithPortal) {
+      return createPortal(<div className="select-overlay" />, document.body);
```
fergusjordantw: Are we sure a new div was needed and we couldn't just move the `:before` element to live relative to the `.dropdown-menu` instead of the `.dropdown` button?

packages/components/src/select/Select.js
Tankenstein: nice

packages/components/src/select/Select.js
sergonius: If it would be inside the dropdown menu along with the list items, it couldn't really be behind the dropdown itself.

packages/components/src/select/Select.spec.js
```diff
@@ -590,27 +589,30 @@ describe('Select', () => {
...
     openSelect();
     expect(createPortal).not.toHaveBeenCalled();
     expect(element('.dropdown-menu--open').exists()).toBe(true);
+    expect(document.querySelector('.select-overlay')).toBeFalsy();
 
     component.setState({ shouldRenderWithPortal: true });
     openSelect();
     expect(createPortal).toHaveBeenCalledWith(expect.anything(), document.body);
     expect(element('.dropdown-menu--open').exists()).toBe(false);
+    setTimeout(() => {
```
Tankenstein: The test does not know about this assertion. Either add `done` or wrap it in a promise.

packages/components/src/select/Select.js
fergusjordantw: Fair, just played around with it in the DOM, figured you'd be able to get it to work with a `z-index: -1` to put it behind the list items but you'd need to change the styling on those a bit as well 👍

PR #182, opened 2020-02-13
feat: add simple search param on search

istvanmakary:
This Pull Request was created to simplify how search is handled in the Select component. 
To cover most of the use cases, there is no need to store the current search value and do the filtering, the Select module does it for you. If you need advance filtering you can specify your search function with any custom logic.
For really complex cases the previous implementation should be used.
Of course it's backward compatible.

## ✅ Checklist


packages/docs/liveEditorCode/select.code.js
```diff
@@ -18,10 +16,9 @@
       disabled={false}
       onChange={v => setSelected(v)}
       required={false}
-      searchValue={searchValue}
       searchPlaceholder="searchplaceholder"
-      onSearchChange={v => setSearchValue(v)}
       dropdownUp={false}
+      search={true}
```
sbrandwoo: In the documentation, is it possible to document what parameters the function takes?
https://transferwise.github.io/neptune-web/branch/improve-select-search/components/Select

packages/components/src/select/Select.js
```diff
@@ -44,6 +47,9 @@ const BOOTSTRAP_DROPDOWN_ANIMATION_TIME = 200;
 
 const defer = fn => setTimeout(fn, 0);
 
+const defaultFilterFunction = (option, keyword) =>
+  option.label && option.label.toLowerCase().includes(keyword.toLowerCase());
```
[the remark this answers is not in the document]
istvanmakary: done

packages/docs/liveEditorCode/select.code.js
istvanmakary: done

packages/components/src/select/Select.js
```diff
@@ -44,6 +47,12 @@ const BOOTSTRAP_DROPDOWN_ANIMATION_TIME = 200;
 
 const defer = fn => setTimeout(fn, 0);
 
+const includesString = (str1, str2) => str1.toLowerCase().includes(str2.toLowerCase());
```
sbrandwoo: Does `includes` get polyfilled for IE?

packages/components/src/select/Select.js
```diff
@@ -44,6 +47,12 @@ const BOOTSTRAP_DROPDOWN_ANIMATION_TIME = 200;
...
+
+const defaultFilterFunction = (option, keyword) =>
```
sbrandwoo: This filters out headers too. I think that's a good thing though.

packages/components/src/select/Select.js
andreapiras: Can we not use indexOf here?

istvanmakary: sure, I'll change it!

packages/components/src/select/Select.js
istvanmakary: As I saw, the default behaviour was to filter out headers. Or not?

packages/components/src/select/Select.js
istvanmakary: Done!

packages/components/src/select/Select.js
istvanmakary: We discussed with @sbrandwoo that the current implementation of this feature is fine.

PR #184, opened 2020-02-24
feat: implement withNextPortal HOC + updata Dimmer, Snackbar

istvanmakary:
## ❓ Context  

When using the `SnackbardProvider` with Crab it's conflicting when the it's doing the hydration on the client side. The result of this conflict is that, it takes a wrong div as snackbar wrapper. To me it looks that it takes the first child div of `SnackbardProvider`.

On the fallowing screenshot you can see a snackbar component rendered on the [users](https://trwi.se/users/) page.
screenshot

## ✅ Checklist


packages/components/src/withNextPortal/withNextPortal.js
```diff
@@ -0,0 +1,14 @@
+import React, { useState, useEffect } from 'react';
+import { createPortal } from 'react-dom';
+
+const withNextPortalWrapper = Component => props => {
```
domhorvathtw: just a personal naming preference, but I like to add HOC as a suffix to Higher Order Components.

packages/components/src/withNextPortal/withNextPortal.spec.js
```diff
@@ -0,0 +1,26 @@
...
+  });
+
+  it('should create a portal with the component', () => {
+    ReactDOM.createPortal.mockReturnValue(() => null);
+    const props = {};
+    const Component = withNextPortal(AnyComponent);
+    const expected = mount(<AnyComponent {...props} />);
+    mount(<Component {...props} />);
+    expect(ReactDOM.createPortal).toBeCalled();
+    const [comp, body] = ReactDOM.createPortal.mock.calls[0];
```
domhorvathtw: 🚀

packages/components/src/dimmer/Dimmer.js
```diff
@@ -85,5 +85,4 @@ Dimmer.defaultProps = {
 
 export { Dimmer };
 
-export default props =>
-  typeof document !== 'undefined' ? createPortal(<Dimmer {...props} />, document.body) : '';
```
domhorvathtw: what was exactly the problem with this solution that we had to replace it with a HOC?

packages/components/src/withNextPortal/withNextPortal.js
```diff
@@ -0,0 +1,14 @@
...
+  const [mounted, setMounted] = useState(false);
```
domhorvathtw: does it mean that we can never SSR a snackbar that is already mounted?

PR #185, opened 2020-02-25
Ds 273 flow navigation restructure

andreapiras:
## 📎 Jira ticket
https://transferwise.atlassian.net/browse/DS-273

## ❓ Context  
Some designs need a subset of FlowNavigation so we decided to separate the existing logic into a different component that can output the desired design without the extra bits coming from flowNavigation.

## 🚀 Changes 

1) Creates Header component
2) Refactor flowNavigation to use Header
3) Update tests minimally
4) Add storybook for header

## 💬 Considerations 
Please check deployment for preview https://transferwise.github.io/neptune-web/branch/DS-273-flow-navigation-restructure/components/FlowHeader

## ✅ Checklist


packages/components/src/header/index.js
```diff
@@ -0,0 +1,3 @@
+import Header from './Header';
+
+export default Header;
```
oliverviljamaa:
```suggestion
export { default } from './Header';
```

packages/components/src/flowNavigation/FlowNavigation.js
```diff
@@ -23,61 +25,53 @@ const FlowNavigation = ({
...
+        'm-l-3',
+        {
+          'close-button-with-avatar': !done,
+        },
+      )}
+      onClick={onClose}
+    />
+  );
+  return (
+    <Header
```
oliverviljamaa: Is the idea that everyone needing the flow navigation without steps and avatar should know about the back button, logo, and header? I love the refactor, but I'd recommend still exposing a composite component.

packages/components/src/header/index.js
andreapiras: Love this! At the moment all components are exported this way so I'd rather change them all in one go :)
https://transferwise.atlassian.net/browse/DS-281

oliverviljamaa: I think there's quite a few `export { default }`s as well, I checked `Modal` and `Alert`, for example. But of course agree that we should unify them.

packages/components/src/flowNavigation/FlowNavigation.js
andreapiras: Hey I think it does make sense to have a sort of FlowHeader which wraps Header and pass the default values. Would you name it FlowHeader?

packages/components/src/header/index.js
andreapiras: yeah it's not very consistent so let's do a general cleanup

packages/components/src/flowHeader/FlowHeader.less
```diff
@@ -0,0 +1,106 @@
+@import 'bootstrap/less/variables';
+
+.tw-flow-navigation {
```
helendurranttw: Has this code been removed from FlowNavigation.less?

packages/components/src/header/Header.js
```diff
@@ -0,0 +1,34 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
```
helendurranttw: Is this necessary here? Might be able to remove it

packages/components/src/flowHeader/FlowHeader.js
```diff
@@ -0,0 +1,69 @@
...
+import classNames from 'classnames';
+// Avatar and Logo to be refactored and exposed DS-274 DS-275
+import Avatar from '../flowNavigation/avatar';
+import Logo from '../flowNavigation/logo';
+import Header from '../header';
+import { Theme } from '../common';
+
+import './FlowHeader.css';
+
+const ProfileType = {
```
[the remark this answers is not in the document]
andreapiras: This should sit in the common/propValue and also attached to the component as static method. But I didn't want to pollute this PR. Can we do this in separate ticket?

packages/components/src/flowHeader/FlowHeader.js
```diff
@@ -0,0 +1,71 @@
...
+  BUSINESS: 'BUSINESS',
+  PERSONAL: 'PERSONAL',
+};
+
+const FlowHeader = ({ avatarUrl, profileType, onClose, theme, showAvatar }) => {
```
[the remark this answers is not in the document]
andreapiras: yep makes sense need also to pair with @gusso and @anttio to check if we can give some guidance around where to use one or the other version

packages/components/src/flowHeader/FlowHeader.spec.js
```diff
@@ -0,0 +1,66 @@
...
+    closeButton().simulate('click');
+    expect(props.onClose).toHaveBeenCalled();
+  });
+
+  it('hides the close button if onClose is not defined', () => {
+    component.setProps({ onClose: undefined });
+    expect(closeButton().exists()).toBe(false);
+  });
+
+  it('inverts the logo if theme is dark', () => {
```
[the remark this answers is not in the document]
andreapiras: makes sense yes. These test will move inside logo once we extract it

packages/components/src/flowHeader/FlowHeader.less
andreapiras: good catch, this needs to be moved to neptune-css

packages/components/src/flowHeader/FlowHeader.js
```diff
@@ -0,0 +1,91 @@
...
+import Avatar from './avatar';
+import { Theme } from '../common';
+import Logo from './logo';
+
+const ProfileType = {
+  BUSINESS: 'BUSINESS',
+  PERSONAL: 'PERSONAL',
+};
+
+const FlowHeader = ({
```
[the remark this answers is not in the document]
andreapiras: LoggedHeader?

packages/components/src/flowHeader/FlowHeader.js
```diff
@@ -0,0 +1,69 @@
...
+import classNames from 'classnames';
+// Avatar and Logo to be refactored and exposed DS-274 DS-275
+import Avatar from '../flowNavigation/avatar';
+import Logo from '../flowNavigation/logo';
+import Header from '../header';
+import { Theme, ProfileType } from '../common';
+
+import '@transferwise/neptune-css/dist/css/flowNavigation.css';
+
+const FlowHeader = ({ avatarUrl, profileType, onClose, theme, showAvatar }) => {
```
helendurranttw: Here is how I would document a component like this, using a more casual approach, and using a more thorough approach in a [JSDoc style](https://jsdoc.app/about-getting-started.html).

```
/**
 * OverlayHeader is a header component that provides a logo, avatar and close button, for use in overlay screens.
 * 
 * The avatar will not show if showAvatar is false. If avatarUrl is falsey, it will show a placeholder avatar.
 * The close button will only show if you provide the onClose callback.
 *
 **/
const OverlayHeader = ({ avatarUrl, profileType, onClose, theme, showAvatar }) => {...
```

```
/**
 * OverlayHeader is a header component that provides a logo, avatar and close button, for use in overlay screens.
 * 
 * @param {string} [avatarUrl] - URL of the avatar. If not present, a placeholder will be used instead.
 * @param {function} [onClose] - Callback for the close button. If not present, no close button will be rendered.
 * @param {string} [theme=THEME.light] - Theme to use
 * @param {boolean} [showAvatar=true] - If false, no avatar will be rendered - even if avatarURL is provided.
 * @param {string} profileType - ...
 *
 * @usage `<OverlayHeader avatarUrl={someUrl}, profileType={PROFILETYPE.BUSINESS}/>`
 **/
const OverlayHeader = ({ avatarUrl, profileType, onClose, theme, showAvatar }) => {

```

andreapiras: I think I prefer the second one which is more datailed

packages/components/src/overlayHeader/OverlayHeader.js
```diff
@@ -0,0 +1,72 @@
...
+import classNames from 'classnames';
+// Avatar and Logo to be refactored and exposed DS-274 DS-275
+import Avatar from '../flowNavigation/avatar';
+import Logo from '../flowNavigation/logo';
+import Header from '../header';
+import { Theme, ProfileType } from '../common';
+
+import '@transferwise/neptune-css/dist/css/flowNavigation.css';
+
+const OverlayHeader = ({ avatarUrl, profileType, onClose, theme, showAvatar }) => {
```
oliverviljamaa: Did we consider having one avatar prop instead of multiple? It's an improvement over `showAvatar`, but I can't see why we need two different props rather than:
`avatar={null | { url: 'https://some-url.com/image.jpg' } | { url: null } }`

oliverviljamaa: What would `showAvatar={true}` and `avatarUrl={undefined}` do? Would it show the placeholder? If yes, I feel we're encouraging the wrong behaviour of just showing the placeholder and not getting the real URL for users who have it 🤔

packages/components/src/flowNavigation/FlowNavigation.js
oliverviljamaa: Good question — I think it's fine to assume the back button is for going back a step, but not entirely sure 🤔

packages/components/src/flowNavigation/FlowNavigation.js
```diff
@@ -23,67 +19,63 @@ const FlowNavigation = ({
...
-          <div
-            className={classNames('logo', 'logo-3', 'hidden-xs', {
-              'logo-primary': theme === Theme.LIGHT,
-              'logo-inverse': theme === Theme.DARK,
-            })}
-          />
+}) => {
+  const avatar = done ? null : <Avatar url={avatarUrl} profileType={profileType} />;
+  const closeButton = onClose && (
+    <button
```
oliverviljamaa: It's great that we have the `OverlayHeader` now, but this part still gets duplicated — couldn't `FlowNavigation` extend `OverlayHeader`?

packages/components/src/overlayHeader/OverlayHeader.spec.js
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
+import { mount } from 'enzyme';
```
oliverviljamaa: Should we use `react-testing-library` for these new components? 🎉

andreapiras: Hey it's a good idea for a second iteration. I don't think is a good idea to make a major refactor and change the tests in the same time. But I'm defo up to update the tests in a second round

oliverviljamaa: Oh, yup, sure, I thought they were new tests 😊

packages/components/src/flowNavigation/FlowNavigation.js
andreapiras: At the moment they both extends Header component. I would leave them separate as the logic in the content is similar but not equal. For example:
In OverlayHeader
`leftContent={
        <div className="m-lg-t-1">
          <Logo theme={theme} onGoBack={null} />
        </div>
      }`
In FlowNavigation
`leftContent={
        <div className="m-lg-t-1">
          <Logo theme={theme} onGoBack={onGoBack} />
          <BackButton steps={steps} activeStep={activeStep} onGoBack={onGoBack} />
        </div>
      }`

packages/components/src/overlayHeader/OverlayHeader.js
andreapiras: What if people needs to show a placeholder then? How would you achieve that if not with showAvatar={true} and avatarUrl={undefined} ?

oliverviljamaa: `avatar={{ url: null }}` — indicating we want to show an avatar, but don't have an image for it.

packages/components/src/flowNavigation/FlowNavigation.js
andreapiras: BackButton and Stepper are tightly coupled since BackButton accept steps and currentStep as props. So it's correct to say that if stepper is not provided BackButton should not be there neither

packages/components/src/overlayHeader/OverlayHeader.js
```diff
@@ -0,0 +1,87 @@
...
+  },
+  profileType: OverlayHeader.ProfileType.PERSONAL,
+  onClose: null,
+  theme: Theme.LIGHT,
+};
+
+OverlayHeader.propTypes = {
+  avatar: Types.oneOfType([
+    Types.shape({
+      url: Types.any.isRequired,
```
oliverviljamaa: I think this shouldn't be required, otherwise we'll always log the error when we don't have an avatar (and when we're showing the placeholder):
```suggestion
      url: Types.string,
```

packages/components/src/overlayHeader/OverlayHeader.js
```diff
@@ -0,0 +1,87 @@
...
+  return (
+    <Header
+      leftContent={
+        <div className="m-lg-t-1">
+          <Logo theme={theme} onGoBack={null} />
+        </div>
+      }
+      rightContent={
+        <div className="m-lg-t-1">
+          {avatar && <Avatar url={avatar.url} profileType={profileType} />}
```
oliverviljamaa: 😍

PR #189, opened 2020-02-26
Add `data-testid` prop for Loader & add react-testing-library

dozortsev:
(better review commit by commit)

## 📎 Jira ticket

- [react-testing-library for components](https://transferwise.atlassian.net/browse/DS-239) 

## ❓ Context  

I have been writing tests in my app and found it annoying to not being able to access Loader component using testing-library [queries](https://testing-library.com/docs/dom-testing-library/cheatsheet#queries) (because it doesn't have label, text, role etc.), and I think the perfect solution for component like Loader should be to expose `data-testid` prop. 🙂

## ✅ Checklist


packages/components/package.json
```diff
@@ -98,6 +98,8 @@
     "jest-fetch-mock": "^2.1.2",
     "jest-resolve": "24.7.1",
     "jest-watch-typeahead": "0.4.0",
+    "@testing-library/jest-dom": "^4.2.4",
```
dozortsev: re. [@testing-library/jest-dom](https://testing-library.com/docs/ecosystem-jest-dom)

danlaush:
I think that's telling us our Loader should actually have some text in it. There are other accessibility things we'd need to do for it to be 100%, but the first step is putting some `visually-hidden` text in it that a screenreader can find.

packages/components/src/loader/Loader.js
```diff
@@ -4,13 +4,17 @@ import classnames from 'classnames';
...
+  const { small, size, classNames } = props;
   const style = className => classNames[className] || className;
 
   const legacySize = small ? Loader.Size.EXTRA_SMALL : size;
 
   return (
-    <div className={classnames(style('tw-loader'), style(`tw-loader--${legacySize}`))}>
+    <div
+      className={classnames(style('tw-loader'), style(`tw-loader--${legacySize}`))}
+      data-testid={props['data-testid']}
```
andreapiras: maybe here we could pass a data object and spread it?

PR #190, opened 2020-02-28
docs: mention about using react-testing-lib

dozortsev:
## ✅ Checklist


packages/components/README.md
```diff
@@ -69,24 +69,29 @@ To start the live-reloading environment for:
...
+
 ```
 yarn test
 ```
 
-- Runs testing suite.
+Please use [react-testing-library](https://github.com/testing-library/react-testing-library) for testing components.
+
+Historically we've used Enzyme so most of tests built using it's API but we encourage you to use react-testing-library when creating a new component
+also, feel free to update tests of existing components (e.g when you do some changes) and flat about this during code review.
```
helendurranttw: Looks good, thanks for adding this! Should "flat about" be "talk about" though?

dozortsev: ah, I actually wanted to use word "flag"

PR #196, opened 2020-03-04
refactor!: deprecated direct use of flex and box

andreapiras:
## ❓ Context  
Flex and Box has been deprecated and from now on usage should restricted only to Layout components

## 🚀 Changes 
Remove export and add deprecation message in docs

## ✅ Checklist


packages/components/src/index.js
```diff
@@ -47,5 +47,5 @@ export { default as OverlayHeader } from './overlayHeader';
 
 // These should be used only internally as block of layout pages. Direct use is discouraged and will easily introduce
 // inconsistencies.
-export { default as Box } from './box';
-export { default as Flex } from './flex';
+// export { default as Box } from './box';
```
helendurranttw: I think you can just remove them, instead of commenting them out. And we can remove the comment.

packages/docs/pages/components/FlexBox.mdx
```diff
@@ -1,16 +1,29 @@
...
 import '../../static/assets/box.docs.css';
 import '@transferwise/neptune-css/dist/css/flex.css';
 import code from '../../liveEditorCode/flexbox.code';
 
+<Alert
+  size={Alert.Size.SMALL}
+  arrow={Alert.ArrowPosition.TOP_LEFT}
+  type={Alert.Type.WARNING}
+>
+  @DEPRECATED <br/>
```
helendurranttw: No need for a deprecation notice, but you should remove the page from the docs so no one is tempted to use it. They won't be able to anyway :)

andreapiras: But what if someone wants to use it to create a layout? Would it not be good to leave the page so people knows that this component exists?

PR #198, opened 2020-03-09
feat: expose inputmode prop for InputWithDisplayFormat component

dozortsev:
## 🚀 Changes 

Expose [`inputmode`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode) attribute via `inputMode` prop in InputWithDisplayFormat component.

Mainly for `inputmode="numeric"` use cases for Mobile Web clients.

## 💬 Considerations 

[Not all browses support this feature](https://caniuse.com/#feat=input-inputmode) but Mobile Web browsers ones (like: iOS Safari, Android Browser, Chrome for Android and Samsung Internet etc.) do. 

## ✅ Checklist


sergonius:
> How are users of the component supposed to know what props are passed through and which are not?

Right now you kinda have to look at the code. IMO this looks a place to spread props and just add and document custom ones.

packages/components/src/inputWithDisplayFormat/InputWithDisplayFormat.spec.js
```diff
@@ -1,16 +1,37 @@
...
+      expect(input).not.toBe(null);
+      expect(input.value).toBe(value);
+    });
 
-  it('should render input with value prop', () => {
-    const component = shallow(<InputWithDisplayFormat {...props} />)
-      .find('WithDisplayFormat')
-      .renderProp('render')({ value: 'test' });
+    it('has inputMode prop when specified', () => {
+      const inputMode = 'numeric';
```
andreapiras: https://caniuse.com/#search=numeric

andreapiras: I don't think we should use this field for validation

dozortsev:
@helendurranttw consumers can check code like @sergonius mentions or they can also check Props table (e.g [Checkbox](https://transferwise.github.io/neptune-web/branch/support-inputmode/components/Checkbox)).

They can also play with component via Editor and inspect DOM 🙂

Screen Shot 2020-03-10 at 11 24 06 AM

PR #203, opened 2020-03-16
fix(tabs)!: update Tabs so the consumer doesn't need to filter disabled tabs and better handling of initial indexes

fergusjordantw:
## 📎 Jira ticket
Updating the API: https://transferwise.atlassian.net/browse/SEND-680
Better handling of disabled tabs: https://transferwise.atlassian.net/browse/SEND-695

## ❓ Context  
While using tabs to implement the local calculator we ran into a few issues with Tabs:

1. We were having to pass in a selected index that accounted for if tabs were disabled. This doesn't make a lot of sense since the consumer knows whats disabled/enabled, it makes more sense to give them all the control instead of letting the component handle half of it.
2. If we were on the second tab (with the first one disabled), and then the first one became enabled, there was a transition flash as the translate value updated.

## 🚀 Changes 

To address 1, I've updated the `MAX_INDEX` property to no longer use the `filteredTabs` length and to process all tabs when rendering (returning null if disabled). We then just account for any tabs before the specified index being disabled when we're setting any `translateX` values.

To address 2, I've added an `instant` argument to the `animatePanel` method which just makes sure we set `isAnimating` as `false` so `Spring` doesn't run any transitions. We pass instant as true in two instances:
1. When we mount, so that if we pass in `selected` as a non-0 index, it doesn't show us as transitioning to that tab
2. When props are updated and the amount of disabled tabs has changed, but the selected tab hasn't changed. This covers when other tabs are becoming enabled and we don't want to transition the updated translateX value. Before and after videos of the behaviour can be found in this JIRA ticket https://transferwise.atlassian.net/browse/SEND-695

## 💬 Considerations 
Breaking change as this slightly changes how consumers might pass in `selected` indexes to `Tabs`

## ✅ Checklist


packages/components/src/tabs/Tabs.js
```diff
@@ -60,18 +60,21 @@ class Tabs extends React.Component {
     const currentSelected = this.props.selected;
     const prevSelected = prevProps.selected;
     const currentSelectedTab = this.props.tabs[currentSelected];
+    const currentSelectedTabIsDiabled = currentSelectedTab && currentSelectedTab.disabled;
```
andreapiras: typo

packages/components/src/tabs/Tabs.js
```diff
@@ -156,33 +167,29 @@ class Tabs extends React.Component {
...
-    while (this.isTabDisabled(nextIndex)) {
-      nextIndex += 1;
-    }
-
-    this.setState({ translateLineX: `${(nextIndex + disabledIndexes) * 100}%` });
+    this.setState({ translateLineX: `${index * 100}%` });
   };
 
-  animatePanel = index => {
+  animatePanel = (index, instant = false) => {
```
helendurranttw: A comment explaining the intention of instant here would be helpful

packages/components/src/tabs/Tabs.js
```diff
@@ -105,18 +108,26 @@ class Tabs extends React.Component {
...
 
   isTabDisabled = index => {
-    const { tabs: allTabs } = this.props;
+    const { tabs } = this.props;
 
-    return allTabs[index] && allTabs[index].disabled;
+    return tabs[index] && tabs[index].disabled;
   };
 
   getTabToSelect = (selected, start, end) => {
```
helendurranttw: A general summary of the logic here would be useful.

PR #205, opened 2020-03-18
perf: use rollup & terser to reduce es bundle size

jackleslie:
## 📎 Jira ticket
None

## ❓ Context  
I was an [intern](https://github.com/jacklesliewise) last summer and I'm joining as a grad in August, I'm wanting to get to grips with Babel, Rollup etc so wanted to check out your setup. I noticed you use `babel-cli` to make the ES bundle and then copy the assets (e.g CSS) manually afterwards with `cpx`. I was wondering if you could use Rollup for this instead? 

I took a go at it and it's reduced the `es/no-polyfill` size from `313kb` to `232kb` and the `es/polyfill` size from `325kb` to `274kb`.

Really interested for input/feedback even if it's not needed, thanks in advance!

## 🚀 Changes 
- Removes the builds using `babel-cli` and removes the `copy-files` scripts
- Uses rollup to create the `es` and `es-nopolyfill` bundles
- Uses `rollup-plugin-terser` to minify ES bundle instead of `rollup-plugin-uglify` (uglify is for [es5](https://github.com/TrySound/rollup-plugin-uglify#rollup-plugin-uglify-) only)

## 💬 Considerations 
- Is there a reason to use `babel-cli` and `cpx` instead of Rollup for ES bundles? The ES bundle previously still has it's file structure intact and was largely readable, is this why?
- Is the code style okay regarding the `env` object in `rollup.config.js`?
- Will more plugins be needed for `svg` and `json` assets as they were being copied over with `cpx`?

## ✅ Checklist

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -28,10 +28,8 @@
     "build:compile-css": "gulp lessCompiler",
     "build:umd": "NODE_ENV=umd rollup -c",
     "build:umd-nopolyfill": "NODE_ENV=umd-nopolyfill rollup -c",
-    "build:es": "NODE_ENV=es babel src -d build/es/polyfill --ignore '**/*.spec.js','**/*.story.js'",
-    "build:es-nopolyfill": "NODE_ENV=es-nopolyfill babel src -d build/es/no-polyfill --ignore '**/*.spec.js','**/*.story.js'",
-    "build:copy-files": "cpx 'src/**/!(db)/*.{css,json,svg}' build/es/polyfill",
-    "build:copy-files2": "cpx 'src/**/!(db)/*.{css,json,svg}' build/es/no-polyfill",
+    "build:es": "NODE_ENV=es rollup -c",
+    "build:es-nopolyfill": "NODE_ENV=es-nopolyfill rollup -c",
     "test": "gulp lessCompiler && jest --env=jsdom --runInBand && yarn test:bundleSize",
     "test:watch": "jest --watch --env=jsdom",
     "test:bundleSize": "bundlesize",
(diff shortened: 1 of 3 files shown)


andreapiras:
Hi Jack!!!!
Thanks for your PR!!!
So the reason why I choose babel over Rollup for es6 are these:

1. As you said the output of babel is much more readable hence debuggable than the main.js that rollup produces.
To be honest multiple if you specify multiple entries you can get multiple output even with rollup.
One thing that I didn't manage to achieve easily with rollup was to have separate css files.
Right now Component.js imports its own Component.css.
Babel has shown limitations if for example we want to switch to css-modules.

2. The choice is also a bit historical because our docs was hot-reloading and babel in watch mode was faster than rollup -w.

So we are in the process of deciding if we'll go for css-module vs css-in-js vs css style, I think based on that we'll also decide if switching to Rollup. The ideal structure ideally will stay the same.

es
   |-button
   |        |-Button.js
   |        |-Button.css
   |-accordion
            |-Accordion.js
            |-Accordion.css

So your PR has nothing wrong but it doesn't follow the structure that we want to achieve. 

I can keep you in the loop if you want and we can collaborate if you want to contribute. 
Looking forward for you to be back in the office :)

jackleslie:
Thank you @andreapiras! 

Yeah I've played about with it some more and while you can get the same module structure with Rollup using [preserveModules](https://rollupjs.org/guide/en/#preservemodules) the same can't be done with respect to CSS files like you said.

And cheers, I'd love to know what you end up deciding to do! Let me know if you document your decisions or whatever anywhere and would be great to contribute 😃

Same to you, looking forward to it!

jackleslie:
Hey Andrea, I've been doing some Webpack stuff recently and came across [mini-css-extract-plugin](https://github.com/webpack-contrib/mini-css-extract-plugin) - would this work for the build structure you're wanting? It creates CSS files per JS file, obviously would require switching to Webpack though.

PR #206, opened 2020-03-19
Fix autofilling for input component

dozortsev:
## 📎 Jira ticket

https://transferwise.atlassian.net/browse/DS-325

## ❓ Context  

See context in Jira ticket ☝️

## 🚀 Changes 

I just removed Event from State and instead of that use Event object provided into event handler functions.

## ✅ Checklist


packages/components/src/withDisplayFormat/WithDisplayFormat.js
```diff
@@ -83,53 +82,43 @@ class WithDisplayFormat extends React.Component {
...
     }
     return null;
   }
 
-  setCursorPosition = select => {
-    const { triggerEvent } = this.state;
-    setTimeout(() => {
-      triggerEvent.target.setSelectionRange(select, select);
-      this.setState({ selectionStart: select, selectionEnd: select });
-    }, 0);
```
dozortsev: I just move this code out of function below

andreapiras: I can see you also remove the event fro onKeydown, that may be what is causing the bug

dozortsev: no, [here it is](https://github.com/transferwise/neptune-web/pull/206/files#diff-8d1c232346c292d8f468fa2c20da72f3R134)

packages/components/src/withDisplayFormat/WithDisplayFormat.js
```diff
@@ -179,34 +167,34 @@ class WithDisplayFormat extends React.Component {
...
-
     if (action === 'Backspace' || action === 'Delete') {
       unformattedValue = this.handleDelete(unformattedValue, action);
     }
 
     const newFormattedValue = formatWithPattern(unformattedValue, displayPattern);
     historyNavigator.add(unformattedValue);
 
-    this.handleCursorPositioning(action);
+    this.handleCursorPositioning(event, action);
```
andreapiras: Then the problem can be this, as the event may be different than the one onKeydown, hence why it was saved in state

dozortsev: yeah, I already figured this out, working on a solution

dozortsev: interesting that [here](https://github.com/transferwise/neptune-web/blob/master/packages/components/src/withDisplayFormat/WithDisplayFormat.js#L162) we store event as [`SyntheticEvent`](https://reactjs.org/docs/events.html) rather than saving it as Native one (`event.nativeEvent`) 🤔 @andreapiras do you know why?

andreapiras: to be honest I'm not sure about it :(

dozortsev:
will be done here https://github.com/transferwise/neptune-web/pull/209

PR #207, opened 2020-03-19
feat(tabs): Separate tab content during transition

fergusjordantw:
## 📎 Jira ticket
https://transferwise.atlassian.net/browse/SEND-741

## ❓ Context  
In send-flow we don't want to have the two Calculators touching each other when changing tabs. We initially had it set so there was 16px of padding inside the tabs, but that looks quite ugly when not swiping/switching tabs.

## 🚀 Changes 
Adds the ability to pass in a `panelTransitionSpacing` number which will apply a spacer of that amount around the tab you currently have selected when swiping. The transition between the old implementation and the new implementation is visually seamless, and is 100% opt-in with a default value of 0 not rendering the spacers at all. 

Additionally, this includes a fix where we weren't applying the drag resistance when swiping toward the a disabled tab when there are no enabled tabs beyond it (or vise-versa toward index 0). 

Videos of the described behaviour can be found in a comment in the JIRA ticket. 

## 💬 Considerations 
Have also refactored a bit of code in here

## ✅ Checklist


packages/components/src/tabs/Tabs.js
```diff
@@ -318,7 +326,14 @@ class Tabs extends React.Component {
...
 
   render() {
-    const { tabs, changeTabOnSwipe, name, selected, className } = this.props;
+    const {
+      tabs,
+      changeTabOnSwipe,
+      name,
+      selected,
+      className,
+      panelTransitionSpacing: spacer,
```
helendurranttw: I think we should constrict the API here a little, rather than letting people pass in an arbitrary number. Couple of options - we could a) have a couple of options (small, large), b) provide a single set spacing boolean (on or off) or c) always apply padding on transition, and not expose it at all.

What are your thoughts? @andreapiras @gusso would appreciate your input too.

packages/components/src/tabs/Tabs.js
```diff
@@ -385,10 +416,10 @@ class Tabs extends React.Component {
         >
           <Spring
             from={{
-              transform: `translateX(${translateFrom})`,
+              transform: `translateX(${translateFrom - spacer}px)`,
```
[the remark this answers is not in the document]
fergusjordantw: Will do tomorrow

packages/components/src/tabs/Tabs.js
fergusjordantw: Definitely not against a small/large option. I'd be wary of adding it as a single spacing/always applying on transition because I think a lot of the time it will depend on what's being rendered in the panel content. Some instances (like the calculator example - set to a 16px difference in the video) which don't have a background colour might need a lot less padding to provide a pleasing visual difference between the two panels. Others might require more based on what they're rendering.

Maybe a `xs/sm/md/lg` corresponding with `8/16/24/32` pixel spacers would provide enough flexibility. Would also remove the need for any negative value safeguarding per your other comment.

gusso: I'd vote for the constrained approach (`xs/sm/md/lg`) you suggested. We're working on a more fixed spacing scale right now, so while I'm not confident we're going to keep using this language to define spacing, t-shirt-sized is what we often use on our components.

helendurranttw: I'm fine with this :)

PR #212, opened 2020-03-21
Image component

andreapiras:
## 📎 Jira ticket
https://transferwise.atlassian.net/browse/DS-318

## ❓ Context  
This PR introduce Image component for dynamic form support.
It also introduces a custom hook that can be used to determine whether an element isintersectingg the viewport and in this case is used for lazy load

## 🚀 Changes 
Add Image component please find preview [here](https://transferwise.github.io/neptune-web/branch/DS-318-image/components/Image)

## 💬 Considerations 
There is a known bug on native Image event not firing on SSR.

https://github.com/facebook/react/issues/15446

We decided to release the component without that fix for now and wait for an official solution from react team, since for now we don’t have any particular use case that requires any event to fire on SSR. 
An example of implementation can be found in the commit history of this PR

## ✅ Checklist


packages/components/src/hooks/useLazy/UseLazy.js
```diff
@@ -0,0 +1,53 @@
...
+  };
+
+  useEffect(() => {
+    let observer;
+    let didCancel = false;
+    if (!isVisible) {
+      if (isValidRef()) {
+        // Check if window is define for SSaR.
+        if (window && window.IntersectionObserver && !didCancel) {
+          observer = new IntersectionObserver(handleOnIntersect, ObserverParams);
```
dozortsev: just out of curiosity, [it's not yet supported by IE 11](https://caniuse.com/#feat=intersectionobserver), do you want to use polyfill?

packages/components/src/imageWithLazy/ImageWithLazy.js
```diff
@@ -0,0 +1,43 @@
+import React, { useRef } from 'react';
+import Types from 'prop-types';
+import { useLazy } from '../hooks';
+import Image from '../image';
+
+const ImageWithLazy = ({ id, src, alt, placeholder, className, onLoad, onError }) => {
```
dozortsev: I would probably avoid having `Image` and `ImageWithLazy` components instead have with prop(s) for "configuring" lazy loading e.g `<Image lazyLoad={ true | false } .. /> `

packages/components/src/hooks/useLazy/UseLazy.js
andreapiras: We can use or polyfill or just don't lazy load images on Ie11. For not supported browser the option will be ---> no lazy loading so we are already providing a suitable solution

packages/components/src/imageWithLazy/ImageWithLazy.js
andreapiras: This can be up to debate. I personally like to extract the non native logic from the component if it's possible. I think Image should do few things and well. Also they may scenarios where people need only images. Why serving extra code in that case? It will be extra size and extra chances of bugs. Initially I had the logic embedded into the component but because of the reasons above I prefer to separate the two

packages/components/src/image/Image.js
```diff
@@ -0,0 +1,67 @@
...
+  const [isVisible] = useIsVisible(imageRef);
+  let imageSrc = src;
+  // Prevent load if image is not visible and lazyload is enabled.
+  if (withLazy) {
+    if (!isVisible) {
+      imageSrc = null;
+    }
+  }
+
+  const handleError = e => onError && onError(e);
```
helendurranttw, line 30: Given that we're defaulting the values to empty functions,  do we need the onError / onLoad checks?

packages/components/src/image/Image.js
```diff
@@ -0,0 +1,67 @@
...
+import { useIsVisible } from '../hooks';
+
+/**
+ * Image.
+ *
+ * @param {string} [alt] - string that contains the alt text.
+ * @param {string} [src] - string that contains the src text.
+ * @param {string} [id] - string that contains the id text.
+ * @param {function} [onLoad] - function that contains load callback.
+ * @param {function} [onError] - function that contains error callback.
```
helendurranttw, line 12: We should probably add a note saying this won't work in SSR

packages/components/src/image/Image.js
```diff
@@ -0,0 +1,67 @@
...
+  const handleLoad = () => onLoad && onLoad();
+
+  return (
+    <img
+      id={id}
+      alt={alt}
+      src={imageSrc}
+      className={className}
+      ref={imageRef}
+      onLoad={() => handleLoad()}
```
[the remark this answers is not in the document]
andreapiras: Yeah this is need I m just used to write it that way because of this

packages/components/src/hooks/useIsVisible/UseIsVisible.js
```diff
@@ -0,0 +1,51 @@
+import { useState, useEffect } from 'react';
+
+const ObserverParams = {
+  threshold: 0.1,
+};
+
+export const useIsVisible = elRef => {
+  const [isVisible, setIsVisible] = useState(() => false);
```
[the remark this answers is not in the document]
andreapiras: When using arrow functions the initial value will be assigned only on the initial render but I don't think is needed in this case.

packages/components/src/image/Image.js
```diff
@@ -0,0 +1,67 @@
...
+ * @usage `<Image alt="alt" src="src" id="id1" onLoad={()=>{}} onError={())=>{}} withLazy={true} />`
+ * */
+
+const Image = ({ id, src, alt, onLoad, onError, className, withLazy }) => {
+  const imageRef = useRef();
+  // This needs to be always included because hooks cannot be called inside conditions/loops.
+  const [isVisible] = useIsVisible(imageRef);
+  let imageSrc = src;
+  // Prevent load if image is not visible and lazyload is enabled.
+  if (withLazy) {
```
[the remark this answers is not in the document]
andreapiras: I would combine just from line 24 as I find it more readable "_"

packages/components/src/hooks/useIsVisible/UseIsVisible.js
```diff
@@ -0,0 +1,51 @@
...
+
+  const isValidRef = () => {
+    return elRef && elRef.current;
+  };
+
+  const handleOnIntersect = (entries, observer) => {
+    entries.forEach(entry => {
+      if (entry.isIntersecting) {
+        setIsVisible(true);
+        observer.unobserve(elRef.current);
```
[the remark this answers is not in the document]
andreapiras: Yeah I'm for changing the name. I don't think we should change the logic to check visibility as I don't see the value here. Which name would u go for?

packages/components/src/image/Image.js
andreapiras: Someone evil can still pass null or false?

packages/components/src/image/Image.js
```diff
@@ -0,0 +1,63 @@
+import React, { useRef } from 'react';
+import Types from 'prop-types';
+import { useHasIntersected } from '../hooks';
+
+/**
+ * Image component provides a wrapper for image tag. Optional lazy loading functionalities can also be enabled setting prop withLazy  to true.
```
helendurranttw, line 6: 🔥

packages/components/src/hooks/useHasIntersected/UseHasIntersected.js
```diff
@@ -0,0 +1,58 @@
...
+      setHasIntersected(true);
+    }
+    if (!didCancel) {
+      observer = new IntersectionObserver(handleOnIntersect, ObserverParams);
+      observer.observe(elRef.current);
+    }
+    // Cleanup function that runs for every render and on unmount and clean the previous render value.
+    return () => {
+      didCancel = true;
+      // on component cleanup, we remove the listener if it has been created.
```
twrdjpalmer, line 51: May I suggest removing these two comments? They seem to document `useEffect`'s API rather than anything specific to TW

packages/components/src/image/Image.js
```diff
@@ -0,0 +1,63 @@
...
+ * @param {function} [onLoad=()=>{}] - function that contains load callback. Doesn't fire on SSR https://github.com/facebook/react/issues/15446
+ * @param {function} [onError=()=>{}] - function that contains error callback. Doesn't fire on SSR https://github.com/facebook/react/issues/15446
+ * @param {boolean} [withLazy=false] - enables lazy loadind callback.
+ *
+ * @usage `<Image alt="alt" src="src" id="id1" onLoad={()=>{}} onError={())=>{}} withLazy={true} />`
+ * */
+
+const Image = ({ id, src, alt, onLoad, onError, className, withLazy }) => {
+  const imageRef = useRef();
+  // This needs to be always included because hooks cannot be called inside conditions/loops.
```
twrdjpalmer, line 21: Same with this comment—this is React's behaviour we're documenting

packages/components/src/image/Image.js
```diff
@@ -0,0 +1,63 @@
...
+  const [hasIntersected] = useHasIntersected(imageRef);
+  let imageSrc = src;
+
+  // Prevent load if image is not visible and lazyload is enabled.
```
twrdjpalmer, line 25: Seems pretty clear to me from the code 😅

packages/components/src/image/Image.js
```diff
@@ -0,0 +1,63 @@
...
+  if (withLazy && !hasIntersected) {
+    imageSrc = null;
+  }
+
+  return (
+    <img
+      id={id}
+      alt={alt}
```
twrdjpalmer, line 33: What's impact of having `Some text` rendered on the page?

packages/components/src/image/Image.js
```diff
@@ -0,0 +1,63 @@
...
+      src={imageSrc}
+      className={className}
+      ref={imageRef}
+      onLoad={onLoad}
+      onError={e => onError(e)}
```
twrdjpalmer, line 38:
```suggestion
      onError={onError}
```

packages/components/src/image/Image.spec.js
```diff
@@ -0,0 +1,63 @@
...
+      expect(onLoad).toHaveBeenCalledTimes(1);
+    });
+
+    it('calls onError when image error event fires', () => {
+      const onError = jest.fn();
+      const { getByRole } = render(<Image {...props} onError={onError} />);
+
+      fireEvent.error(getByRole('img'));
+      expect(onError).toHaveBeenCalledTimes(1);
+    });
```
twrdjpalmer, line 41: What confidence do these tests give? Seems like we're testing React's API

packages/components/src/image/Image.js
```diff
@@ -0,0 +1,63 @@
...
+ * @param {function} [onError=()=>{}] - function that contains error callback. Doesn't fire on SSR https://github.com/facebook/react/issues/15446
+ * @param {boolean} [withLazy=false] - enables lazy loadind callback.
+ *
+ * @usage `<Image alt="alt" src="src" id="id1" onLoad={()=>{}} onError={())=>{}} withLazy={true} />`
+ * */
+
+const Image = ({ id, src, alt, onLoad, onError, className, withLazy }) => {
+  const imageRef = useRef();
+  // This needs to be always included because hooks cannot be called inside conditions/loops.
+  const [hasIntersected] = useHasIntersected(imageRef);
```
twrdjpalmer, line 22: Should `withLazy` be passed into the hook, so we can early out of the hook's behaviour if its false?

twrdjpalmer: Am sure this has been discussed already, so please ignore if so...

I'm not sure about `withLazy`—is there a way to make it use the [`loading=""` API](https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading#Images) so it's as close to what we'll use in the future anyway?

andreapiras: Hey, so the attribute loading probably will replace this hook in the future but the coverage is not great at the moment https://caniuse.com/#feat=loading-lazy-attr

packages/components/src/image/Image.js
andreapiras: This is a good point. I have found some work around that initialize the src with a small transparent src 'data:image/gif;base64,R0lGODlhAQABAAAAACH5BAEKAAEALAAAAAABAAEAAAICTAEAOw=='
WDYT

packages/components/src/image/Image.js
andreapiras: Regarding the implementation is pretty much the same as the one suggested in this [blog post](https://developers.google.com/web/fundamentals/performance/lazy-loading-guidance/images-and-video)

packages/components/src/hooks/useIsVisible/UseIsVisible.js
```diff
@@ -0,0 +1,51 @@
...
+    });
+  };
+
+  useEffect(() => {
+    let observer;
+    let didCancel = false;
+    if (!isVisible) {
+      if (isValidRef()) {
+        // Check if window is define for SSaR.
+        if (window && window.IntersectionObserver && !didCancel) {
```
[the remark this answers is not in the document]
andreapiras: Just for reference this would always cause SSR to alway return false

packages/components/src/image/Image.spec.js
andreapiras: @twrdjpalmer Yeah to be honest they are not great. Initially the component was performing extra tasks on load and on error but now it doesn't anymore so probably they don't make sense. What would you test for ?

packages/components/src/hooks/useHasIntersected/UseHasIntersected.js
```diff
@@ -0,0 +1,54 @@
+import { useState, useEffect } from 'react';
```
sbrandwoo, line 1: Minor, but while is this file named `UseHas` rather than `useHas`?

packages/components/src/image/Image.js
```diff
@@ -0,0 +1,65 @@
+import React, { useRef } from 'react';
+import Types from 'prop-types';
+import { useHasIntersected } from '../hooks';
+
+/**
+ * Image component provides a wrapper for image tag. Optional lazy loading functionalities can also be enabled passing prop withLazy = true.
```
sbrandwoo, line 6: Why would you use this component with `withLazy` as `false`? You could just use a regular ``? In which case you could calls this `LazyLoadedImage` to be clearer what it's used for.

danlaush: I'd be in favour of the default being `withLazy={true}`. I like the idea of `Image` billing itself as "Like ``, but better". Support for lazy loading out of the box, and long-term would be cool to support `srcset`; similar to `gatsby-image` - give it an object with all the image options, it'll give you the markup to support them all.

packages/components/src/hooks/useHasIntersected/UseHasIntersected.js
andreapiras: all the rest of the repo follow this convention, should we change the case for custom hooks?

packages/components/src/image/Image.js
andreapiras: First iteration was supporting extra functionalities, like onError callbacks to set optional placeholder that we decide not to proceed with. To be honest your comments makes total sense. If someone is not looking for more than  tag functionality it should not use this component in first place.

packages/components/src/image/Image.spec.js
twrdjpalmer: I would just ignore these tests entirely—prop-types should catch issues with them.

Have you looked at [making them fail tests yet](https://github.com/transferwise/homepage/blob/5a225ffd53b6b4fb3244950f2f5afaf10502f0dc/config/setupTests.js#L8-L16)?

packages/components/src/image/Image.js
twrdjpalmer: Yeah, just thinking it would be best to use the same prop api as we will in future to maximise familiarity and make switching to that as easy as changing `Image` to `img` in our code, e.g.:

```jsx
<Image loading="lazy" />
// becomes 👇when coverage is good
<img loading="lazy" />
```

packages/components/src/hooks/useHasIntersected/UseHasIntersected.js
oliverviljamaa: I think the idea is to have the same casing as the exported member — which, for components, is `PascalCase`, but for hooks, `camelCase`.

packages/components/src/image/Image.spec.js
andreapiras: Oh that's cool thanks for sharing this

packages/components/config/jest/setupTests.js
```diff
@@ -4,4 +4,15 @@ const Adapter = require('enzyme-adapter-react-16');
 
 global.fetch = require('jest-fetch-mock');
 enzyme.configure({ adapter: new Adapter() });
+
+// https://github.com/esphen/jest-prop-type-error/blob/master/index.js
+// https://github.com/esphen/jest-prop-type-error/blob/master/index.js
```
twrdjpalmer, line 9:
```suggestion
```

packages/components/src/image/Image.js
```diff
@@ -0,0 +1,62 @@
...
+ * Once element intersects viewport image loads.
+ *
+ * @param {string} [alt=''] - string that contains the alt text.
+ * @param {string} [src=''] - string that contains the src text.
+ * @param {string} [id=''] - string that contains the id text.
+ * @param {function} [onLoad=()=>{}] - function that contains load callback. Doesn't fire on SSR https://github.com/facebook/react/issues/15446
+ * @param {function} [onError=()=>{}] - function that contains error callback. Doesn't fire on SSR https://github.com/facebook/react/issues/15446
+ * @param {boolean} [loading="eager"] - This attribute follows the loading API attribute. If set to eager image will load immediately just like a normal image tag otherwise it will lazy laod.
+ *
+ * @usage `<Image alt="alt" src="src" id="id1" onLoad={()=>{}} onError={())=>{}} loading={lazy} />`
```
twrdjpalmer, line 16:
```suggestion
 * @usage `<Image alt="alt" src="src" id="id1" onLoad={()=>{}} onError={())=>{}} loading="lazy" />`
```

packages/components/src/image/Image.js
andreapiras: We discuss this component within the team and some very good points are that this component enforce some the use of some attributes props like alt and src. We may introduce some changes in future (placeholder or ssr native events or particular class) and it would be good to roll all these changes and get them applied to all places where the component is used

packages/components/src/image/Image.spec.js
```diff
@@ -0,0 +1,49 @@
...
+const props = {
+  id: 'id',
+  src: 'https://www.a-valid-src.png/',
+  alt: 'test',
+  className: 'className',
+  loading: 'eager',
+};
+
+describe('Image', () => {
+  describe('when loading is not enabled', () => {
```
dozortsev, line 17: do you mean "when **lazy** loading is not enabled", right?

packages/components/src/image/Image.spec.js
```diff
@@ -0,0 +1,49 @@
...
+    it(`renders image`, () => {
+      const { getAllByRole } = render(<Image {...props} />);
+      const images = getAllByRole('img');
+      const image = images[0];
+      expect(images.length).toBe(1);
```
dozortsev, line 22: for such things better use [`.toHaveLength` function](https://jestjs.io/docs/en/expect#tohavelengthnumber)

and this `expect` should be one line above (before you fetch image from array of images)

packages/components/src/ssr.spec.js
```diff
@@ -65,6 +65,8 @@ describe('Server side rendering', () => {
       type: 'string',
       label: 'hello',
     },
+    src: 'src',
+    alt: 'alt',
```
dozortsev, line 69: could you please remind what this test check for?

packages/docs/liveEditorCode/image.code.js
```diff
@@ -0,0 +1,9 @@
+<Image
+  alt="test"
+  id="id1"
+  className="a-class-name"
+  src="//i.picsum.photos/id/1025/300/200.jpg"
+  withLazy
```
dozortsev, line 6: obsolete property

yarn.lock
```diff
@@ -1771,6 +1771,14 @@
     core-js "^2.6.5"
     regenerator-runtime "^0.13.2"
 
+"@babel/runtime-corejs3@^7.7.4":
```
dozortsev, line 1774: interesting, why all this updates here? I would try clean up this by fully removing `node_modules/`  & `yarn install`

packages/components/src/common/Option/Option.spec.js
```diff
@@ -65,7 +65,7 @@ describe('Option', () => {
   });
 
   it('renders as the tag that you pass it', () => {
-    component.setProps({ as: 'a' });
+    component.setProps({ as: 'a', href: 'https://example.com' });
```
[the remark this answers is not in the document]
andreapiras: because test was failing due to new props check

packages/components/src/ssr.spec.js
andreapiras: This test checks if the components renders on Server Side

yarn.lock
andreapiras: I think it's dependencies from ```'@testing-library/react-hooks'```
image

packages/components/config/jest/setupTests.js
```diff
@@ -4,4 +4,15 @@ const Adapter = require('enzyme-adapter-react-16');
 
 global.fetch = require('jest-fetch-mock');
 enzyme.configure({ adapter: new Adapter() });
+
+// https://github.com/esphen/jest-prop-type-error/blob/master/index.js
```
dozortsev, line 8: please remove this link

packages/components/src/image/Image.js
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * @param {string} [alt=""] - string that contains the alt text. Required.
+ * @param {string} [src=""] - string that contains the src text. Required.
+ * @param {string} [id=""] - string that contains the id text.
+ * @param {function} [onLoad=()=>{}] - function that contains load callback. Doesn't fire on SSR https://github.com/facebook/react/issues/15446
+ * @param {function} [onError=()=>{}] - function that contains error callback. Doesn't fire on SSR https://github.com/facebook/react/issues/15446
+ * @param {boolean} [loading="lazy"] - This attribute follows the loading API attribute. If set to eager image will load immediately just like a normal image tag otherwise it will lazy load.
+ *
+ * @usage `<Image alt="alt" src="src" id="id1" onLoad={()=>{}} onError={())=>{}} loading="lazy" />`
+ * */
```
dozortsev, line 17: why you decide to write API docs? I'm not against this but this is not something we do across our components now, in anycase for now I would delete it

packages/components/src/image/Image.js
dozortsev: > ...Why would you use this component with withLazy as false? You could just use a regular ?

@sbrandwoo no 🙂I totally agree with Andrea's last comment, 

having components even for things like `` , `<button>` `<a >` etc will help all of us easier rollout new future features and changes thus drive consistency, for instance atm Image component don't apply any of our CSS classes but it might will at some point (e.g we decide make corners of all our images slightly rounded 😄) then we don't want ask you and all other consumers to start use Image component over `img` tag it will be a more expensive to do that rather than just bump up version of components library to get keep your UI up to date according latest DS requirements

dozortsev: > ..I'd be in favour of the default being withLazy={true}.

Note: `withLazy` is already obsolete approach, Andrea introduced `loading={ 'lazy' | 'eager' }` prop instead

@danlaush wh

packages/docs/pages/components/Image.mdx
```diff
@@ -0,0 +1,43 @@
...
+Image component renders an img tag with some optional lazy load feature. "alt" and "src" props are required to enforce accessibility good practices.
+
+<LiveEditorBlock code={code} scope={{ Image }} />
+<GeneratePropsTable componentName="Image" />
+<>
+<Image
+  alt="test"
+  id="id1"
+  className={'m-t-5'}
+  src="//i.picsum.photos/id/1025/300/200.jpg"
```
dozortsev, line 14: please change demo image to something meaningful, maybe use some of our TW illustration etc.

packages/components/src/image/Image.js
sbrandwoo: > easier rollout new future features and changes thus drive consistency, for instance atm Image component don't apply any of our CSS classes but it might at some point (e.g we decide make corners of all of our images slightly rounded)

That would be a massive breaking change, so I would hope not

dozortsev: sorry, @sbrandwoo I didn't get what exactly you mean, in the example above which I describe consumers only will need bump up version of components lib that's it

PR #234, opened 2020-03-26
feat(backbutton): accept node label

sergonius:
Related to a previous PR, where `Stepper` and `FlowNavigation` got the same treatment. Missed this one.


sergonius:
> I think at this point it should be a fix instead?
Not from the library point of view I guess.

sergonius:
> I think at this point it should be a fix instead?

Not from the component's point of view I think.

mikknarusk:
> Not from the component's point of view I think.

My thinking is the original implementation of the PR had a bug (you can't use the new feature without these warnings), meaning this is a bugfix

PR #252, opened 2020-03-31
Select hot fix

andreapiras:
## ❓ Context  

A recent [change](https://github.com/transferwise/neptune-web/pull/235) has been introduced and caused some css issues around flags margins.
## 🚀 Changes 
Removes some redundant classes.
Fix nesting rules

## ✅ Checklist


packages/components/src/select/Select.js
```diff
@@ -502,7 +502,7 @@ export default class Select extends Component {
     const { open, shouldRenderWithPortal } = this.state;
 
     if (open && shouldRenderWithPortal) {
-      return createPortal(<div className="tw-select select-overlay" />, document.body);
+      return createPortal(<div className="select-overlay" />, document.body);
```
twrdjpalmer: Removing this means `select-overlay` won't get applied per the CSS. As it's a portal, it's rendered outside the context of its parent:

Screenshot 2020-03-31 at 11 27 04

packages/components/src/select/Select.less
```diff
@@ -32,16 +34,18 @@ html.ios-click {
...
-    position: fixed;
-    top: 0;
-    bottom: 0;
-    right: 0;
-    left: 0;
-    display: block !important;
-    background: rgba(0, 0, 0, 0.3);
-    z-index: 999;
-    animation: twSelectFadeIn 0.2s cubic-bezier(0.6, 0.2, 0.1, 1) both;
+  .tw-select {
```
twrdjpalmer: Per above, once we've readded the class, this should still be `.tw-select.select-overlay`

packages/components/src/select/Select.js
andreapiras: right right I'll revert

PR #276, opened 2020-04-06
docs: remove references to components src files from neptune web docs

andreapiras:
## 📎 Jira ticket
https://transferwise.atlassian.net/browse/DS-331

## ❓ Context  

Remove reference to src flex and box from layout.
Move flex and box inside docs and remove emotion from components and added to docs. 
## 🚀 Changes 
Remove emotion from components.
Install emotion into docs
Move flex and box from components to docs
Create ThreeColumnLayout for docs

## ✅ Checklist


packages/docs/components/layout/common/flexProperties.js
```diff
@@ -11,3 +11,19 @@ export const AlignItems = ['flex-start', 'flex-end', 'center', 'stretch', 'basel
...
+
+export const Size = {
+  EXTRA_SMALL: 'xs',
+  SMALL: 'sm',
+  MEDIUM: 'md',
+  LARGE: 'lg',
+  EXTRA_LARGE: 'xl',
+};
+
+export const Breakpoint = {
```
dozortsev, line 23: I think we better expose these common constants out from components package too

andreapiras: These so far were only used inside flex and box component. So for now I rather remove them from components and add them back if we need in components

PR #293, opened 2020-04-15
Fix Custom media max values

nabaroa:
## 📎 Jira ticket: https://transferwise.atlassian.net/browse/DS-369

The custom media max values, must be `min value -1 of the right bigger size` and right now they are the same as `min-value`

- @screen-xs-max:  (@screen-sm-min - 1);
- @screen-sm-max:  (@screen-md-min - 1);
- @screen-md-max:  (@screen-lg-min - 1);
- @screen-lg-max:  (@screen-xl-min - 1);

## ✅ Checklist


nabaroa:
Oh god! I messed it up with the merge. I close this one.  I'll remake this code change in another PR, taking into account that the documentation final approach is going to be different as well. Sorry for that 😬

PR #298, opened 2020-04-21
feat: markdown component

andreapiras:
## 📎 Jira ticket
https://transferwise.atlassian.net/browse/DS-304

## ❓ Context  
Add markdown component

## 🚀 Changes 
Introduce markdown component

## ✅ Checklist


packages/components/package.json
```diff
@@ -140,6 +142,7 @@
     "@transferwise/icons": "^1.0.0",
     "@transferwise/neptune-css": "^0.13.1",
     "classnames": "^2.2.6",
+    "commonmark": "^0.29.1",
```
dozortsev, line 145: @andreapiras will be this added in bundle of consumer app if consumer app doesn't use Markdown component? in another words will consumer app package builder (e.g webpack) claver enough to not add commonmark lib to the bundle incase app doesn't `import` Markdown component?

packages/components/src/markdown/Markdown.js
```diff
@@ -0,0 +1,41 @@
...
+const Markdown = ({ as: Element, value }) => {
+  if (!value) {
+    return null;
+  }
+
+  const createMarkup = () => {
+    const parsed = reader.parse(value);
+    const __html = writer.render(parsed);
+
+    return { __html };
```
helendurranttw, line 26: Let's not add an eslint exception just for this one case. You can also remove the lint error by using:

```
const html = writer.render(parsed);

return { __html: html };
```

packages/components/src/markdown/Markdown.js
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Markdown accepts a markdown string and returns the corresponding html markup.
+ *
+ * @param {object} [as] - html wrapper for the parsed markdown.
+ * @param {string} [value] - the string markdown to be converted.
+ *
+ * @usage `<Markdown value={'some markdown'} />`
+ * */
+
+const Markdown = ({ as: Element, value }) => {
```
helendurranttw, line 17: Is `as` a standard variable name in this case? I find it a bit confusing. Could we use something more descriptive like `outputElement` or similar?

packages/components/src/markdown/Markdown.spec.js
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import Markdown from './Markdown';
+import { render } from '@testing-library/react';
+
+const sublistMarkDown = `- sublist
+   - sublist`;
+
+const expetecSublistMarkDown = `<ul>
```
helendurranttw, line 8: Typo: there are a few "expetec" instead of "expect"s :)

packages/components/rollup.config.js
```diff
@@ -44,7 +46,9 @@ const plugins = [
     extract: pkg.style,
     extensions: ['.css'],
   }),
+  json(),
   uglify(),
+  visualizer(),
```
dozortsev, line 51: I would mention about this in README.md file

packages/components/src/ssr.spec.js
```diff
@@ -77,15 +77,7 @@ describe('Server side rendering', () => {
...
-    Box: {
-      size: {
-        xs: 1,
-        sm: 1,
-        md: 1,
-        lg: 1,
-        xl: 1,
-      },
-    },
+    Markdown: { value: '# Mardown string' },
```
dozortsev, line 80: you missed `k` in `Mardown` 🙂

packages/docs/pages/components/Markdown.mdx
```diff
@@ -0,0 +1,12 @@
+import { LiveEditorBlock, GeneratePropsTable } from '../../utils';
+import { Markdown } from '@transferwise/components';
+import code from '../../liveEditorCode/markdown.code';
+
+Renders markdown strings and returns html markup.
```
dozortsev, line 5: please mention here that Markdown "engine" is based on https://www.npmjs.com/package/commonmark

packages/components/src/markdown/Markdown.js
dozortsev: @helendurranttw 

> ..Is as a standard variable name in this case?

I would say so, I seen it a few other component libraries

dozortsev: we already use same naming in serval Neptune components too

andreapiras: Yes I'd leave it like that for consistency too

twrdjpalmer, line 17: Any reason you couldn't use `children` instead of `value`?

andreapiras: I feel that children should be used more when you want to pass different types to the container https://reactjs.org/docs/composition-vs-inheritance.html. In this case we want to pass only string. Also it's more in line with rest of the library api where we use more often value

dozortsev: I don't have strong opinion here but I guess writing like this would be cool

```jsx
<Markdown>
  ### Header

  Some list 
   - item 1
   - item 2
   - item 3
</Markdown>
```

dozortsev: I guess for such specific component `children` could be `Types.string.isRequired` 🤔

packages/components/rollup.config.js
andreapiras: I mentioned it in the about bundle page in docs where all the info about bundle are collected

packages/components/src/markdown/Markdown.js
twrdjpalmer: @andreapiras strings are valid HTML though, is the same as writing `<p>some content</p>`. If you imagine a native `<markdown />` element, I'd expect it to work per @dozortsev's example.

Would definitely force it to `Types.string` too 👍

andreapiras: Ok I m not strongly behind one or the other but since you guys are two I ll do it :)

PR #301, opened 2020-04-22
feat: Card component

ekwonye-richard:
## 📎 Jira ticket

https://transferwise.atlassian.net/browse/DS-376
## ❓ Context  
Introduces a new card component
Demo link: https://transferwise.github.io/neptune-web/branch/card-component/components/Card
https://www.figma.com/file/f1sB2JKuWY6W2GU61RYqxt/Card?node-id=249%3A10

## 🚀 Changes 
Adds a new card component and adds some customisations to `Options` component to make t compatible with the design

## ✅ Checklist


sbrandwoo:
Does this also satisfy the requirements of the Activity screen?

packages/components/src/card/Card.js
```diff
@@ -0,0 +1,79 @@
...
+import Chevron from '../chevron';
+import Option from '../common/Option';
+import './Card.css';
+
+const Card = ({ as: Element, isExpanded, title, details, children, onToggle, cardIcon }) => {
+  const isOpen = !!(isExpanded && children);
+
+  return (
+    <Element className={classNames('tw-card list-group-item p-a-0', { active: isOpen })}>
+      <div
```
andreapiras, line 17: Do we need this extra div? Can we not pass the onClick and className directly to Option?

ekwonye-richard: Moving the `.p-a-panel` to the Option doesn't have any visual difference is this use case. But I'll like to know of by practice `.p-a-panel` should be a parent of `.media` and not have both classes in the same element.

The components I made reference to has `.p-a-panel` as the parent.

https://github.com/transferwise/styleguide-components/blob/master/src/layout/cards/card.html#L7

ekwonye-richard:
> Does this also satisfy the requirements of the Activity screen?

No, I was told the activity screen will have a different design

packages/components/src/card/Card.js
andreapiras: Seems like the p-a-panel is seeting just the padding and I could not find any rule as p-a-panel { .media

andreapiras: but maybe we can leave is not a big deal

packages/components/src/card/Card.spec.js
```diff
@@ -0,0 +1,168 @@
...
+  });
+
+  it('renders as a tag passed to it', () => {
+    const { container } = renderCard({ as: 'li' });
+    expect(container.querySelector('.tw-card').tagName).toBe('LI');
+  });
+
+  it('renders the card title', () => {
+    const { container } = renderCard();
+    const titleWrapper = container.querySelector('h5');
```
andreapiras, line 33: Normally we should try to test in a way that resemble how users interact with your code. Maybe getByText here is more appropriate also according to this priority [rules](https://testing-library.com/docs/guide-which-query)

packages/components/src/card/Card.spec.js
```diff
@@ -0,0 +1,168 @@
...
+      cleanup();
+
+      const expandedProps = { children, isExpanded: true };
+      expect(contentPanel(expandedProps)).not.toBe(null);
+    });
+
+    it('renders children when expanded', () => {
+      const expandedProps = { children, isExpanded: true };
+      const { container } = renderCard(expandedProps);
+      const chidrenWrapper = container.querySelector('.media-body div');
```
andreapiras, line 134: typo

packages/docs/pages/components/Card.mdx
```diff
@@ -0,0 +1,54 @@
...
+<ul className="list-group panel-list-group list-group-slide-out list-group-inactive">
+  <Card
+    as="li"
+    title="Collapsed card"
+    details="details"
+    isExpanded={false}
+    onToggle={() => alert('toggled')}
+    cardIcon={<FastFlagIcon />}
+  >
+    <div />
```
andreapiras, line 22: probably forgot here

packages/components/src/card/Card.spec.js
```diff
@@ -0,0 +1,168 @@
...
+    const titleWrapper = container.querySelector('h5');
+    expect(titleWrapper.innerHTML).toEqual(defaultProps.title);
+  });
+
+  it('renders the card details', () => {
+    const { container } = renderCard();
+    const panelBody = container.querySelector('.tw-card__panel .media-body');
+    const { container: renderedCardDetails } = render(defaultProps.details);
+
+    expect(panelBody.innerHTML.includes(renderedCardDetails.innerHTML)).toBe(true);
```
andreapiras, line 42: not an error but have you tried to use ```toContainHTML```

packages/components/src/card/Card.spec.js
ekwonye-richard: Thanks `getByText` will be more appropriate.

packages/docs/pages/components/Card.mdx
ekwonye-richard: I just wanted to pass an empty div as children since it will not be expanded in the docs

packages/components/src/card/Card.spec.js
ekwonye-richard: Thanks!

andreapiras:
Well done !!! Very nice code.

packages/components/src/card/Card.spec.js
ekwonye-richard: Details can be a string.
`toContainHTML` expects an HTMLElement or an SVGElement

packages/components/src/card/Card.js
```diff
@@ -0,0 +1,79 @@
+/* eslint-disable jsx-a11y/click-events-have-key-events */
+/* eslint-disable jsx-a11y/interactive-supports-focus */
```
dozortsev, line 2: could you please better place it on exact line there this we violate these rules

packages/components/src/card/Card.js
```diff
@@ -0,0 +1,79 @@
...
+import React from 'react';
+import Types from 'prop-types';
+import requiredIf from 'react-required-if';
+import classNames from 'classnames';
+
+import Chevron from '../chevron';
+import Option from '../common/Option';
+import './Card.css';
+
+const Card = ({ as: Element, isExpanded, title, details, children, onToggle, cardIcon }) => {
```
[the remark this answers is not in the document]
dozortsev: I guess, I would also expose `id` prop for this component

packages/components/src/card/Card.spec.js
```diff
@@ -0,0 +1,167 @@
...
+  const children = <div />;
+  const renderCard = (props = {}) => {
+    return render(<Card {...defaultProps} {...props} />);
+  };
+
+  afterEach(cleanup);
+
+  it('renders as a div by default', () => {
+    const { container } = renderCard();
+    expect(container.querySelector('.tw-card').tagName).toBe('DIV');
```
dozortsev, line 23: when you access component via `.querySelector` or similar this is a sign that you component is not accessible, ideally you should use one of these queries https://testing-library.com/docs/dom-testing-library/cheatsheet#queries

dozortsev: if there is not why todo that then at at least I would expose `data-testid` prop

packages/components/src/common/propsValues/position.js
```diff
@@ -17,3 +17,10 @@ export const ArrowPosition = {
   BOTTOM: 'down-center',
   BOTTOM_RIGHT: 'down-right',
 };
+
+export const MediaPosition = {
+  RIGHT: 'media-right',
+  BOTTOM: 'media-bottom',
+  LEFT: 'media-left',
+  MIDDLE: 'media-middle',
```
dozortsev, line 25: can't we just reuse `Position.RIGHT`, etc.?

dozortsev: because this looks pretty custom not common one

packages/components/src/card/Card.js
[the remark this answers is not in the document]
ekwonye-richard: Thanks for your feedback.
I have some thoughts on the `expandable`. It comes like a question, meaning can this card be expanded? Which is what the `dismissible` prop actually does. if false, then the Alert can't be dismissed. 
That's not the purpose of our `isExpanded` (or any other name we use), rather it controls the expanded state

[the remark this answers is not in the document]
ekwonye-richard: Title
Components across the app uses the title approach. [Accordion](https://github.com/transferwise/neptune-web/blob/master/packages/components/src/accordion/AccordionItem/AccordionItem.js#L10)
[Option](https://github.com/transferwise/neptune-web/blob/master/packages/components/src/common/Option/Option.js#L33)
These are just 2 I came across

I assume we would want to keep things a bit uniform across the code base

packages/components/src/card/Card.less
```diff
@@ -0,0 +1,22 @@
+@import 'bootstrap/less/variables';
```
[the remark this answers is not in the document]
ekwonye-richard: Thanks, I'll update

packages/components/src/card/Card.spec.js
ekwonye-richard: I actually started using data-testId, but my concern or question will be how many of these will we have? We have a lot of elements we want to query, so it feels like we are going to expose a lot of props for us to use this route

packages/components/src/common/propsValues/position.js
ekwonye-richard: I didn't introduce these classes, `media-right` and the others exists in the bootstrap media component

packages/components/src/card/Card.js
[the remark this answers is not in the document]
dozortsev: > ..I assume we would want to keep things a bit uniform across the code base

yeah for sure, we definitely what that

yeah, but that could be mistake of previous code review, ok let it be `title` 🙂

packages/components/src/common/propsValues/position.js
dozortsev: yeah, but can we manipulate them using `Position` instead of `MediaPosition`

packages/components/src/card/Card.spec.js
dozortsev: > ..but my concern or question will be how many of these will we have?

only one on top element

> ..We have a lot of elements we want to query, so it feels like we are going to expose a lot of props for us to use this route

ideally those parts should be queried by other queries like `getByLabel*`, `getByPlaceholder*`, `getByRole*` etc

packages/components/src/common/Option/Option.js
```diff
@@ -15,28 +16,39 @@ const Option = ({
...
     href={href}
     onClick={onClick}
     htmlFor={htmlFor}
   >
     {media && (
       <div className="media-left">
-        <div className="circle circle-sm circle-inverse text-primary">{media}</div>
+        <div
+          className={classNames('circle circle-sm text-primary', {
+            'circle-inverse': inverseMediaCircle,
```
dozortsev, line 35: according to Figma mockup, shouldn't be this part alway grey (with `.circle-inverse`)?
Screen Shot 2020-04-23 at 4 04 48 PM

dozortsev: cc @anttio

ekwonye-richard: According to the figma design it shouldn't.

you are viewing this component on a white background, that's the conflict 😃

packages/components/src/common/propsValues/position.js
ekwonye-richard: Explain better please.
`Position.BOTTOM` for instance is a string with "bottom", so it's a bit unclear how we want to manipulate this in a clean way, let's not forget we are only adding this alignment if the props is passed.

It may start to look hacky trying to combine strings based on a condition.
Also Position doesn't have middle by default.

This isn't different from [ArrowPosition](https://github.com/transferwise/neptune-web/blob/b0434b1a696512ed707427b7f18f35eec5c6bdb8/packages/components/src/common/propsValues/position.js#L12)

packages/components/src/common/Option/Option.js
dozortsev: but then on white background it should be grey, even in collapsed state, no?

packages/components/src/card/Card.spec.js
ekwonye-richard: When we want to tests if certain classes are available?
Which seem to be necessary in this for this component

ekwonye-richard: I'll update the top element to use testId

packages/components/src/common/Option/Option.js
ekwonye-richard: The component is transparent when collapsed. I believe the component will ideally be used in a grey background (Like we have in the settings page)

packages/components/src/common/propsValues/position.js
ekwonye-richard: The figma design has been updated and the chevron icon is now at the top, so this logic has been removed.

packages/components/src/card/Card.js
```diff
@@ -0,0 +1,95 @@
...
+const Card = ({
+  as: Element,
+  isExpanded,
+  title,
+  details,
+  children,
+  onClick,
+  icon,
+  id,
+  dataTestId,
```
dozortsev, line 19: I would try to not change style of property name, so when consumer use this prop it use as native one

```jsx
<Card .. data-testid={..} />
```

packages/components/src/card/Card.js
```diff
@@ -0,0 +1,95 @@
...
+        className={classNames('p-a-panel tw-card__panel', {
+          'tw-card__panel--inactive': !children,
+        })}
+        role={children ? 'button' : null}
+        onClick={() => children && onClick(!isExpanded)}
+      >
+        <Option
+          media={icon}
+          title={title}
+          content={details}
```
dozortsev, line 40: let's call here this prop as `content` too

packages/components/src/card/Card.js
dozortsev: same as we have in [`Loader`](https://transferwise.github.io/neptune-web/components/Loader)

packages/components/src/card/Card.js
```diff
@@ -0,0 +1,95 @@
...
+  );
+};
+
+const hasChildren = ({ children }) => children;
+
+Card.propTypes = {
+  as: Types.elementType,
+  // eslint-disable-next-line
+  isExpanded: requiredIf(Types.bool, hasChildren),
+  title: Types.string.isRequired,
```
dozortsev, line 78: `title` should be `node` as people will pass there some translation components, e.g:

```jsx
title={<FormattedMessage id="message.translation.key" />}
```

packages/components/src/card/Card.js
dozortsev: I think you can archive this way (of course code can be polished in some places):

```jsx
const Card = ({ as: Element, isExpanded, title, details, icon, ...rest }) => {
  const { children } = rest;
  const isOpen = !!(isExpanded && children);
  const nativeAttrs = {
    id: rest.id,
    'data-testid': rest['data-testid'],
    onClick: () => children && rest.onClick(!isExpanded),
  };
  return (
    <Element
      className={classNames('tw-card list-group-item p-a-0', { active: isOpen }, rest.className)}
      {...nativeAttrs}
    >
```

dozortsev: note: [every* component should also expose `className` prop](https://github.com/transferwise/neptune-web/blob/master/packages/docs/pages/rfc/0001-always-pass-classname.md), above example cover it was well (passing `rest.className` in `className` attr)

packages/components/src/card/Card.less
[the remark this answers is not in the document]
dozortsev: you mention that it doesn't compile if you only add `_colors.less`,

I think the fix would be use CSS custom properties, so instead of `@brand-white` you have to write `var(--brand-white)`, could you please try?

packages/components/src/card/Card.js
ekwonye-richard: But that's not the content of the card.
Content makes sense in the context of the Option component because it has no children, but the card component also has children.

This is actually the details of the card, as seen in the figma designs

packages/components/src/card/Card.js
ekwonye-richard: The Option component already declares title as a string, This is a shared props.
Also @andreapiras advised that we should use a string here so as not to expose the component to a lot of variant styling.

packages/components/src/card/Card.js
ekwonye-richard: I'll expose the native `data-testid` prop, but I don't necessary have to create a nativeAttrs object.
onClick isn't called on the top element. I can just update the `data-testid` prop.

ekwonye-richard: Updated! I've added className too

packages/components/src/card/Card.less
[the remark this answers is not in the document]
ekwonye-richard: It works, thanks

packages/components/src/card/Card.js
ekwonye-richard: Are we suggesting we update the common `Option` component to now render nodes?

anttio:
Looking good and I'll leave here few general comments:

* border-top on the card has wrong colour, should be smoke-light, #e2e6e8
* there’s 1px extra margin-top, and 2px extra border-left. These are coming from `.list-group-item`.
* card content shouldn’t be inside `label` html element, I don't believe this is valid html. This is probably due the use of the `option` component as base.
* I would question the use of `h5` html element as it might not be semantically `h5` in the order of headers on the page.
* title font-size is wrong, should be 18px instead of 16px
* body content font-size is wrong, should be 16px instead of 14px

ekwonye-richard:
> Looking good and I'll leave here few general comments:
> 
> * border-top on the card has wrong colour, should be smoke-light, #e2e6e8
> * there’s 1px extra margin-top, and 2px extra border-left. These are coming from `.list-group-item`.
> * card content shouldn’t be inside `label` html element, I don't believe this is valid html. This is probably due the use of the `option` component as base.
> * I would question the use of `h5` html element as it might not be semantically `h5` in the order of headers on the page.
> * title font-size is wrong, should be 18px instead of 16px
> * body content font-size is wrong, should be 16px instead of 14px

Thanks for your feedback Antti, I'll address these.

danlaush:
@anttio how are the font sizes wrong if he hasn't set any in his styles? @ekwonye-richard I might check whether extra styles are coming from other elements (like the label) before adding styles specifically for those font sizes.

ekwonye-richard:
> @anttio how are the font sizes wrong if he hasn't set any in his styles? @ekwonye-richard I might check whether extra styles are coming from other elements (like the label) before adding styles specifically for those font sizes.
The font sizes and `h5` elements are actually set from the `Option` component.
I was advised to use this component for the Card collapsed panel. And it seems we'll require a lot of customisations for it to fit our design.
I also tried to avoid introducing new styles, because I believe the existing classes used in this component are based on the design system spec

cc @anttio

anttio:
@danlaush @ekwonye-richard Actually never mind about the font-sizes being wrong. I was comparing them to the AngularJS implementation of the card on the settings page which has bigger font-sizes. But the Option component in React is more recent and more used, so let's go with those styles.

ekwonye-richard:
> @danlaush @ekwonye-richard Actually never mind about the font-sizes being wrong. I was comparing them to the AngularJS implementation of the card on the settings page which has bigger font-sizes. But the Option component in React is more recent and more used, so let's go with those styles.

Okay @anttio 
I've pushed `updated` the component to use a `label` not a `div` and corrected the borders and margins

anttio:
LGTM

packages/components/src/card/Card.js
andreapiras: Actually @ekwonye-richard my bad, the Option accepts node so here it should be the same because of the translations. ```title: Types.node.isRequired,```

packages/components/src/card/Card.js
dozortsev: cool, thank you

> ..onClick isn't called on the top element.

just out of curiosity, why?

packages/components/src/card/Card.js
```diff
@@ -0,0 +1,99 @@
...
+      {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events,jsx-a11y/interactive-supports-focus, jsx-a11y/no-static-element-interactions */}
+      <div
+        className={classNames('p-a-panel tw-card__panel', {
+          'tw-card__panel--inactive': !children,
+        })}
+        role={children ? 'button' : null}
+        onClick={() => children && onClick(!isExpanded)}
+      >
+        <Option
+          as="div"
```
dozortsev, line 39: just a little thing, here you don't need pass string `"div"`, pass `as` prop there as it's `"div"` by default 🙂

ekwonye-richard: It's actually a Label by default
https://github.com/transferwise/neptune-web/blob/master/packages/components/src/common/Option/Option.js#L55

dozortsev: ah, it'a my mistake, ignore this comment

packages/components/src/card/Card.js
ekwonye-richard: @dozortsev If it's on the top element, it will be triggered when you click on any part of the expanded component since it's the parent. Meaning if any element of the children is clicked, the card will collapse immediately.
The onClick is on the top panel which controls the open and close state just like accordion component

PR #303, opened 2020-04-23
chore: update eslint and apply formatting changes

helendurranttw:
Update Eslint and associated dependencies

## 📎 Jira ticket

[DS-373](https://transferwise.atlassian.net/browse/DS-373)

## ❓ Context 

Part of a wider dependency update.

## 🚀 Changes

* Updates Eslint and associated deps
* Turns off one new rule that caused hundreds of errors
* Applied formatting changes

## ✅ Checklist


packages/docs/package.json
```diff
@@ -53,15 +53,15 @@
     "react-live": "^2.2.2",
     "refractor": "^3.0.0",
     "rehype": "^10.0.0",
-    "rehype-slug": "^3.0.0"
+    "rehype-slug": "^3.0.0",
+    "react": "16.8.6",
```
andreapiras, line 57: can this be in root ? and also version 16.13.1?

packages/docs/package.json
```diff
@@ -36,7 +36,7 @@
     "@transferwise/icons": "^1.5.0",
     "@transferwise/neptune-css": "^0.13.1",
     "@zeit/next-css": "^1.0.1",
-    "bootstrap": "transferwise/bootstrap",
+    "bootstrap": "github:transferwise/bootstrap#^v5.20.0",
```
andreapiras, line 39: Should this be in root?

packages/docs/.eslintrc
```diff
@@ -1,10 +1,14 @@
 {
```
andreapiras, line 1: We need to look into centralise this file too

packages/css/package.json
```diff
@@ -30,8 +30,7 @@
     "docs": "gulp lessWatcher",
     "copy:images": "cpx 'src/img/**' dist/img",
     "copy:fonts": "cpx 'src/fonts/*' dist/fonts",
-    "lint": "npm-run-all lint:*",
-    "lint:prettier": "prettier --find-config-path --write src/less/*.less",
+    "format": "prettier --write src/less/*.less",
```
andreapiras, line 33: should we have this also in root? I'd leave prettier in the naming even ```format:prettier```

helendurranttw: I'll rename it to prettier - I prefer to leave the : for when there are multiple tasks or sub tasks.

PR #319, opened 2020-04-30
feat: Summary component

sbrandwoo:
##  ❓ Context  

A Summary component, for displaying a list of requirements, some instructions, or perhaps some general information. Always has an icon and title, can also have a body text and a help icon with some popover text.

Usually will be used as a group:

image

## 🚀 Changes 

Adds the Summary component.

## 💬 Considerations 

The code is rebased from a while ago - let me know if any of the naming conventions etc are out of date.

## ✅ Checklist


packages/components/src/summary/Summary.story.js
```diff
@@ -0,0 +1,51 @@
...
+              you’re verified.
+            </>
+          ),
+        }}
+      />
+      <Summary icon={<IdIcon size="md" />} title="You uploaded you identity documents" />
+      <Summary
+        icon={<HomeIcon size="md" />}
+        title="You entered your card delivery address"
+        body={
```
oliverviljamaa, line 35: Could this be `content` as well to align with `help` props?

packages/components/src/summary/Summary.less
```diff
@@ -0,0 +1,36 @@
+@import 'bootstrap/less/variables.less';
```
oliverviljamaa, line 1: Can we get those variables from Neptune somehow? 🤔

packages/components/src/summary/Summary.less
```diff
@@ -0,0 +1,36 @@
...
+
+.summary {
+  &__main {
+    display: flex;
+    align-items: center;
+  }
+
+  &__icon {
+    width: 24px;
```
oliverviljamaa, line 10: Do we have a variable for these as well?

packages/components/src/summary/Summary.less
```diff
@@ -0,0 +1,36 @@
...
+    height: 24px;
+  }
+
+  &__help {
+    height: 16px;
+
+    .tw-icon {
+      color: @brand-light-blue;
```
oliverviljamaa, line 18: And this?

packages/components/src/summary/Summary.story.js
sbrandwoo: :+1:

packages/components/src/summary/Summary.less
```diff
@@ -0,0 +1,36 @@
...
+    }
+  }
+
+  &__body {
+    margin-left: @spacer * 5;
+    font-size: @font-size-small;
+  }
+
+  & + .summary {
+    margin-top: @spacer * 3;
```
sbrandwoo, line 34: This should be `* 4` (32px)

packages/components/src/summary/Summary.js
```diff
@@ -0,0 +1,46 @@
...
+      </div>
+      {help && (
+        <div className="summary__help m-l-2">
+          <Popover preferredPlacement="bottom-left" title={help.title} content={help.content}>
+            <HelpCircleIcon size="sm" />
+          </Popover>
+        </div>
+      )}
+    </div>
+    {body && <div className="summary__body m-t-1">{body}</div>}
```
andreapiras, line 25: Is there any reason why you didn't use children here?

packages/components/src/summary/Summary.less
andreapiras: Hey Stephen you can the variables from here https://github.com/transferwise/neptune-web/blob/master/packages/css/src/less/variables/_colors.less

packages/components/src/summary/Summary.story.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
```
andreapiras, line 1: Nice one! I think it would cool to add some knobs to make it more interactive

packages/components/src/summary/Summary.less
andreapiras: I'm not very familiar with this syntax. Why didn't you use m-t-3 in the main class?

packages/components/src/summary/Summary.less
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  &__help {
+    height: 16px;
+
+    .tw-icon {
+      color: @brand-light-blue;
+    }
+
+    [data-toggle='popover']:not(.btn) {
```
andreapiras, line 21: Optional: This can also be ``` span[role='button'] ``` which I find more readable

packages/components/src/summary/Summary.less
andreapiras: https://github.com/transferwise/neptune-web/blob/master/packages/css/src/less/variables/_colors.less

packages/components/src/summary/Summary.less
```diff
@@ -0,0 +1,36 @@
...
+      // Bootstrap popover has !important dotted border bottom
+      // This doesn't make sense on a help icon
+      border-bottom: none !important;
+    }
+  }
+
+  &__body {
+    margin-left: @spacer * 5;
```
andreapiras, line 29: Spacer can be found here :  https://github.com/transferwise/neptune-web/blob/master/packages/css/src/less/variables/_spacing.less

andreapiras: Also this can be replaced by using m-l-5 on the body :)

packages/components/src/summary/Summary.story.js
rerdos: Wouldn't this be better as children?w

packages/components/src/summary/Summary.less
```diff
@@ -0,0 +1,36 @@
...
+    font-size: @font-size-small;
```
andreapiras, line 30: This can be found here https://github.com/transferwise/neptune-web/blob/master/packages/css/src/less/variables/_typography.less

packages/components/src/summary/Summary.story.js
sbrandwoo: There's 4 things that might be included in the component, two of them are optional. `body` / `content` is an optional one and might be reasonably uncommon. I don't agree that we should use `children` for `body` and leave the the rest as props, it feels inconsistent.

packages/components/src/summary/Summary.less
sbrandwoo: This is the gap between two `.summary` components next to each other.

packages/components/src/summary/Summary.js
```diff
@@ -0,0 +1,46 @@
...
+import React from 'react';
+import Types from 'prop-types';
+import HelpCircleIcon from '@transferwise/icons/react/help-circle';
+
+import Popover from '../popover';
+
+import './Summary.css';
+
+const Summary = ({ className, icon, title, help, body }) => (
+  <div className={classNames('summary', className)}>
```
andreapiras, line 11: I imagine this could be at some point be used inside a list? If that's the case maybe it's worth it to render it as `<Element` with `as` prop default to div 
You can find an example here
https://github.com/transferwise/neptune-web/blob/master/packages/components/src/common/Option/Option.js

packages/components/src/summary/Summary.less
sbrandwoo: @andreapiras Should I add a CSS variables file for icons? `_icons.less`?

packages/components/src/summary/Summary.story.js
sbrandwoo: I'm not sure what knobs would be useful. The docs include three examples that cover all of the options.

packages/components/src/summary/Summary.less
sbrandwoo: This exactly matches the CSS that is is overriding. `span[role='button']` wouldn't be specific enough.

andreapiras: Hey I did try and it worked, but as I said it's optional :)

packages/components/src/summary/Summary.less
dozortsev: cc @nabaroa @anttio

packages/components/src/summary/Summary.story.js
andreapiras: Yeah I saw that and that's great but if someone wants to test out a different text to have an idea of the visual impact won't be able to do it without modifying the code

packages/docs/pages/components/Summary.mdx
```diff
@@ -0,0 +1,17 @@
+
+import { Summary } from '@transferwise/components';
+import MoneyIcon from '@transferwise/icons/react/money';
+import IdIcon from '@transferwise/icons/react/id';
+import HomeIcon from '@transferwise/icons/react/home';
+
+import { LiveEditorBlock, GeneratePropsTable } from '../../utils';
+import code from '../../liveEditorCode/summary.code';
+
+<div>
```
dozortsev, line 10: could you please add here a bit of documentation, e.g the one which is PR seems good:

> A Summary component, for displaying a list of requirements, some instructions, or perhaps some general information. Always has an icon and title, can also have a body text and a help icon with some popover text.

packages/components/src/summary/Summary.js
```diff
@@ -0,0 +1,48 @@
...
+import React from 'react';
+import Types from 'prop-types';
+import HelpCircleIcon from '@transferwise/icons/react/help-circle';
+
+import Popover from '../popover';
+
+import './Summary.css';
+
+const Summary = ({ as: Element, className, icon, title, help, content }) => (
+  <Element className={classNames('summary', className)}>
```
andreapiras, line 11: Nice! I tried out on docs and spotted this 
image
maybe we can add some css to prevent the bullet to show

packages/components/src/summary/Summary.less
dozortsev: this CSS class adds fix the alignment the icon

Screen Shot 2020-05-01 at 1 04 23 PM

here how it looks like if you remove it, seem icon and text a aligned on same line rather then centrally (vertically)

Screen Shot 2020-05-01 at 2 17 28 PM

packages/components/src/summary/Summary.js
dozortsev: please mention how did you reproduce this, asking because I don't see this on docs

sbrandwoo: The dot is part of a parent `ul`, we can't remove it from the component.

packages/components/src/summary/Summary.less
anttio: Do we need these hardcoded values here? Could you just set `vertical-align: top` for the `.tw-icon` when it is inside `.summary`?

packages/components/src/summary/Summary.js
```diff
@@ -0,0 +1,48 @@
...
+      </div>
+      {help && (
+        <div className="summary__help m-l-2">
+          <Popover preferredPlacement="bottom-left" title={help.title} content={help.content}>
+            <HelpCircleIcon size="sm" />
+          </Popover>
+        </div>
+      )}
+    </div>
+    {content && <div className="summary__content m-t-1 m-l-5">{content}</div>}
```
twrdjpalmer, line 25: Any reason to use `content` rather than `children`?

packages/components/src/summary/Summary.less
sbrandwoo: The reason for the width and height is to make sure the component aligns properly when the icon is not what we expect. Without it, you'd get a result like this:
image

packages/components/src/summary/Summary.js
sbrandwoo: We pass four items into the component, two of them are required and two are optional. The `content` is one of the optional ones, and didn't feature at all in the first version of this component. I think it would inconsistent to use children for `content` and to have the other ones as props.

Is there a benefit to using children?

dozortsev: it should be named `content`, in this case it's very similar to [Option](https://github.com/transferwise/neptune-web/blob/196dd8e623213911ccf2f5044a36acc733fc5037/packages/components/src/common/Option/Option.js#L52) style e.g [RadioOption](https://transferwise.github.io/neptune-web/branch/summary-component/components/CheckboxOption)

packages/components/src/summary/Summary.less
anttio: Makes sense. Let's have the fixed dimensions here for now. We'll might introduce these as variables when revisiting the icons. But let's not have this as a blocker now.

gusso:
Was there a decision not to use the badge on the bottom right of the icon?

gusso:
The container of the icon should be 44x44 to support displaying an image also and maintaining text alignment if they're displayed on the same screen, like on the screenshot below. I don't think the image support needs to be here necessarily, but it would be good if it was already prepared for it.

Lets get setup card

packages/components/src/summary/Summary.js
gusso: This one should have a `4px` margin top, that's not an option on bootstrap, right? It will be available on the new core, can we make the change manually on CSS until then?

packages/components/src/summary/Summary.spec.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { shallow } from 'enzyme';
```
dozortsev, line 2: please use react-testing-library, https://github.com/transferwise/neptune-web/blob/master/packages/components/CONTRIBUTING.md#testing

sbrandwoo:
@dozortsev @andreapiras updated with react testing library

PR #336, opened 2020-05-01
feat: SelectWithAlternatives - gives additional values that can be searched on with a Select component

pkaszubinski-transferwise:
## 📎 Jira ticket
https://transferwise.atlassian.net/browse/USD-5940

## ❓ Context  
While reviewing NorthAm plastic onboarding flow, it was discovered that the state dropdown used in profile address and the one used in delivery address do not have the same search functionality. One would not let you to search on the state abbreviation. As dropdowns of this variety are somewhat common (and abbreviations for values is not a US specific use-case, nor address specific), a common select with a common search functionality should be added in order to encourage reuse and avoid reimplementations.

## 🚀 Changes 
- Adds SelectWithAlternatives, which behaves similarly to Select but gives a common search functionality across the options, including value, label, and an optional array of alternatives, which could be things like abbreviations or nicknames.
- Adds SelectUsaStates - Wraps SelectWithAlternatives to have a common component with a common data set with a standardized order.

## ✅ Checklist


packages/components/src/selectUsaStates/data/usStates.js
```diff
@@ -0,0 +1,54 @@
+const usStates = [
+  { value: 'AL', label: 'Alabama' },
```
sbrandwoo, line 2: Will these need to be translated?

packages/components/src/selectUsaStates/SelectUsaStates.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
```
sbrandwoo, line 1: I think this component might be too specific for Neptune.

StevePole:
The way this works in the angular select box, is that you supply a 'searchable' string within the select options:
```
<tw-select
  options="[{
    value: "EUR",
    label: "Euro",
    searchable: "France, Germany, Spain ... "
  },{
    value: "GBP",
    label: "Pound Sterling",
    searchable: "England, Wales, Scotland, ..."
  }]" />
```
You then don't need a domain specific component, the consumer just adds the searchable content to the options, the select box checks against this when searching.

packages/components/src/selectUsaStates/SelectUsaStates.js
pkaszubinski-transferwise: I added it here based on a comment in a thread made by @oliverviljamaa on if it this was an appropriate location for this component or not.
https://transferwise.slack.com/archives/C2SGDKP4Y/p1580854536098900

I don't disagree that it may be too specific, but I can also understand the opposite viewpoint as well.

PR #337, opened 2020-05-04
feat: allows searchable alternatives to be given to options

pkaszubinski-transferwise:
## 📎 Jira ticket
https://transferwise.atlassian.net/browse/USD-5940

## ❓ Context  
Feedback from https://github.com/transferwise/neptune-web/pull/336

## 🚀 Changes 
Extends default search functionality of Select to support passing in alternatives to options that can be searched on out of the box. Usage of this component will no longer require a custom search function to be reimplemented each time this type of functionality is desired.

## 💬 Considerations 
Because defaultFilterFunction uses includesString, meaning any alternative value only needs to contain the search query, representing alternatives as a single string rather than an array would have achieved the same functionality. By making it an array, it allows for custom search functions to have easier control over the individual alternatives. An example would be if a custom search function was implemented to only filter based on if the search value started with the search query, the array makes this more straightforward to implement.

## ✅ Checklist


packages/components/src/select/Select.js
```diff
@@ -52,7 +52,9 @@ const includesString = (str1, str2) => str1.toLowerCase().indexOf(str2.toLowerCa
 
 const defaultFilterFunction = (option, keyword) =>
   (option.label && includesString(option.label, keyword)) ||
-  (option.currency && includesString(option.currency, keyword));
+  (option.currency && includesString(option.currency, keyword)) ||
+  (option.alternatives &&
+    option.alternatives.findIndex((alternative) => includesString(alternative, keyword)) > -1);
```
andreapiras, line 57: Hey, the support for [findIndex](https://caniuse.com/#search=findIndex) doesn't cover our browser support, can you please use an alternative?

packages/components/src/select/Select.js
```diff
@@ -52,7 +52,9 @@ const includesString = (str1, str2) => str1.toLowerCase().indexOf(str2.toLowerCa
 
 const defaultFilterFunction = (option, keyword) =>
   (option.label && includesString(option.label, keyword)) ||
-  (option.currency && includesString(option.currency, keyword));
+  (option.currency && includesString(option.currency, keyword)) ||
+  (option.alternatives &&
```
helendurranttw, line 56: Maybe we should use the name `searchable` instead of `alternatives`, to match the existing property in the Angular version. Theirs is a string, but I think having an array here is fine.

packages/components/src/select/Select.spec.js
```diff
@@ -583,6 +583,29 @@ describe('Select', () => {
       document.activeElement.blur();
     });
 
+    it('should filter the options with the default filter function in their alternatives attribute', async () => {
```
helendurranttw, line 586: Perhaps "should include alternatives in option search if present" might be clearer.

packages/components/src/select/Select.js
andreapiras: Also I saw that @StevePole  pointed out to the angular implementation. 
  ```
<tw-select
  options="[{
    value: "EUR",
    label: "Euro",
    searchable: "France, Germany, Spain ... "
  },{
    value: "GBP",
    label: "Pound Sterling",
    searchable: "England, Wales, Scotland, ..."
  }]" />
```
Is there any reason why we are not replicating this?

File: packages/components/src/select/Select.story.js [hunk 4]
```diff
@@ -55,6 +55,12 @@ export const basic = () => {
         { value: 6, label: 'Euro', currency: 'eur' },
         { separator: true },
         { value: 7, label: 'Something else' },
+        { header: 'Alternate Search Options' },
```
andreapiras, line 58: Can you please also document the changes in the docs folder ?. https://github.com/transferwise/neptune-web/blob/master/packages/docs/pages/components/Select.mdx

This is what generates the interactive demo https://transferwise.github.io/neptune-web/branch/USD-5940-add-alts-to-base-select/components/Select

packages/components/src/select/Select.js
pkaszubinski-transferwise: I chose to use the array because I felt that each searchable term should be considered on an individual basis for filtering rather than as one single item. As it stands now, the commas used to separate the searchables are also included when filtering. It's a detail that is generally hidden from users and could be mitigated by not using commas at all, but the general point is that delimiters should not be part of a search option. If this sort of functionality was truly desired, an array of a single large element could be used instead.

I also feel that when writing a custom search function, it's highly likely that the developer would want to make their comparisons based on the individual elements. I don't think that we should force a developer to first to split or tokenize the searchable in order to do so.

Thank you for pointing out the issue with findIndex. I will look into an alternative.

packages/components/src/select/Select.js
```diff
@@ -50,9 +50,19 @@ const defer = (fn) => setTimeout(fn, 0);
 
 const includesString = (str1, str2) => str1.toLowerCase().indexOf(str2.toLowerCase()) > -1;
 
+const arrayIncludesString = (arrayToSearch, keyword) => {
```
helendurranttw, line 53: Perhaps we could simplify this using some:

```
const arrayIncludesString = (arrayToSearch, keyword) => arrayToSearch.some((str) => includesString(str, keyword));
```

PR #338, opened 2020-05-05
refactor!: move formControl and Field to dynamic-flows

andreapiras:
## 📎 Jira ticket
https://transferwise.atlassian.net/browse/DS-409

## ❓ Context  
We decided to group all dynamic-flows and forms in one separate repo. As part of this grouping makes sense that FormControl lives in dynamic-flows.

## 🚀 Changes 
1) Moved FormControl to dynamic-flows

## ✅ Checklist


packages/components/src/field/Field.js
```diff
@@ -1,326 +0,0 @@
-import React, { Component } from 'react';
```
andreapiras, line 1: Field has been replaced by JsonSchemaForm that lives inside dynamic-flows

packages/components/src/index.js
```diff
@@ -21,7 +21,6 @@ export { default as Upload } from './upload';
 export { default as DefinitionList } from './definitionList';
 export { default as DynamicFieldDefinitionList } from './dynamicFieldDefinitionList';
 export { default as Button } from './button';
-export { default as FormControl } from './formControl';
```
andreapiras, line 24: FormControl has been moved to dynamic-flows package

sbrandwoo:
I don't understand this change. What if we want to use a simple `Field`, but don't want to use dynamic flows? Or why would I want to use `JsonSchemaForm` instead, when it's a much more complicated component?

andreapiras:
> I don't understand this change. What if we want to use a simple `Field`, but don't want to use dynamic flows? Or why would I want to use `JsonSchemaForm` instead, when it's a much more complicated component?

Hey @sbrandwoo , at the moment I ran a quick search and no one is suing Field. where would you use Field if not inside a form? I feel it would be risky to let people basically create Fields in two different ways as then we'll have to maintain both, and they output the same UI. I feel that maybe if we need a "reduced" version of JsonSchemaForm we can reimplement Field on top of it, but only if a use case is provided. Which use case were you thinking about?

andreapiras:
> I don't understand this change. What if we want to use a simple `Field`, but don't want to use dynamic flows? Or why would I want to use `JsonSchemaForm` instead, when it's a much more complicated component?

Hey @sbrandwoo I just spoke with @helendurranttw and it's probably better to either introduce a deprecation process and/or refactor this component to share the logic with JsonSchemaForm

StevePole:
@sbrandwoo the thing with this Field component is it relies on our JSON schema spec to function, so I think it belongs with that package.

In another chat with @andreapiras I was highlighting that you probably wouldn't have a JSON schema based interface for a general use case field component.  I can see a case for a general Field component that is designed to be easily consumed without dynamic forms (perhaps with the ability to pass in the form component you want to use), and a JSON schema variant within the dynamic flows package.  The latter would possibly consumer the former, just exposing the interface that makes most sense for that domain.

sbrandwoo:
@StevePole / @andreapiras 
My understanding (which might be wrong) of how Neptune would be structured is as a stack of small building blocks, moving upwards to larger components and finally our own applications. For example:

- `<input />`, the raw element defined by your browser
- `class="form-control"`, styles for inputs defined by Bootstrap / Neptune CSS
- `<Field />`, a convenience wrapper of the above, defined by Neptune Components 
- `<JsonSchemaForm />`, a way of automating Fields, defined by Neptune Components (or maybe DF)
- Dynamic flows, super crazy fancy stuff, defined by Neptune Dynamic Flows

There's a path from CSS to Components to Dynamic Flows.

Every step builds on the ones above it, and the application pulls in only the level it needs. So in my case, I just want to show a nice input field to the user, so I believe there should be a convenient component for doing that - I thought `Field` was this component.

If my application has to pull in dynamic flows in order to work, I'd consider that a bug in Neptune - that would be an unnecessary increase in bundle size.

If I need to write JSON to use an input box in my app, I'd also consider this to be a bug in Neptune - it's unnecessary complexity - which likely will lead to bugs and headaches around breaking changes in the future.



If this is what the plan is, then cool. Removing `Field` would be a breaking change though, with no benefit to the end consumers of Neptune.

StevePole:
Yeah, I think that's a fair summary.  One little thing I'd add is that the dynamic flow package is still in the same repo, and you should still be able to engage with it at the level you want.  So if you wanted only the json schema Field, that would be fine.  But as you say, a Field component for the general use case shouldn't require you to use a relatively complex JSON config.

packages/docs/pages/components/Field.mdx
```diff
@@ -1,5 +1,5 @@
 import { LiveEditorBlock, GeneratePropsTable } from '../../utils';
-import { Field, Button } from '@transferwise/components';
+import { Field, Button } from '@transferwise/dynamic-flows';
```
helendurranttw, line 2: Button should still be coming from components.

helendurranttw:
Hey @sbrandwoo,

To be clear, we definitely want to provide a Field component that provides validation on top of inputs, as part of Neptune Web. 
 
We just want to deprecate our existing Field, and move it to `dynamic-flows` (as its implementation is tied to FormControl). So there will be a period where you have to import Field from `dynamic-flows`, until we reimplement Field in `components`. At that point, there will be an upgrade step, probably needing to change what you pass in to the Field, but then you'll get the same (or improved!) behaviour of form control validation that the Field currently provides.

Hope that makes sense. cc/ @StevePole @andreapiras

sbrandwoo:
I think in the short term we can use the CSS versions directly.

PR #341, opened 2020-05-06
feat(components): add click events to Accordion

tiborsaas-tw:
## 📎 Jira ticket
https://transferwise.atlassian.net/browse/DS-416

## ❓ Context  

We missed an open event when we wanted to track with mixpanel if the customer opened an item.

## 🚀 Changes 
Adds `onClick` event to the `Accordion` component.

## ✅ Checklist


packages/components/src/accordion/Accordion.js
```diff
@@ -25,8 +29,21 @@ class Accordion extends PureComponent {
...
+        this.props[cb](value);
+      }
+    };
+
     this.setState((prevState) => {
-      return { indexOpen: prevState.indexOpen === index ? -1 : index };
+      const newIndex = prevState.indexOpen === index ? -1 : index;
+
+      if (newIndex === -1) {
+        callbackProxy('onClose', index);
```
sbrandwoo, line 42: I'd expect to get an `onClose` when any element closed, not just when they are all closed?

packages/components/src/accordion/Accordion.js
```diff
@@ -25,8 +29,21 @@ class Accordion extends PureComponent {
...
+      } else {
+        callbackProxy('onOpen', index);
```
dozortsev, line 44: what the point invoking `onClose` and `onOpen` via `callbackProxy`?

sbrandwoo:
For your use case you only need the `onOpen`? It could fire `onOpen(null)` when no items are open?

(Just thinking out loud)

packages/components/src/accordion/Accordion.js
tiborsaas-tw: That's a good point, I'll think of something.

packages/components/src/accordion/Accordion.js
```diff
@@ -13,10 +13,14 @@ class Accordion extends PureComponent {
         content: Types.node.isRequired,
       }),
     ).isRequired,
+    onOpen: Types.func,
+    onClose: Types.func,
```
dozortsev, line 17: do you need fire Mixpanel events on `onClose` and `onOpen` events?

tiborsaas-tw:
I added `onClose` since it looked like a nice to have feature, but I don't use it at the moment. Passing the new index in all cases makes sense though and consumers could rely on it. 

The event name should change then.

packages/components/src/accordion/Accordion.js
tiborsaas-tw: No, onClose is there because it looked like a nice to have feature.

packages/components/src/accordion/Accordion.js
tiborsaas-tw: Avoiding nested if-else blocks

packages/components/src/accordion/Accordion.js
andreapiras: Have you tried to return the index instead of using open and close and comparing it in your app to check what's been opened or closed and then manage all this logic in you app rather than inside the component?

dozortsev: @tiborsaas-tw I agree with Andrea here, I think it would make more sense to have on `onClick` prop we pass state (closed or opened)

tiborsaas-tw: I'll go with this approach

dozortsev: @tiborsaas-tw do I understand correctly that index value indicate whether component is opened or closed?

tiborsaas-tw: yes, if -1 then it's closed

tiborsaas-tw:
I updated the PR with a single `onClick` that always returns the new index.

packages/components/src/accordion/Accordion.spec.js
```diff
@@ -61,18 +60,14 @@ describe('Accordion', () => {
...
       component = createComponent(props);
     });
 
     it('onOpen event fires then onClose event fires', () => {
       component.find(AccordionItem).simulate('click');
-      expect(openClickMock).toHaveBeenCalled();
-
-      component.find(AccordionItem).simulate('click');
-      expect(closeClickMock).toHaveBeenCalled();
+      expect(onClickMock).toHaveBeenCalled();
```
andreapiras, line 70: it would be nice here to use `toHaveBeenCalledWith()` and `toHaveBeenCalledTimes`

andreapiras: or maybe ths expect(drink).toHaveBeenNthCalledWith(1, 'lemon');

tiborsaas-tw: It would and I tried it, but it didn't work, I got undefined :) Any idea why?

tiborsaas-tw: I know why :)

tiborsaas-tw: Nope. Didn't work again :( I get undefined.

tiborsaas-tw: Fixed it with adding a second parameter to `.simulate()`

packages/components/src/accordion/Accordion.js
```diff
@@ -26,8 +28,10 @@ class Accordion extends PureComponent {
 
   handleOnClick = (index) => {
     this.setState((prevState) => {
-      return { indexOpen: prevState.indexOpen === index ? -1 : index };
+      const newIndex = prevState.indexOpen === index ? -1 : index;
+      return { indexOpen: newIndex };
     });
+    this.props.onClick(index);
```
dozortsev, line 34: we should invoke this function only if it's not `null`

packages/components/src/accordion/Accordion.js
```diff
@@ -26,8 +28,10 @@ class Accordion extends PureComponent {
 
   handleOnClick = (index) => {
     this.setState((prevState) => {
-      return { indexOpen: prevState.indexOpen === index ? -1 : index };
+      const newIndex = prevState.indexOpen === index ? -1 : index;
+      return { indexOpen: newIndex };
```
dozortsev, line 32: I guess extracting this change now not necessary in this PR 🙂

packages/components/src/accordion/Accordion.js
tiborsaas-tw: Aaah, the proxy took care of that, but I removed it :)

packages/components/src/accordion/Accordion.js
tiborsaas-tw: One line over all!

PR #344, opened 2020-05-07
docs: Mixpanel setup & Cookie banner & page tracking

dozortsev:
**Link to deployed changes: https://transferwise.github.io/neptune-web/branch/mixpanel-setup/about/Home**

## ❓Context  

We want to start tracking Neptune Docs usage.

## 🚀 Changes 

Add Mixpanel, configure it and enable tracking for all pages.

Example of events: https://mixpanel.com/s/2w49R1

Plus, add yummy cookie banner so users can opt in or out using cookies usage, for that we use same script which on transferwise.com domain the [cookie-consent-js](https://github.com/transferwise/cookie-consent-js).

---

Tracking is happens on clicking by element with `.Nav__Link` CSS class.

## ✅ Checklist

### DIFF
--- a/packages/docs/next.config.js
+++ b/packages/docs/next.config.js
@@ -18,9 +18,9 @@ const assetPrefix =
   process.env.NODE_ENV === 'production'
     ? `/neptune-web${branch !== 'master' ? `/branch/${branch}` : ''}`
     : '';
-const isProdInstance =
-  process.env.NODE_ENV === 'production' && (branch === 'master' || branch === 'mixpanel-setup');
-// const isProdInstance = true;
+// const isProdInstance =
+//   process.env.NODE_ENV === 'production' && (branch === 'master' || branch === 'mixpanel-setup');
+const isProdInstance = true;
 
 module.exports = () =>
   withTM(
--- a/packages/docs/next.config.js
+++ b/packages/docs/next.config.js
@@ -13,14 +13,13 @@ const withTM = require('next-transpile-modules');
 
 const pageExtensions = ['js', 'mdx'];
 
+const isProdMode = process.env.NODE_ENV === 'production';
+
 const branch = getBranch.sync();
-const assetPrefix =
-  process.env.NODE_ENV === 'production'
-    ? `/neptune-web${branch !== 'master' ? `/branch/${branch}` : ''}`
-    : '';
-// const isProdInstance =
-//   process.env.NODE_ENV === 'production' && (branch === 'master' || branch === 'mixpanel-setup');
-const isProdInstance = true;
+const isMasterBranch = branch === 'master';
+
+const assetPrefix = isProdMode ? `/neptune-web${isMasterBranch ? '' : `/branch/${branch}`}` : '';
+const isProdInstance = isProdMode && isMasterBranch;
 
 module.exports = () =>
   withTM(
@@ -33,7 +32,7 @@ module.exports = () =>
             assetPrefix,
             env: {
               ASSET_PREFIX: assetPrefix,
-              isProdInstance,
+              IS_PROD_INSTANCE: isProdInstance,
             },
             webpack: (config) => {
               config.module.rules.push({
--- a/packages/docs/next.config.js
+++ b/packages/docs/next.config.js
@@ -18,8 +18,9 @@ const assetPrefix =
   process.env.NODE_ENV === 'production'
     ? `/neptune-web${branch !== 'master' ? `/branch/${branch}` : ''}`
     : '';
-const isProdInstance =
-  process.env.NODE_ENV === 'production' && (branch === 'master' || branch === 'mixpanel-setup');
+// const isProdInstance =
+//   process.env.NODE_ENV === 'production' && (branch === 'master' || branch === 'mixpanel-setup');
+const isProdInstance = true;
 
 module.exports = () =>
   withTM(
--- a/packages/docs/next.config.js
+++ b/packages/docs/next.config.js
@@ -19,7 +19,8 @@ const branch = getBranch.sync();
 const isMasterBranch = branch === 'master';
 
 const assetPrefix = isProdMode ? `/neptune-web${isMasterBranch ? '' : `/branch/${branch}`}` : '';
-const isProdInstance = isProdMode && isMasterBranch;
+// const isProdInstance = isProdMode && isMasterBranch;
+const isProdInstance = true;
 
 module.exports = () =>
   withTM(
--- a/packages/docs/next.config.js
+++ b/packages/docs/next.config.js
@@ -18,6 +18,8 @@ const assetPrefix =
   process.env.NODE_ENV === 'production'
     ? `/neptune-web${branch !== 'master' ? `/branch/${branch}` : ''}`
     : '';
+// const isProdInstance = process.env.NODE_ENV === 'production' && branch === 'master';
+const isProdInstance = true;
 
 module.exports = () =>
   withTM(
@@ -30,6 +32,7 @@ module.exports = () =>
             assetPrefix,
             env: {
               ASSET_PREFIX: assetPrefix,
+              isProdInstance,
             },
             webpack: (config) => {
               config.module.rules.push({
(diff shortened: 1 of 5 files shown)


anttio:
Using Mixpanel requires us to introduce cookie consent mechanism, a.k.a. cookie banner. So I would add that change to this PR.

dozortsev:
good call Antti, will add as part of this PR

packages/docs/pages/_app.js
```diff
@@ -25,6 +29,7 @@ class MyApp extends App {
       <>
         <Head>
           <title>Neptune Design System — TransferWise</title>
+          <script src="https://transferwise.com/cookie-consent.js" />
```
dozortsev, line 32: this is the only why how to consume this library https://github.com/transferwise/cookie-consent-js#consumption

it's not published on any kind of package registry like NPM

packages/docs/pages/_app.js
```diff
@@ -16,6 +19,8 @@ class MyApp extends App {
     if (pathname === '/') {
       Router.push(`${process.env.NODE_ENV === 'production' ? '/neptune-web/' : '/'}about/Home`);
     }
+    initMixpanel();
+    trackLink('.Nav__Link', Events.PAGE_VIEWED, { page: window.location.pathname });
```
dozortsev, line 23: we init mixpanel with disabled tracking by default (see mixpanel configs below), until user confirms consent with cookies on cookie banner

packages/docs/utils/mixpanel/index.js
```diff
@@ -0,0 +1,51 @@
...
+    // We add "opt_out_tracking_by_default=true" and "opt_out_persistence_by_default=true" to mixpanel.init
+    // so we don't set cookies and don't track customers before they give us their consent
+    opt_out_tracking_by_default: true,
+    opt_out_persistence_by_default: true,
+    // secure_cookie: true,
+    debug: true,
+  });
+
+  window.addEventListener('accepttwcookieconsent', () => {
+    mixpanel.opt_in_tracking();
```
dozortsev, line 25: this is necessary for catching event about user confirming cookies send by `https://transferwise.com/cookie-consent.js` script to enable Mixpanel tracking

packages/docs/utils/mixpanel/index.js
```diff
@@ -0,0 +1,51 @@
...
+  });
+}
+
+export function trackEvent(name, props) {
+  if (IS_PROD_INSTANCE) {
+    // https://developer.mixpanel.com/docs/javascript-full-api-reference#mixpaneltrack
+    mixpanel.track(eventNamePrefix(name), props);
+  } else {
+    // eslint-disable-next-line no-console
+    console.log(`Track Event '${name}', props`, props);
```
dozortsev, line 35: this log here is just for debugging reasons when you run docs locally

packages/docs/next.config.js
```diff
@@ -13,11 +13,14 @@ const withTM = require('next-transpile-modules');
...
+
 const branch = getBranch.sync();
-const assetPrefix =
-  process.env.NODE_ENV === 'production'
-    ? `/neptune-web${branch !== 'master' ? `/branch/${branch}` : ''}`
-    : '';
+const isMasterBranch = branch === 'master';
+
+const assetPrefix = isProdMode ? `/neptune-web${isMasterBranch ? '' : `/branch/${branch}`}` : '';
+// const isProdInstance = isProdMode && isMasterBranch;
```
twrdjpalmer, line 22: Comment

packages/docs/pages/_app.js
twrdjpalmer: Would it be better to track page views as opposed to nav link clicks?

packages/docs/utils/mixpanel/index.js
```diff
@@ -0,0 +1,51 @@
...
+
+export const Events = {
+  /**
+   * Event for tracking pages which user opens
+   */
+  PAGE_VIEWED: 'Page Viewed',
+};
+
+export function initMixpanel() {
+  console.log('initMixpanel');
```
twrdjpalmer, line 14: stray log

packages/docs/utils/mixpanel/index.js
```diff
@@ -0,0 +1,51 @@
...
+  mixpanel.init('8ba4a7a5182f05e0a79ded57d5d2f051', {
```
twrdjpalmer, line 15: Which mixpanel token is this?

packages/docs/utils/mixpanel/index.js
```diff
@@ -0,0 +1,51 @@
...
+    // We add "opt_out_tracking_by_default=true" and "opt_out_persistence_by_default=true" to mixpanel.init
+    // so we don't set cookies and don't track customers before they give us their consent
+    opt_out_tracking_by_default: true,
+    opt_out_persistence_by_default: true,
+    // secure_cookie: true,
+    debug: true,
```
twrdjpalmer, line 21: Does this need cleaning up?

packages/docs/pages/_app.js
domhorvathtw: +1 for page views

packages/docs/pages/_app.js
domhorvathtw: Here is how we did it in the help center: 
https://github.com/transferwise/help-center/blob/master/src/common/cookie-consent/cookie-consent-script.tsx

maybe, you need the `type="application/javascript` part?

packages/docs/utils/mixpanel/index.js
domhorvathtw: it's the same that we use in Help Center, should we use the same?

packages/docs/pages/_app.js
dozortsev: it's JavaScript by default https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attr-type

packages/docs/utils/mixpanel/index.js
dozortsev: yes, and same as in many other TW apps https://github.com/search?q=org%3Atransferwise+%228ba4a7a5182f05e0a79ded57d5d2f051%22&type=Code

packages/docs/utils/mixpanel/index.js
dozortsev: yes

packages/docs/pages/_app.js
dozortsev: seems not necessary as it's JavaScript by default https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attr-type

packages/docs/pages/_app.js
dozortsev: @twrdjpalmer you mean fire event on "component did mount"?

packages/docs/next.config.js
dozortsev: yes, this is here now so we can play (send some events) with Mixpanel from non prod instance

dozortsev: will remove this right before merging PR

packages/docs/utils/mixpanel/index.js
twrdjpalmer: For your sake, might be better to have your project so design systems stuffs are totally separate from the core business tracking?

packages/docs/pages/_app.js
twrdjpalmer: Yeah, exactly. You can probably encapsulate the logic into a component rendered by `_app.js`, e.g.:

```
// Tracker.js
import Router from 'next/router';

export default function PageViewTracker() {
  const handleRouteChange = url => {
    // tracking code
  }

  useEffect(() => {
    Router.events.on('routeChangeStart', handleRouteChange);

    return () => Router.events.off('routeChangeStart', handleRouteChange);
  }, []);

  return null;
}

// _app.js
return (
  <>
    <Component />
    <PageViewTracker />
  </>
)
```

helendurranttw:
Closing for now, will reopen when we're ready to revisit mixpanel

PR #345, opened 2020-05-07
fix: remove button role from modal

paul-frybarger:
## ❓ Context  

Modal component has the 'role="button"' property set, which makes the cursor a pointer over everything.

## 🚀 Changes 

Before:
Before

After:
After

## 💬 Considerations 

This one is hard to test, although it shouldn't af

## ✅ Checklist


packages/components/src/modal/Modal.js
```diff
@@ -98,7 +98,7 @@ class Modal extends Component {
           <div
             className={`tw-modal fade ${className}`}
             tabIndex="-1"
-            role="button"
+            role="dialog"
```
paul-frybarger, line 101: `role="dialog"` is also set on the child component, not sure if this needs to be set on both. Since this has a click handler, it needs a role.

paul-frybarger:
Alright, so this is still failing the build and I'm not sure what the appropriate fix is ☹️

dozortsev:
@paul-frybarger it's failing because of eslint error

```
/home/circleci/neptune-web/packages/components/src/modal/Modal.js
 
  98:11  error  Non-interactive elements should not be assigned mouse or keyboard event listeners  jsx-a11y/no-noninteractive-element-interactions
 
```

packages/components/src/modal/Modal.js
helendurranttw: Ahh I see what's happened. The build is failing because the linter doesn't like that there is a click handler on a non-interactive element. I'm guessing the first time around, someone decided to add role button to that modal to mark it as interactive. We'll need another solution.

PR #360, opened 2020-05-14
feat: implement secondary text on radio & checkbox

StevePole:
##  📎 Jira ticket
😕 

## ❓ Context 
The interface for these components already supported the feature, but it was never implemented.

## 🚀 Changes 
Adds an optional second line of text in radio and checkboxes
image
image

## 💬 Considerations 
n/a

## ✅ Checklist


packages/components/src/checkbox/Checkbox.js
```diff
@@ -9,18 +9,29 @@ const Checkbox = ({
...
   readOnly,
   label,
+  secondary,
   onChange,
   onFocus,
   onBlur,
 }) => {
   const handleClick = () => (!disabled && !readOnly ? onChange(!checked) : null);
   const errorClass = required && !disabled && !readOnly && !checked ? 'has-error' : '';
+
```
helendurranttw, line 19: Could use `classnames` here to tidy up these conditionals.

packages/components/src/checkbox/Checkbox.js
```diff
@@ -53,6 +65,7 @@ Checkbox.defaultProps = {
   required: false,
   disabled: false,
   readOnly: false,
+  secondary: '',
```
helendurranttw, line 68: In Radio, the default value is null. We should probably be consistent and change it to string.

packages/components/src/checkbox/Checkbox.js
```diff
@@ -9,18 +9,29 @@ const Checkbox = ({
   disabled,
   readOnly,
   label,
+  secondary,
```
helendurranttw, line 12: In Radio option and Checkbox option the secondary text is called "content". I agree that secondary is a sensible name, but do we want to be consistent?

packages/components/src/checkbox/Checkbox.js
StevePole: Not sure I follow what you mean.  Originally in used classNames in an inline fashion, but it wasn't readable so I broke it out.

packages/components/src/checkbox/Checkbox.js
StevePole: I chose secondary as it's already in the interface for the Radio component: https://github.com/transferwise/neptune-web/blob/master/packages/components/src/radio/Radio.js#L23

As it didn't actually do anything before, we could change it.  Though I'm not a big fan of 'content', it's too generic.  Would also be happy with `description`.

packages/components/src/checkbox/Checkbox.js
helendurranttw: Sorry - `classnames` is a package that lets you easily construct class lists like this:

```
const classList = classNames({
    'has-error' : required && !disabled && !readOnly && !checked,
   'checkbox-lg': secondary, 
    disabled: disabled
  });
```

packages/components/src/checkbox/Checkbox.js
helendurranttw: Yeah I'm with you on "content". Let's just leave it as is, I think secondary makes sense.

PR #361, opened 2020-05-14
Add support for const schemas, select boxes for oneOf schemas

helendurranttw:
## ❓ Context  
We would like to be able to support a native JSON schema approach for select boxes
```
{
  title: "Favourite number?",
  oneOf: [
    { title: "One", const: 1 }
    { title: "Two", const: 2 }
  ]
}
```

## 🚀 Changes 
This changes oneOf schemas to display as select boxes when there are more options, and support const schemas which broadcast their value when chosen.

## 💬 Considerations 
n/a

## ✅ Checklist


packages/dynamic-flows/src/common/validation/type-validators/spec.js
```diff
@@ -94,4 +103,30 @@ describe('Given a library for validating data types', () => {
...
+      expect(isNull(0)).toBe(false);
+      expect(isNull(false)).toBe(false);
+      expect(isNull([])).toBe(false);
+      expect(isNull({})).toBe(false);
+      expect(isNull(undefined)).toBe(false);
+    });
+  });
+
+  describe('when validating an undefined', () => {
+    it('should return true when the value is null', () => {
```
helendurranttw, line 121: Typo: undefined, not null

packages/dynamic-flows/src/common/validation/type-validators/spec.js
```diff
@@ -94,4 +103,30 @@ describe('Given a library for validating data types', () => {
...
+      expect(isUndefined(undefined)).toBe(true);
+    });
+    it('should return false when the value is not null', () => {
```
helendurranttw, line 124: Typo: undefined, not null

packages/dynamic-flows/src/formControl/FormControl.js
```diff
@@ -232,6 +232,14 @@ export default class FormControl extends PureComponent {
     return selectedOption;
   };
 
+  mapOption = (option) => {
+    return {
+      ...option,
+      disabled: option.disabled || this.props.disabled,
+      readOnly: this.props.readOnly,
```
helendurranttw, line 239: I guess it doesn't make sense for radioButtons, but should we also prefer `option.readOnly` over `props.readOnly` here?

packages/dynamic-flows/src/jsonSchemaForm/validation/valid-model/spec.js
```diff
@@ -76,6 +76,63 @@ describe('Given a library for returning the valid parts of a model based on a sc
...
+      it('should return the original model', () => {
+        expect(result).toBe(1);
+      });
+    });
+
+    describe('if the model is not part of the enum', () => {
+      beforeEach(() => {
+        result = getValidModelParts(3, schema);
+      });
+      it('should return undefined', () => {
```
helendurranttw, line 97: Undefined, or null?

packages/dynamic-flows/src/jsonSchemaForm/validation/valid-model/spec.js
```diff
@@ -76,6 +76,63 @@ describe('Given a library for returning the valid parts of a model based on a sc
...
+        expect(result).toBe(1);
+      });
+    });
+
+    describe('if the model does not match the const', () => {
+      beforeEach(() => {
+        result = getValidModelParts(2, schema);
+      });
+      it('should return undefined', () => {
+        expect(result).toBeNull();
```
helendurranttw, line 122: Undefined, or null?

packages/dynamic-flows/src/formControl/FormControl.js
StevePole: I don't think there is a concept of a readonly option, as options are never editable.  I presume that by passing it here, we're actually just disabling the option as part of broader control.  However, I was really just preserving the existing behaviour.

packages/dynamic-flows/src/jsonSchemaForm/validation/valid-model/spec.js
StevePole: null, good spot.

PR #368, opened 2020-05-20
Add missing text colours to tokens

anttio:
## 📎 Jira ticket
https://transferwise.atlassian.net/browse/DS-428

## ❓ Context  
Tokens were missing one colour, the `text-control` one.

## 🚀 Changes 
Adds missing colours to the tokens. Also adds deprecation notice to deprecated colours.


packages/css/src/less/variables/_colors.less
```diff
@@ -59,13 +60,12 @@
   --color-background-positive: @color-base-green-fade;
   --color-background-negative: @color-base-red-fade;
   --color-background-warning: @color-base-orange-fade;
-}
 
-:root {
+  // Deprecated
   --brand-white: @color-base-white;
```
helendurranttw, line 65: Should we make an announcement that these are deprecated?

PR #369, opened 2020-05-20
fix: tooltip position when using bottom position

anttio:
## 📎 Jira ticket
https://transferwise.atlassian.net/browse/DS-240

## ❓ Context  
Using tooltip position bottom positioned the tooltip wrongly.

## 🚀 Changes 
Calculate the position of the tooltip based on the parent element height.

## ✅ Checklist


packages/components/src/tooltip/Tooltip.js
```diff
@@ -23,7 +23,7 @@ function getTooltipStyle(parent, tooltip, position, offset) {
       };
     case Position.BOTTOM:
       return {
-        top: `${parent.offsetTop + tooltip.offsetHeight + offset}px`,
+        top: `${parent.offsetTop + parent.offsetHeight + offset}px`,
```
andreapiras, line 26: Is there any test covering this?

anttio: No. I think the tests on this component has never been that thorough.

andreapiras: We need to give some love to this component

helendurranttw: @andreapiras we have a ticket for that :)

PR #372, opened 2020-05-21
chore: add minimum required node version

helendurranttw:
BREAKING CHANGE: Adding minimum required node version

## 📎 Jira ticket

DS-410

## 💬 Considerations

Neptune web will not build if your node version is older than the minimum required version, so marking as breaking.


package.json
```diff
@@ -21,6 +21,9 @@
   "workspaces": [
     "packages/*"
   ],
+  "engines": {
+    "node": ">=12.16.3"
```
dozortsev, line 25: will one see an error message in terminal if their node version is lower?

PR #374, opened 2020-05-21
feat: Allow generating html id tag for each accordion item

transferwise-yizhang:

## ❓ Context  
So that each accordion item can have an id. This would allow deep linking into that accordion's section from the url.

## ✅ Checklist


packages/components/src/accordion/Accordion.js
```diff
@@ -38,6 +39,7 @@ class Accordion extends PureComponent {
   render() {
     return this.props.items.map((item, index) => (
       <AccordionItem
+        id={this.props.idFunc && this.props.idFunc(item)}
```
sergonius, line 42: You can use optional chaining here

paul-frybarger:
Hm, looks like optional chaining isn't supported:
`SyntaxError: /home/circleci/neptune-web/packages/components/src/accordion/Accordion.js: Support for the experimental syntax 'optionalChaining' isn't currently enabled (42:23)`

paul-frybarger:
How about this:
- add `id` to the accordion item
- if it exists, use that as the id as well as the react key
  - added benefit of keeping the item order if rearranged

sergonius:
> Hm, looks like optional chaining isn't supported:
> `SyntaxError: /home/circleci/neptune-web/packages/components/src/accordion/Accordion.js: Support for the experimental syntax 'optionalChaining' isn't currently enabled (42:23)`

Yeah, `@babel/preset-env` in this repo needs to be upgraded for that :/

PR #375, opened 2020-05-22
feat: Allow accordion open item to react to parent state change

transferwise-yizhang:

## ✅ Checklist


packages/components/src/accordion/Accordion.js
```diff
@@ -27,6 +27,13 @@ class Accordion extends PureComponent {
     indexOpen: this.props.indexOpen,
   };
 
+  /* eslint-disable react/no-did-update-set-state */
+  componentDidUpdate(prevProps) {
```
sergonius, line 31: Did you consider `getDerivedStateFromProps`?

transferwise-yizhang: Hey thanks for the suggestion. I tried to use `getDerivedStateFromProps` just now but it didn't work. With accordions, it's a bit more complex. What I imagined the follow is that the indexOpen should be if the parent prop changes, but within the accordion itself, clicking each item open and close should still be valid, which I couldn't get `getDerivedStateFromProps` to work.

Anything else I could try?

PR #378, opened 2020-05-27
fix:  babel minify renaming error

andreapiras:
## ❓ Context  
Babel minify plugin is outputting a wrong code 
Original code: 
```
const countSeparatorsAfterCursor = (newPattern, position) => {
  let separators = 0;
  while (positionIsSeparator(newPattern, position + separators)) {
    separators += 1;
  }
  return separators;
};
```

Generated code:
```
countSeparatorsAfterCursor = function (a, b) {
    for (var c = 0; positionIsSeparator(a, b + c); ) c += 1;
    return separators;
  },
```

separator is undefined

## 🚀 Changes 
Add an exception for separator variable in variables renaming. 

## 💬 Considerations 
I specifically excluded renaming separator since the issue is happening only in that file and may be fixed in next versions of minify, but it can be disabled for all renames.

## ✅ Checklist


packages/dynamic-flows/package.json
```diff
@@ -23,7 +23,7 @@
     "dev:storybook": "start-storybook",
     "docs": "npm-run-all --parallel docs:*",
     "docs:babel": "NODE_ENV=es babel src -d build/es/polyfill --ignore '**/*.spec.js','**/*.story.js' -w",
-    "docs:watch-less": "gulp lessWatcher --dest=src",
+    "docs:watch-less": "gulp lessWatcher --gulpfile=../../node_modules/@transferwise/less-config --dest=src",
```
andreapiras, line 26: add missing path for gulpfile

packages/components/package.json
```diff
@@ -22,7 +22,7 @@
     "dev:watch-less": "gulp lessWatcher --gulpfile=../../node_modules/@transferwise/less-config --dest=src",
     "dev:storybook": "start-storybook -p 3001",
     "docs": "npm-run-all --parallel docs:*",
-    "docs:babel": "NODE_ENV=es babel src -d build/es/polyfill --ignore '**/*.spec.js','**/*.story.js,' -w",
+    "docs:babel": "NODE_ENV=es babel src -d build/es/polyfill --ignore '**/*.spec.js','**/*.story.js' -w",
```
andreapiras, line 25: Remove extra comma

packages/components/babel.config.js
```diff
@@ -47,12 +47,7 @@ const esConfig = {
         modules: false,
       },
     ],
-    [
-      'minify',
-      {
-        builtIns: false,
-      },
-    ],
+    ['minify', { builtIns: false, mangle: { exclude: { separators: true } } }],
```
andreapiras, line 50: variables separator does't get replaced. We can exclude all mangles but I think it will be too much. This fix may be fixed in babel-minify at some point

andreapiras: `https://github.com/babel/minify/issues/966`

dozortsev: what source code now gets generated?

dozortsev: make sure to leave your vote in this issue https://github.com/babel/minify/issues/966 so maybe owners & contributors fix it faster

andreapiras: `countSeparatorsAfterCursor = function (a, b) {
    for (var separators = 0; positionIsSeparator(a, b + separators); ) separators += 1;
    return separators;
  },`

dozortsev: but `separators` is only inside of scope of `for` loop, this function will just return `undefined` no?

andreapiras: The var statement declares a function-scoped or globally-scoped variable

PR #382, opened 2020-06-01
feat: change popover interface to allow close action from inside content

rerdos:

## ❓ Context
It's not possible to close the Popover from its content, however if we need 'customer action' inside the Popover (for example a button), we should provide ability to trigger a close on Popover.

## 🚀 Changes
- Be able to pass a function as `content`, which will receive `close` and `isOpen` props.
- Be able to set `tabIndex` of Popover, so customers could focus a focusable trigger within the Popover, and interact with that. (Which will still open the Popover due to event bubbling). This would enable us to eliminate the double focus we currently have.
- Make Popover toggle itself, rather than just open, so it is aligned with Popover from Bootstrap (where the trigger toggles the popover).

## ✅ Checklist


packages/components/src/popover/Popover.js
```diff
@@ -48,15 +50,16 @@ export default class Popover extends Component {
...
+    if (!popoverClicked && !triggerClicked) {
       this.close();
     }
   };
 
-  handleKeyUp = ({ keyCode }) => {
-    if (keyCode === KeyCodes.ENTER) {
-      this.open();
+  handleKeyUp = (event) => {
+    if (event.target.nodeName !== 'BUTTON' && event.keyCode === KeyCodes.ENTER) {
```
rerdos, line 61: There was a bug where `this.toggle` (previously `this.open`) were triggered twice, as JavaScript automatically maps a `keypress` `ENTER` to a `click` event on a button. So when we had a button within Popover the keypress and click bubbled up called `this.toggle` (`this.open`) twice. Which was not an issue as previously it was only setting the state of the Popover to open.

rerdos: So we can ignore keyUp events from buttons, as they will trigger onClick regardless.

StevePole:
Should popovers contain interactive elements?  I would have thought they were reserved for information display.  With modals providing an alternative where interaction is possible?  

Could you share a screenshot of your use case?

packages/components/src/popover/Popover.js
```diff
@@ -18,7 +18,8 @@ export default class Popover extends Component {
   static propTypes = {
     children: Types.oneOfType([Types.element, Types.string]).isRequired,
     title: Types.oneOfType([Types.element, Types.string]),
-    content: Types.oneOfType([Types.element, Types.string]).isRequired,
+    content: Types.oneOfType([Types.element, Types.string, Types.func]).isRequired,
```
helendurranttw, line 21: I'd like to understand your thinking here - did you consider just passing in an onClose function that could be run on close?

packages/components/src/popover/Popover.js
```diff
@@ -34,6 +35,7 @@ export default class Popover extends Component {
 
   static defaultProps = {
     title: null,
+    tabIndex: null,
```
helendurranttw, line 38: Why the need to change the tab index here?

packages/components/src/popover/Popover.js
helendurranttw: Oh.. wait. Ignore me. I just realised of course you need the ability to trigger close yourself 🤦‍♀️

packages/components/src/popover/Popover.js
rerdos: If the `children` that opens the Popover has a button inside itself, that could be focused, we should provide a way to disable the `tabIndex` on the outer wrapper, so users with keyboard-control would not encounter 'double focus'.

For example:
https://transferwise.github.io/neptune-web/components/Popover

When we `tab` onto the Popover, firstly, it focuses the wrapper in a non-visible way, but it's able to open the Popover when we press Enter. The second `tab` focuses the button inside the wrapper (this one has a :focused state), and again we are able to open the Popover thanks to event bubbling.

I also changed the default `tabIndex` to `0` in my most recent commit. So it's enabled by default with the option to disable.

helendurranttw: Gotcha. Thanks heaps for the explanation. That's definitely a bug with the existing code.

So - I would like to try and fix this without a need to change the API, as we're hoping to reimplement this component soon, and I'd like to not have to support the tabIndex prop forever.

How about this for an idea:

```
const config = child.props && typeof child.props.onClick === 'function' ? {
    tabIndex: -1,
} : {
    role: 'button',
};

return cloneElement(wrappedChild, {
    ...config,
   // and so on
},
    });
```

So.. if there is a click handler, we don't allow focus of the span, and we don't set the role. We still attach the click handler, as this will get triggered by propagation of the child event handler.

What do you think?

packages/components/src/popover/Popover.js
```diff
@@ -118,7 +129,9 @@ export default class Popover extends Component {
         >
           {title && <h3 className={classNames(this.style('popover-title'))}>{title}</h3>}
           <p className={classNames(this.style('popover-content'), this.style('m-b-0'))}>
-            {content}
+            {typeof content === 'function'
```
andreapiras, line 132: I played a bit with the content and tried this:

 {isContentInteractive
              ? React.cloneElement(<span>{content}</span>, { onClick: this.close })
              : content}
That would prevent us to add line 21

maybe we can add a prop isContentInteractive
WDYT

rerdos: That's an interesting idea, @andreapiras . Although I can see two issues:

- If the content contains a block level element (like `div` or `p`) it would make the markup incorrect. A solution could be to change the `span` to a `div`.
- We might not want to close the Popover on `any` click within the `content`. I think the current approach (passing `close` to `content`), gives way more control to the consumer.

Otherwise, I'd prefer that approach. 👍

rerdos: So after playing around with the aforementioned suggestion, I couldn't find a way that would fit our use-case, where we only want to close the popover, when a specific element is clicked within content. I would go with the original change @andreapiras , unless you have any other ideas / suggestions regarding this. 👍

andreapiras:
Hey @rerdos code seems ok to me, Can I just ask you to add the new prop in the live code editor
https://github.com/transferwise/neptune-web/blob/master/packages/docs/liveEditorCode/popover.code.js
Thanks

PR #383, opened 2020-06-01
fix: improve schema validation utils

StevePole:
## 📎 Jira ticket
n/a

## ❓ Context  
Improves JSON schema validation to better meet spec and cover additional edge cases

## 🚀 Changes 
JSON schema validation now expects 'minimum' and 'maximum', not 'min' and 'max' as per the specification. 

isNumber now return false for NaN

We reuse the our type checkers in more places to better cover edge cases.

## 💬 Considerations 
n/a

## ✅ Checklist


packages/dynamic-flows/src/common/validation/type-validators/index.js
```diff
@@ -2,7 +2,7 @@ function isString(value) {
   return typeof value === 'string';
 }
 function isNumber(value) {
-  return typeof value === 'number';
+  return typeof value === 'number' && !isNaN(value); // eslint-disable-line
```
helendurranttw, line 5: Is there a reason to use `isNaN` over `Number.isNaN` here? I think since you're checking typeof `Number.isNaN` is equivalent, and you can remove the warning.

packages/dynamic-flows/src/jsonSchemaForm/validation/schema-validators/index.js
```diff
@@ -121,3 +121,5 @@ export function isValidSchema(value, schema) {
       return false;
   }
 }
+
+export { isValidSchema }; // eslint-disable-line
```
helendurranttw, line 125: You shouldn't need this eslint disable I don't think, what error were you getting?

packages/dynamic-flows/src/jsonSchemaForm/validation/valid-model/spec.js
```diff
@@ -52,6 +52,30 @@ describe('Given a library for returning the valid parts of a model based on a sc
...
+      it('should return the original integer', () => {
+        expect(result).toBe(12345);
+      });
+    });
+
+    describe('with any non integer model', () => {
+      beforeEach(() => {
+        result = getValidModelParts(12.34, schema);
+      });
+      it('should return undefined', () => {
```
helendurranttw, line 73: should return null

packages/dynamic-flows/src/common/validation/type-validators/index.js
StevePole: No reason, good point 👍

packages/dynamic-flows/src/jsonSchemaForm/validation/schema-validators/index.js
StevePole: Unfortunately this same code also exists in the angular implementation, and for some reason I needed it there.  Works fine here without 👍

PR #400, opened 2020-06-03
feat: add Neptune tokens

gusso:
## 🚀 Changes
Adds [Neptune Tokens](https://github.com/transferwise/neptune-tokens) as a dependency to move them to a common source of truth with other platforms and design tools.

## 💬 Considerations
The output on Neptune Web should continue being the same in terms of exposing the Less variables and CSS properties for tokens, but we can reconsider how we expose them later, as generating multiple versions of our tokens is very easy now.


anttio:
There are two custom properties for colours missing that were in the old manually handled tokens: `--color-text-link-hover` and `--color-text-link-active`.

Should we have these in tokens, or maybe implement the interactive states differently?

PR #357, opened 2020-05-14
feat: migration on new icons

dozortsev:
## ❓ Context  

Migrating on new [`@transferwise/icons`](https://github.com/transferwise/icons) v2.

## 🚀 Changes 

Driven by **[Migration Checklist](https://transferwise.atlassian.net/wiki/spaces/DS/pages/1355256863/Checklist+of+Migration+on+New+Icons)**.

- update `@transferwise/icons` package to `v2.0.0` in components and docs packages
- replaced old React icons on new ones
- use new React icons instead of `.icon .icon-*`, `.glyphicon .glyphicon-*`, and `.caret` CSS classes
- use Chevron icon inside Chevron React component
- delete `.caret` CSS class
- delete `iconfont.less` file
- stopped explicitly import `glyphicons.less` and `icons.less` from Bootstrap
- renamed `.icon` on `.tw-icon` in styles
- and of course after all of that I had to update loads of tests and adjust styles 🙄

## 💬 Considerations 

Bases code review it needs visual review, here is list of components and places in docs which have got new icons:

Components Pages:
- [x] new [`Alert`](https://transferwise.github.io/neptune-web/branch/migrate-componenents-on-new-icons-set/components/Alert) vs current [one](https://transferwise.github.io/neptune-web/components/Alert)
- [x] new [`Accordion`](https://transferwise.github.io/neptune-web/branch/migrate-componenents-on-new-icons-set/components/Accordion) vs current [one](https://transferwise.github.io/neptune-web/components/Accordion)
- [x] new [`Card`](https://transferwise.github.io/neptune-web/branch/migrate-componenents-on-new-icons-set/components/Card) vs current [one](https://transferwise.github.io/neptune-web/components/Card)
- [x] new [`Chevron`](https://transferwise.github.io/neptune-web/branch/migrate-componenents-on-new-icons-set/components/Chevron) vs current [one](https://transferwise.github.io/neptune-web/components/Chevron)
- [x] new [`CheckboxOption`](https://transferwise.github.io/neptune-web/branch/migrate-componenents-on-new-icons-set/components/CheckboxOption) vs current [one](https://transferwise.github.io/neptune-web/components/CheckboxOption)
- [x] new [`Chip`](https://transferwise.github.io/neptune-web/branch/migrate-componenents-on-new-icons-set/components/Chip) vs current [one](https://transferwise.github.io/neptune-web/components/Chip)
[...]

PR #401, opened 2020-06-03
Ds 461 switch component

andreapiras:
**[Demo](https://transferwise.github.io/neptune-web/branch/DS-461-switch-component/components/Switch)** in docs.

## 📎 Jira ticket
https://transferwise.atlassian.net/browse/DS-461

## ❓ Context  
Add switch components. Please find designs here
https://www.figma.com/file/nQMZNbhGV0QkbNypQksuGH/Switcher---web?node-id=1%3A352

## 🚀 Changes 
Adds Switch components

## ✅ Checklist

win8_firefox_62 0
win8 1_chrome_71 0
win7_ie_11 0
win7_chrome_71 0
win10_ie_11 0
macmav_chrome_36 0
macelc_safari_9 1
6 0_Samsung-Galaxy-S7_portrait_real-mobile
6 0_Google-Nexus-6_portrait_real-mobile
4 4_Google-Nexus-5_portrait_real-mobile
11_iPhone-8-Plus_portrait_real-mobile
10_iPhone-7_portrait_real-mobile


packages/components/src/switch/Switch.js
```diff
@@ -0,0 +1,41 @@
...
+  return (
+    <span
+      className={classnames('switch', {
+        'switch--unchecked': !checked,
+        'switch--checked': checked,
+      })}
+      onClick={onClick}
+    >
+      <span className="switch--thumb">
+        {checked ? <CheckCircle /> : <CrossCircle filled hasFillVariant />}
```
dozortsev, line 25: React icons components doesn't have prop like `hasFillVariant`, what is it for here?

packages/components/src/switch/Switch.js
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Switch docs.
+ *
+ * @param {type} [propname=""] - propsDescription.
+ *
+ * @usage '<Switch/>'
+ **/
+
+const Switch = ({ checked, onClick }) => {
```
dozortsev, line 15: will this component have some `label`?

dozortsev: asking because I think that if not then this component should be similar to `CheckboxButton` and `RadioButton` ones which are used in [`Checkbox`](https://github.com/transferwise/neptune-web/blob/master/packages/components/src/checkbox/Checkbox.js#L36) and [`Radio`](https://github.com/transferwise/neptune-web/blob/master/packages/components/src/radio/Radio.js#L10) accordingly, so I'm saying that we probably should have a `SwitchButton` component (atom) and `Switch` (molecule) where is last one use `SwitchButton` and adds `label` etc.

andreapiras: I think that we should have a unique and consistent way of adding labels and that should be via Field or any other wrapper. So all the label will have one style and looks consistent

package.json
```diff
@@ -59,7 +59,7 @@
     "test": "lerna run test --stream",
     "test:watch": "cd packages/components && yarn test:watch",
     "clean": " npm-run-all clean:*",
-    "clean:modules": " lerna exec -- rm -rf ./node_modules && rm -rf node_modules  && rm -rf .next",
```
andreapiras, line 62: This command wasn't working

packages/components/src/switch/Switch.js
```diff
@@ -0,0 +1,62 @@
...
+      onClick={onClick}
+      onKeyDown={handleKeyDown}
+      tabIndex={0}
+      role="checkbox"
+      aria-checked={checked}
+    >
+      <span className="switch--thumb">
+        {checked ? (
+          <>
+            <CheckCircle filled className="hidden-xs" />
```
andreapiras, line 39: There's no way to change pass the size to icon based on window size as that would cause problem on SSR initial load

packages/components/src/switch/Switch.less
```diff
@@ -0,0 +1,60 @@
...
+    width: 44px;
+    &--thumb {
+      .tw-icon {
+        width: 16px;
+        height: 16px;
+      }
+    }
+    &--checked {
+      .switch--thumb {
+        transform: translateX(18px);
```
andreapiras, line 56: could not find a better way to manage this value

packages/components/src/switch/Switch.js
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Switch component provides a simple switch button.
+ *
+ * @param {boolean} [checked=""] - boolean to set the switch status.
+ * @param {function} [onClick=""] - function that manage the check logic. Required.
+ *
+ * @usage '<Switch checked={checked} onClick={() => setCheck(!checked)} />'
+ * */
+
+const Switch = ({ checked, onClick }) => {
```
dozortsev, line 17: please add `className` prop

StevePole:
I clicked on this one because I wasn't sure what a 'Switch' was.  Is that what it's called on mobile?  My instinct was that this would have been called 'Toggle', which may better describe how we expect it to be used?

andreapiras:
> I clicked on this one because I wasn't sure what a 'Switch' was. Is that what it's called on mobile? My instinct was that this would have been called 'Toggle', which may better describe how we expect it to be used?


image
Hey Steve, the naming on Android for the whole component is (purple rect): Toggle and 
the control type (pink rect): Switch, Checkbox, Radio, the pink Rectangle is called UISwitch natively on iOS

packages/components/src/switch/Switch.less
```diff
@@ -0,0 +1,44 @@
+@import '@transferwise/neptune-css/src/less/variables/_transitions.less';
+@import '@transferwise/neptune-css/src/less/variables/_colors.less';
+@import '@transferwise/neptune-css/src/less/variables/_custom-media.less';
+@import '@transferwise/neptune-css/src/less/variables/_spacing.less';
+
+.switch {
+  display: inline-flex;
+  width: 50px;
+  overflow: hidden;
+  border-radius: 16px;
```
dozortsev, line 10: this should be defined as variable

packages/components/src/switch/Switch.less
```diff
@@ -0,0 +1,44 @@
...
+  padding: 2px;
```
dozortsev, line 11: and I guess this one as well

packages/components/src/switch/Switch.less
andreapiras: This seems a very specific case, normally we define variable in neptune-css or tokens , I don't think we should not put it for now in those files

packages/components/src/switch/Switch.less
andreapiras: Same as above

packages/components/src/switch/Switch.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import Types from 'prop-types';
+import classnames from 'classnames';
+import './Switch.css';
+import { CheckCircle, CrossCircle } from '@transferwise/icons';
+import KeyCodes from '../common/keyCodes';
+
+/**
+ * Switch component provides a simple switch button.
```
helendurranttw, line 9: Lovely docs thank you 😌you can remove the "Required" - this is implied by the brackets around the parameter name. You can make a few changes to make them match the signature:

`@param {boolean} [checked=false]` - means "A parameter called `checked` that is optional and defaults to false
`@param {function} onClick` - means "A parameter called `onClick` which is required (has no default).

Also we need to add one for the new ariaLabelledby :)

packages/components/src/switch/Switch.less
```diff
@@ -0,0 +1,42 @@
...
+  cursor: pointer;
+  border: 1px solid var(--brand-smoke);
```
helendurranttw, line 13: Hmmmm we haven't decided yet if we want to use CSS props here. I've scheduled a conversation for tomorrow to discuss this.

packages/components/src/switch/Switch.less
```diff
@@ -0,0 +1,42 @@
...
+    }
+  }
+  &--checked {
+    border: 1px solid var(--color-accent);
+    background: var(--color-accent);
+    .switch--thumb {
+      transform: translateX(20px);
+    }
+  }
+  > input {
```
helendurranttw, line 34: Probably don't need the `>` here.

packages/components/src/switch/Switch.less
```diff
@@ -0,0 +1,42 @@
+@import '@transferwise/neptune-css/src/less/variables/_transitions.less';
+@import '@transferwise/neptune-css/src/less/variables/_colors.less';
+@import '@transferwise/neptune-css/src/less/variables/_custom-media.less';
```
helendurranttw, line 3: Is custom media being used here?

packages/components/src/switch/Switch.story.js
```diff
@@ -0,0 +1,18 @@
...
+  component: Switch,
+  title: 'Switch',
+};
+
+export const basic = () => {
+  const [checked, setCheck] = useState(false);
+
+  return (
+    <>
+      <div id="myNameId">A Possible label</div>
```
helendurranttw, line 14: Use `<label>` 😍

packages/components/src/switch/Switch.js
andreapiras: I did add ariaLabelledBy because is going to be needed once we add the label

packages/components/package.json
```diff
@@ -74,7 +74,7 @@
   "dependencies": {
     "@babel/runtime": "^7.9.2",
     "@transferwise/formatting": "^2.1.0",
-    "@transferwise/icons": "^1.0.0",
+    "@transferwise/icons": "^2.0.0",
```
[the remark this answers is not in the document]
andreapiras: This branch should be released after @dozortsev icon branch
https://github.com/transferwise/neptune-web/pull/357/files#diff-fcafbf9322186d9a2998fd8c67903791R77

PR #405, opened 2020-06-03
fix: removed gulpwatcher redundant .less regex

andreapiras:

## ❓ Context  
gulp lessWatcher wasn't working because of an extra `.less` in config

## ✅ Checklist


helendurranttw:
The less watching is broken in a more substantial way than this. I have this fix (among others) in another branch which I'll be committing soon.

PR #409, opened 2020-06-04
fix: fixing docs css reload

helendurranttw:

## ✅ Checklist


packages/css/src/less/components/_component-animations.less
```diff
@@ -1,7 +1,8 @@
 // Component animations
 .fade {
+  transition: opacity 0.15s linear;
```
helendurranttw, line 3: These mixins just added vendor prefixes.

packages/components/.gulp.json
```diff
@@ -0,0 +1,5 @@
+{
+  "flags": {
+    "gulpfile": "../../node_modules/@transferwise/less-config/gulpfile.js"
```
andreapiras, line 3: Very nice

andreapiras: Shall we maybe mention this config inside less-config package README?

yarn.lock
```diff
@@ -9,6 +9,14 @@
   dependencies:
     cross-fetch "3.0.4"
 
+"@ampproject/toolbox-core@^2.5.0":
```
dozortsev, line 12: where this dependencies coming from?

helendurranttw: Good point, I think I reverted some deps but not the lock file. I'll revert this.

PR #417, opened 2020-06-08
feat: avatar component

rerdos:
[**Demo**](https://transferwise.github.io/neptune-web/branch/DS-275-avatar-component/components/Avatar) on docs.

## 📎 Jira ticket
[DS-275](https://transferwise.atlassian.net/browse/DS-275)

## ❓ Context
We want to have an Avatar component in Neptune.

## 🚀 Changes
Adding Avatar component.

## 💬 Considerations
We want to have badges on the Avatar component, so we have to think how a Badge component would with the Avatar.

## ✅ Checklist


packages/components/src/avatar/Avatar.less
```diff
@@ -0,0 +1,65 @@
...
+  font-weight: bold;
+  border-radius: 50%;
+  overflow: hidden;
+  user-select: none;
+
+  & &__content {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    height: calc(100% + @line-height-offset);
```
rerdos, line 27: We need to do this, as if text/emoji is passed to Avatar, then the content seems to be off.

packages/components/src/avatar/Avatar.less
```diff
@@ -0,0 +1,65 @@
+@background-color: #f2f5f7;
```
rerdos, line 1: Any idea where could I get the colour from?

Also, how would it sound to have dynamic background colours with a prop on Avatar?

packages/components/src/avatar/Avatar.less
```diff
@@ -0,0 +1,65 @@
...
+@base-diameter-sm: 3rem;
+@base-diameter-md: 4rem;
```
rerdos, line 3: Arbitrary numbers, unsure if we have defined sizes for this case already?

packages/components/src/avatar/Avatar.js
```diff
@@ -0,0 +1,36 @@
...
+import Types from 'prop-types';
+import classnames from 'classnames';
+
+import './Avatar.css';
+
+import { Size } from '../common';
+
+const BASE_CLASS_NAME = 'avatar';
+
+const Avatar = ({ size, children, outlined }) => {
```
dozortsev, line 11: I was kinda also expected this component having `url` prop for image rather then passing it via `children`

packages/components/src/avatar/Avatar.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import Types from 'prop-types';
+import classnames from 'classnames';
+
+import './Avatar.css';
+
+import { Size } from '../common';
+
+const BASE_CLASS_NAME = 'avatar';
```
andreapiras, line 9: Hey we think this may be more readable inline and also is not used very often.

packages/components/src/avatar/Avatar.js
```diff
@@ -0,0 +1,36 @@
...
+
+const Avatar = ({ size, children, outlined }) => {
+  const classNames = classnames(BASE_CLASS_NAME, `${BASE_CLASS_NAME}--${size}`, {
+    [`${BASE_CLASS_NAME}--outlined`]: outlined,
```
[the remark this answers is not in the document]
andreapiras: It's necessary because it's a dynamic property.

packages/components/src/avatar/Avatar.js
```diff
@@ -0,0 +1,36 @@
...
+  });
+
+  return (
+    <div className={classNames}>
```
andreapiras, line 17: I'd prefer to have the classes inline here to make it more readable

packages/components/src/avatar/Avatar.less
andreapiras: We still in the process of figuring out whether this variable should be coming from less or custom properties. Can you please just add a TODO with the Jira number DS-453 and we'll be happy to address it in a second iteration.

packages/components/src/avatar/Avatar.less
andreapiras: Can you please double check with the Designer. We imagine that it will be fixed size for each breakpoint

packages/components/src/avatar/Avatar.less
```diff
@@ -0,0 +1,65 @@
...
+
+@font-size-sm: @base-diameter-sm / 1.6;
```
andreapiras, line 5: Same as above, we probably have fixed font-size. Plus do we have text on avatar?

packages/components/src/avatar/Avatar.less
```diff
@@ -0,0 +1,65 @@
...
+@font-size-md: @base-diameter-md / 1.6;
+
+@avatar-base-name: avatar;
+
+@line-height-offset: 0.25rem;
+
+.@{avatar-base-name} {
+  margin: 0 auto 0.5rem auto;
```
andreapiras, line 13: Why do we have a margin hardcoded for the avatar?

packages/components/src/avatar/Avatar.less
```diff
@@ -0,0 +1,65 @@
...
+  background-color: @background-color;
+  width: @base-diameter-sm;
+  height: @base-diameter-sm;
+  font-size: @font-size-sm;
+  font-weight: bold;
```
andreapiras, line 18: Are we passing text that is not emoji to Avatar? And if so, does it have to be bold? I'd say that in case you need to pass text, you can also pass it wrapped in a `span` or any other element and add your styling via that

packages/components/src/avatar/Avatar.less
andreapiras: Can you please add the docs so we can check all this scenarios? You just need to add a code snippet here https://github.com/transferwise/neptune-web/tree/master/packages/docs/liveEditorCode and an mdx file here
https://github.com/transferwise/neptune-web/tree/master/packages/docs/pages/components

andreapiras: So we ll finish CR the css once we'll have the demo as it will help us understand better

andreapiras:
Hey @rerdos thanks for contributing, I'm happy to discuss the comment I left. It looks very good overall!! Just couple of things to discuss

sbrandwoo:
As Andrea says, would be good to have a test page. The component should support:
- Profile images
- Initials
- Emoji
- Other images
And at various sizes, with or without badges.

packages/components/src/avatar/Avatar.less
rerdos: The design document says 72px with 1px border and 3px gap between border and the content.
So that is: 72 - (2 * 1px border) - (2 * 3px gap) = 64px.

With the assumption our base font size is 16px, 64 / 16 = 4 rem.

Also I think it makes sense to scale the size with font-size, given we can have text (emojis) inside. So they would scale together, for example when we go to mobile view, where the base font size is 15px, the height & width would be still 4 times the size of the text in it.

packages/components/src/avatar/Avatar.less
rerdos: Done. 👍

packages/components/src/avatar/Avatar.less
rerdos: We do need emojis, which will come as text and affected by font-size.

packages/components/src/avatar/Avatar.js
rerdos: Yes, I agree. `url` would make sense, however we want to be able to pass text to the Avatar (for example an emoji).

So the current interface allows us to pass text, or an `` tag as children. If we pass an `` then the css rules autoscale the image to fit the Avatar component.

I've updated the storybook so now you can test how it behaves when we pass an image.

packages/docs/liveEditorCode/avatar.code.js
```diff
@@ -0,0 +1,12 @@
+<>
+  <div style={{ marginBottom: 16 }}>
+    <Avatar size={Avatar.Size.MEDIUM} outlined>
+      💰
```
andreapiras, line 4: Should we maybe update this to be 
```<span role="img" aria-label="emoji-money-icon">💰</span>```

File: packages/docs/liveEditorCode/avatar.code.js [hunk 12]
```diff
@@ -0,0 +1,12 @@
+<>
+  <div style={{ marginBottom: 16 }}>
```
andreapiras, line 2: Not essential : className="m-b-2" would give the same result :)

packages/docs/liveEditorCode/avatar.code.js
andreapiras: Not essential: I'd also add an examle with text and one without outline

packages/components/src/avatar/Avatar.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import Types from 'prop-types';
+import classnames from 'classnames';
+
+import './Avatar.css';
+
+import { Size } from '../common';
+
+const Avatar = ({ size, children, outlined }) => (
```
dozortsev, line 9: could you please expose [`className` prop](https://github.com/transferwise/neptune-web/blob/master/packages/docs/pages/rfc/0001-always-pass-classname.md)

packages/docs/liveEditorCode/avatar.code.js
```diff
@@ -0,0 +1,12 @@
...
+    <Avatar size={Avatar.Size.MEDIUM} outlined>
+      💰
+    </Avatar>
+  </div>
+  <div>
+    <Avatar size={Avatar.Size.MEDIUM} outlined>
+      <img src="http://placehold.it/1000x1000" />
```
dozortsev, line 9: please also add example with using one of [icon](https://github.com/transferwise/icons)

packages/components/src/avatar/Avatar.js
dozortsev: make sense 👍

gusso:
I think the component should occupy the same space independently if it's outlined or not. A good test for this would be that 2 avatars next to each other should look aligned. That could be either by making the outline absolutely positioned, or having the border be always present but transparent, what do you think?

gusso:
This logic is not there, is it? For web only the light inset outline is relevant, but it should be shown when you pass a full-bleed image to the Avatar.
image

packages/components/src/avatar/Avatar.less
gusso: Even if emoji is technically text, users perceive the Avatar as a graphic element, so I wouldn't scale it with text changes as it will just take more space when people are asking for text to be prioritised.

packages/docs/liveEditorCode/avatar.code.js
twrdjpalmer: One thing to note here is the `aria-label` should be human friendly, e.g., `Money bag` rather than `emoji-money-icon`

packages/components/src/avatar/Avatar.less
```diff
@@ -0,0 +1,192 @@
...
+
+@avatar-thumbnail-border-width: 1px;
+
+@avatar-outline-mask-size-lg: 4px;
+@avatar-outline-mask-size-md: 3px;
+@avatar-outline-mask-size-sm-mobile: 3px;
+@avatar-outline-mask-size-sm-desktop: 2px;
+
+.avatar {
+  display: inline-block;
```
andreapiras, line 24: Why not using display:inline-flex here and justify-content/align-item center so we can take advantage of flex centering functionalities

packages/components/src/avatar/Avatar.less
```diff
@@ -0,0 +1,192 @@
+@import 'bootstrap/less/variables.less';
+
+// TODO: https://transferwise.atlassian.net/browse/DS-453
+// Still in the process of figuring out whether this variable should be coming from less or custom properties
+@background-color: #f2f5f7;
+@font-color: #2e4369;
+
+@avatar-diameter-lg-desktop: 64px;
```
andreapiras, line 8: Nice very readable

packages/components/src/avatar/Avatar.less
```diff
@@ -0,0 +1,192 @@
...
+.avatar::before {
+  border: none;
+  position: absolute;
+  content: '';
+  border-radius: 50%;
+}
+
+.avatar--outlined {
+  &::before {
+    border: 1px solid #00b9ff;
```
andreapiras, line 136: What's the idea of applying the border to pseudo? Could we not add the border to avatar and the use a padding to give the "border-content space" effect?

andreapiras: Unless there is a particular reason, this would allow us to remove all this :after and :before styling and just manage the outline as border

packages/components/src/avatar/Avatar.less
```diff
@@ -0,0 +1,192 @@
...
+
+  &--thumbnail {
+    & * {
+      width: 100%;
+      height: 100%;
+      object-fit: cover;
+    }
+  }
+
+  &--thumbnail,
```
andreapiras, line 49: do we need thumbnail again here?

packages/components/src/avatar/Avatar.less
```diff
@@ -0,0 +1,192 @@
...
+  &.avatar--initials,
+  &.avatar--emoji {
+    color: @font-color;
+
+    // There is a problem how emojis & text vertically aligned between Windows and Mac. Setting font family to Arial seems to resolve it.
+    // Slack discussion: https://transferwise.slack.com/archives/CJNBX9CP6/p1592926116250200
+    font-family: Arial;
+  }
+
+  &--sm {
```
andreapiras, line 71: I think our component, unless specific scenarios, should not be provided with hardcoded margins. Margin is normally something related to the context in which avatar is used hence the context should place it

packages/components/src/avatar/Avatar.less
```diff
@@ -0,0 +1,192 @@
...
+
+    &.avatar--thumbnail::after {
+      width: @avatar-size;
+      height: @avatar-size;
+      transform: translateY(-@avatar-size);
+    }
+  }
+}
+
+.avatar::before {
```
andreapiras, line 127: As said above I think we can get rid of :before and : after unless I m ignoring something here

packages/components/src/avatar/Avatar.spec.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+
+import Avatar from './';
+
+describe('Avatar', () => {
+  it('renders a medium sized thumbnail Avatar with NO outline', async () => {
```
andreapiras, line 7: async await are not necessary here. I'd refactor this way:
`describe('Avatar', () => {
  it('renders a medium sized thumbnail Avatar with NO outline', () => {
    const { getByText } = render(
      <Avatar size={Avatar.Size.MEDIUM}>
        <span role="img" aria-label="Person with sunglasses emoji">
          😎
        </span>
      </Avatar>,
    );

    expect(getByText('😎').parentElement.parentElement).toHaveClass(
      'avatar avatar--thumbnail avatar--md',
    );
  });

  it('renders a small sized emoji Avatar with outline', () => {
    const { getByText } = render(
      <Avatar type={Avatar.Type.EMOJI} size={Avatar.Size.SMALL} outlined>
        <span role="img" aria-label="Money bag emoji">
          💰
        </span>
      </Avatar>,
    );

    expect(getByText('💰').parentElement.parentElement).toHaveClass(
      'avatar avatar--emoji avatar--sm avatar--outlined',
    );
  });
});`

packages/components/src/avatar/Avatar.story.js
```diff
@@ -0,0 +1,23 @@
...
+import Avatar from './Avatar';
+import { boolean, select, text } from '@storybook/addon-knobs';
+
+export default {
+  component: Avatar,
+  title: 'Avatar',
+};
+
+export const basic = () => {
+  const size = select('size', Object.values(Avatar.Size), Avatar.Size.LARGE);
```
andreapiras, line 11: Knobs are good but having all the different options explicitly next to each other can help understand quicker what the component can do

packages/components/src/overlayHeader/OverlayHeader.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import Types from 'prop-types';
 import classNames from 'classnames';
 // Avatar and Logo to be refactored and exposed DS-274 DS-275
-import Avatar from '../flowNavigation/avatar';
+import AvatarWrapper from '../flowNavigation/AvatarWrapper';
```
andreapiras, line 5: Do you mind having it lowercase?

packages/components/src/avatar/Avatar.less
rerdos: There is no immediate child to this that would benefit from this being a flex. So as far as I can see, it doesn't matter if it's a block or flex.

packages/components/src/avatar/Avatar.less
rerdos: True! At the initial phase of development, we had a different purpose & alignment for the outline, so it made sense for it to be pseudo, but by now it actually makes a lot more sense to have it as you just described. 👍 Pushed new commits where I fixed this. Very good catch. 🙌

packages/components/src/avatar/Avatar.less
rerdos: As we were using pseudo element before, we needed to have margin to be aligned with what Gusso suggested ("I think the component should occupy the same space independently if it's outlined or not.")

As we moved away from having the outline as pseudo, the margins are no longer needed. 🎉

packages/components/src/avatar/Avatar.spec.js
rerdos: Good one! Changed it. 👍

packages/components/src/overlayHeader/OverlayHeader.js
rerdos: Done!

packages/components/src/avatar/Avatar.story.js
rerdos: Changed them to radio buttons.

packages/components/src/avatar/Avatar.less
rerdos: So I moved the outline away from being pseudo. However we still need the `::after` pseudo for the thumbnail border, as it has to go over the "image" with 0.08 opacity (which we do with rgba background). See the image what Gusso posted above in this PR.

packages/components/src/avatar/Avatar.less
```diff
@@ -0,0 +1,136 @@
...
+@avatar-outline-mask-size-md: 2px;
+@avatar-outline-mask-size-sm-mobile: 2px;
+@avatar-outline-mask-size-sm-desktop: 1px;
+
+@avatar-outline-border-width: 1px;
+@avatar-outline-border-color: #00b9ff;
+
+.avatar {
+  display: inline-block;
+  text-align: left;
```
andreapiras, line 28: What is this used for?

packages/components/src/avatar/Avatar.spec.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+
+import Avatar from './';
+
+describe('Avatar', () => {
+  it('renders a medium sized thumbnail Avatar with NO outline', () => {
+    const { getByText } = render(
```
andreapiras, line 8: 😍

packages/components/src/avatar/Avatar.less
```diff
@@ -0,0 +1,136 @@
...
+
+@avatar-outline-mask-size-lg: 3px;
+@avatar-outline-mask-size-md: 2px;
+@avatar-outline-mask-size-sm-mobile: 2px;
+@avatar-outline-mask-size-sm-desktop: 1px;
+
+@avatar-outline-border-width: 1px;
+@avatar-outline-border-color: #00b9ff;
+
+.avatar {
```
andreapiras, line 26: Also wondering since we are really focused on having avatar width and height fixed no matter what border or content is applied maybe it would a good idea to use box-sizing: border-box; So once we set the width and height we don't have to stress over margin and padding affecting it WDYT?

packages/components/src/flowNavigation/FlowNavigation.js
```diff
@@ -34,7 +34,7 @@ const FlowNavigation = ({
   onGoBack,
   theme,
 }) => {
-  const avatar = done ? null : <Avatar url={avatarUrl} profileType={profileType} />;
+  const avatar = done ? null : <AvatarWrapper url={avatarUrl} profileType={profileType} />;
```
andreapiras, line 37: The Avatar is slightly smaller now (I guess it's the new dimension of Avatar) but the margin right is a bit off
image
vs live
image

packages/components/src/avatar/Avatar.less
```diff
@@ -0,0 +1,136 @@
...
+  &.avatar--thumbnail::after {
+    @thumbnail-diameter: @avatar-size - 2 * (@mask-size + 1px);
+
+    width: @thumbnail-diameter;
+    height: @thumbnail-diameter;
+    transform: translateY(-@thumbnail-diameter);
+  }
+}
+
+.avatar--thumbnail::after {
```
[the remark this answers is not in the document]
rerdos: So we need two "borders". One is the blue outline (which you can enable with `outlined` prop). Previously as you know this was `::before` pseudo, but now it has been moved (which removed like 60 lines of code 😍)

The second border is needed when the `type` is `thumbnail`. It should look something like this:

Screenshot 2020-06-25 at 09 37 58

Based on Gusso's comment, that 'border' has to be `rgba(0, 0, 0, 0.08)`. Problem with that, a simple `border` will not go 'over' the content of the box. It will just squeeze the content by 1px, and the border would just appear as light grey `border` over white background.

So we're using `::after` pseudo to 'float' it over the photo, so we have a nice 0.08 opacity border around the Avatar.

packages/components/src/avatar/Avatar.less
rerdos: As we're using `::after` pseudo for the light grey border, if an outer element has `text-align: right` on it, then it will align all the children to the right, including the `.avatar`'s `::after`. This way we can ensure the border will *always* appear at the correct place over the Avatar. 👍

packages/components/src/flowNavigation/FlowNavigation.js
rerdos: Very good catch! 👍

packages/components/src/avatar/Avatar.less
rerdos: It's already inheriting `box-sizing: border-box` from somewhere, but I don't mind specifying it just to be safe. I will add it. 👍

packages/components/src/flowNavigation/avatarWrapper/AvatarWrapper.js
```diff
@@ -0,0 +1,43 @@
+import React, { useState, useEffect } from 'react';
```
andreapiras, line 1: Something has changed in the logic here:
When selecting `FlowNavigation.ProfileType.PERSONAL` and providing empty `avatarUrl` the result is
image
vs live 
image

also avatar + close moved done by 2/3 px

packages/components/src/flowNavigation/avatarWrapper/AvatarWrapper.js
```diff
@@ -0,0 +1,43 @@
...
+export default function AvatarWrapper({ url, profileType }) {
+  const [hasImageLoadError, setImageLoadError] = useState(false);
+  const renderImage = url && !hasImageLoadError;
+  const isBusinessProfile = profileType === ProfileType.BUSINESS;
+
+  // Reset the errored state when url changes
+  useEffect(() => setImageLoadError(false), [url]);
+
+  return (
+    <div className="tw-avatar-wrapper">
```
andreapiras, line 21: These logic has been removed `<div
      className={classnames('circle circle-responsive circle-inverse', {
        'tw-avatar--with-image': !isBusinessProfile && renderImage,
      })}
    >`

packages/components/src/flowNavigation/avatarWrapper/AvatarWrapper.js
```diff
@@ -0,0 +1,43 @@
...
+      {isBusinessProfile ? (
+        <Avatar type={Avatar.Type.ICON} size={Avatar.Size.MEDIUM}>
+          <div className="icon icon-bank" />
+        </Avatar>
+      ) : (
+        <Avatar type={Avatar.Type.THUMBNAIL} size={Avatar.Size.MEDIUM}>
```
andreapiras, line 27: This has been removed <div className="tw-avatar__icon-container icon icon-profile"> and it's causing the default icon to disappeared.

packages/components/src/flowNavigation/avatarWrapper/AvatarWrapper.js
andreapiras: also the icon is no longer aligned with the cross :)

rerdos: tumblr_f08f2683afc93ad74ee43e38837bdebe_aac17afd_250
I am going to push a fix.

packages/components/src/flowNavigation/avatarWrapper/AvatarWrapper.js
rerdos: I have added this `div` to be able to apply `margin-right` to even out the spacing from the 'separator line' in the middle.

packages/components/src/flowNavigation/avatarWrapper/AvatarWrapper.js
rerdos: This is now fixed in 1a20b74.

PR #419, opened 2020-06-09
feat: add badge component

aurbano:
Demo: https://transferwise.github.io/neptune-web/branch/DS-468-badge-component/components/content/Badge

## 📎 Jira ticket
[DS-468](https://transferwise.atlassian.net/browse/DS-468)

## ❓ Context 

We want to have a Badge component in Neptune.

## 🚀 Changes

Add a Badge component

## 💬 Considerations 

The `Badge` started as a part of the [Avatar component](https://github.com/transferwise/neptune-web/pull/417), but we decided to separate it into its own component.

## ✅ Checklist


aurbano:
@andreapiras things that need clarification:

1. Tests fail because docs `import Badge from @transferwise/components` - but it won't be there until this PR is merged, so I left the docs page there until I understand how this is typically done
1. I've hardcoded a badge size in `Badge.less` - `@badge-diameter-sm: 2rem;` but I assume this should come from some global variable?
1. Right now I didn't add a `size` prop, just wanted to get some feedback on point 2. first as it kinda depends on that.

packages/components/src/badge/Badge.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import Types from 'prop-types';
+import classNames from 'classnames';
+
+import './Badge.css';
+
+const BASE_CLASS_NAME = 'tw-badge';
```
andreapiras, line 7: As in the Avatar component we feel this doesn't help readability and doesn't provide a great benefit for now, so I'd put the class inline

packages/components/src/badge/Badge.js
```diff
@@ -0,0 +1,37 @@
...
+
+const Badge = ({
+  badge,
+  className,
+  children,
+  ...rest
+}) => {
+  const classes = classNames(BASE_CLASS_NAME, className);
```
andreapiras, line 15: We do limit use of rest props to lockdown the API of the component as much as we can unless there is a particular usecase. So if there's no specific one I'd get rid of it

packages/components/src/badge/Badge.less
```diff
@@ -0,0 +1,38 @@
+@import 'bootstrap/less/variables';
+
+@badge-base-name: tw-badge;
+
+@badge-diameter-sm: 2rem;
```
andreapiras, line 5: We should reach out designers in the channel and ask to provide the exact dimension for breakpoint. we don't normally have scalable dimensions

packages/components/src/badge/Badge.less
```diff
@@ -0,0 +1,38 @@
...
+
+  & > &__content {
+    position: absolute;
+    bottom: -@badge-diameter-sm / 2;
+    right: -@badge-diameter-sm / 2;
+    width: @badge-diameter-sm;
+    height: @badge-diameter-sm;
+    display: flex;
+    justify-content: center;
+    text-align: center;
```
andreapiras, line 20: will badge contain text too? Should we also add align-items Just in case

packages/components/src/badge/Badge.spec.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import { render, cleanup } from '@testing-library/react';
```
andreapiras, line 2: Noice!

packages/docs/pages/components/Badge.mdx
```diff
@@ -0,0 +1,10 @@
+import { LiveEditorBlock, GeneratePropsTable } from '../../utils';
+import { Badge } from '@transferwise/components';
```
andreapiras, line 2: Hey this doesn't work because the export from components is missing `https://github.com/transferwise/neptune-web/blob/master/packages/components/src/index.js`. We do use `yarn workspaces` so the packages are linked locally, you don't need to publish it in order to have it available

packages/components/src/badge/Badge.spec.js
```diff
@@ -0,0 +1,38 @@
...
+
+  it('renders the badge content', () => {
+    const badgeText = 'badge-text';
+    const badge = (
+      <div>
+        {badgeText}
+      </div>
+    );
+    const { getByText } = renderBadge({ badge });
+    expect(getByText(badgeText).tagName).toBe('DIV');
```
andreapiras, line 31: maybe we can check also the class so we are sure that element is rendered inside right div. I know it won't provide much of meaningful test right now but it will prevent people changing accidentally classes etc

andreapiras:
Hey it looks very good!! Nice and clean code. Not bad for your first PR. I just left couple of comments. Once docs deploy get fixed I can also have a visual regression. Please don't forget to test on browsers

packages/docs/pages/components/Badge.mdx
aurbano: Aah makes sense! I'll fix this now then

packages/components/src/badge/Badge.less
aurbano: Makes sense! I've sent a message on #design-system

packages/components/src/badge/Badge.js
aurbano: The only additional use I can think is if people want to use a custom `style` prop, but I don't see any other components supporting it so I imagine we don't allow it.

I'll remove `rest` then

sbrandwoo:
Docs page: https://transferwise.github.io/neptune-web/branch/DS-468-badge-component/components/Badge

I believe the size is wrong, the white mask should be outside of the 16x16 icon, not inside. It'd be good to have some Avatars on the docs page too, so we can see it being used in its common use case.

packages/components/src/badge/Badge.js
```diff
@@ -0,0 +1,42 @@
...
+
+const Badge = ({ badge, className, size, outline, variant, children }) => {
+  const classes = classNames(
+    'tw-badge',
+    {
+      outline,
+      [variant]: outline,
+    },
+    className,
+    size,
```
andreapiras, line 15: I'd prefix these classes with tw-badge-sm or tw-badge-outline

packages/components/src/badge/Badge.less
```diff
@@ -0,0 +1,65 @@
...
+@badge-gap: 2px;
+@badge-border: 1px;
+@badge-border-light: rgba(255, 255, 255, 0.08);
+@badge-border-dark: rgba(0, 0, 0, 0.08);
+
+.tw-badge {
+  position: relative;
+  display: inline-block;
+
+  --badge-size: @badge-size-small;
```
andreapiras, line 15: not sure what the benefits of having custom props here are

packages/components/src/badge/Badge.js
```diff
@@ -0,0 +1,42 @@
...
+  );
+
+  return (
+    <div className={classes}>
+      <div className="tw-badge__children">{children}</div>
+      <div className="tw-badge__content">{badge}</div>
+    </div>
+  );
+};
+
```
andreapiras, line 25: ``import { Size } from '../common';
Can we please define the size as Badge.Size = Size;
and theme as Badge.Variant = {LIGHT:'light',DARK:'dark'}
```

File: packages/components/src/badge/Badge.js [hunk 11]
```diff
@@ -0,0 +1,42 @@
...
+Badge.propTypes = {
+  badge: Types.node.isRequired,
+  outline: Types.bool,
+  size: Types.oneOf(['sm', 'lg']),
```
andreapiras, line 29: Badge.Size.SMALL,Badge.Size.LARGE

andreapiras, line 29: Do we need small and large? Can small just be the default?

packages/components/src/badge/Badge.js
```diff
@@ -0,0 +1,42 @@
...
+  variant: Types.oneOf(['light', 'dark']),
```
andreapiras, line 30: Same here do we need light? can it be the default via css?

packages/components/src/badge/Badge.js
andreapiras: Actually seems that there is no class for 'sm' or 'light' so I'd just remove them

packages/docs/liveEditorCode/badge.code.js
```diff
@@ -0,0 +1,10 @@
+<Badge
+  outline
+  variant="light"
+  size="sm"
+  badge={ <div style={{ background: 'var(--color-positive)', width: '100%', height: '100%', color: 'white', fontSize: '0.8em' }}>B</div> }
```
andreapiras, line 5: Shall we reuse the same css here https://github.com/transferwise/neptune-web/pull/419/files#diff-576beb94c4575c12839f843671b06518R22

sbrandwoo:
I don't like the prop name `outline` as it does something very different to Avatar's `outlined`

aurbano:
> I don't like the prop name `outline` as it does something very different to Avatar's `outlined`

@andreapiras Thoughts/suggestions? `bordered`?

packages/components/src/badge/Badge.less
aurbano: Do you mean the `--badge-size` variable or the `position` & `display`?

The variable means that switching between small & large only requires updating the var, and the position and display are required for the positioning of the badge.

andreapiras:
> > I don't like the prop name `outline` as it does something very different to Avatar's `outlined`
> 
> @andreapiras Thoughts/suggestions? `bordered`?

withBorder? bordered?

aurbano:
@andreapiras I've combined `bordered` and `variant` making the API cleaner I believe (it didn't really make sense before, because `variant` only had the `dark` option, which would do nothing if `bordered` was false. So now with only `border=[true | light | dark]` that issue is gone.

packages/components/src/badge/Badge.less
```diff
@@ -0,0 +1,66 @@
...
```
andreapiras, line 15: Hey apparently this custom properties will get injected everytime you instantiate a component. Can we please use something else?

packages/components/src/badge/Badge.js
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import Types from 'prop-types';
+import classNames from 'classnames';
+import { Size } from '../common';
+
+import './Badge.css';
+
+const Badge = ({ badge, className, size, border, children }) => {
+  const borderVariant = typeof border === 'string' ? border : Badge.Border.LIGHT;
```
andreapiras, line 9: What's the idea behind this ?

andreapiras: maybe we can have borderInverse: Type:bool

aurbano: The idea was to have a cleaner API, before we had `bordered: boolean`, and `variant: LIGHT | DARK`, but `variant` would be ignored if `bordered` was false.

So with this we have `border: boolean | LIGHT | DARK`, with `border: true` defaulting to `LIGHT`

packages/components/src/badge/Badge.less
aurbano: Oh does that mean that if there are two badges, one small and one large they essentially overwrite each other?

I'm used to working with css modules so didn't realise this might happen here

packages/components/src/badge/Badge.js
aurbano: As discussed, change to null | LIGHT | DARK

sbrandwoo:
Please can you add an example to the demo page using an Avatar, since that will be a really common use case.

It looks like the sizes are still wrong, based on the current demo page.

andreapiras:
> Please can you add an example to the demo page using an Avatar, since that will be a really common use case.
> 
> It looks like the sizes are still wrong, based on the current demo page.
> 
> If you need some examples, here's how we currently use Avatars + Badges: https://transferwise.github.io/balance-avatar-web/

Hey Steve, my bad the demo link at the top was incorrect. Please check this one
https://transferwise.github.io/neptune-web/branch/DS-468-badge-component/components/content/Badge

sbrandwoo:
The expected sizes and positions are here:

https://www.figma.com/file/hnFCwe04Isg4c7LZkReUb8/Avatars-and-badges?node-id=256%3A94

PR #420, opened 2020-06-09
perf: remove propTypes from components and dynamic-flows build 

jackleslie:

## 📎 Jira ticket
N/A

## ❓ Context  
[This babel plugin](https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types) removes `propTypes` declarations from production builds (it was featured in [this Google Chrome list](https://github.com/GoogleChromeLabs/webpack-libs-optimizations#react) of recommendations for webpack library optimisations) and thought I'd make a PR. It's also used internally by [Next.js](https://github.com/vercel/next.js/blob/e125d905a0dd93d247c6122d349c2c90268f0713/packages/next/build/babel/preset.ts#L176) to reduce build sizes (hence why it was already in the `yarn.lock` file)

## 🚀 Changes 
Adds `babel-plugin-transform-react-remove-prop-types` plugin to `babel.config.js` in the `components` and `dynamic-flows` packages. It reduces bundles sizes (in filesystem) as follows:

### `components`
- `es/no-polyfill`: `265kb` -> `242kb`
- `es/polyfill`: `274kb` -> `251kb`
- `umd/no-polyfill`: `341kb` -> `324kb`
- `umd/polyfill`: `382kb` -> `365kb`

### `dynamic-flows`
- `es/no-polyfill`: `126kb` -> `117kb`
- `es/polyfill`: `130kb` -> `120kb`

## 💬 Considerations 
~~This only applies to functional components, class components which use `static` propTypes won't benefit from this~~. In the Google Chrome list it is flagged as "safe to use by default". ~~I didn't do any testing beyond symlinking locally and running storybook~~ I managed to get my tests running locally and fixed the failing test.

@andreapiras Apologies if this has become more effort than it's worth 😅 ~~now it doesn't apply the babel plugin to the `testConfig` envs in the babel files - is this okay? It means that the test environment deviates slightly from the production environments~~ the plugin applies to all builds, and the test referring to `propTypes` now uses `.props` (this also removed the ESLint warning for `forbid-foreign-prop-types`). It required a bit of faff in the test though as it needed to differentiate between a component selector and a wrapper, let me know if it's okay.

## ✅ Checklist

### DIFF
--- a/packages/components/babel.config.js
+++ b/packages/components/babel.config.js
@@ -19,7 +19,6 @@ const umdConfig = {
       },
     ],
   ],
-  plugins: ['transform-react-remove-prop-types'],
 };
 
 const umdConfigNoPolyfill = {
@@ -33,7 +32,6 @@ const umdConfigNoPolyfill = {
       },
     ],
   ],
-  plugins: ['transform-react-remove-prop-types'],
 };
 
 const esConfig = {
@@ -51,7 +49,6 @@ const esConfig = {
     ],
     ['minify', { builtIns: false, mangle: { exclude: { separators: true } } }],
   ],
-  plugins: ['transform-react-remove-prop-types'],
 };
 
 const esConfigNoPolyfill = {
@@ -59,7 +56,6 @@ const esConfigNoPolyfill = {
     ['@babel/preset-env', { useBuiltIns: false, modules: false }],
     ['minify', { builtIns: false, mangle: { exclude: { separators: true } } }],
   ],
-  plugins: ['transform-react-remove-prop-types'],
 };
 
 const testConfig = {
@@ -83,6 +79,7 @@ module.exports = {
     '@babel/plugin-proposal-class-properties',
     '@babel/plugin-transform-react-jsx',
     '@babel/plugin-transform-runtime',
+    'transform-react-remove-prop-types',
   ],
   env: {
     test: testConfig,
--- a/packages/components/babel.config.js
+++ b/packages/components/babel.config.js
@@ -79,6 +79,7 @@ module.exports = {
     '@babel/plugin-proposal-class-properties',
     '@babel/plugin-transform-react-jsx',
     '@babel/plugin-transform-runtime',
+    'transform-react-remove-prop-types',
   ],
   env: {
     test: testConfig,
--- a/packages/components/babel.config.js
+++ b/packages/components/babel.config.js
@@ -19,6 +19,7 @@ const umdConfig = {
       },
     ],
   ],
+  plugins: ['transform-react-remove-prop-types'],
 };
 
 const umdConfigNoPolyfill = {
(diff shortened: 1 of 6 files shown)


jackleslie:
@andreapiras should I be targetting the `develop` branch instead?

andreapiras:
> @andreapiras should I be targetting the `develop` branch instead?

Hey @jackleslie , thanks for taking time to look into that. Can you explain a bit better how that would solve the issue?

jackleslie:
@andreapiras so once `babel-plugin-transform-react-remove-prop-types` has been introduced the crux of the problem is [the line](https://github.com/transferwise/neptune-web/blob/f0bc02b052af5ef4fd626ddf82606e5b94b7ce84/packages/dynamic-flows/src/formControl/FormControl.spec.js#L484) in the function `getPropsToPassDown()` in `FormControl.spec.js` where you use `customComponent.propTypes` to get the component's props - this causes the ESLint warning [`react/forbid-foreign-prop-types`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/forbid-foreign-prop-types.md). There are two solutions (that I can see):

1. Alter the test to access props using something other than `.propTypes` - this will also remove the ESLint warning
2. Don't alter the test, and use a different Babel config/build when testing which doesn't include the plugin - this won't remove the ESLint warning

This PR does the former; by utilising that when you use JSX in the test file e.g `<Select/>` rather than just the reference e.g `Select` you can access the props with `.props`. 

I can do whichever solution you prefer, assuming you still want to utilise it - sorry this has ended up being such a pain! My local yarn setup was causing lots of issues trying to run tests locally but I think it's fine now.

andreapiras:
> @andreapiras so once `babel-plugin-transform-react-remove-prop-types` has been introduced the crux of the problem is [the line](https://github.com/transferwise/neptune-web/blob/f0bc02b052af5ef4fd626ddf82606e5b94b7ce84/packages/dynamic-flows/src/formControl/FormControl.spec.js#L484) in the function `getPropsToPassDown()` in `FormControl.spec.js` where you use `customComponent.propTypes` to get the component's props - this causes the ESLint warning [`react/forbid-foreign-prop-types`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/forbid-foreign-prop-types.md). There are two solutions (that I can see):
> 
> 1. Alter the test to access props using something other than `.propTypes` - this will also remove the ESLint warning
> 2. Don't alter the test, and use a different Babel config/build when testing which doesn't include the plugin - this won't remove the ESLint warning
> 
> This PR does the former; by utilising that when you `mount` in the test file rather than just use the reference e.g `Select` you can access the props with `.props()`.
> 
> I can do whichever solution you prefer, assuming you still want to utilise it - sorry this has ended up being such a pain! My local yarn setup was causing lots of issues trying to run tests locally but I think it's fine now.

Hey @jackleslie no worries at all!, actually thanks for contributing :). I think we'll go for the solution you implemented right now as we need to refactor these tests anyway to use React testing library. I'll create a branch with your changes and open a PR so we can run CI on it.I'll reference this PR

andreapiras:
Merged via https://github.com/transferwise/neptune-web/pull/432

PR #425, opened 2020-06-11
chore: add lint to docs

andreapiras:
## ❓ Context  

Lint is now running via jest. Added jest lint runner to docs
## ✅ Checklist


packages/docs/pages/_app.js
```diff
@@ -1,4 +1,6 @@
 /* eslint-disable react/jsx-props-no-spreading */
+/* eslint-disable import/no-extraneous-dependencies */
```
andreapiras, line 2: 'react-axe' should be listed in docs dependency but we have it in root

PR #426, opened 2020-06-12
fix: WCAG AA compliant color for grey backgrounds

nabaroa:
## 📎 Jira ticket: https://transferwise.atlassian.net/browse/DS-471

imagen

The .bg-default background and colour contrast is not enough for AA in normal text size in CSS core.
(I am aware that there is an accessibility colour task going on, but that could take some time, and the intention of this fix is going to solve only this particular use case in web, so we can relese the new css core without this issue).

Here conversation thread for context: https://transferwise.slack.com/archives/GHW2Y4B5H/p1591030310007400 

**Before**

imagen

**After**

imagen

## ✅ Checklist


packages/css/src/less/addons/_background-utilities.less
```diff
@@ -3,7 +3,10 @@
 // Contextual backgrounds
 
 .bg-default {
+  --color-text-secondary: #4a5860; /* Temporary WCAG AA compliant color proposal while getting accesible color palette. Related ticket: https://transferwise.atlassian.net/browse/DS-471 */
```
anttio, line 6: Is there a reason of introducing new colour and not using something from our existing colour palette? How would Grey Dark from base colours work here?

nabaroa: Yes, you're totally right. Makes a lot of sense. I was using the former web color we had for this cases, sorry. 

imagen

imagen

PR #429, opened 2020-06-16
feat: add responsive features to image

andreapiras:

## 📎 Jira ticket
https://transferwise.atlassian.net/browse/DS-465

## ❓ Context  
We want to add responsive features to normal image so they'll have the ability to stretch and shrink based on container

## 🚀 Changes 
Add stretch and shrink props 

## ✅ Checklist


packages/components/src/image/Image.js
```diff
@@ -19,7 +21,7 @@ import { useHasIntersected } from '../common/hooks';
 export const EmptyTransparentImage =
   'data:image/gif;base64,R0lGODlhAQABAAAAACH5BAEKAAEALAAAAAABAAEAAAICTAEAOw==';
 
-const Image = ({ id, src, alt, onLoad, onError, className, loading }) => {
+const Image = ({ id, src, alt, onLoad, onError, className, loading, stretch, shrink }) => {
```
helendurranttw, line 24: What do you think about combining stretch and shrink into one prop? I could imagine we might want to add more fancy ways of behaving in the future. We could have a prop called `responsive`? It could be an object with stretch and shrink props?

packages/docs/pages/components/Image.mdx
```diff
@@ -7,37 +7,68 @@ Lazy loading functionality is currently not supported on IE11. If you are need t
 
 <LiveEditorBlock code={code} scope={{ Image }} />
 <GeneratePropsTable componentName="Image" />
-<>
+
+
```
helendurranttw, line 11: Thanks for the great doc updates 😀

packages/components/src/image/Image.less
```diff
@@ -0,0 +1,11 @@
+.tw-image {
+  min-width: fit-content;
+  width: auto;
+  &__shrink {
+    max-width: 100%;
+    min-width: auto;
+  }
+  &__stretch {
```
sergonius, line 8: Do we have a utility class that could be used instead of a new prop for this?

packages/components/src/image/Image.less
```diff
@@ -0,0 +1,11 @@
+.tw-image {
+  min-width: fit-content;
+  width: auto;
+  &__shrink {
+    max-width: 100%;
```
sergonius, line 5: Is this not a "good practice" for default image styles?

packages/components/src/image/Image.js
StevePole: Andrea and I were chatting about this.  The use case I had in mind, was something like out 'success screen', where we typically have an image (an illustration) that is centred.  On desktop you're probably happy for it to be it's native size, so you get pixel perfection.  But you would want to make sure it didn't overflow it's container (or the screen) on a narrow portrait phone.  So `<Image shrink={true} stretch={false} ... />`

The alternative is just to go for the container size 100% of the time, using that to dictate width on desktop.  I think that's normally ok as well, although you could still do it with `<Image shrink={true} stretch={true} ... />`

We also debated default behaviours.  I feel like shrink=true is something you would almost always want, hence might make sense as a default, but I always think the absence of something should indicate false e.g. `<Image ... />` should equal `<Image shrink={false} ... />` _not_ `<Image shrink={true} ... />`.  Obviously if you can find a suitable antonym that can still work, but I couldn't...

Can you think of any other 'fancy' ways we might want to size?

packages/components/src/image/Image.less
andreapiras: I'm up to discuss that but I feel that by default an Image should not be doing anything different from an  tag.

packages/components/src/image/Image.less
andreapiras: I guess we can use/create a utility class, the convenience here is that people don't have to know the class name but they can just play with the interface without have to know about a class stored somewhere

packages/components/src/image/Image.less
```diff
@@ -0,0 +1,11 @@
+.tw-image {
+  min-width: fit-content;
```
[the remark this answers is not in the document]
andreapiras: Actually good spot, this should have been `min-content` to prevent the shrink when stretch is true and `width: 100%;`

packages/components/src/image/Image.less
sergonius: Well, do we ever want images to break out of the screen or their containers?

andreapiras: I don't think we'll have cases in our website,in fact I'm favour of having shrink set by default to true. My argument is more about since we are returning an img tag that should behave like a native img tag where possible.

packages/components/src/image/Image.js
helendurranttw: Hey @StevePole. I've had a chat with Andrea about this. I was in favour at first of a simple "always stretch to the bounds of your container" approach, but after discussion I've come around to shrink / stretch + sensible defaults.

The `fancy` behaviours I was thinking was around object-fit - i.e changing between cover / contain / fit to change the behaviour of images on stretch. If you were working with a fixed height, you may want to choose between stretching, cropping etc depending on the image.

I'd like to keep the API as clean as possible, so I worry about the need in the future to just add more and more props. I like the idea of containing some configuration properties inside an object.. but this isn't really a pattern we follow elsewhere (at least not yet). For example, here we could have something like `sizingOpts` or `responsiveConfig` to contain these options. 

But I'm not 100% sure. What do you think?

StevePole: It depends if you believe this component is going to be used for background images, or for images where we don't care about maintaining aspect ratio.

This approach makes no allowance for height constraints or aspect ratio control, but _typically_ I think that's most relevant for backgrounds. If, for example, we wanted to implement a traditional landing page hero unit with copy over the top, you'd probably want that height bound.  That's where 'cover' becomes useful, but I think we'd implement that through a different component.  Maybe a `<BackgroundImage src={} fit="cover"> { your content here } </BackgroundImage>` or something.

PR #431, opened 2020-06-17
fix: change dates in the date lookup to use buttons

helendurranttw:
## 🖼 Context

The empty hrefs in the anchors used inside the date lookup component were causing issues inside a new transfer flow screen. We should be using buttons anyway, as they're better for a11y.

## 🚀 Changes

* Changed anchors in the calendar to buttons
* Removed the use of the `table-calendar` class, and deprecated that class in Neptune CSS (it is only used here and in `styleguide-components`, and should never be used in isolation)

**Unrelated**

* Fixed some prop-type errors in storybook

## 🤔 Considerations

There are some dubious a11y things happening here - you can tab to the header and use the down key to open the calendar, but a screen reader would not be able to instruct the user to do this, as the element is a div. This needs fixing but was beyond the scope of this ticket. 

## ✅ Checklist


packages/components/src/dateLookup/header/Header.js
```diff
@@ -7,6 +7,7 @@ const Header = ({ label, onLabelClick, onPreviousClick, onNextClick }) => (
   <div className="text-xs-center p-t-1 p-b-2 clearfix">
     <div className="pull-xs-left">
       <button onClick={onPreviousClick} className={btnClasses}>
+        <span className="sr-only">Back in time</span>
```
andreapiras, line 10: Where is this copy coming from? Should this be translatable?

packages/components/src/dateLookup/header/Header.js
```diff
@@ -17,6 +18,7 @@ const Header = ({ label, onLabelClick, onPreviousClick, onNextClick }) => (
     )}
     <div className="pull-xs-right">
       <button onClick={onNextClick} className={btnClasses}>
+        <span className="sr-only">Forward in time</span>
```
andreapiras, line 21: Same as above

packages/components/src/dateLookup/DateLookup.less
```diff
@@ -0,0 +1,39 @@
...
+  > tbody > tr > td {
+    padding: @space-4;
+
+    button {
+      width: 100%;
+      padding: @space-4 0;
+
+      color: @color-base-blue-light;
+      border: transparent;
+      border-radius: 3px;
```
andreapiras, line 19: We do a custom property for this https://github.com/transferwise/neptune-web/blob/master/packages/css/src/less/neptune-tokens.less

packages/components/src/dateLookup/DateLookup.less
```diff
@@ -0,0 +1,39 @@
...
+  // SIIIIIIIGGGGHHHHH - blame Bootstrap
+  > tbody > tr > td {
+    padding: @space-4;
+
+    button {
+      width: 100%;
+      padding: @space-4 0;
+
+      color: @color-base-blue-light;
+      border: transparent;
```
andreapiras, line 18: interested in understanding transparent vs none, is it to keep the width of the button?

packages/components/src/dateLookup/header/Header.js
helendurranttw: Let's add them as part of the a11y ticket.

packages/components/src/dateLookup/DateLookup.less
helendurranttw: As of the conversation yesterday, I'm not going to use custom props in here (because there's no guarantee they'd be on the page). I'll replace it when we've solved the custom props issue.

packages/components/src/dateLookup/DateLookup.less
helendurranttw: Yep :)

PR #435, opened 2020-06-18
fix: reduce margin for horizontal definition lists

luujam2:
## 📎 Jira ticket
https://transferwise.atlassian.net/browse/HT-534

## ❓ Context  
We want to align the style in `DefinitionList` with the review card in the transfer flow. Currently we use it in `rate-review-flow` but there's too much spacing in between the rows.

Rate review flow (using existing `DefinitionList`):
image-20200504-104116

Transfer flow (not using neptune):
image-20200504-104131

## 🚀 Changes 
- Reduce margin bottom for horizontal `dd` elements on screen sizes larger than `screen-sm` var

## 💬 Considerations 
- I've made a change to storybook config to point to locally built css bundle instead of bootstrap
- Not sure whether the changes to storybook should be in a separate commit

Before(larger than screen-sm):
Screenshot 2020-06-18 at 11 56 45

After(larger than screen-sm):
Screenshot 2020-06-18 at 11 54 30

No difference for smaller screens:
Before(smaller than screen-sm):
Screenshot 2020-06-18 at 11 55 09

After(smaller than screen-sm):
Screenshot 2020-06-18 at 11 56 51

## ✅ Checklist


helendurranttw:
Hey @luujam2 - @anttio suggested trying 12px instead of 8px - he thought it looked a little cramped. I also saw Steve made a comment on the original thread about in vertical orientation it needing to be the opposite. Do you need to make changes here for vertical layouts?

luujam2:
Thanks for the response, @helendurranttw This is what the 12px spacing looks like:

Screenshot 2020-06-19 at 13 33 22
Screenshot 2020-06-19 at 13 36 36

As for the vertical layouts, I've not touched the styles for them and they should remain the same. If the 12px spacing is fine, I can go ahead and push those changes, cc @mrmurtz @anttio

mrmurtz:
checked in with @monicacru @twlleon - we think it still feels likes 8px gives us the best use of space. The review card often needs to fit a fair amount of info and we want to avoid scrolling for info as much as possible.

gusso:
Screenshot-2020-06-24-at-10 59 34
For reference, this is what it look like applied with 8px, which is what we currently have in the Send flow. I'd follow the current application, so I'm leaning towards 8 as well, I just want to make sure we're not making another part of the product feel cramped.

@luujam2 Did you find any instances of the component being used elsewhere so we know how it will affect the rest of the product?

luujam2:
@gusso I'm not sure where the `DefinitionList` pattern is used outside of this particular use case. We use the pattern in the context of a review card but it could be the case that `DefinitionList`s could be used for something else?

luujam2:
Some use cases of `DefinitionList`. These use the default `VERTICAL_TWO_COLUMN` layout which won't be affected by the changes. The only usage I see of `HORIZONTAL_JUSTIFIED` layout is in our use case (`rate-review-flow`) and a component `DynamicFieldDefinitionList` which isn't used in any other repos.

Screenshot 2020-07-01 at 13 59 25 

Screenshot 2020-07-01 at 13 42 25

PR #437, opened 2020-06-18
Ds 425 field refactor

andreapiras:

## ✅ Checklist


packages/dynamic-flows/src/field/Field.new.js
```diff
@@ -0,0 +1,214 @@
...
+    >
+      {label && (
+        <label className="control-label" htmlFor={id}>
+          {label}
+        </label>
+      )}
+      {React.cloneElement(children, fieldProps)}
+      {showAlert && (
+        <Alert
+          type={alertContent.type}
```
StevePole, line 179: Would be good to think about the accessibility of this content.  Could we use https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute to make sure the control references this? 

Also is there a way to use aria-invalid to convey when we're in an error state? 
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-invalid_attribute

packages/dynamic-flows/src/field/Field.new.js
```diff
@@ -0,0 +1,214 @@
...
+import classNames from 'classnames';
+import { Alert } from '@transferwise/components';
+import { FormControlType } from '../common';
+import { getValidationFailures } from '../common/validation/validation-failures';
+
+// @TODO translations for default?
+const DefaultValidationMessages = {
+  Required: 'Required',
+  pattern: 'Incorrect format',
+  minLenght: 'The value is too short',
```
StevePole, line 13: typo

packages/dynamic-flows/src/field/Field.new.js
```diff
@@ -0,0 +1,214 @@
...
+
+  const validate = (newValue) => {
+    // If no type of validation is specified don't validate.
+    if (!field.validation.type) {
+      return null;
+    }
+
+    const { validation } = field;
+
+    const schema = { type: validation.type };
```
StevePole, line 111: I don't really understand what's happening in this method.  I you want to know which validation messages to show, you need the schema that contains the rules.  You can't really just 'make' a schema that only contains the type, you need minimum/maximum/minLength/maxLength/pattern etc.

packages/dynamic-flows/src/field/Field.new.js
```diff
@@ -0,0 +1,214 @@
...
+      const alertInfo = getAlertInfo();
+      if (alertInfo) {
+        setAlertContent(alertInfo);
+        setShowAlert(true);
+      }
+    } else if (!isValueValid) {
+      const validationMessages = validate(value);
+      if (validationMessages) {
+        setAlertContent(getAlertError(validationMessages));
+        setShowAlert(true);
```
StevePole, line 47: The logic for showing validation should be:

messages.length > 0 && (formSubmitted || (valueChanged && controlBlurred))

packages/docs/liveEditorCode/field.code.js
```diff
@@ -1,46 +1,30 @@
...
-        validationMessages: {
-          required: 'Number is required',
-          minimum: 'Must be 10 or greater',
-          maximum: 'Must be 99 or less',
+        help: (
+          <InstructionsList dos={['good', 'much better']} donts={['uhm not so good', 'erm no!']} />
+        ),
+        validation: {
+          type: 'string',
+          minLength: { value: 10, message: 'Must be 10 or greater' },
```
StevePole, line 15: Just spotted this, which helps explain the validate method.  This in reinventing what JSON schema already does with a slightly different syntax.  Not sure that's a good idea.

StevePole:
I'm not sure about the responsibilities we're giving it.  For validation we're reinventing what JSON schema does with a very slightly different interface.  But if we're going to use a language to tell the fields what to do, we may as well bite the bullet and use JSON schema, but then we're just reimplementing the dynamic form version.  Alternatively, we could delegate validation logic to the consumer, just allowing them to pass a list of validation messages, and negating the need for the validation language inside field

Then you need to decide if the field should know about interaction states, should the field decide when to show the validation messages based on when the user has interacted with the control or submitted the form.  Or should that also be the preserve of the consumer?  i.e. just pass in the messages when relevant.  I'm not sure what's best, but validation also needs to understand when the form is submitted so that's something to consider.

Basically it feels to me like a if we want a component that does everything for us, we need a way to tell it how to behave.  But we already have that through JSON schema, and we already have a component that implements it.   I'm not sure there's a point it having another component that's so similar.  I see a role for a 'dumber' less stateful component to own the visual logic of how we lay out fields and associated messaging.  That component could even be consumed by the dynamic form variant, leaving the dynamic form variant in charge of validation logic as it already understands that, but taking over control of presentation logic.

StevePole:
Something a bit more like this
```
<Field 
  label={...} 
  hasFocus={true}
  hasError={false}
  errorMessages={[...]}
  helpMessages={[...]}>
  <input ... />
</Field>
```

StevePole:
I'm also thinking of cases where we have to make an async request to validate a Field.  Not sure how you would facilitate that if validation is all internal?

andreapiras:
> of the consumer

Hey @StevePole, thanks for your feedback.
- So this component is meant to cover those scenarios where people need some fields with validation (sometimes just one) and don't want to use JsonSchemaForm which is very heavy as it uses under the hood FormControl which imports a lot of  components.
- The validation API is up to discussion but I find it cleaner, rather than have to match error and errorMessage via string that can lead to unwanted errors. This component shares the validation files with JsonSchema so is not reimplementing them.
- I think if we go for this implementation we can reuse it inside jsonSchema  [here]([https://github.com/transferwise/neptune-web/blob/master/packages/dynamic-flows/src/jsonSchemaForm/basicTypeSchema/BasicTypeSchema.js#L92) to take care about the visual
We can have a chat about it if you want

StevePole:
I see the value in passing the component, but not so much the interface change.

In JSON schema you define rules like this
```
{
  type: "string",
  minLength: 10,
  maxLength: 20,
  pattern: "^[A-Z]$"
}
```
In our dynamic form system we allow you to also supply validation messages relating to these:
```
{
  type: "string",
  minLength: 10,
  maxLength: 20,
  pattern: "^[A-Z]$",
  validationMessages: {
    minLength: "The value must be at least 10 characters",
    minLength: "The value can be no more than 20 characters",
    pattern: "Capital letters only",
  }
}
```
But you're changing this to
```
{
  type: 'string',
  minLength: { value: 10, message: 'The value must be at least 10 characters' },
  maxLength: { value: 20, message: 'The value can be no more than 20 characters' },
  pattern: { value: "^[A-Z]$", message: 'Capital letters only' },
}
```
It's not simpler, nor semantically different, it's just a different structure for the same information.   By changing the format you're making it difficult to use the validation library, you'll have to convert it back to JSON schema to get it to work.  By all means don't adopt the whole of json schema, but you could an interface including `validation-schema' and 'validation-messages' or similar

I would take a look at how it's done in the ControlFeedback component.  You're going to need something very similar to this, as presenting error/validation/help information also has logic:

https://github.com/transferwise/neptune-web/blob/master/packages/dynamic-flows/src/jsonSchemaForm/controlFeedback/ControlFeedback.js

The logic is all documented in the tests: 
https://github.com/transferwise/neptune-web/blob/master/packages/dynamic-flows/src/jsonSchemaForm/controlFeedback/ControlFeedback.spec.js

An explicitly passed error message (thanks to feedback from a form submission or asynchronous validation) takes precedence over client side validation, which takes precedence over help information.  In this component we pass in the interaction state, the available messages, and the currently failing validation keys, and let the component control the logic of what to show.  The interface is like this because the parent also needs to know if it's in an error state so needs access to the validation keys and interaction state, we didn't want to calculate validation state twice, but it greatly cleaned up the parent by delegating the presentation logic.

That's consumed by the basicTypeSchema:
https://github.com/transferwise/neptune-web/blob/master/packages/dynamic-flows/src/jsonSchemaForm/basicTypeSchema/BasicTypeSchema.js#L106

Basic type schema should end up using your Field, but you either need the flexibility to be able to receive manual feedback in addition to validation and the associated logic, or you need to give the power to the consumer to control what messages are shown. 

If you want it to control all the messaging you need to replicate the logic around form submitted state as well as field interactions state, showing manual errors vs validation errors vs help and when to show each. I'd refer to the dynamic forms implementation/tests to understand what needs to be considered.

StevePole:
This is an example of the feedback logic in action.  We start at the point we've just submitted a form and received back an error that relates to this field, something that was manually returned from the server.  

Once we change the value that error is no longer relevant, because it refers to an old value.  So then we carry on with real time validation, and we highlight that the value is too short.  

Once the validation problem has been addressed we can go back to showing the help text.  All 3 of these were eligible to be shown from the start, but there's a hierarchy of importance.

feedback-logic

StevePole:
This one is the validation logic.  This shows a fresh form.  Initially, as we type the first character, we don't immediately see an error messages as soon, even though it's invalid, as that would be really annoying.  We stay on the help text.  

Once the user blurs the control, we start validating, and show a message that it's too short.  (Note: it also doesn't start validating if you focus then immediately blur without interacting).

When the user returns focus to fix it, we start validating in real time to help them correct the problem.

validation-logic

andreapiras:
Hey @StevePole, I took into account your comments and implement a version reusing ControlFeedback and BasicSchema.
https://github.com/transferwise/neptune-web/pull/437/files#diff-d872a9a283bc62689fcf557ac1b8e15cR1.
The implementation of BasicSchema should change to become
https://github.com/transferwise/neptune-web/pull/437/files#diff-e68550a2ecdf496fc6ebe1f564d34ee5R10
It's still a heavy WIP but what do you think about this implementation

packages/dynamic-flows/src/field/Field.BasicShema.js
```diff
@@ -0,0 +1,184 @@
...
+const Field = (props) => {
+  const [id, setId] = useState('');
+  const [model, setModel] = useState(props.model);
+  const [lastModel, setLastModel] = useState(props.model);
+  const [changed, setChanged] = useState(false);
+  const [focused, setFocused] = useState(false);
+  const [blurred, setBlurred] = useState(false);
+  const [validations, setValidations] = useState([]);
+
+  const getValueFromEmitted = (event) => {
```
StevePole, line 19: This method feels quite fragile.  The 'else' clauses feel like they making some assumptions e.g. if the type is not object we're dealing with a checkbox?  That doesn't seem to follow...

It looks like you're detecting if this is a native JS event or something generated by our components, perhaps you can do that with something like this instead
```
if (event instanceOf Event) {
  ...
}
```
Then it may also be useful to use our utility type function to know how to treat different data type e.g.
```
if (isNumber(event)) {
   ... 
} else if (isBoolean(event)) {
   ...
} else if (isString(event)) {

}
```
That will save you having to do thing like '=== 0' checks

packages/dynamic-flows/src/field/Field.BasicShema.js
```diff
@@ -0,0 +1,184 @@
...
+    let newValue;
+
+    if (event && typeof event === 'object') {
+      if (event.target) {
+        // This is a SyntheticEvent coming from React
+        // Input type number target value is a string and needs to be a number.
+        if (props.schema.type === 'number') {
```
StevePole, line 26: Within JSON schema you also have integer types.

packages/dynamic-flows/src/field/controlFeedback/ControlFeedback.js
```diff
@@ -0,0 +1,88 @@
+import React from 'react';
+import Types from 'prop-types';
+
+import { Alert } from '@transferwise/components';
+
+const ControlFeedback = (props) => {
+  // Use validation messages from the schema if possible.
+  const validationMessages = {
```
StevePole, line 8: Maybe we should do this at a higher level?  Previously I had considered this component to just be an internal implementation detail of basic schema.  But it could be something more generalised that is not "schema aware"

That is, it could just accept one list of validation messages, one list of error message etc. Letting the parent component compile the list from the two possible sources.  The help messages could also be a top level prop, rather than something we pluck out of the schema.

packages/dynamic-flows/src/field/Field.BasicShema.js
```diff
@@ -0,0 +1,184 @@
...
+
+    setId(generateId());
+  };
+
+  const onModelChange = () => {
+    setValidations(getValidationKeys(model));
+  };
+
+  const isConst = props.schema.const || (props.schema.enum && props.schema.enum.length === 1);
+  const isHidden = props.schema.hidden || isConst;
```
StevePole, line 107: Think this should probably be delegated to the parent.  So BasicTypeSchema would just choose not render a Field if the schema should be hidden.  Similarly, the 'const' handling (broadcasting the new value when receiving a const schema) might make more sense in the parent.  If you were using a field in isolation, it would be weird to pass in a const to get it to broadcast and not render anything.

packages/dynamic-flows/src/field/Field.story.js
```diff
@@ -23,47 +26,86 @@ export const basic = () => {
...
+        <input type="text" value="default" />
+      </FieldNew> */}
+      <Field
+        model="intialValue"
+        schema={{
+          title: 'title',
+          type: 'string',
+          minLength: 3,
+          maxLength: 3,
+          validationMessages: {
```
StevePole, line 96: I think probably worth having a separate prop for these. In dynamic forms there are potentially a global set of generic errors messages, and more specific ones for a field.  However, maybe the consumer can deal with that and just pass in a separate 'validation-messages' prop to this component?

packages/dynamic-flows/src/field/Field.BasicShema.js
```diff
@@ -0,0 +1,184 @@
...
+    'has-error':
+      (!changed && props.errors) ||
+      ((props.submitted || (changed && blurred)) && validations.length),
+    'has-info': focused && props.schema.help,
+  };
+
+  const showLabel = props.schema.format !== 'file' && props.schema.type !== 'boolean';
+  const fieldProps = {
+    className: 'form-control',
+    value: model,
```
StevePole, line 123: Does this make the assumption that all component use 'value' as their prop name?  There are cases where that is not true, though it would be good to standardise.

image

image

image

packages/dynamic-flows/src/field/Field.BasicShema.js
```diff
@@ -0,0 +1,184 @@
...
+    schema: props.schema,
```
StevePole, line 124: Do we need to pass the schema to the child component?

packages/dynamic-flows/src/field/Field.BasicShema.js
andreapiras: Hey @StevePole, you are correct  is not safe to assume that they all use value so this needs to be looked at

packages/dynamic-flows/src/field/Field.BasicShema.js
andreapiras: It seems that SchemaFormcontrol neeeds it. But I'm going to rework this completely and figure out for alternatives

packages/dynamic-flows/src/field/controlFeedback/ControlFeedback.js
andreapiras: Yes, I totally agree, I'd make this component as dumb as possible

packages/dynamic-flows/src/field/Field.BasicShema.js
StevePole: ah, yeah, that makes sense.  In the regular case you wouldn't need it, but for that it does.  Potentially the responsibilities of SchemaFormControl could be passed over to BasicTypeSchema, which could use a switch to pass down the right component to the field.  Only thing I'm wary of there is how much logic it contains.

packages/dynamic-flows/src/field/Field.BasicShema.js
andreapiras: I copied this method from FormControl, we need to change it there too

PR #438, opened 2020-06-19
feat: improve NavigationOption semantics

twrdjpalmer:

## 📎 Jira ticket

N/A

## ❓ Context  

By default, the `NavigationOption` component renders something like this:

```jsx
<label onClick={() => {}}>
  <h5 />
  <button />
</label>
```

The header element is redundant, and I assume only for styling. Having a button element nested inside a label, or an anchor (when the `href` prop is supplied) is a potential accessibility concern.

## 🚀 Changes 

1. Updates `NavigationOption` to render a `button` by default. This is to demonstrate to the user that the element is clickable.
2. Replaces the `h5` with a div with the same styles applied
3. Replaces the `button` with a `span` so the icon still renders

```jsx
<button|a|as onClick={() => {}}>
  <div />
  <span />
</button|a|as>
```

## 💬 Considerations 

How should we handle default focus states?

## ✅ Checklist


StevePole:
If optimising for accessibility I would also consider if each option (there are typically multiple) should be an entry in a list.

packages/components/src/common/Option/Option.js
```diff
@@ -38,7 +38,7 @@ const Option = ({
       </div>
     )}
     <div className="media-body">
-      <h5>{title}</h5>
+      <div className="h5">{title}</div>
```
andreapiras, line 41: Can we use `<Element` and `as` prop` here

twrdjpalmer: Element and `as` are used at the top level. Do you want something like `titleAs`? What cases are we catering for here?

twrdjpalmer:
@StevePole Companion component makes sense. Can add that in a separate PR 👍

packages/components/src/common/Option/Option.js
andreapiras: Yes actually you are right, let's leave this way I was thinking about maybe some SEO benefits but the whole thing is wrapped in Element so no point of adding here too

packages/components/src/navigationOption/NavigationOption.spec.js
```diff
@@ -40,6 +40,12 @@ describe('Navigation option', () => {
     expect(option().prop('as')).toBe('a');
   });
 
+  it('tells option to render as the `as` proper when it is passed', () => {
```
fergusjordantw, line 43:
```suggestion
  it('tells option to render as the `as` prop when it is passed', () => {
```

PR #440, opened 2020-06-19
fix: styles for public-navigation

dozortsev:
## ❓ Context  

This fix fixes this visual bugs on [public-navigation with new icons](https://transferwise.github.io/public-navigation/branch/migrate-on-new-icons/index.html) ([PR](https://github.com/transferwise/public-navigation/pull/153)).

## ✅ Checklist


packages/css/src/less/link-callout.less
```diff
@@ -1,17 +1,14 @@
...
   -webkit-font-smoothing: antialiased;
   font-size: 10px;
   margin-left: 8px;
-  transition: transform 0.2s ease-in-out;
-  transform: none;
   pointer-events: none;
-  color: @link-color;
 }
 
 .link-callout:hover::after,
```
domhorvathtw, line 14: what about these styles?  why are we removing the transform transition completely?

dozortsev: > what about these styles?

I'm not sure I guess they can be deleted

> ..why are we removing the transform transition completely?

Think we don't need animation when you on hover of the chevron, see this animation on prod instance https://transferwise.github.io/public-navigation/
Screen Shot 2020-06-22 at 7 59 53 PM

packages/css/src/less/link-callout.less
```diff
@@ -1,17 +1,14 @@
 @import (reference) 'bootstrap/less/variables.less';
 
-.link-callout::after {
+.link-callout .tw-icon {
   display: inline-block;
   font-style: normal;
   font-weight: normal;
   line-height: 1;
   -webkit-font-smoothing: antialiased;
   font-size: 10px;
```
domhorvathtw, line 9: I think these styles won't be needed, since it's an SVG now, not an icon font

packages/css/src/less/link-callout.less
domhorvathtw: I think currently that's the only hover effect for those links. I would think twice before we remove it

packages/css/src/less/link-callout.less
dozortsev: yeah, ok, you mean?

```css
  font-style: normal;
  font-weight: normal;
  line-height: 1;
  -webkit-font-smoothing: antialiased;
  font-size: 10px;
```

domhorvathtw: yeap, I think anything related to font styling is not needed anymore, but please check it visually

dozortsev: done

packages/css/src/less/icons.less
```diff
@@ -1,18 +0,0 @@
-// TODO: remove bootstrap dependencies
-
-@import './variables/_spacing.less';
```
dozortsev, line 3: this fixes visual bug in user-app-shell

Screen Shot 2020-06-24 at 12 50 16 PM

packages/css/src/less/link-callout.less
```diff
@@ -1,22 +1,14 @@
...
-.link-callout:hover::after,
-.link-callout:focus::after,
-.callout-container:hover .link-callout::after,
-.callout-container:focus .link-callout::after {
-  transform: translateX(4px);
+.link-callout:hover > .tw-icon-chevron-up,
+.link-callout:focus > .tw-icon-chevron-up,
+.callout-container:hover .link-callout > .tw-icon-chevron-up,
+.callout-container:focus .link-callout > .tw-icon-chevron-up {
+  transform: translateX(4px) rotate(+90deg);
```
dozortsev, line 13: fix for 

Screen Shot 2020-06-22 at 7 54 19 PM

PR #443, opened 2020-06-23
feat: allow NavigationOption to render illustrations

twrdjpalmer:

## 🖼 Context

We'd like to use the [NavigationOption component with our illustrations](https://www.figma.com/file/LNyDzbKAdYxHqT4NopHhZV/Intent-picker?node-id=2255%3A800). Currently the NavigationOption makes an assumption that the media should be rendered within a circle.

Depends on #438 

## 🚀 Changes

1. Add a new prop to turn off the circle background
2. Add a new prop to enable showing media at all sizes

Screenshot 2020-06-23 at 10 38 11

Screenshot 2020-06-23 at 10 38 41

## 🤔 Considerations

When using an illustration that is taller than `48px`, the padding becomes overly large. See this [Slack thread for more context](https://transferwise.slack.com/archives/CJNBX9CP6/p1592839932238700). I'm happy to address this in a separate PR.

## ✅ Checklist


packages/components/src/navigationOption/NavigationOption.spec.js
```diff
@@ -40,6 +41,12 @@ describe('Navigation option', () => {
     expect(option().prop('as')).toBe('a');
   });
 
+  it('tells option to render as the `as` proper when it is passed', () => {
```
fergusjordantw, line 44:
```suggestion
  it('tells option to render as the `as` prop when it is passed', () => {
```

andreapiras:
I would be cool if you can also add some documentation in the docs

PR #448, opened 2020-06-24
feat: add dynamic layout and associated components

StevePole:
## 🖼 Context
The needs and specification are discussed here
https://transferwise.atlassian.net/wiki/spaces/DF/pages/1303972575/Dynamic+layout

## 🚀 Changes
Added a number of components to allow a flow page layout to be specified programmatically.  Forms, decisions and buttons can all be laid out using this system.

## 🤔 Considerations

I'm opening this just to see visualise the diff, I may break up into smaller pieces.

## ✅ Checklist


packages/dynamic-flows/src/layout/DynamicLayout.e2e.spec.js
```diff
@@ -0,0 +1,80 @@
...
+        schema.properties.an,
+      );
+    });
+  });
+
+  describe('when a button element is clicked', () => {
+    beforeEach(() => {
+      component.find('button').simulate('click');
+    });
+    it('should broadcast onModelChange', () => {
```
StevePole, line 76: Whoops, will fix.

packages/dynamic-flows/src/layout/list/index.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
```
StevePole, line 1: I'm going to remove this file.  Not ye convinced about the direction of this component.

packages/dynamic-flows/src/layout/columns/spec.js
```diff
@@ -0,0 +1,97 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import DynamicColumns from '.';
+import DynamicLayout from '../DynamicLayout';
+
+describe('Given a component for dynamically rendering image', () => {
```
helendurranttw, line 7: *columns

packages/dynamic-flows/src/layout/heading/index.js
```diff
@@ -0,0 +1,36 @@
...
+import { getTextClasses } from '../utils';
+
+const DynamicHeading = (props) => {
+  const heading = props.component;
+
+  const classes = getTextClasses(heading);
+
+  switch (heading.size) {
+    case 'xs':
+      return <h5 className={classes}> {heading.text} </h5>;
```
helendurranttw, line 14: Hmm.. this is a bit tricky. To make the page semantically correct we'd want to ensure the right hierarchy of <h*> elements, and this is always tricky in a dynamic context. Perhaps we could either get more specific (specify h* instead of size), or cheat and not use <h*> elements at all (you can use the h* classes instead on a div).

packages/dynamic-flows/src/layout/review/index.js
```diff
@@ -0,0 +1,70 @@
...
+const DynamicReview = (props) => {
+  const review = props.component;
+
+  const mapDefinition = ({ label, value }, index) => {
+    return { key: String(index), title: label, value };
+  };
+
+  const getReviewLayout = (orientation) => {
+    if (orientation === 'horizontal') {
+      return DefinitionList.Layout.HORIZONTAL_JUSTIFIED;
```
helendurranttw, line 18: Do you want `HORIZONTAL_JUSTIFIED` or `HORIZONTAL_RIGHT_ALIGNED` here?

packages/dynamic-flows/src/layout/review/index.js
```diff
@@ -0,0 +1,70 @@
...
+    return (
+      <a href={action.url} className="pull-right" onClick={onActionClick} role="button">
+        {action.label}
+      </a>
+    );
+  };
+
+  return (
+    <>
+      <h6 className="m-b-2">
```
helendurranttw, line 40: We shouldn't hardcode this as an h6, or we might break the hierarchy of h elements.

helendurranttw: You can use class=h6 to get the same styling :)

packages/dynamic-flows/src/layout/review/index.js
```diff
@@ -0,0 +1,70 @@
...
+};
+
+DynamicReview.propTypes = {
+  onAction: Types.func.isRequired,
+  component: Types.shape({
+    text: Types.string,
+    definitions: Types.arrayOf(
+      Types.shape({
+        label: Types.string,
+        value: Types.string,
```
helendurranttw, line 61: The definition list also accepts node values, so you can do something like this: `<span className="h3">89.45GBP</span>` to achieve this:

Screenshot 2020-07-20 at 16 34 15

packages/dynamic-flows/src/layout/utils.js
```diff
@@ -0,0 +1,34 @@
...
+      return 'text-xs-right';
+    case 'center':
+      return 'text-xs-center';
+    case 'left':
+    default:
+      return '';
+  }
+};
+
+const getTextClasses = (component) => {
```
helendurranttw, line 30: Maybe something like get `textAlignmentClasses`? `getTextPositioningClasses`? I was pretty unsure what classes that method was fetching when I saw it in use.

packages/dynamic-flows/src/layout/heading/index.js
StevePole: Yeah, I did think about this.  The layout is effectively a tree structure, so it would be possible to pass down the closest parent heading size and ensure correct hierarchy, implementing the desired visual size with classes.  But it seemed quite complicated, especially given the size of the PR 😬 .  Perhaps something to follow up on.

packages/dynamic-flows/src/layout/alert/index.js
```diff
@@ -0,0 +1,50 @@
...
+import { sizeModel, marginModel, contextModel } from '../models';
+import { getMarginBottom } from '../utils';
+
+import { Alert, Markdown } from '@transferwise/components';
+
+const DynamicAlert = (props) => {
+  const alert = props.component;
+
+  const getAlertSize = (size) => {
+    switch (size) {
```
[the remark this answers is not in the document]
StevePole: The way I was looking at it, from a spec point of view, we support five sizes, it's up to the client to decide what that means for rendering.  In this case, we decide not to implement 5 different sizes but map some of them.  This could be something we limit in the backend library, but felt it was better to behave sensibly if we receive these other sizes.

packages/dynamic-flows/src/layout/DynamicLayout.js
```diff
@@ -0,0 +1,102 @@
...
+
+  return <>{components.map(renderComponent)}</>;
+};
+
+DynamicLayout.propTypes = {
+  onAction: Types.func.isRequired,
+  onModelChange: Types.func.isRequired,
+  components: Types.arrayOf(componentModel).isRequired,
+  submitted: Types.bool.isRequired,
+  errors: Types.oneOfType([Types.string, Types.number, Types.object, Types.array, Types.bool]),
```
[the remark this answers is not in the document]
StevePole: You're right, I went this way because the schema could be any of these types, and the error format matches the schema format.  But actually in the case of a simple type schema, the error would just be a string.  So it should be string, object or array.

packages/dynamic-flows/src/layout/review/index.js
StevePole: The thing is, it should be a heading _of some sort_.  So I could style it with H6, but without going the whole hog and automating heading hierarchy, we'll still have the same problem...

packages/dynamic-flows/src/layout/review/index.js
StevePole: I wanted justified, as that's the one I've seen in common usage.  There's a debate to be had about the spec, and whether we should support both in dynamic flows.  I choose orientation as it felt a bit less arbitrary than the existing 'layout' enum, and would be useful for horizontal layout forms as well.  If we need another variant, I would include another prop (e.g. align) in the dynamic layout spec.

packages/dynamic-flows/src/layout/review/index.js
StevePole: Interesting.  I think in this case we only want strings, but we may make use of that to do emphasised values.

packages/dynamic-flows/src/layout/DynamicLayout.e2e.spec.js
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
+import { mount } from 'enzyme';
```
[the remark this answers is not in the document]
StevePole: Not going to change that right now, probably quite a big refactor...

PR #452, opened 2020-06-29
fix: replace bootstrap variables with neptune tokens

anttio:

## 🖼 Context
We are deprecating Bootstrap so there's a need to start using Neptune tokens and variables.

## 🚀 Changes
Replaced all imported Bootstrap variables with Neptune Tokens and LESS variables.

## ✅ Checklist


packages/components/src/navigationOption/NavigationOption.less
```diff
@@ -1,4 +1,5 @@
-@import 'bootstrap/less/variables';
+@import '@transferwise/neptune-css/src/less/variables/_colors.less';
+@import '@transferwise/neptune-css/src/less/variables/_typography.less';
```
helendurranttw, line 2: Looks like maybe this isn't needed any more?

packages/components/src/phoneNumberInput/PhoneNumberInput.less
```diff
@@ -1,16 +1,18 @@
-@import 'bootstrap/less/variables';
+@import '@transferwise/neptune-css/src/less/variables/_custom-media.less';
```
helendurranttw, line 1: Think you can remove this now too

packages/components/src/stepper/Stepper.less
```diff
@@ -1,33 +1,35 @@
-@import 'bootstrap/less/variables';
+@import '@transferwise/neptune-tokens/colors-base.less';
+@import '@transferwise/neptune-css/src/less/variables/_custom-media.less';
```
helendurranttw, line 2: I think these two can go

packages/components/src/tabs/Tabs.less
```diff
@@ -1,5 +1,5 @@
-@import 'bootstrap/less/variables.less';
 @import '@transferwise/neptune-css/src/less/variables/_colors.less';
+@import '@transferwise/neptune-css/src/less/variables/_custom-media.less';
```
helendurranttw, line 2: And this?

packages/components/src/navigationOption/NavigationOption.less
anttio: There's the link hover and active state colours that are not yet part of the tokens package, thus the `_colors.less` is needed.

`_typography.less` is needed for font-weight change in the media-body.

These imports can be removed when we have them in the tokens package.

packages/components/src/phoneNumberInput/PhoneNumberInput.less
anttio: I think this is still needed to get the media queries to work.

packages/components/src/stepper/Stepper.less
anttio: Still needed until everything is in tokens.

packages/components/src/tabs/Tabs.less
anttio: I think this is still needed to get the media queries to work.

packages/components/src/dynamicFieldDefinitionList/FormattedValue/FormattedValue.less
```diff
@@ -1,5 +1,3 @@
-@import 'bootstrap/less/variables';
-
 .formatted-value__h3-custom-alignment {
   margin-bottom: 16px;
   line-height: 24px;
```
dozortsev, line 3: shouldn't those two be using variables?

packages/components/src/flowNavigation/backButton/BackButton.less
```diff
@@ -1,4 +1,4 @@
-@import 'bootstrap/less/variables';
+@import '@transferwise/neptune-tokens/colors-base.less';
 
 @transition-timing-function: ease-in-out;
 @transition-duration: 0.3s;
```
dozortsev, line 4: what about this one? we have `--transition-duration: 350ms;` in `_transitions.less` file

packages/components/src/summary/Summary.less
```diff
@@ -1,5 +1,3 @@
-@import '@transferwise/neptune-css/src/less/variables/_colors.less';
-@import '@transferwise/neptune-css/src/less/variables/_spacing.less';
 @import '@transferwise/neptune-css/src/less/variables/_typography.less';
```
dozortsev, line 1: why we removing those two and leaving `_typography.less` ?

packages/components/src/typeahead/Typeahead.less
```diff
@@ -1,14 +1,19 @@
-@import 'bootstrap/less/variables';
+@import '@transferwise/neptune-css/src/less/variables/_custom-media.less';
+@import '@transferwise/neptune-css/src/less/variables/_sizes.less';
+@import '@transferwise/neptune-css/src/less/variables/_zindex.less';
+
 @animation-timing-function: ease-in-out;
 @animation-duration: 0.03s;
```
dozortsev, line 6: I guess those two can be delete and instead use variables from `_transitions.less`

packages/components/src/dynamicFieldDefinitionList/FormattedValue/FormattedValue.less
anttio: Margin-bottom replaced with space-16. For line-heights we don't have fixed sized variables.

packages/components/src/flowNavigation/backButton/BackButton.less
anttio: We can revisit these LESS variables later. The variables in `_transitions.less` are almost the same, but might affect the animations.

packages/components/src/typeahead/Typeahead.less
anttio: Let's revisit these LESS variables later.

packages/components/src/select/Select.less
```diff
@@ -10,27 +11,27 @@ html.ios-click {
...
-        margin-right: @spacer;
+        margin-right: var(--space-8);
       }
     }
 
     &--divider {
       border-bottom-style: solid;
       border-bottom-width: 1px;
-      border-bottom-color: @brand-smoke-plus-20;
+      border-bottom-color: @color-base-smoke-fade;
```
andreapiras, line 21: Can we not use `--color-background-neutral` here from @import '@transferwise/neptune-tokens/colors.css'; and import @transferwise/neptune-tokens/colors.css with the custom plugin u added?

postcss.config.js
```diff
@@ -1,14 +1,20 @@
...
+require('@arshaw/postcss-custom-properties');
 require('postcss-custom-media');
 require('cssnano');
 
 module.exports = {
   plugins: {
     'postcss-import': {},
     autoprefixer: {},
-    'postcss-custom-properties': {},
+    '@arshaw/postcss-custom-properties': {
```
andreapiras, line 11: 😍

packages/components/src/select/Select.less
anttio: We should use the `--color-border-*` semantic colours here, but we don't have that lighter colour there yet.

PR #455, opened 2020-06-30
feat!: migrate to icons v2

domhorvathtw:

## 🖼 Context
[See PR description for beta pre-release](https://github.com/transferwise/neptune-web/pull/357)


packages/css/src/less/alerts.less
```diff
@@ -8,3 +8,9 @@
 @import (reference) 'bootstrap/less/mixins/alerts.less';
 
 @import 'bootstrap/less/alerts.less';
+
+.alert {
+  .close {
```
helendurranttw, line 13: I think it would be better to move this to the component less, so it doesn't impact people using basic HTML alerts

packages/dynamic-flows/babel.config.js
```diff
@@ -53,6 +53,7 @@ module.exports = {
     '@babel/plugin-proposal-class-properties',
     '@babel/plugin-transform-react-jsx',
     '@babel/plugin-transform-runtime',
+    'transform-react-remove-prop-types',
```
helendurranttw, line 56: Same concern as above.

PR #459, opened 2020-07-03
fix!: definition list improvements for wrapping and alignment

helendurranttw:
## 🖼 Context

At the moment the label is cut off after 156px in horizontal layouts. Also, the justified option actually right aligns the text 🤔

## 🚀 Changes

Refactored the markup a bit. Moved CSS into the component. Deprecated the CSS in Neptune CSS. Increased the label width to 30%.

**BREAKING CHANGES:** I changed the HORIZONTAL_JUSTIFIED layout from right aligning to actually justifying, and I added a new HORIZONTAL_RIGHT_ALIGNED. Also, long labels will now wrap, not truncate. 

## ✅ Checklist


package.json
```diff
@@ -114,14 +114,14 @@
...
-        2,
+        1,
         "never"
       ],
       "type-case": [
         0
       ],
       "type-empty": [
-        2,
+        1,
```
dozortsev, line 124: what this does?

packages/components/.storybook/config.js
```diff
@@ -5,7 +5,8 @@ import { withInfo } from '@storybook/addon-info';
 import { withA11y } from '@storybook/addon-a11y';
 
 import theme from './common/theme';
-import 'bootstrap/dist/css/bootstrap.css';
+import '@transferwise/neptune-css/dist/css/neptune.css';
```
dozortsev, line 8: 🙇‍♂️

packages/css/src/less/core/_scaffolding.less
```diff
@@ -286,6 +287,7 @@ dd {
   }
 }
 
+/* Deprecated - use the Neptune Web DefinitionList component */
```
dozortsev, line 290: why can't we just delete this code?

postcss.config.js
```diff
@@ -9,7 +9,9 @@ module.exports = {
     'postcss-import': {},
     autoprefixer: {},
     'postcss-custom-properties': {},
-    'postcss-custom-media': {},
+    'postcss-custom-media': {
+      stage: 1,
```
dozortsev, line 13: what this for? :)

package.json
helendurranttw: Ugggh, the commit lint started complaining about the commits in the branch with !s in them, even though we were under the impression they were valid :( So I've moved them to warnings for now. [This is the relevant issue](https://github.com/conventional-changelog/commitlint/issues/878).

packages/css/src/less/core/_scaffolding.less
helendurranttw: We agreed to give people the chance to remove their use of deprecated components (or classes) for a while after deciding to deprecating them, before we remove them.

PR #460, opened 2020-07-06
chore: updating dependencies

helendurranttw:
## 🖼 Context

Updating dependencies within range and removing an unnecessary plugin.

## 🚀 Changes

Updating all deps within range, also removed one plugin being used in docs that was causing the import of js-yaml, which had a vulnerability (now resolved).

## 🤔 Considerations

Since a patch for js-yaml was released the removal of the inline svg plugin wasn't necessary any more, but it was also a bit overkill for a single SVG so I removed it anyway.

## ✅ Checklist


helendurranttw:
Closing because I decided to merge the docs update straight into master (I didn't realise it was actually pulling in a vulnerable version of js-yaml before I removed the plugin).

PR #464, opened 2020-07-07
feat: Fix Avatar sizes, add inverse and backgroundColor props

sbrandwoo:
## 🖼 Context

The original sizes of the Avatar didn't match Figma, so these have been updated.

I also added two properties:
- `inverse`: For when the avatar is on a dark background, this changes the inset border colour, and the default background colour
- `backgroundColor`: Optional colour to use instead of the defaults

## ✅ Checklist


packages/components/src/avatar/Avatar.js
```diff
@@ -35,13 +38,17 @@ Avatar.propTypes = {
   ]),
   children: Types.node,
   outlined: Types.bool,
+  inverse: Types.bool,
```
rerdos, line 41: I wonder if this should be `theme` to be aligned with other components.

Like: `theme={FlowNavigation.Theme.LIGHT}`

sbrandwoo: Done!

packages/components/src/avatar/Avatar.js
```diff
@@ -2,21 +2,25 @@ import React from 'react';
...
 import classnames from 'classnames';
 
-import './Avatar.css';
+import { Size, Theme } from '../common';
 
-import { Size } from '../common';
+import './Avatar.css';
 
-const Avatar = ({ size, type, children, outlined, className }) => (
+const Avatar = ({ size, type, children, outlined, theme, backgroundColor, className }) => (
```
dozortsev, line 9: can we avoid creating `backgroundColor` prop and use `className` one for this purposes?

```jsx
<Avatar .. className="custom-background-color" />
```

dozortsev: note: this CSS class then can be reused for other `Avatar`s

sbrandwoo: We're setting colours programmatically, so CSS classes are no good to us.

If we don't want the component to support the use case then we'd need to put the background in the content instead. I'd prefer if the component did what we needed.

sbrandwoo: Setting a background colour is a product feature that Avatar needs to support.

dozortsev: okey

PR #470, opened 2020-07-09
fix: allow empty month in dateInput

andreapiras:

## 🖼 Context

User should not always be provided with a default choice for month in order not to pollute user data
## 🚀 Changes
This PR set month to none if no value is passed

 

## ✅ Checklist


gusso:
@andreapiras Considering the deployments are not working, could you post a few screenshots of what it looks like before you click it and while the select is open? Or a gif of it working?

dozortsev:
@gusso it works, you just have to manually construct the link

💁‍♂️ https://transferwise.github.io/neptune-web/branch/DS-99-dateinput-firstmonth/components/inputs/DateInput

gusso:
I think we should have a better default for the month placeholder, as it will most certainly almost never fit. What do you think of `Month`?

packages/docs/pages/_app.js
```diff
@@ -24,7 +24,7 @@ if (process.env.NODE_ENV !== 'production' && typeof window !== 'undefined') {
 class MyApp extends App {
   componentDidMount() {
     const { pathname } = Router;
-    if (pathname === '/') {
+    if (pathname === '/' || pathname === '/_error') {
```
helendurranttw, line 27: These changes should be in their own PR.

helendurranttw:
When I load the page I see a flash where the month changes from 12 to December. Could you look into that @andreapiras ?

PR #475, opened 2020-07-10
chore: update docs for typeahead

andreapiras:

## ✅ Checklist


packages/components/src/typeahead/Typeahead.story.js
```diff
@@ -39,31 +65,16 @@ export const createable = () => {
...
-    {
-      label: 'Euro',
-    },
-    {
-      label: 'Something else',
-    },
-  ]);
+  const [options, setOptions] = React.useState(OriginalOptions);
+
+  const handleOnSearch = (value) => {
```
helendurranttw, line 70: Thanks for this! I'm wondering though if we should provide this functionality built in. Or even by default? I feel like I'd want my typeahead to filter by default. I guess we wouldn't want to break existing functionality, but we could enable this with a flag. What do you think? `filterOptions={true}`?

andreapiras: I thought about this, but then I also thought that people may want different types of filtering? I'd leave it out to be honest

packages/components/src/typeahead/Typeahead.js
```diff
@@ -68,6 +68,7 @@ export default class Typeahead extends Component {
     onFocus: Types.func,
     chipSeparators: Types.arrayOf(Types.string),
     size: Types.oneOf([Typeahead.Size.MEDIUM, Typeahead.Size.LARGE]),
+    enableDefaultFiltering: Types.bool,
```
helendurranttw, line 71: I've got an idea that might make it easier to support both filtering and custom filtering. How about we have a `filterOptions` prop (name up for debate), which has three values:

- `false|undefined|null` - no option filtering (default)
- `true` - apply default filtering
- `function` - apply this custom filtering function

What do you think? Might be more intuitive then filtering the options externally.

1.

andreapiras: @helendurranttw  Hey do we need the default filtering. maybe we can have 
```
filterOptions: Types.shape({
      filter: Types.bool,
      customFilter: Types.func,
    }),
```
so if

- `filter === false` ===> no filtering
- `filter === true && customFilter=== null ` ===> default filtering
- `filter === true && customFilter`=== func => custom filtering

WDYT?

packages/components/src/typeahead/Typeahead.js
```diff
@@ -363,6 +365,22 @@ export default class Typeahead extends Component {
     );
   };
 
+  filterOptions = (query) => {
+    const { options, filterOptions } = this.props;
+    let filteredOptions = [];
```
helendurranttw, line 370: Nice, exactly what I was imagining. We could make a couple of minor optimisations:

1. We could keep a list of the options mapped to their lower case equivalents, and only change it when options changes. This would remove`option.label.toLowerCase()` from the loop
2. We could set our filterFunction outside of this method. We know from filterOptions whether there is a custom function or not. If we set it, we only need to update it when props.filterOptions changes (which is probably hardly ever). That would remove the if / else.

packages/components/src/typeahead/Typeahead.js
```diff
@@ -363,6 +365,22 @@ export default class Typeahead extends Component {
...
+      if (typeof filterOptions === 'function') {
+        filteredOptions = filterOptions(options);
+      } else {
+        filteredOptions = options.filter(
+          (option) => option.label.toLowerCase().indexOf(query.toLowerCase()) >= 0,
+        );
+      }
+    }
+
+    return filteredOptions.length ? filteredOptions : options;
```
helendurranttw, line 381: Do we want to show all options if there are no results? Seems like we'd want to show none.

packages/components/src/typeahead/Typeahead.spec.js
```diff
@@ -328,13 +328,40 @@ describe('Typeahead', () => {
...
+      expect(options).toHaveLength(props.options.length);
     });
 
     it('does not show options if query is too short', () => {
       expect(menu().is('.open')).toBe(false);
       input().simulate('change', { target: { value: 'test' } });
       expect(menu().is('.open')).toBe(true);
     });
+
+    it('does filter results by default', () => {
```
helendurranttw, line 340: Could you clarify the wording here? I'm not sure what this is doing

packages/components/src/typeahead/Typeahead.spec.js
```diff
@@ -328,13 +328,40 @@ describe('Typeahead', () => {
...
+      component.setProps({
+        filterOptions: false,
+      });
+      input().simulate('change', { target: { value: 'Test1' } });
+
+      const options = option().map((optNode) => optNode.text());
+      expect(options).toEqual(['Test', 'Test1', 'Test2', 'Test3']);
+    });
+
+    it(`use custom filter if provided and enbale filter is true`, () => {
```
helendurranttw, line 357: Typos: "uses" and "enable"

packages/components/src/typeahead/Typeahead.js
```diff
@@ -92,6 +93,7 @@ export default class Typeahead extends Component {
     onInputChange: null,
     onFocus: null,
     validateChip: () => true,
+    filterOptions: true,
```
helendurranttw, line 96: We have to set this to `false` to be backwards compatible.

packages/components/src/typeahead/Typeahead.story.js
```diff
@@ -101,6 +102,7 @@ export const basic = () => {
       addon={<i className="input-group-text icon icon-search" />}
       onBlur={() => {}}
       options={options}
+      enableDefaultFiltering={enableDefaultFiltering}
```
helendurranttw, line 105: filterOptions? 🤔

packages/docs/pages/components/inputs/Typeahead.mdx
```diff
@@ -5,6 +5,19 @@ import code from '../../../liveEditorCode/typeahead.code';
 <LiveEditorBlock code={code} scope={{ Typeahead }} display="vertical" />
 <GeneratePropsTable componentName="Typeahead" />
 
+By default Typeahead does provide basic filter functionalities. If you want to disable filtering please set `filterOptions= false`. Filtering can also be customized by passing a filtering function
```
helendurranttw, line 8: "By default Typeahead provides basic filtering functionality."

... except we'll want to update that once you change the default to false. Also should there be a line break here?

packages/components/src/typeahead/Typeahead.js
```diff
@@ -68,6 +68,7 @@ export default class Typeahead extends Component {
     onFocus: Types.func,
     chipSeparators: Types.arrayOf(Types.string),
     size: Types.oneOf([Typeahead.Size.MEDIUM, Typeahead.Size.LARGE]),
+    filterOptions: Types.oneOfType([Types.bool, Types.func]),
```
StevePole, line 71: This signature is a bit nasty.  Ideally
```
<thing /> == <thing whatever="false" />
```
i.e. the absence of an attribute should be functionally equivalent to setting that attribute to false.  Is that the case here?  I feel like it's not...

andreapiras: To make it equivalent we should set the default to false, but that means that people will have to opt in if they want this behavior.

andreapiras:
After investigation, we might want to investigate further the commonalities between select and typeahead

PR #479, opened 2020-07-13
fix: new icon alignments for select and money input component

domhorvathtw:
#### Before
image
#### After
image

#### Before
image
#### After
image

## 🚀 Changes
- CSS fixes for `MoneyInput` component
- CSS fixes for `Select` component

## 🤔 Considerations
- keep this CSS fixes close to the React components

## ✅ Checklist


packages/css/src/less/buttons.less
```diff
@@ -47,5 +47,9 @@ a.btn {
   }
   &.btn-sm .tw-icon {
     right: 8px;
+    top: 8px;
+  }
+  &.btn-lg .tw-icon {
+    top: 27px;
```
andreapiras, line 53: Is this valid for all buttons sitewide???

PR #481, opened 2020-07-13
Adding className prop to Navigation option

gittymcgit:

## 🚀 Changes
Added className prop to NavigationOption
 

## ✅ Checklist


packages/components/src/navigationOption/NavigationOption.js
```diff
@@ -36,7 +38,7 @@ const NavigationOption = ({
 
   return (
     <Option
-      className="tw-navigation-option"
+      className={classNames('tw-navigation-option', className)}
```
dozortsev, line 41: `Option` is core component for this one so I think you just have to pass it down there, it already has `className` prop

andreapiras:
Hey @gittymcgit would u mind adding the classname on docs file too please?
https://github.com/transferwise/neptune-web/blob/master/packages/docs/pages/components/navigation/NavigationOption.mdx

packages/components/src/navigationOption/NavigationOption.js
dozortsev: I mean just pass it to `Option` via `sharedProps` constant

PR #485, opened 2020-07-15
feat: value normalizer

andreapiras:

## 🖼 Context

Our components return values in different format. This wrapper aims to normalized the different return and make it consistent and predictable

## ✅ Checklist


StevePole:
Feels like we're working around other problems, rather than addressing the core issue, which is that our components return inconsistent things.

What about if we changed all component to return a native compliant event with a target and value?  Or alternatively what if we implemented a React `<Input ... />` that broadcast just the value?

StevePole:
With the type validators, I wouldn't implement isNumeric or isEvent in this file as these aren't JS types.  There could possibly be an isClass, or isInstanceOf (that has two params).

Keeping inline with JSON schema we might consider 'format validators' that check, for example, if a string is: numeric, an iso date, or a base64url.

packages/dynamic-flows/src/WithNormalizer/WithNormalizer.js
```diff
@@ -0,0 +1,44 @@
...
+const WithNormalizer = ({ children }) => {
+  const handleOnChange = (event) => {
+    let value = event;
+
+    if (isEvent(event)) {
+      value = isNumeric(event.target.value) ? parseFloat(event.target.value) : event.target.value;
+    } else if (event && (event.value || isNumber(event.value))) {
+      value = event.value;
+    } else if (event.status) {
+      value = event.status;
```
dozortsev, line 14: why `event` would have `value`/`status` props?

dozortsev: it feels like to start use it everywhere we will need grow complexity of these checks, e.g tomorrow you would need add check for `data` prop and then for `initialValue` etc etc.

andreapiras:
> We better refactor our components to consistent return value, instead of fixing "mess" with wrapper/abstraction.
> 
> I haven't done much thought about this and didn't do any research but atm I'm thinking in favor of [native Event object](https://developer.mozilla.org/en-US/docs/Web/API/Event) as it's data could be useful for consumers in their business logic / UI.

This is on our roadmap, we'll centralize this function first and then update our component's API to emit consistently events

packages/dynamic-flows/src/WithNormalizer/validation/event-utilities/index.js
```diff
@@ -0,0 +1,26 @@
...
+
+const normalizeEvent = (event, type) => {
+  let value = event;
+
+  if (event) {
+    if (isSyntheticEvent(event)) {
+      value = type === 'number' ? parseFloat(event.target.value) : event.target.value;
+    } else if (isNumber(event.value) || event.value) {
+      // Checkbox,Select
+      value = event.value;
```
dozortsev, line 20: what about `event.status`?

andreapiras: There was no handling in the previous logic so we decided to remove it. We can add it in a second iteration if needed

dozortsev: 👌

PR #489, opened 2020-07-16
fix: for Select component

dozortsev:
## 🖼 Context

Migrating styleguide-components on `neptune.css` and have found this issue(s) in `ngSelect` component.

## 🚀 Changes

See comments in changes.

## ✅ Checklist


packages/css/src/less/circles.less
```diff
@@ -13,7 +13,8 @@
 
 li,
 li > a {
-  > .circle:first-child {
+  > .circle:first-child,
+  > .tw-icon:first-child {
     margin-right: var(--spacer);
```
dozortsev, line 18: Screen Shot 2020-07-16 at 3 16 45 PM

packages/css/src/less/neptune.bundle.less
```diff
@@ -25,6 +25,7 @@
 @import './list-group.less';
 @import './navs.less';
 @import './navbar.less';
+@import './select.less';
```
dozortsev, line 28: I need select styles to be part of `neptune.css` bundle, [fix which Domi made for React component](https://github.com/transferwise/neptune-web/pull/479) now necessary for Angular one

Screen Shot 2020-07-16 at 4 03 20 PM

PR #490, opened 2020-07-16
feat: expose complete step of upload component

arikanmstf-tw:
## 🖼 Context
In some cases (e.g. batch payments), upload steps needed to be used individually. In previous version of neptun we did this by importing from direct path. ( `import UploadCompleteStep "@transferwise/components/lib/upload/steps/completeStep/completeStep";` )

With new build set up, this is not possible. So we have came with the solution exporting those component in Upload.

[See](https://transferwise.slack.com/archives/CJNBX9CP6/p1594906799316700) slack discussion for more.

## ✅ Checklist


packages/components/src/upload/Upload.js
```diff
@@ -374,4 +374,8 @@ Upload.defaultProps = {
   usPlaceholder: 'Drag and drop a file less than 5MB',
 };
 
+Upload.UploadImageStep = UploadImageStep;
```
andreapiras, line 377: Why would u expose this one? is it not the default start state?

arikanmstf-tw: with `<Upload />` , step can change depends on the state,
however `< Upload.UploadImageStep />` would always show uploading step

andreapiras: Ok, but what would it be the scenario of this use case. Same for processing actually. I think we should just expose completed and then if we need it we can expose the others

PR #497, opened 2020-07-21
Merge branch 'master' into beta

helendurranttw:
## 🖼 Context

Lots of small updates in both branches so want to make sure I haven't messed up this merge..

## ✅ Checklist


packages/components/src/common/Option/Option.less
```diff
@@ -1,3 +1,7 @@
+.media .circle .tw-icon {
+  margin-top: var(--space-8);
+}
```
dozortsev, line 3: this shouldn't be here

dozortsev:
will be done in https://github.com/transferwise/neptune-web/pull/501

PR #498, opened 2020-07-22
feat: add Tile and Decision components

hendrikmolder:
## 🖼 Context

We identified this pattern is already used in multiple places across web and there are teams looking to implement this pattern. Hence, these two components are introduced.

[Demo](https://transferwise.github.io/neptune-web/branch/REC-1178_add-tile/components/navigation/Decision)
[Figma](https://www.figma.com/file/VGORCloYyujoJ4XYfLPoa5/Picker?node-id=0%3A1)
[Proposal](https://transferwise.atlassian.net/wiki/spaces/DS/pages/1401191329/Picker+navigation+option)
[Slack](https://transferwise.slack.com/archives/C016ABG4HT8)

## 🚀 Changes
 
- Add `Tile`
- Add `Picker`
- Docs and stories for new components

## 🤔 Considerations

`Picker` relies heavily on `css-element-queries` – testing the CSS changes on various breakpoints is not straightforward due to that.

## ✅ Checklist


hendrikmolder:
@gusso @helendurranttw This is an extremely quick draft I put together after our meeting. Would you like to give some quick feedback? I'll work on this tomorrow afternoon, so would be useful 😬

packages/components/src/tile/Tile.js
```diff
@@ -0,0 +1,44 @@
...
+import classNames from 'classnames';
+
+import './Tile.css';
+
+export const Tile = ({ as: Element, className, title, content, illustration, onClick }) => {
+  return (
+    <Element
+      className={classNames('tile', className)}
+      onClick={onClick}
+      onKeyDown={(e) => {
```
hendrikmolder, line 12: 😬 😬

packages/components/src/picker/Picker.less
```diff
@@ -0,0 +1,21 @@
...
+    justify-content: center;
+  }
+
+  &__navigation-option-container {
+    display: none;
+  }
+}
+
+// Should we use media queries for switching?
+@media (max-width: 768px) {
```
helendurranttw, line 13: Hmm hmm hmm. I went down a rabbit hole thinking about this. Ideally these wouldn't be tied to the screen size, but the width of the container. But how would we do this? We could use the ResizeObserver but it's not supported in IE11. I guess we could degrade to always show one or the other in this case..

Should we even be in charge of this? Maybe the consumer could choose when to swap between them?!

Let's chat about this 😁

packages/components/src/tile/Tile.js
```diff
@@ -0,0 +1,44 @@
...
+  as: Types.elementType,
+  title: Types.node.isRequired,
+  content: Types.node,
+  media: Types.node, // TODO: illustration or icon?
+  onClick: Types.func.isRequired,
+  className: Types.string,
+};
+
+Tile.defaultProps = {
+  as: 'div', // TODO: clickable div – accessibility? should it be a button? or a link? Does Button support having illustrations?
```
helendurranttw, line 38: Do you want to keep it configurable? I think for accessibility a link is probably best? 🤔

packages/components/src/tile/Tile.less
```diff
@@ -0,0 +1,22 @@
+@import '@transferwise/neptune-css/src/less/neptune-tokens.less';
```
helendurranttw, line 1: You don't need to import these :) The tokens will automatically be resolved when we build.

packages/docs/pages/components/content/Tile.mdx
```diff
@@ -0,0 +1,26 @@
+import { LiveEditorBlock, GeneratePropsTable } from '../../../utils';
+import tileCode from '../../../liveEditorCode/tile.code';
+import pickerCode from '../../../liveEditorCode/picker.code';
+
+import { Tile, Picker } from '@transferwise/components';
+
+<LiveEditorBlock code={tileCode} scope={{ Tile }} />
+<GeneratePropsTable componentName="Tile" />
+
+## Usage
```
helendurranttw, line 10: I think the Picker probably deserves it's own page, and we can link to it from here 🙂

packages/components/src/tile/Tile.less
```diff
@@ -0,0 +1,22 @@
...
+  flex-direction: column;
+  width: 340px;
+  padding: var(--space-16);
+  border: 1px solid var(--color-border-neutral);
+  border-radius: 3px;
+  text-align: center;
+
+  &:hover {
+    // TODO: do we want border and drop shadow on hover?
+    cursor: pointer;
```
helendurranttw, line 14: Will wait for @gusso to weigh in on this

packages/components/src/tile/Tile.js
helendurranttw: Yeah, if we use a link this won't be necessary :) We can make the whole thing a link, though [this article](https://inclusive-components.design/cards/) has a cool trick that allows you to use more semantic markup (only the title is the link) and yet keeping the whole element clickable.

packages/components/src/picker/Picker.less
helendurranttw: Oooooh https://www.npmjs.com/package/css-element-queries

packages/components/src/tile/Tile.js
hendrikmolder: Left the parent element as `div` and title as `a`

packages/docs/pages/components/content/Tile.mdx
hendrikmolder: ✅ Added a separate page for Picker

packages/components/src/tile/Tile.js
hendrikmolder: ✅

packages/components/src/tile/Tile.less
hendrikmolder: Imo we should definitely have some focus/hover, but not sure how it should look like exactly 🙈 Took this from one of the designs/prototypes that was shared on the channel... However, I think we found from the first research, that some users still didn't think they can click on the tiles, despite the hover/focus...

hendrikmolder:
Kapture 2020-07-31 at 17 03 48

@gusso @helendurranttw what do you think of having a min-height for `media` or positioning the text towards the bottom of the tile? Not sure, if we'd want the titles & content text to be on the same line? Example above ☝️

packages/components/src/picker/Picker.js
```diff
@@ -0,0 +1,69 @@
...
+    <div className={classNames('tw-picker', className)}>
+      <div className="tw-picker__tile-container p-x-2">
+        {items.map(({ title, content, media, key }) => (
+          <Tile
+            key={key}
+            className="m-x-1"
+            title={title}
+            content={content}
+            media={media}
+            size={small && Tile.Size.SMALL}
```
andreapiras, line 25: I'd change this to be size={size} as we may introduce more sizes in future maybe we can specify a default prop

packages/components/src/picker/Picker.js
```diff
@@ -0,0 +1,69 @@
...
+import Tile from '../tile';
+
+import './Picker.css';
+
+ElementQueries.listen();
+
+const Picker = ({ className, items, onClick, small }) => {
+  return (
+    <div className={classNames('tw-picker', className)}>
+      <div className="tw-picker__tile-container p-x-2">
```
andreapiras, line 17: Can we not reuse utility classes here such as `.visible-xs-*` etc

packages/components/src/picker/Picker.less
```diff
@@ -0,0 +1,18 @@
...
+    display: flex;
+    justify-content: center;
+  }
+
+  &__navigation-option-container {
+    display: none;
+  }
+}
+
+.tw-picker[max-width~="768px"] > .tw-picker__tile-container {
```
andreapiras, line 12: See utility classes for toggling content across viewport breakpoints

andreapiras: https://github.com/transferwise/neptune-web/blob/main/packages/css/src/less/addons/_display-utilities.less

packages/components/src/tile/Tile.js
```diff
@@ -0,0 +1,58 @@
...
+    <div
+      className={classNames('tw-tile', className, {
+        'tw-tile--small': size === Tile.Size.SMALL,
+        'tw-tile--large': size === Tile.Size.LARGE || size !== Tile.Size.SMALL,
+      })}
+    >
+      <div className="m-y-2 m-x-auto">{media}</div>
+      {/* using anchor as a button */}
+      {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
+      <a
```
andreapiras, line 22: Why not using a button here?

packages/components/src/tile/Tile.less
```diff
@@ -0,0 +1,58 @@
...
+    &:focus {
+      outline: none;
+      text-decoration: none;
+    }
+
+    &:hover {
+      text-decoration: none;
+    }
+
+    &::after {
```
andreapiras, line 39: I don't quite get what is this used for

andreapiras:
Please do not forget tests files

packages/components/src/tile/Tile.story.js
```diff
@@ -0,0 +1,30 @@
...
+
+export const basic = () => {
+  const title = text('title', 'Hank Miller');
+  const body = <p>{text('body', 'Click here to send money to Mr. Miller')}</p>;
+  const size = select('Size', Tile.Size, Tile.Size.LARGE);
+
+  return (
+    <Tile
+      title={title}
+      content={body}
```
andreapiras, line 20: not biggie...but why not calling the knob with the same name of prop?

packages/components/src/tile/Tile.js
andreapiras: <button className="tw-tile__link btn-link"

packages/components/src/tile/Tile.js
```diff
@@ -0,0 +1,58 @@
...
+        onClick={onClick}
+        tabIndex="0"
+        role="button"
+        onKeyDown={(e) => {
+          if (e.key === 'Enter' || e.key === 'Spacebar') {
+            onClick();
+          }
+        }}
+      >
+        <h4 className="m-b-1">{title}</h4>
```
andreapiras, line 33: we should not hardcode heading as they may cause SEO problems on the space, please use a span with .h4 class

packages/components/src/tile/Tile.js
```diff
@@ -0,0 +1,58 @@
...
+import { Size } from '../common';
+
+import './Tile.css';
+
+const { LARGE, SMALL } = Size;
+
+export const Tile = ({ className, title, content, media, size, onClick }) => {
+  return (
+    <div
+      className={classNames('tw-tile', className, {
```
andreapiras, line 14: Should the whole element be clickable hence wrapped in a button?

packages/components/src/tile/Tile.js
andreapiras: This can be actually removed if we wrap the whole element in a button tag

packages/components/src/tile/Tile.less
```diff
@@ -0,0 +1,58 @@
...
+
+  &--small {
+    width: 180px;
+    padding: var(--space-16);
+
+    h4 {
+      font-size: var(--font-size);
+    }
+
+    > .tw-tile__content {
```
andreapiras, line 22: tw-tile__content  has not enough contrast to meet our accessibility requirements. Please check accessibility tab on storybook
image

packages/components/src/tile/Tile.less
```diff
@@ -0,0 +1,58 @@
...
+    width: 340px;
+    padding: var(--space-32);  
+  }
+
+  &--small {
+    width: 180px;
+    padding: var(--space-16);
+
+    h4 {
+      font-size: var(--font-size);
```
andreapiras, line 19: We should switch utility class rather than changing the font size of the element. We can use .h4 class and maybe remove it if small is passed

packages/components/src/tile/Tile.less
```diff
@@ -0,0 +1,58 @@
+.tw-tile {
+  position: relative;
```
andreapiras, line 2: Does it need to be relative/?

packages/components/src/tile/Tile.less
hendrikmolder: I think that's the CSS trick @helendurranttw suggested – it basically covers the whole Tile with the link, so one could click the link by clicking anywhere on the Tile

packages/components/src/tile/Tile.less
```diff
@@ -0,0 +1,58 @@
...
+  text-align: center;
+
+  &--large {
+    width: 340px;
+    padding: var(--space-32);  
+  }
+
+  &--small {
+    width: 180px;
+    padding: var(--space-16);
```
andreapiras, line 16: we can use utility classes and switch based on small m-x-2 m-y-2

packages/components/src/tile/Tile.less
hendrikmolder: This is only needed for the CSS trick for the `link` to work, I think

packages/components/src/tile/Tile.less
```diff
@@ -0,0 +1,58 @@
...
+    h4 {
+      font-size: var(--font-size);
+    }
+
+    > .tw-tile__content {
+      font-size: var(--font-size-small);
+    }
+  }
+
+  &__link {
```
andreapiras, line 27: we may be able to remove all this if we switch to button

packages/components/src/picker/Picker.js
```diff
@@ -0,0 +1,69 @@
...
+            className="m-x-1"
+            title={title}
+            content={content}
+            media={media}
+            size={small && Tile.Size.SMALL}
+            onClick={() => onClick(key)}
+          />
+        ))}
+      </div>
+      <div className="tw-picker__navigation-option-container">
```
andreapiras, line 30: we can remove this container and tarrget `tw-navigation-options-list` instead

packages/components/src/tile/Tile.less
andreapiras: Yes but if we wrap the whole component in a button we don't need to style the after

packages/components/src/tile/Tile.js
hendrikmolder: ✅

packages/components/src/tile/Tile.less
hendrikmolder: ✅

packages/components/src/tile/Tile.less
hendrikmolder: ✅

packages/components/src/picker/Picker.js
hendrikmolder: ✅

packages/components/src/picker/Picker.js
hendrikmolder: ✅

gusso:
Hello hello. Some design feedback based on [Figma](https://www.figma.com/file/VGORCloYyujoJ4XYfLPoa5/Picker?node-id=194%3A422):

- The tile should have 40px padding vertically, and 32px horizontally
- The title font should be 18px
- The tiles should be the same width, either controlled by the picker parent, or allowed to be set individually

hendrikmolder:
> Hello hello. Some design feedback based on [Figma](https://www.figma.com/file/VGORCloYyujoJ4XYfLPoa5/Picker?node-id=194%3A422):
> 
> * The tile should have 40px padding vertically, and 32px horizontally
> * The title font should be 18px
> * The tiles should be the same width, either controlled by the picker parent, or allowed to be set individually

🙈  For some reason after `master` was changed to `main` the build started failing... with no obvious reason – so whatever is deployed from this branch is probably missing the latest updates 😕  @andreapiras do you know why it might be failing? Haven't been able to figure that out

- ✅  Fixed the paddings
- 👍  Font size is 18px for large and 16px for small tile
- 👍  The tiles currently have a set width – 340px for large and 180px for small size.. we can maybe make it adjustable – I guess one can override the styles with a custom class at the moment as well

andreapiras:
> 🙈 For some reason after `master` was changed to `main` the build started failing... with no obvious reason – so whatever is deployed from this branch is probably missing the latest updates 😕 @andreapiras do you know why it might be failing? Haven't been able to figure that out
 Hey, the error is related to `css-element-queries/src/ElementQueries`, trying to fix it

packages/components/src/picker/Picker.spec.js
```diff
@@ -0,0 +1,55 @@
...
+        <Picker
+          items={[anItem(), anItem({ title: 'Hank Miller', key: 'HANK_MILLER' })]}
+          onClick={onClick}
+        />
+      </div>,
+    );
+
+    await screen.findByText('Melipe Forales');
+
+    expect(tileContainer(container)).toBeVisible();
```
hendrikmolder, line 23: Actually this doesn't work ([link](https://github.com/testing-library/dom-testing-library/issues/196#issuecomment-628268484)) 😬

hendrikmolder: Actually, it's probably just the `ElementQueries` that isn't working in the test env 🤔

packages/components/src/tile/Tile.less
gusso: This is using `color-text-secondary` correctly, it's the colour token's fault, not the implementation.

gusso:
Few more things regarding paddings:

- Paragraph has extra margin below, adding to the padding (image attached)
- Same thing with the media element, we should remove the margin above it so it doesn't add to the container padding (image attached)
- The paddings are inverted, it should be `size-40` on top and bottom, `size-32` on left and right
- Space between image and title should be `size-12` on the small size
- Content font size should be `12px` on the small size

Screenshot 2020-09-24 at 11 00 11 am
Screenshot 2020-09-24 at 11 04 47 am

hendrikmolder:
> Few more things regarding paddings:
> 
> * Paragraph has extra margin below, adding to the padding (image attached)
> * Same thing with the media element, we should remove the margin above it so it doesn't add to the container padding (image attached)
> * The paddings are inverted, it should be `size-40` on top and bottom, `size-32` on left and right
> * Space between image and title should be `size-12` on the small size
> * Content font size should be `12px` on the small size

Awesome – fixed everything except the first one... `<p>` has a padding and I think it's important we keep accepting passing `Node`s as `title`/`content` (usually needed when using translations etc). I updated the docs instead to not pass a paragraph to not encourage doing that 🙈

packages/components/src/picker/Picker.js
```diff
@@ -0,0 +1,75 @@
...
+import classNames from 'classnames';
+import ElementQueries from 'css-element-queries/src/ElementQueries';
+
+import NavigationOption from '../navigationOption';
+import NavigationOptionList from '../navigationOptionsList';
+import Tile from '../tile';
+
+import './Picker.css';
+
+const Picker = ({ className, items, onClick, small }) => {
```
dozortsev, line 12: please better introduce `size` prop (with two size) rather then bool prop, as tomorrow we might decide add one more size for this component which will force us introduce breaking change

packages/components/package.json
```diff
@@ -80,6 +80,7 @@
     "classnames": "^2.2.6",
     "commonmark": "0.29.1",
     "core-js": "^3.6.5",
+    "css-element-queries": "^1.2.3",
```
dozortsev, line 83: never used this library, could you please explain why we need it?

packages/components/src/picker/Picker.js
hendrikmolder: ✅

packages/components/package.json
hendrikmolder: https://github.com/transferwise/neptune-web/pull/498#discussion_r460115451

gusso:
Screenshot 2020-09-28 at 12 40 16 pm
I think this is the last visual one, distance between image and text on small variant should be 12 instead of 16.

hendrikmolder:
> I think this is the last visual one, distance between image and text on small variant should be 12 instead of 16.

✅

packages/components/src/decision/Decision.js
```diff
@@ -0,0 +1,97 @@
...
+        title={title}
+        content={description}
+        onClick={onClick}
+        media={list}
+        showMediaAtAllSizes
+        showMediaCircle
+      />
+    ));
+  }
+  return <></>;
```
andreapiras, line 62: docgen for docs requires something to be returned
```
Error: No suitable component definition found.
    at parse (neptune-web/node_modules/react-docgen/dist/parse.js:74:9)
```

packages/components/src/decision/Decision.js
```diff
@@ -0,0 +1,99 @@
...
+          key={key}
+          items={[
+            <NavigationOption
+              complex={false}
+              content={description}
+              href={href}
+              media={list}
+              onClick={onClick}
+              showMediaAtAllSizes
+              showMediaCircle
```
hendrikmolder, line 33:
```suggestion
```

I think having `showMediaCircle` here means we wouldn't be able to show illustrations within `NavigationOption` without the grey background? Maybe consumers could just use `Avatar` component if they need the grey circle and we won't have to pass it here?


Screenshot 2020-10-27 at 09 32 43

packages/docs/liveEditorCode/decision.code.js
```diff
@@ -0,0 +1,28 @@
...
+            <Avatar size="md" theme="light" type="initials" className="text-xs-center">
+              HM
+            </Avatar>
+          ),
+          block: (
+            <img src="https://transferwise.com/public-resources/assets/bank-details/bank-details-flow/finish.svg" />
+          ),
+        },
+        title: 'Hank Miller',
+        content: (
```
hendrikmolder, line 16:
```suggestion
        description: (
```

😝

packages/components/src/common/customProps/elementOfType/elementOfType.js
```diff
@@ -0,0 +1,47 @@
...
+
+export const errorReturn = ({ propName, typeToCheck, componentName, allowedType }) =>
+  new Error(
+    `Invalid prop ${propName} of type ${typeToCheck} provided to ${componentName}. Only ${allowedType} allowed`,
+  );
+
+const createChainableTypeChecker = (validate) => (allowedType) => {
+  const checkType = (isRequired, props, propName, componentName) => {
+    const prop = props[propName];
+    if (prop == null) {
```
helendurranttw, line 32: Or is undefined?

packages/components/src/common/customProps/elementOfType/elementOfType.js
```diff
@@ -0,0 +1,47 @@
...
+  const prop = props[propName];
+
+  if (prop && allowedType) {
+    const { type } = prop;
+
+    if (!type) {
+      return errorReturn({ propName, typeToCheck: prop, componentName, allowedType });
+    }
+    const typeToCheck = type.displayName || type;
+    // This can be improved by checking against the actual component decalration but we need to investigate which way is the most performant.
```
helendurranttw, line 13: Typo: decalration

packages/components/src/common/customProps/elementOfType/elementOfType.spec.js
```diff
@@ -0,0 +1,68 @@
+import { elementOfType, errorReturn } from './elementOfType';
+
+describe('elementOfType', () => {
+  it('returns null if prop is not defined', () => {
+    const props = {
+      myProp: {},
+    };
+    const propName = ' A Prop';
```
helendurranttw, line 8: extra space

packages/components/src/common/customProps/elementOfType/elementOfType.js
```diff
@@ -0,0 +1,47 @@
+import { isString } from '@transferwise/neptune-validation';
+
+const elementOfTypeChecker = (allowedType) => (props, propName, componentName) => {
```
helendurranttw, line 3: Do you want to consider using an object for props? In case we ever want to add more in the future? I.e. `({props, propName, componentName})`.

packages/components/src/common/customProps/elementOfType/elementOfType.js
```diff
@@ -0,0 +1,47 @@
...
+    return validate(allowedType)(props, propName, componentName);
+  };
+
+  const chainedCheckType = checkType.bind(null, false);
+  chainedCheckType.isRequired = checkType.bind(null, true);
+
+  return chainedCheckType;
+};
+
+export const elementOfType = createChainableTypeChecker(elementOfTypeChecker);
```
helendurranttw, line 47: Excited to introduce this checker!

packages/components/src/decision/Decision.js
```diff
@@ -0,0 +1,99 @@
...
+    }),
+  ).isRequired,
+  /**  Handles the display mode of the component */
+  presentation: Types.oneOf([
+    Decision.Presentation.LIST,
+    Decision.Presentation.LIST_BLOCK,
+    Decision.Presentation.LIST_BLOCK_SMALL,
+  ]),
+
+  /** The element type to be rendered in the list */
```
helendurranttw, line 90: Probably worth a comment here explaining our intention to extend this

packages/components/src/decision/Decision.js
```diff
@@ -0,0 +1,99 @@
...
+  options: Types.arrayOf(
+    Types.shape({
+      media: Types.shape({
+        block: elementOfType(['Avatar', 'img']).isRequired,
+        list: elementOfType(['Avatar', 'img']).isRequired,
+      }),
+      title: Types.oneOfType([elementOfType(['Message']), Types.string]).isRequired,
+      description: Types.oneOfType([elementOfType(['Message']), Types.string]).isRequired,
+      onClick: Types.func.isRequired,
+      href: requiredIf(Types.string, (props) => props.type === Decision.Type.NAVIGATION),
```
helendurranttw, line 80: We need a disabled prop here, and also Gusso said description should always be optional.

packages/components/src/decision/Decision.spec.js
```diff
@@ -0,0 +1,102 @@
...
+
+  let container;
+
+  describe(`when presentation is ${Decision.Presentation.LIST_BLOCK}`, () => {
+    beforeEach(() => {
+      resetClientWidth(500);
+      ({ container } = render(<Decision {...props} />));
+    });
+
+    it('renders Navigation Option before breakpoint', () => {
```
helendurranttw, line 45: Might be nice to say "breakpoint small" (or whatever it actually is) for future reference.

packages/components/src/decision/Decision.spec.js
```diff
@@ -0,0 +1,102 @@
...
+      expect(getNavigationOption()).toBeInTheDocument();
+    });
+
+    it('renders Tile after breakpoint', async () => {
+      resetClientWidth(900);
+      fireEvent(window, new Event('resize'));
+      await waitFor(() => {
```
helendurranttw, line 52: Are you sure you need async await here?

packages/components/src/decision/Decision.spec.js
```diff
@@ -0,0 +1,102 @@
...
+
+  describe(`when presentation is ${Decision.Presentation.LIST_BLOCK_SMALL}`, () => {
+    beforeEach(() => {
+      resetClientWidth(500);
+      ({ container } = render(
+        <Decision {...props} presentation={Decision.Presentation.LIST_BLOCK_SMALL} />,
+      ));
+    });
+
+    it('renders Navigation Option before breakpoint', () => {
```
helendurranttw, line 67: I think this is redundant

packages/components/src/decision/Decision.story.js
```diff
@@ -0,0 +1,85 @@
...
+          onClick: action('clicked'),
+          media: {
+            block: (
+              <img
+                src="https://transferwise.com/public-resources/assets/bank-details/bank-details-flow/finish.svg"
+                alt="alt"
+              />
+            ),
+            list: (
+              <Avatar size="md" theme="light" type="initials" className="text-xs-center">
```
helendurranttw, line 35: Why do you have text-xs-center here? Hopefully that's not needed... is it? 😬

packages/components/src/decision/Decision.story.js
```diff
@@ -0,0 +1,85 @@
...
```
helendurranttw, line 54: Same same

packages/components/src/sizeswapper/SizeSwapper.js
```diff
@@ -0,0 +1,42 @@
...
+
+import { Breakpoint } from '../common';
+import useClientWidth from './hooks';
+
+const SizeSwapper = ({ breakpoints, items, debouncingDelay }) => {
+  const elRef = useRef();
+  const [clientWidth] = useClientWidth({ elRef, debouncingDelay });
+
+  return (
+    <div className="tw-size-swapper" ref={elRef}>
```
helendurranttw, line 12: Let's use `np-` 😍

packages/components/src/sizeswapper/SizeSwapper.js
```diff
@@ -0,0 +1,42 @@
...
+    Types.oneOf([
+      SizeSwapper.Breakpoint.EXTRA_SMALL,
+      SizeSwapper.Breakpoint.SMALL,
+      SizeSwapper.Breakpoint.MEDIUM,
+      SizeSwapper.Breakpoint.LARGE,
+      SizeSwapper.Breakpoint.EXTRA_LARGE,
+    ]),
+  ),
+  /** Applied to window resize event, determine the frequency of the resize callback call */
+  debouncingDelay: Types.number,
```
helendurranttw, line 32: I don't think we need to expose this.

packages/components/src/sizeswapper/SizeSwapper.spec.js
```diff
@@ -0,0 +1,57 @@
...
+      />,
+    );
+
+    expect(getByText('element1')).toBeInTheDocument();
+    expect(queryByText('element2')).not.toBeInTheDocument();
+    expect(queryByText('element3')).not.toBeInTheDocument();
+
+    resetClientWidth(breakpoints[0]);
+    fireEvent(window, new Event('resize'));
+    await waitFor(() => {
```
helendurranttw, line 35: Same thing, I feel like you shouldn't need async here 🤔

packages/components/src/sizeswapper/SizeSwapper.story.js
```diff
@@ -0,0 +1,24 @@
...
+  title: 'SizeSwapper',
+};
+
+export const basic = () => {
+  const breakpoints = [
+    SizeSwapper.Breakpoint.EXTRA_SMALL,
+    SizeSwapper.Breakpoint.SMALL,
+    SizeSwapper.Breakpoint.MEDIUM,
+  ];
+  const items = [0, ...breakpoints].map((bp) => <div>Element show from {bp}px</div>);
```
helendurranttw, line 16: Typo: shown or shows

packages/components/src/sizeswapper/hooks/useClientWidth.js
```diff
@@ -0,0 +1,26 @@
+import { useLayoutEffect, useState } from 'react';
+import debounce from 'lodash.debounce';
+
+const DEBOUNCE_DELAY = 5;
+
+export const useClientWidth = ({ elRef, debouncingDelay = DEBOUNCE_DELAY }) => {
```
helendurranttw, line 6: Wait a minute.. don't you want throttle, not debounce? Yeah - you definitely want throttle, not debounce.

helendurranttw: And that would explain why you need the async, I guess!

packages/components/src/tile/Tile.js
```diff
@@ -0,0 +1,69 @@
...
+      <a
+        className="tw-tile__link text-no-decoration "
+        onClick={onClick}
+        tabIndex="0"
+        role="button"
+        onKeyDown={(e) => {
+          if (
+            e.keyCode === KeyCodes.ENTER ||
+            e.keyCode === KeyCodes.SPACE ||
+            e.code === 32 ||
```
helendurranttw, line 31: Duplicate lines

packages/components/src/tile/Tile.js
```diff
@@ -0,0 +1,69 @@
...
+  className: Types.string,
+  description: Types.oneOfType([elementOfType(['Message']), Types.string]).isRequired,
+  media: Types.node.isRequired,
+  onClick: Types.func.isRequired,
+  size: Types.oneOf([Tile.Size.LARGE, Tile.Size.SMALL]),
+  title: Types.oneOfType([elementOfType(['Message']), Types.string]).isRequired,
+};
+
+Tile.defaultProps = {
+  className: null,
```
helendurranttw, line 65: I think this should be ''

packages/components/src/tile/Tile.spec.js
```diff
@@ -0,0 +1,58 @@
...
+import { render, screen, fireEvent } from '@testing-library/react';
+
+import KeyCodes from '../common/keyCodes';
+
+import Tile from '.';
+
+describe(Tile, () => {
+  const onClick = jest.fn();
+  const defaultProps = {
+    description: 'Receive money from friends and familty like a local',
```
helendurranttw, line 11: Typo family

packages/components/src/tile/Tile.story.js
```diff
@@ -0,0 +1,31 @@
...
+import Tile from '.';
+import Avatar from '../avatar';
+
+export default {
+  component: Tile,
+  title: 'Tile',
+};
+
+export const basic = () => {
+  const description = text('description', 'Click here to send money to Mr. Miller');
```
helendurranttw, line 14: Lucky Mr Miller

packages/components/src/tile/Tile.story.js
```diff
@@ -0,0 +1,31 @@
...
+  const size = select('Size', Tile.Size, Tile.Size.LARGE);
+  const title = text('title', 'Hank Miller');
+
+  return (
+    <Tile
+      description={description}
+      media={
+        <Avatar size="md" theme="light" type="initials" className="text-xs-center">
```
helendurranttw, line 22: Same thing with the text-center

packages/docs/pages/components/content/Tile.mdx
```diff
@@ -0,0 +1,15 @@
...
+import code from '../../../liveEditorCode/tile.code';
+
+import { Tile } from '@transferwise/components';
+
+<LiveEditorBlock code={code} scope={{ Tile }} />
+<GeneratePropsTable componentName="Tile" />
+
+## Groups of tiles
+
+It's recommended to use [`Picker`](https://transferwise.github.io/neptune-web/components/content/Picker) for displaying groups of tiles.
```
helendurranttw, line 11: Needs updating

packages/docs/pages/components/navigation/Decision.mdx
```diff
@@ -0,0 +1,139 @@
...
+
+import { Decision, Avatar, Tile, NavigationOption } from '@transferwise/components';
+
+<LiveEditorBlock code={code} scope={{ Decision, Avatar }} display="vertical" />
+<GeneratePropsTable componentName="Decision" />
+
+## Usage
+
+Decision can be used for illustrating major decisions for the user.
+It is encouraged to keep the number of options in `items` under three.
```
helendurranttw, line 12: `options`

packages/docs/pages/components/navigation/Decision.mdx
```diff
@@ -0,0 +1,139 @@
...
+      <td>
+        <NavigationOption
+          id="navigation-option"
+          name="navigation-option"
+          className="custom-navigation-option"
+          title="Navigation option"
+          content="Button and icon are vertically centered."
+          onClick={() => alert('clicked')}
+          media={
+            <Avatar size="md" theme="light" type="initials" className="text-xs-center">
```
helendurranttw, line 38: More text-xs-centers

packages/components/src/decision/Decision.js
```diff
@@ -0,0 +1,99 @@
...
+              onClick={onClick}
+              showMediaAtAllSizes
+              showMediaCircle
+              title={title}
+            />,
+            <Tile
+              description={description}
+              media={block}
+              onClick={onClick}
+              size={small ? Tile.Size.SMALL : null}
```
hendrikmolder, line 40: I guess we might as well inline `presentation === LIST_BLOCK_SMALL` for now?

packages/components/src/decision/Decision.spec.js
```diff
@@ -0,0 +1,102 @@
...
+  });
+
+  describe(`when presentation is ${Decision.Presentation.LIST}`, () => {
+    beforeEach(() => {
+      resetClientWidth(500);
+      ({ container } = render(<Decision {...props} presentation={Decision.Presentation.LIST} />));
+    });
+
+    it('renders Navigation Option before breakpoint', () => {
+      expect(getNavigationOption()).toBeInTheDocument();
```
hendrikmolder, line 88: Shall we check Tile as well?
```suggestion
      expect(getNavigationOption()).toBeInTheDocument();
      expect(getTile()).not.toBeInTheDocument();
```

packages/components/src/decision/Decision.spec.js
```diff
@@ -0,0 +1,102 @@
...
+    });
+
+    it('renders Navigation Option after breakpoint', async () => {
+      resetClientWidth(900);
+      fireEvent(window, new Event('resize'));
+      await waitFor(() => {
+        expect(getNavigationOption()).toBeInTheDocument();
```
hendrikmolder, line 95: And here too?
```suggestion
        expect(getNavigationOption()).toBeInTheDocument();
        expect(getTile()).not.toBeInTheDocument();
```

packages/components/src/decision/Decision.spec.js
andreapiras: I replaced with debounce mock jest.mock('lodash.debounce', () => jest.fn((fn) => fn));

packages/components/src/tile/Tile.js
andreapiras: You are right, I normally put null because if you are not using classNames function if you use '' instead of null the dom generated is:
`<div class></div>`. => with ''
vs
`<div></div>` => with null

packages/components/src/sizeswapper/SizeSwapper.spec.js
andreapiras: mocked debounce

packages/components/src/decision/Decision.js
```diff
@@ -0,0 +1,115 @@
...
+const DecisionPresentantion = {
+  LIST: 'LIST',
+  LIST_BLOCK: 'LIST_BLOCK',
+  LIST_BLOCK_SMALL: 'LIST_BLOCK_SMALL',
+};
+
+const Decision = ({ options, presentation, type }) => {
+  if (type === DecisionType.NAVIGATION) {
+    const { LIST_BLOCK, LIST_BLOCK_SMALL } = DecisionPresentantion;
+    if (presentation === LIST_BLOCK || LIST_BLOCK_SMALL) {
```
helendurranttw, line 19: This won't work, I think you mean

```suggestion
    if (presentation === LIST_BLOCK || presentation === LIST_BLOCK_SMALL) {
```

packages/components/src/decision/Decision.js
```diff
@@ -0,0 +1,115 @@
...
+          layout: SizeSwapper.Layout.COLUMN,
+        },
+        {
+          items: [],
+          breakpoint: SizeSwapper.Breakpoint.MEDIUM,
+        },
+      ];
+
+      options.forEach(
+        ({ description, disabled, href, media: { block, list }, onClick, title }, key) => {
```
helendurranttw, line 32: I see and appreciate the alphabetisation here 😄

packages/components/src/decision/Decision.js
```diff
@@ -0,0 +1,115 @@
...
+          items[0].items.push(
+            <NavigationOption
+              content={description}
+              disabled={disabled}
+              href={href}
+              // eslint-disable-next-line react/no-array-index-key
+              key={`nav-${key}`}
```
helendurranttw, line 39: `no-array-index-key` is there for a reason - how about a generated id, or using the href?

packages/components/src/decision/Decision.js
```diff
@@ -0,0 +1,115 @@
...
+              media={list}
+              onClick={onClick}
+              title={title}
+            />,
+          );
+          items[1].items.push(
+            <Tile
+              className={key % 2 !== 0 ? 'm-x-3' : null}
```
helendurranttw, line 47: What's happening here?

packages/components/src/sizeswapper/SizeSwapper.js
```diff
@@ -0,0 +1,51 @@
...
+  const [clientWidth] = useClientWidth({ ref });
+
+  // If all breakpoints are specified and clientWidth never > breakpoint itemsToRender can be undefined.
+  // Do not use deconstruct here to get items and layout.
+  const itemsToRender = items
+    .filter(({ breakpoint = 0 }) => clientWidth >= breakpoint)
+    .slice(-1)[0];
+
+  // Always return parent container even if there are no items to display to
+  // keep the ref on DOM and let clientWidth be calculated properly.
```
helendurranttw, line 24: Useful ✅

packages/components/src/sizeswapper/hooks/useClientWidth.js
```diff
@@ -0,0 +1,26 @@
+import { useLayoutEffect, useState } from 'react';
+import throttle from 'lodash.throttle';
+
+const DEBOUNCE_DELAY = 10;
```
helendurranttw, line 4: Throttle

helendurranttw: You could also bump up to at least 100 I'd imagine

packages/components/src/tile/Tile.js
```diff
@@ -0,0 +1,63 @@
...
+  return (
+    <div
+      className={classNames('np-tile', 'd-flex', 'flex-column', 'text-xs-center', className, {
+        'np-tile--small p-a-2': size === Tile.Size.SMALL,
+        'np-tile--large p-y-5 p-x-4': size !== Tile.Size.SMALL,
+        disabled,
+      })}
+    >
+      <div className="np-tile__media d-flex justify-content-center">{media}</div>
+      <button
```
helendurranttw, line 20: Why a button and not a link?

packages/components/src/tile/Tile.less
```diff
@@ -0,0 +1,70 @@
...
+      &::after {
+        cursor: not-allowed;
+      }
+    }
+  }
+
+  &:not(.disabled) {
+    &:hover,
+    &:focus-within {
+      border: 1px solid var(--color-border-focus);
```
helendurranttw, line 62: Looks nice but it's different from the focus styling on the navigation option 😬 Is this the one Gusso advised to move ahead with?

packages/docs/pages/components/navigation/Decision.mdx
```diff
@@ -0,0 +1,139 @@
...
+
+import { Decision, Avatar, Tile, NavigationOption } from '@transferwise/components';
+
+<LiveEditorBlock code={code} scope={{ Decision, Avatar }} display="vertical" />
+<GeneratePropsTable componentName="Decision" />
+
+## Usage
+
+Decision can be used for illustrating major decisions for the user.
+It is encouraged to keep the number of options in `options` under three.
```
helendurranttw, line 12:
```suggestion
It is encouraged to keep the number of options in `options` under four.
```

packages/components/src/decision/Decision.js
andreapiras: oops pushed this by error

packages/components/src/decision/Decision.js
andreapiras: I think in this case it's pretty safe as key must only be unique among siblings which means they don't need to be unique globally. We need to be careful and investigate about using generated id, as react expects stable ids to avoid unnecessary re-renders. I can use href here but I see this topic recurring so I think we should create a pattern solution and apply everywhere

andreapiras: If for any reason we have two identical href (evenif by mistake) the render will be wrong

packages/components/src/tile/Tile.js
andreapiras: I didn't know it could accept an href , reverted to a

packages/components/src/tile/Tile.less
andreapiras: Removed this and  used media and decision class as it's used in navigationOption and let that manage the styles.

packages/components/src/decision/Decision.less
```diff
@@ -0,0 +1,4 @@
+.np-decision {
```
andreapiras, line 1: This compensate the m-x-1 of the tiles

packages/components/src/decision/Decision.js
```diff
@@ -0,0 +1,114 @@
...
+      />
+    ));
+  }
+  return <></>;
+};
+
+Decision.Presentation = {
+  LIST: 'LIST',
+  LIST_BLOCK: 'LIST_BLOCK',
+  LIST_BLOCK_SMALL: 'LIST_BLOCK_SMALL',
```
mikknarusk, line 79: what's the reason for having the `size` be included in the `presentation` prop? Aren't they usually separated?

packages/components/src/sizeswapper/SizeSwapper.js
```diff
@@ -0,0 +1,51 @@
+import React, { useRef } from 'react';
```
oliverviljamaa, line 1: Should the directory casing be `sizeSwapper`?

packages/components/src/decision/Decision.js
```diff
@@ -0,0 +1,114 @@
...
+import Types from 'prop-types';
+import requiredIf from 'react-required-if';
+import classNames from 'classnames';
+
+import SizeSwapper from '../sizeswapper';
+import NavigationOption from '../navigationOption';
+import Tile from '../tile';
+import './Decision.css';
+
+const Decision = ({ options, presentation, type }) => {
```
oliverviljamaa, line 11: Where can I learn more about the decision (no pun intended) to use separate components, rather than style the navigation options as tiles on larger viewports? I'm not against it, would just like to understand better.

packages/components/src/decision/Decision.js
```diff
@@ -0,0 +1,114 @@
...
+        );
+        items[1].items.push(
+          <Tile
+            description={description}
+            disabled={disabled}
+            href={href}
+            key={`tile-${href}`}
+            media={block}
+            onClick={onClick}
+            size={presentation === LIST_BLOCK_SMALL ? Tile.Size.SMALL : Tile.Size.LARGE}
```
oliverviljamaa, line 47: Where do we use the small tiles in the product?

packages/components/src/sizeswapper/SizeSwapper.js
```diff
@@ -0,0 +1,51 @@
...
+    >
+      {itemsToRender && itemsToRender.items}
+    </div>
+  );
+};
+
+SizeSwapper.Breakpoint = Breakpoint;
+SizeSwapper.Layout = Layout;
+
+SizeSwapper.propTypes = {
```
oliverviljamaa, line 40: Was this interface modelled after something? From another component library or something?

packages/components/src/decision/Decision.js
```diff
@@ -0,0 +1,114 @@
...
+        {
+          items: [],
+          breakpoint: SizeSwapper.Breakpoint.MEDIUM,
+        },
+      ];
+
+      options.forEach(({ description, disabled, href, media: { block, list }, onClick, title }) => {
+        items[0].items.push(
+          <NavigationOption
+            content={description}
```
oliverviljamaa, line 30: Should this have `complex={false}`, like the `LIST`, as well?

packages/components/src/decision/Decision.story.js
```diff
@@ -0,0 +1,94 @@
...
+  );
+  const disabled = boolean('disabled', false);
+
+  return (
+    <>
+      <Decision
+        options={[
+          {
+            description:
+              "Click here to send money to Hank Miller. Money will be sent directly to Han Miller's multi-currency account.",
```
oliverviljamaa, line 27:
```suggestion
              "Click here to send money to Hank Miller. Money will be sent directly to Hank Miller's multi-currency account.",
```
(elsewhere, too)

packages/components/src/index.js
```diff
@@ -53,4 +53,6 @@ export { default as Markdown } from './markdown';
 export { default as Card } from './card';
 export { default as Summary } from './summary';
 export { default as Switch } from './switch';
+export { default as Tile } from './tile';
```
oliverviljamaa, line 56: Do we need to expose this at all? When do we want to use tiles, not decisions?

packages/components/src/sizeswapper/SizeSwapper.js
andreapiras: Nope, we build it from scratch why?

StevePole:
Should we invert the avatar circle background on hover so it doesn't disappear?
image

packages/components/src/decision/Decision.js
```diff
@@ -0,0 +1,114 @@
...
+  }
+  return <></>;
+};
+
+Decision.Presentation = {
+  LIST: 'LIST',
+  LIST_BLOCK: 'LIST_BLOCK',
+  LIST_BLOCK_SMALL: 'LIST_BLOCK_SMALL',
+};
+Decision.Type = { NAVIGATION: 'NAVIGATION' };
```
dozortsev, line 81: why only one type? I guess for one we don't need this to be a prop at all

packages/components/src/sizeswapper/SizeSwapper.story.js
```diff
@@ -0,0 +1,36 @@
...
+    SizeSwapper.Breakpoint.EXTRA_LARGE,
+  ];
+
+  const items = breakpoints.map((bp, key) => {
+    const isColumn = (key + 1) % 2 === 0;
+
+    return {
+      items: [...Array(4)].map((val, key2) => (
+        // eslint-disable-next-line react/no-array-index-key
+        <div className="p-a-3 well m-x-2" key={`el-${key2}-${key}`}>
```
helendurranttw, line 26: What if you had two size swappers on the screen? These keys will not be unique.

packages/components/src/sizeswapper/hooks/useClientWidth.js
```diff
@@ -0,0 +1,26 @@
+import { useLayoutEffect, useState } from 'react';
+import throttle from 'lodash.throttle';
+
+const DEBOUNCE_DELAY = 100;
```
helendurranttw, line 4: Throttle ✨

packages/components/src/sizeswapper/hooks/useClientWidth.js
```diff
@@ -0,0 +1,26 @@
...
+
+    updateClientWidth();
+
+    return () => window.removeEventListener('resize', updateClientWidth);
+  }, []);
+
+  return [clientWidth];
+};
+
+useClientWidth.DEBOUNCE_DELAY = DEBOUNCE_DELAY;
```
helendurranttw, line 26: Throttle ✨

packages/components/src/sizeswapper/SizeSwapper.js
dozortsev: I guess @oliverviljamaa want learn more about this, and proof himself that this is best solution and perhaps gain some confidence in it 🙂

packages/components/src/decision/Decision.js
hendrikmolder: I think there was an idea to use them for recipients (in send flow?)

packages/components/src/sizeswapper/SizeSwapper.js
dozortsev: I'm not really a fan of this solution, I'm not sure this is solid and long term solution...

Why creating components for only few specific cases (from Tile -> *Option), I would much rather provide **API which is use case agnostic**, API which allows consumer easily define when to switch from on component to another

Have you looked into libraries like https://www.npmjs.com/package/react-responsive? Not saying we should use it and alike libs but IMHO but their approach look much more convenient, easier to understand.

dozortsev (commented):
https://github.com/transferwise/neptune-web/pull/498/files#r520575672

packages/components/src/sizeswapper/SizeSwapper.story.js
andreapiras: This is just an example in the story, I don't think we'll add more but I can still change it

packages/components/src/sizeswapper/SizeSwapper.js
andreapiras: The problem we were trying to solve is slightly different from the one solved in react-responsive as we want to adapt the content to a parent with rather than a windows width. The closest analogy would be [css-element-queries](https://github.com/marcj/css-element-queries) which has 44 issues open and some limitations due to the fact that part of its logic rely on css which makes it hard to share values between css and js

packages/components/src/decision/Decision.js
andreapiras: Hey @oliverviljamaa, can you please give me more details? At the moment Decision only provides the behavior that you just mentioned and nothing else. Were you thinking about using it with other components?

packages/components/src/decision/Decision.js
andreapiras: I think also in send money flow for new recipients. Please check in the slack channel

packages/components/src/sizeswapper/SizeSwapper.js
dozortsev: > ..we want to adapt the content to a parent with rather than a windows width..

could you please explain this more perhaps with an example? and why you want this?

hendrikmolder:
Should small tile be _smaller_ than the large one? I think it was before...

Screenshot 2020-11-11 at 10 02 52

andreapiras:
> Should small tile be _smaller_ than the large one? I think it was before... maybe something has broken
> 
> Screenshot 2020-11-11 at 10 02 52

Hey @hendrikmolder , with @gusso we decided to remove the size of the boxes to make them fluid and adapt better to spacing so the diff between large and small is in the font-size and spacing between title/desc/media

packages/components/src/sizeswapper/SizeSwapper.js
andreapiras: I think this is the perfect example. FlowNavigation's elements currently switch to mobile version on window sm breakpoint. This will work perfectly if FN container width (green) is equal to window width (blue)

image
but if the container is not full width the responsive behaviour (base on the width size) won't be as good

image
If FN would respond to green size rather than blue the layout will work with or without sidebar. Hope it makes sense, if not please DM and we can go through more in details
image

dozortsev: okey, thanks,

we still should be able to implement "use case agnostic solution" where API consumer could provider ref to parent (`window` or some component), wdyt?

andreapiras: If you are using window I'd probably use the media-queries and normal breakpoints. This component assume ref is the ref of the sizeSwapper I cannot imagine a scenario where you want to switch content in an element based on something that is not the parent. You can still use `useClientWidth` hook to monitor any other size if that scenario occurs

packages/components/src/decision/Decision.js
andreapiras: The way we base on what's shown for breakpoint in this case:
LIST: LIST -> Mobile, LIST-> Desktop
LIST_BLOCK: LIST -> Mobile, BLOCK -> Desktop
LIST_BLOCK_SMALL: LIST -> Mobile, BLOCK_SMALL-> Desktop

It's referring to the BLOCK only rather than both elements. Are you suggesting to pass a `size` prop ?

packages/components/src/sizeSwapper/SizeSwapper.spec.js
```diff
@@ -0,0 +1,51 @@
...
+    const { getByText, queryByText } = render(<SizeSwapper items={items} />);
+
+    expect(getByText('element-1-0')).toBeInTheDocument();
+    expect(queryByText('element-2-0')).toBeInTheDocument();
+
+    expect(queryByText(`element-1-${breakpoints[1]}`)).not.toBeInTheDocument();
+    expect(queryByText(`element-2-${breakpoints[1]}`)).not.toBeInTheDocument();
+
+    resetClientWidth(breakpoints[1]);
+    fireEvent(window, new Event('resize'));
```
dozortsev, line 44: why here we dispatch event to `window` why not to SizeSwapper?

andreapiras: the evenListener inside `useClientWidth` is attached to the window so you need to fire an window resize to trigger the clientWidth calculations

packages/components/src/sizeswapper/SizeSwapper.js
dozortsev: I would suggest change API of SizeSwapper to something like following, where consumer set parent element to [context component](https://reactjs.org/docs/context.html#contextconsumer) and then tell _when_ and _what_ to render:

```jsx
const currentParent = useRef(null);

...
<ElementMedia ref={currentParent}>
  (elementWidth) => {
    if (elementWidth > Breakpoint.EXTRA_SMALL) {
      return (
        <div className="flex-row">
          <NavigationOption ... />
          <NavigationOption ... />
          ...
        </div>
      )
    } else if (elementWidth > Breakpoint.SMALL) {
      return <WhatEverComponent className="flex-column" />
    } else {
      return (
        <div className="flex-column">
          <Tile ... />
          <Tile ... />
          ...
        </div>
      )
    }
  }
</ElementMedia>
```

Note: ElementMedia component uses `useClientWidth` hook

IMHO this is more intuitive approach, consumer almost won't need learn any API, the API of ElementMedia is pretty simple (compare to [SizeSwapper one](https://github.com/transferwise/neptune-web/pull/498/files#diff-1374b695c71a82de34aebf21e745e36e9377a36fc750823ac70da33f7fb0fc05R42-R48)).

re. ElementMedia component, if you don't like the name we can think of better one 🙂

dozortsev: or actually we seem don't need `useClientWidth` hook, [it's logic](https://github.com/transferwise/neptune-web/pull/498/files#diff-d90d88f58418bc79762e6f7593921ba986da30856a1979cae8b3bad5bfe1c545R7-R21) could be inside ElementMedia component

packages/components/src/decision/Decision.js
dozortsev: by wrapping logic of switching from Tile to *Option in component I don't think we simple consumers life, actually vise-versa, now instead of just using components Tile and e.g NavigationOption they have deal with it's abstraction, which they have to learn how to use 🙂 this component brings more complexity than value.

you mentioned that you want enforce this as a pattern, I would suggest we expose API which help consumers switch components bases on parents element width (either SizeSwapper, ElementMedia, or useClinetWidth hook) and _document_ this pattern rather than _implement_ it as some component.

packages/components/src/sizeSwapper/SizeSwapper.spec.js
dozortsev: ok

packages/components/src/tile/Tile.less
```diff
@@ -0,0 +1,50 @@
...
+  &__title {
+    margin-top: var(--size-16);
+    font-size: 1.333333rem; // 20px
+    line-height: calc(24 / 20); // 24px
+    color: var(--color-text-primary);
+  }
+
+  &__description {
+    margin-top: var(--size-12);
+    // font-size: @font-size-base; //16px
```
helendurranttw, line 25: Remove comment

packages/components/src/tile/Tile.story.js
```diff
@@ -0,0 +1,51 @@
...
+        href="#href1"
+        media={
+          <img
+            src="https://transferwise.com/public-resources/assets/bank-details/bank-details-flow/finish.svg"
+            alt="illustration"
+            style={{ width: '100%' }}
+          />
+        }
+        onClick={action('onClick')}
+        size={large ? Tile.Size.LARGE : Tile.Size.SMALL}
```
helendurranttw, line 33: Update to MEDIUM / SMALL

packages/components/src/tile/Tile.story.js
```diff
@@ -0,0 +1,51 @@
...
+        description={description}
+        disabled={disabled}
+        href="#href1"
+        media={
+          <Avatar size="md" type="initials">
+            HM
+          </Avatar>
+        }
+        onClick={action('onClick')}
+        size={large ? Tile.Size.LARGE : Tile.Size.SMALL}
```
helendurranttw, line 46: Update to MEDIUM / SMALL

packages/components/src/tile/Tile.js
```diff
@@ -0,0 +1,66 @@
...
+Tile.Size = Size;
+
+Tile.propTypes = {
+  className: Types.string,
+  description: Types.node,
+  disabled: Types.bool,
+  href: Types.string.isRequired,
+  media: Types.node.isRequired,
+  onClick: Types.func.isRequired,
+  size: Types.oneOf([Tile.Size.LARGE, Tile.Size.SMALL]),
```
helendurranttw, line 55: Update to MEDIUM / SMALL

packages/docs/liveEditorCode/decision.code.js
```diff
@@ -0,0 +1,71 @@
+() => (
+  <Decision
+    options={[
+      {
+        description:
+          "Click here to send money to Hank Miller. Money will be sent directly to Han Miller's multi-currency account.",
```
helendurranttw, line 6:
```suggestion
          "Click here to send money to Hank Miller. Money will be sent directly to Hank Miller's multi-currency account.",
```

packages/docs/liveEditorCode/decision.code.js
```diff
@@ -0,0 +1,71 @@
...
+            <Avatar size="md" type="initials">
+              HM
+            </Avatar>
+          ),
+        },
+        title: 'Hank Miller',
+      },
+      {
+        description:
+          "Click here to send money to Hank Miller. Money will be sent directly to Han Miller's multi-currency account.",
```
helendurranttw, line 27:
```suggestion
          "Click here to send money to Hank Miller. Money will be sent directly to Hank Miller's multi-currency account.",
```

packages/docs/liveEditorCode/decision.code.js
```diff
@@ -0,0 +1,71 @@
...
```
helendurranttw, line 48:
```suggestion
          "Click here to send money to Hank Miller. Money will be sent directly to Hank Miller's multi-currency account.",
```

packages/docs/pages/components/navigation/Decision.mdx
```diff
@@ -0,0 +1,14 @@
+import { LiveEditorBlock, GeneratePropsTable } from '../../../utils';
+import code from '../../../liveEditorCode/decision.code.js';
+
+import { Decision, Avatar } from '@transferwise/components';
+
+Decision can be used for illustrating major decisions for the user.
+It is encouraged to keep the number of options in `options` under four.
```
helendurranttw, line 7:
```suggestion
Because blocks will not wrap onto multiple lines, we suggest keeping the number of options under four when using presentation `LIST_BLOCK`.
```

packages/components/src/decision/Decision.js
andreapiras: Hey @oliverviljamaa so at the beginning we though the Tile could have been a component on its own. We found out that this is not the case and it would make total sense for it to be incorporated NavigationOption. To do that we need to refactor a bit NavigationOption and we'll come back to this later.

helendurranttw (approved):

PR #505, opened 2020-07-24
fix: fix icon alignment and move component specific CSS into components

helendurranttw:
## ✅ Checklist


packages/components/src/checkboxButton/CheckboxButton.js
```diff
@@ -2,6 +2,8 @@ import React from 'react';
 import Types from 'prop-types';
 import classNames from 'classnames';
 
+import './Checkbox.css';
```
dozortsev, line 5: should be `CheckboxButton.css`

PR #506, opened 2020-07-29
feat(pagination): add component

laczikt:

## 🖼 Context

Pagination is still being used on the History and Recipients pages, this component aims to get rid of code duplication in these repos. Link to the component in Bootstrap: [https://bootstrap.transferwise.com/components/#pagination](https://bootstrap.transferwise.com/components/#pagination)

## 🚀 Changes

 

I've added the Pagination component along with tests, documentation, and a storybook entry.

## ✅ Checklist


packages/components/src/pagination/Pagination.js
```diff
@@ -0,0 +1,135 @@
...
+  FULL: 'full',
+};
+
+const Pagination = ({
+  display,
+  numberOfPages,
+  currentPageIndex,
+  onSelectPage,
+  canFetchMorePages,
+  inverse,
```
dozortsev, line 21: each component should have [`className` prop](https://github.com/transferwise/neptune-web/blob/master/packages/docs/pages/rfc/0001-always-pass-classname.md), see examples in [Summary](https://github.com/transferwise/neptune-web/blob/master/packages/components/src/summary/Summary.js), [NavigationOption](https://github.com/transferwise/neptune-web/blob/master/packages/components/src/navigationOption/NavigationOption.js)..

packages/components/src/pagination/Pagination.js
```diff
@@ -0,0 +1,135 @@
...
+  ARROWS_ONLY: 'arrowsOnly',
+  FULL: 'full',
+};
+
+const Pagination = ({
+  display,
+  numberOfPages,
+  currentPageIndex,
+  onSelectPage,
+  canFetchMorePages,
```
dozortsev, line 20: sorry, could you please explain for what use case this prop for?

packages/components/src/pagination/Pagination.js
```diff
@@ -0,0 +1,135 @@
...
+import './Pagination.css';
+
+const Display = {
+  NUMBERS_ONLY: 'numbersOnly',
+  ARROWS_ONLY: 'arrowsOnly',
+  FULL: 'full',
+};
+
+const Pagination = ({
+  display,
```
dozortsev, line 16: I would name `type` or [`mode` like in DateInput](https://github.com/transferwise/neptune-web/blob/master/packages/components/src/dateInput/DateInput.js#L25)

packages/components/src/pagination/Pagination.js
```diff
@@ -0,0 +1,135 @@
...
+  numberOfPages,
+  currentPageIndex,
+  onSelectPage,
+  canFetchMorePages,
+  inverse,
+}) => {
+  const displayArrows = display !== Display.NUMBERS_ONLY;
```
dozortsev, line 23: I would use word "show" here (`displayArrows`), [we tend to use it more for such vars](https://github.com/transferwise/neptune-web/search?p=1&q=%22show%22&unscoped_q=%22show%22)

packages/components/src/pagination/Pagination.js
```diff
@@ -0,0 +1,135 @@
...
+        }
+        if (numberOfPages > 1) {
+          pages.push({ index: numberOfPages - 2 });
+        }
+        pages.push({ index: numberOfPages - 1, active: true });
+        break;
+      default:
+        pages.push({ index: currentPageIndex - 1 });
+        pages.push({ index: currentPageIndex, active: true });
+        pages.push({ index: currentPageIndex + 1 });
```
dozortsev, line 51: JFYI [`.push`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push) function accepts multiple values which you want to add in the array

packages/components/src/pagination/Pagination.spec.js
```diff
@@ -0,0 +1,127 @@
+import React from 'react';
+import { shallow } from 'enzyme';
```
dozortsev, line 2: 😬 

please use RTL for testing components, try to get rid of Enzyme

https://github.com/transferwise/neptune-web/blob/master/packages/components/CONTRIBUTING.md#testing

packages/components/src/pagination/Pagination.js
laczikt: When this prop is true, an additional "..." button is added behind the last page number. An example for using this is the recipients page, where we have no way of knowing how many recipients the user has in total, therefore we don't know the total number of pages before reaching the last one. (When fetching a page, we only know if there are more recipients or not, therefore if we can fetch more pages)

packages/components/src/pagination/Pagination.less
```diff
@@ -0,0 +1,98 @@
+@import '@transferwise/neptune-tokens/colors-base.less';
```
helendurranttw, line 1: You can not bother about importing these and use custom properties instead: [Colours](https://transferwise.github.io/neptune-web/styles/tokens/Colors) and [Spacing](https://transferwise.github.io/neptune-web/styles/tokens/Spacing). They'll be automatically assigned fallbacks as part of the build process.

packages/components/src/pagination/Pagination.less
```diff
@@ -0,0 +1,98 @@
...
+@import '@transferwise/neptune-tokens/spacing.less';
+
+.pagination {
+  display: inline-block;
+  padding-left: 0;
+  margin-bottom: @space-24;
+
+  > li {
```
helendurranttw, line 9: You don't need such specific selectors when we know the markup is pretty limited in what it's ever going to contain :)

packages/components/src/pagination/Pagination.less
```diff
@@ -0,0 +1,98 @@
...
+      float: left; // Collapse white-space
+      line-height: (@space-24 * 2);
+      height: (@space-24 * 2);
+      text-decoration: none;
+      color: @color-base-navy-mid;
+      margin: 0 4px;
+      width: (@space-24 * 2);
+      text-align: center;
+      border-radius: 50%;
+      font-weight: normal;
```
helendurranttw, line 23: Quite a few of these props seem unnecessary

packages/components/src/pagination/Pagination.less
```diff
@@ -0,0 +1,98 @@
...
+      outline: none;
+    }
+  }
+
+  > .active > button,
+  > .active > span {
+    &,
+    &:hover,
+    &:focus {
+      z-index: 2;
```
helendurranttw, line 56: We should call out the active page in a different way from the focussed elements. Maybe make it bold?

packages/components/src/pagination/Pagination.js
```diff
@@ -0,0 +1,140 @@
...
+          </button>
+        </li>
+      )}
+
+      {pages.map((page) => {
+        const label = page.label || page.index + 1;
+        return (
+          <li key={page.index} className={classNames({ active: page.active })}>
+            {page.active ? (
+              <span>{label}</span>
```
helendurranttw, line 98: I think it will be cleaner to (once you've changed to links) have all items as links, even when they're not selectable. You can just disable them, or make the onclick handler ignore them. This will mean you don't need to apply interaction styles to spans in the CSS

packages/components/src/pagination/Pagination.less
```diff
@@ -0,0 +1,98 @@
...
+    }
+  }
+
+  > li > button {
+    &:hover,
+    &:focus {
+      z-index: 3;
+      color: @color-base-navy-mid;
+      background-color: @color-base-smoke-fade;
+      outline: none;
```
helendurranttw, line 47: We shouldn't be turning off the focus styles - they're important for accessibility. We can just leave the browser default.

packages/components/src/pagination/Pagination.less
```diff
@@ -0,0 +1,98 @@
...
+    &:last-child > button,
+    &:last-child > span {
+      margin-right: 0;
+    }
+  }
+
+  > li > button {
+    &:hover,
+    &:focus {
+      z-index: 3;
```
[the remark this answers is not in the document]
laczikt: I took it from the legacy bootstrap stylesheet, but it really seems unnecessary here 🤔

packages/components/src/pagination/Pagination.js
```diff
@@ -0,0 +1,140 @@
...
+          { index: currentPageIndex, active: true },
+          { index: currentPageIndex + 1 },
+        );
+        break;
+    }
+
+    const firstPageIndex = pages[0].index;
+    if (firstPageIndex > 0) {
+      if (firstPageIndex > 2) {
+        pages.unshift({ label: '...', index: firstPageIndex - 1 });
```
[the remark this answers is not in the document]
laczikt: I was playing around with it a bit, but I think that keeping the active state and the label in the same object as the index is a more reliable solution, and expresses it more clearly that these are closely related properties

PR #516, opened 2020-08-10
feat(Tabs): add non-full width headers option

ksunandp:

## 🖼 Context
From slack conversation - https://transferwise.slack.com/archives/CJNBX9CP6/p1596458576365500
We (activity team) want to use the `Tabs` component but without full width headings. Design is in the slack conversation above.

## 🚀 Changes
Allows a new prop `headerWidth` to be supplied to the component to make tab headers either full width or width based on its content. 
To calculate `state.translateX` with the new feature. I've implemented using `refs`. These are created on the parent component and forwarded onto each tab. With this, we can calculate each individual tabs width and make the appropriate calculations for animation.
 

## 🤔 Considerations
The default for the prop `headerWidth` is as current implementation (full width headers).

## ✅ Checklist


packages/components/src/tabs/Tabs.js
```diff
@@ -37,9 +37,11 @@ class Tabs extends React.Component {
     isSwiping: false,
     isScrolling: false,
     lastSwipeVelocity: 0,
+    fullWidthTabs: this.props.headerWidth === 'auto',
```
dozortsev, line 40: the implementation should be a bit vice-versa I think

`block` value should mean full width, like `block={true | false}` prop in [Button component](https://transferwise.github.io/neptune-web/branch/AT-1693-tabs-width/components/inputs/Button) (when it's `true` the button has full width in the container) while `auto` is width based on header text

ksunandp: Makes sense, will switch them

packages/components/src/tabs/Tabs.js
```diff
@@ -399,7 +458,13 @@ class Tabs extends React.Component {
...
               className={classNames('tabs__line')}
               style={{
-                width: `${(1 / tabs.length) * 100}%`,
+                width: fullWidthTabs
+                  ? `${(1 / tabs.length) * 100}%`
+                  : `${
+                      (
+                        this.tabRefs[selected] || this.tabRefs[this.tabRefs.length - 1]
+                      ).getBoundingClientRect().width
+                    }px`,
```
dozortsev, line 467: this logic would be better to be extracted in function to keep JSX simpler

packages/components/src/tabs/Tab.js
```diff
@@ -2,51 +2,54 @@ import React, { useEffect, useRef } from 'react';
...
-      onKeyDown={disabled ? null : onKeyDown}
-      onClick={onClick}
-      style={style}
-    >
-      {children}
-    </li>
-  );
-};
+const Tab = React.forwardRef(
+  ({ children, id, disabled, panelId, selected, onKeyDown, onClick, style, focusTab }, ref) => {
```
dozortsev, line 6: could you please explain this change (`React.forwardRef`) not that I'm again it but just don't see how this related to non-full width headers and there is not info about it in PR description

packages/components/src/tabs/Tabs.js
```diff
@@ -116,6 +127,45 @@ class Tabs extends React.Component {
...
+    return this.tabRefs
+      .map((ref) => {
+        return ref.getBoundingClientRect().width;
+      })
+      .reduce((a, b) => a + b, 0);
+  };
+
+  getDistanceToSelectedTab = (index) => {
+    return this.tabRefs
+      .filter((_, idx) => idx < index)
```
dozortsev, line 140: if you rename the above variable (`index`) it would also make sense to avoid using underscore here in a following way:

```js
.filter(({ index }) => index < selectedTabIndex)
```

packages/components/src/tabs/Tab.js
ksunandp: The way the component is built is that the highlighted line under active tab uses `transform: translateX(n)`. When in full width, that's easy to determine `${index * 100}%` however with auto width headers, you don't know the width of each header. So creating a ref on the parent component for each header and forwarding it onto the tab, we can calculate the header width and calculate the distance the highlighted line needs to translate by.

ksunandp: Added a bit to the PR description

ksunandp:
@dozortsev Made changes based on your review

PR #520, opened 2020-08-12
fix: delete redundant .media + .media

dozortsev:
## 🚀 Changes

See comments below with description.

## 🤔 Considerations

All React components which somehow uses `.media` class are not affected:

- [Accordion](https://transferwise.github.io/neptune-web/branch/delete-media/components/content/Accordion)
- [Card](https://transferwise.github.io/neptune-web/branch/delete-media/components/content/Card)
- [CheckboxOption](https://transferwise.github.io/neptune-web/branch/delete-media/components/content/CheckboxOption)

## ✅ Checklist


packages/css/src/less/media.less
```diff
@@ -8,10 +8,6 @@
   display: flex;
 }
 
-.media + .media {
-  margin-top: calc(var(--spacer) * 3);
-}
```
dozortsev, line 13: this fix removes gap, for context: https://github.com/transferwise/verification-components/pull/233#issuecomment-671851585

Screen Shot 2020-08-12 at 12 12 55 PM

helendurranttw:
I'm a bit concerned we might break other instances of people using components with .media. Have you looked through styleguide-components to see where they use media in other components?

dozortsev:
@helendurranttw just have done, there are fine, those two components use `.media` class, here is them using `neptune.css` with out the rule:

Screen Shot 2020-08-12 at 1 44 56 PM

Screen Shot 2020-08-12 at 1 44 21 PM

here is with the rule, so as you can see it's not even gets applied:

Screen Shot 2020-08-12 at 1 53 07 PM

StevePole:
I can't think of a reason for this rule to exist.  Generally speaking media elements should be contiguous.  If there is a case where they should not be, that should be the exception and there should be a special rule to provide space in that case.

PR #522, opened 2020-08-13
fix: Chevron in Select

dozortsev:

## ✅ Checklist


packages/css/src/less/select.less
```diff
@@ -29,10 +29,25 @@
     }
   }
 
+  .dropdown-toggle.btn-input {
+    &.btn-sm {
+      padding-right: calc(var(--space-8) * 4);
```
anttio, line 34: No need to do the calc() as we have spacing tokens.

```suggestion
      padding-right: var(--space-32);
```

packages/css/src/less/select.less
```diff
@@ -29,10 +29,25 @@
...
+    }
+
+    &.btn-md {
+      padding-right: calc(var(--space-8) * 5);
```
anttio, line 38:
```suggestion
      padding-right: var(--space-40);
```

packages/css/src/less/select.less
```diff
@@ -29,10 +29,25 @@
...
+    }
+
+    &.btn-lg {
+      padding-right: calc(var(--space-8) * 6);
```
anttio, line 42:
```suggestion
      padding-right: var(--space-48);
```

packages/css/src/less/select.less
dozortsev: thanks @anttio

PR #524, opened 2020-08-14
feat: add a layout service for converting dynamic flow steps without layouts to DynamicLayout format

StevePole:
## 🖼 Context

Layout is not a required property of dynamic flow steps so we need to handle layout of two different approaches.  Rather than do this, we will simply map steps without layouts to layout format using a service.  From then on, we can treat all dynamic flow steps as layouts and we need only one approach for the layout of our template.

## 🚀 Changes

We add a service that maps three step types (form, decision & final) to a dynamic layout specification

## 🤔 Considerations

The DynamicFlow component that will use the service is yet to be written.

We do not mimic the logic from V2 that narrows/widens the form based on the width of the fields inside the form, as in V3, it is not straightforward to interrogate the fields. 

We do not convert "upload steps" into side by side layouts, as in V2.  The intent is to add a layout to the step to achieve these more complex layouts.

## ✅ Checklist

### DIFF
--- a/packages/dynamic-flows/src/flow/layoutService/LayoutService.js
+++ b/packages/dynamic-flows/src/flow/layoutService/LayoutService.js
@@ -1,0 +1,176 @@
+function convertStepToLayout(step) {
+  switch (step.type) {
+    case 'final':
+      return convertFinalStepToDynamicLayout(step);
+    case 'decision':
+      return convertDecisionStepToDynamicLayout(step);
+    case 'form':
+      return convertFormStepToDynamicLayout(step);
+    default:
+      throw new Error('invalid step type');
+  }
+}
+
+function convertCommonComponents(step) {
+  const layout = [];
+
+  if (step.title) {
+    layout.push(convertStepTitleToDynamicHeading(step.title));
+  }
+  if (step.image) {
+    const image = convertStepImageToDynamicImage(step.image);
+    layout.push(dynamicBox([image]));
+  }
+  if (step.description) {
+    layout.push(convertStepDescriptionToDynamicParagraph(step.description));
+  }
+
+  return layout;
+}
+
+function convertFormStepToDynamicLayout(step) {
+  let layout = convertCommonComponents(step);
+
+  if (step.reviewFields) {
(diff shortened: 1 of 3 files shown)


packages/dynamic-flows/src/flow/layoutService/LayoutService.spec.js
```diff
@@ -0,0 +1,237 @@
...
+        data: {
+          exitValue: 'value',
+        },
+      };
+
+      const finalStep = {
+        type: 'final',
+        key: 'thing-final',
+        title: 'We create the thing!',
+        description: 'You now do stuff with the thing',
```
helendurranttw, line 81: 💪

packages/dynamic-flows/src/flow/layoutService/LayoutService.js
```diff
@@ -0,0 +1,177 @@
...
+    layout.push(convertStepDecisionToDynamicDecision(step.options));
+  }
+
+  return layout;
+}
+
+function dynamicBox(components, size) {
+  return {
+    type: 'box',
+    width: size || 'sm',
```
helendurranttw, line 74: Should this default to sm? Not md?

packages/dynamic-flows/src/flow/layoutService/LayoutService.js
```diff
@@ -0,0 +1,177 @@
...
+
+function convertFormStepToDynamicLayout(step) {
+  let layout = convertCommonComponents(step);
+
+  if (step.reviewFields) {
+    layout.push(convertStepReviewToDynamicReview(step.reviewFields));
+  }
+
+  if (step.schemas) {
+    layout = layout.concat(getSchemaLayout(step));
```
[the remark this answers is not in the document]
StevePole: Love it, that's what I was looking for and couldn't think of :-)

packages/dynamic-flows/src/flow/layoutService/LayoutService.js
StevePole: Yeah, that's fair.  I only went with that as my first use case needed 'sm', I subsequently extended it with an override. 'md' is a better default.

PR #527, opened 2020-08-18
docs: add translation rfc proposal

andreapiras:
## 🖼 Context

RFC suggestion to improve consistency and provide full internalisation to components.
## ✅ Checklist


rfc/0003-internalisations-component-api.md
```diff
@@ -0,0 +1,68 @@
...
+  ​
+
+# Summary
+
+Currently there's not a consistent API to manage internalisation in neptune web components. This RFC aims to provide an overview on the suggested solution and on the deprecation strategy for the new translation API.
+
+# Motivation
+
+- Some of our components have hardcoded strings which makes impossible to achieve a full internalisation.
+- The way we manage strings is not very scalable and can be improved by grouping all the texts into an object.
```
twrdjpalmer, line 12: Can the example cover some harder use cases than the placeholder one? E.g., the nested close button inside the modal component? Or where we might have multiple `alt` attributes

Gave the former a crack, and see it now, the translations object is definitely better, should we allow it to contain nested objects 👍

```js
<Modal closeButtonLabel="Close" />
// 👇 
<Modal translations={{ close: { label: "Close" }} />
```

rfc/0003-internalisations-component-api.md
```diff
@@ -0,0 +1,68 @@
...
+```
+<MoneyInput>
+    <Select placeholder="Select an option..."/>
+</MoneyInput>
+```
+
+The above example will turn into the following one where consumers can pass placeholder in any language they need.
+
+```
+<MoneyInput translations={{placeholder:"your placeholder translated"}}>
```
twrdjpalmer, line 27: Thing to consider here is the MoneyInput will rerender every render of the parent now, as we're creating a new object every time.

Would have to recommend memoizing the translations object to avoid that:

```js
const translations = React.useMemo(() => ({
  placeholder: translate('input.placeholder')
}), [translate])

<MoneyInput={translations} />
```

twrdjpalmer: Would it be an improvement too, to have the components handle translating the keys into the actual values themselves? So, instead of having to call `translate` in the parent of `MoneyInput`, it would handle it for us?

Probably not, as it would force us into use one translations lib

sbrandwoo:
I prefer the old syntax of `<Search placeholder={myString}>`. I don't think of this as a translation, I think of it as a value that I'm passing into the component. In some components this is more than a string too, it could be a `<>Something <strong>else</strong></>`. It also mirrors `<input placeholder />`

There are a few cases where the consistency of a `translations` param wouldn't make sense. For example in `Summary` we'd still want separate items for `title` and `content`, but we are still passing translated content in there. For `Select`, we still need to set a translated `label` on our options directly too.

For Neptune I think it'd make sense to remove all of the existing English, to force consumers to pass in translated value. Currently it's easy to accidentally miss a string that needs to be translated.

andreapiras:
Closing as it seems that we are heading to this direction https://github.com/transferwise/neptune-web/pull/545

PR #529, opened 2020-08-18
fix: footer padding and icon underline

nabaroa:
## 🖼 Context

Ticket: [https://transferwise.atlassian.net/browse/DS-608](https://transferwise.atlassian.net/browse/DS-608)
In the spacing refactor we cleaned code, one of it was a utility class used in the footer to give a big padding to it.

## 🚀 Changes

I add the padding to the footer itself, so we don't have to use a utility for the padding.

**Before**
imagen

**After**
imagen

Fixed the little underline issue of the icon as well.

## ✅ Checklist


packages/css/src/less/footer.less
```diff
@@ -23,6 +32,10 @@
     }
   }
 
+  .link-icon {
+    text-decoration: none;
+  }
```
sergonius, line 37: Instead of adding classes like this, wouldn't it make more sense to use an already defined utility class in the template?

nabaroa: I understand the use of an utility class to make an exception, as in this case, an `a` without the default underline fulfills that law. But this class is expecific for footer icon that already has a class name and I think that makes sense to add this property in the CSS. And it can be used in several templates (at least 2 repos as I can remember), so the simplest and cleanest change is the one in the source.

andreapiras (approved):

PR #530, opened 2020-08-18
refactor!: a11y & i18n fixes for switch

helendurranttw:
## 🖼 Context

Fixing up a11y / i18n issues for the Switch component. 

## 🚀 Changes

* Added a new translations prop to pass down aria labels / other translations. See [this RFC](https://github.com/transferwise/neptune-web/pull/527) for more.
* We also realised while fixing eslint errors that we needed to add React Testing Library as a peer dep, because you can't rely on the dependencies of a dependency. Also updated `test-config` and `dynamic-flows` to fix this.

## 🤔 Considerations

We were debating the use of prop spreading: we will likely submit an idea for increased flexibility in the rule to the `@transferwise/eslint-config` repo. For now we just took it out.

## ✅ Checklist


packages/components/src/switch/Switch.js
```diff
@@ -76,6 +79,9 @@ Switch.propTypes = {
   onClick: Types.func.isRequired,
   className: Types.string,
   id: Types.string,
+  translations: Types.shape({
+    ariaDescription: Types.string.isRequired,
```
twrdjpalmer, line 83: As there is also [`aria-describedby`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute), it may be better to use `ariaLabel` or just `label` for the translation?

packages/components/src/switch/Switch.js
```diff
@@ -59,8 +61,9 @@ const Switch = ({ checked, onClick, className, id }) => {
       onClick={onClick}
       onKeyDown={handleKeyDown}
       tabIndex={0}
-      role="checkbox"
+      role="switch"
       aria-checked={checked}
+      aria-label={translations.ariaDescription}
```
dozortsev, line 66: Usually Switch component will have some `<label />` "assigned" to it and in this case we should not use `aria-label`

See MDN docs:
- [Using the aria-label attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute)
- [Using the aria-labelledby attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute)

packages/components/src/switch/Switch.js
```diff
@@ -31,15 +33,17 @@ const CrossCircle = () => (
...
+ * @param {string} [className=''] - String that contains extra-classes.
+ * @param {string} [id=''] - Element Id.
+ * @param {function} onClick - Function that manages the check logic.
  *
- * @usage '<Switch checked={checked} onClick={() => setCheck(!checked)} id='id' className='extra-class-name' />'
+ * @usage '<Switch ariaLabelledBy='A switch with a label' checked={checked} className='extra-class-name' labelId='the label id' id='switchID' onClick={() => setCheck(!checked)} />'
  * */
 
-const Switch = ({ checked, onClick, className, id }) => {
+const Switch = ({ ariaLabel, ariaLabelledBy, checked, className, id, onClick }) => {
```
dozortsev, line 46: could you please use same syntax for those aria props as they are in HTML (`aria-label`,  `aria-describedby`)

we do same for `data-testid` attr in components which have this prop

dozortsev, line 46: why you need write all those native HTML props and then also define propTypes and defaultProps for them if you could use spread operator here to collect props

```js
const Switch = ({ ...props }) => {
```

later in JSX you **explicitly pass only necessary props**:

```js
onClick={props.onClick}
aria-checked={props.checked}
aria-labelledby={props['aria-labelledby']}
aria-label={props['aria-label']}
```

dozortsev: note, you will still will be able to mark some props as required

```js
Switch.propTypes = {
  'aria-label': requiredIf(Types.string, props => props['aria-labelledby'] === ''),
```

helendurranttw:
Closing as out of date. We need to commit this to `next`, as there are breaking changes.

PR #532, opened 2020-08-19
demo for RFC dedicated for i18y of component's copy and aria attrs

dozortsev:
it's a demo PR for a feedback for RFC https://github.com/transferwise/neptune-web/pull/527

### Description

Example of small component with suggested in RFC approach (**grouping copy props in one "translations" prop**):

```js
<Switch onClick={() => setCheck(!checked)} translations={{ ariaLabel: 'Some label' }} />
```

☝️ This interface doesn't look natural/semantic at all, why introducing custom things (`translations={{ ariaLabel: '..' }}`) in places where you can avoid this and just pass label value via old well know `aria-label` syntax?.

Example of a more complex component:

```js
<FlowNavigation
     translations={{
         avatar: { label: 'Some text', ariaLabel: '...' },
         closeButton: { label: 'Hello there!', ariaLabel: '...' },
     }}
     avatarUrl="..."
     steps={{
        steps: [
          {
            label: 'Amount',
            ariaLabel: '...'
          },
          {
            label: 'My details',
            ariaLabel: '...'
          },
          ...
        ],
      }}
      onClose={...}
      onGoBack={...}
/>
```

Here this approach has a drawback incase of big components where we render multiple items (e.g steps (in flow navigation), options (in select), items (in summary) ...) as part of copy will be passed via `translations` prop and some of them via `items` prop so it's not actually fully grouped.

---

Suggesting use `[innerComponentName]Props` prop (e.g `avatarProps`) or just `[innerComponentName]` prop (e.g `avatar`) (this PR uses first one).

I call this approach **props grouping per sub component**:

```js
// Example of small and not complex component (a few elements and doesn't consist of other Neptune Web components)
<Switch aria-label="Some label" onClick={...} />

// Example of big and complex component (contains several other components: Header, Stepper, CloseButton, Avatar..)
<FlowNavigation
      avatarProps={{
        url: 'https://github.com/transferwise.png',
        'aria-label': 'Some text',
      }}
      stepperProps={{
        activeStep,
        steps: [
          {
            label: 'Amount',
            'aria-label': 'Amount',
          },
          {
            label: 'My details',
            'aria-label': 'My details',
          },
          ...
        ],
      }}
      closeButtonProps={{ onClick: () => { ... }, 'aria-label': 'Hello there!' }}
      onGoBack={...}
/>
```

Semantic and familiar interface for aria props is saved and incase big components props of sub component are nicely grouped. IMHO this interface makes much more sense and it's truly scalable.


sbrandwoo:
I like this approach. For individual components, I like that our components are used in the same way as regular HTML elements:

```
<button aria-label="Close">X</button>
<Button aria-label="Close">X</button>
```

And for bigger components like `FlowNavigation`, I agree that it makes sense to group together the props that you are passing directly into the sub-components like `Avatar`. This way I don't need to learn two different interfaces.

dozortsev:
Closing it for now.

PR #533, opened 2020-08-20
Ds 425 field refactor

andreapiras:
## 🖼 Context
Demo: https://transferwise.github.io/neptune-web/branch/DS-425-field-refactor/lab/Field
Ticket: https://transferwise.atlassian.net/browse/DS-425

Old implementation of field component was really heavyweight and basically imported all the components in all instances. We decide to reimplement Field in a more lightweight solution in order to provide basic validations to wrapped elements and styled messages in case of failure

## 🚀 Changes
(Breaking) Remove old Field implementations from dynamic flows
(Feat) Add new Field implementation in components

 

## 🤔 Considerations
This field will be released as part of experimental feature with support for input type='text|number' , checkbox and DateInput. We don't recommend using it in production yet.
There is an known a11y issue due to low contrast for placeholder in select

## ✅ Checklist


packages/validation/src/type-validators/index.js
```diff
@@ -19,8 +19,22 @@ function isArray(value) {
 function isNull(value) {
   return value === null;
 }
+
+function isFalse(value) {
+  return value === false;
+}
```
sergonius, line 25: All these utility functions could be replaced by a single `import { is } from 'an existing library'`. Why aren't we using something like it here?

packages/components/src/lab/field/Field.js
```diff
@@ -0,0 +1,193 @@
...
+
+/**
+ * Field component performs basic validations and provides a styled message to wrapped components.
+ *
+ * @param {boolean} [checked=false] - Initial value for the field type checkbox.
+ * @param {boolean} [submitted=false] - Flag that indicates whether the field has been submitted or not.
+ * @param {boolean} [type] - Type of component wrapped. This prop switches the type of validation perfomed.
+ * @param {function} [onChange] - Handler for onChange event.
+ * @param {object} [children] - Element that needs to be validated.
+ * @param {object} [errors=null] - Translatable message for manual error messages. If specified it has priority over validation errors.
```
sbrandwoo, line 17: Don't think this should be plural `errors` unless it accepts an array.

packages/components/src/lab/field/Field.js
```diff
@@ -0,0 +1,193 @@
...
+ * @param {object} [validation] - Object that contains the value to be validated against and the translatable message that will appear on validation failure.
+ * @param {string|number} [model=null] - Initial value for the field type text or number.
```
sbrandwoo, line 19: `initialValue`?

packages/components/src/lab/field/Field.js
```diff
@@ -0,0 +1,193 @@
...
+  const { messageType, message, formGroupClasses } = getMessage();
+
+  const childProps = {
+    onChange,
+    onBlur,
+    onFocus,
+    id: props.id,
+  };
+
+  if (props.type === 'boolean') {
```
sbrandwoo, line 144: Do we need both `checked` and `model`?

sbrandwoo (commented):
I'd be interested to see how this integrates with a library such as Formik. We use Formik for validations and  we'd need the `touch` state and error messages to work correctly.

packages/docs/utils/sections.js
```diff
@@ -76,4 +76,8 @@ export default [
     ],
     searchable: true,
   },
+  {
+    title: 'Lab 👩‍🔬',
```
dozortsev, line 80: why adding an emoji?

packages/components/src/lab/field/Field.spec.js
```diff
@@ -0,0 +1,264 @@
+import React from 'react';
+import '@testing-library/jest-dom';
+import { render, screen, cleanup } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import Field from './';
+// import Checkbox from '../../checkbox';
```
dozortsev, line 6: comment

packages/components/src/lab/field/Field.spec.js
```diff
@@ -0,0 +1,264 @@
...
+  // Checkbox needs to be revisited.
+  // it('broadcast model onChange for type checkbox', () => {
+  //   const { container } = render(
+  //     <Field {...props} type="checkbox" checked={false}>
+  //       <Checkbox label="label" onChange={jest.fn()} />
+  //     </Field>,
+  //   );
+  //   userEvent.click(container.querySelector('button'));
+  //   expect(props.onChange).toHaveBeenCalledWith(false);
+  // });
```
dozortsev, line 256: comment

packages/docs/pages/lab/Field.mdx
```diff
@@ -0,0 +1,135 @@
+import { LiveEditorBlock, GeneratePropsTable } from '../../utils';
```
dozortsev, line 1: I would also add a page for which describes the concept of lab components (in `Lab.mdx`)
Screen Shot 2020-08-27 at 1 14 21 PM

packages/docs/utils/sections.js
dozortsev: if you still would like have some stuff there then I would better pick [icon](https://transferwise.github.io/icons/)

packages/components/src/lab/field/Field.js
```diff
@@ -0,0 +1,182 @@
+import React, { useState, useEffect } from 'react';
+import Types from 'prop-types';
+import classNames from 'classnames';
+
+import Alert from '../../alert';
+import { getValidationFailures } from './utils/getValidationFailures';
+import WithNormaliser from '../../withNormaliser';
+
+/**
+ * Field component performs basic validations and provides a styled message to wrapped components.
```
helendurranttw, line 10: typo: validation

packages/components/src/lab/field/Field.js
```diff
@@ -0,0 +1,182 @@
...
+ *
+ * @param {boolean} [submitted=false] - Flag that indicates whether the field has been submitted or not.
```
helendurranttw, line 12: In jsdoc this indicates something is optional: `@param {boolean} [type]`. To indicate a required field, leave out the square brackets: `@param {boolean} type`. Also, I think this can be more than just a boolean?

Lastly- I'm not quite sure by this description what type is, or what I would pass in if I was using field. What does a boolean or numeric type indicate? What is an example?

packages/components/src/lab/field/Field.js
```diff
@@ -0,0 +1,182 @@
...
+ * @param {boolean} [type] - Type of component wrapped. This prop switches the type of validation perfomed.
+ * @param {function} [onChange] - Handler for onChange event.
+ * @param {object} [children] - Element that needs to be validated.
```
helendurranttw, line 15: What can this be? We have some restrictions on this at the moment right? So maybe we should explain that here?

packages/components/src/lab/field/Field.js
```diff
@@ -0,0 +1,182 @@
...
+ *           maximum: { value: 10, message: 'The value should smaller than 10' },
+ *           required: { value: true, message: 'This field is required' },
+ *          }}
+ *       >
+ *         <input type="number" className="form-control" />
+ *     </Field>`
+ * */
+
+const Field = (props) => {
+  const getValidations = () => {
```
helendurranttw, line 41: May as well move this function out of Field?

packages/components/src/lab/field/Field.js
```diff
@@ -0,0 +1,182 @@
...
+ * @param {boolean} [type] - Type of component wrapped. This prop switches the type of validation perfomed.
+ * @param {function} [onChange] - Handler for onChange event.
+ * @param {object} [children] - Element that needs to be validated.
+ * @param {object} [errors=null] - Translatable message for manual error messages. If specified it has priority over validation errors.
+ * @param {object} [validation] - Object that contains the value to be validated against and the translatable message that will appear on validation failure.
+ * @param {string|number} [initValue=null] - Initial value for the field type text or number.
+ * @param {string} [id] - Field id. Required for accessibility in order to couple label and field with via htmlFor attribute.
+ * @param {string} [label=''] - Label associated to field.
+ *
+ * @usage `<Field
```
helendurranttw, line 22: I really appreciate the docs and the usage example! But I think you're missing at least one prop (type is required, right?)

packages/components/src/lab/field/Field.js
```diff
@@ -0,0 +1,182 @@
...
+      newModel,
+      props.type,
+      validations.validationRules,
+    );
+    const isValid = newValidationFailures.length === 0;
+    setValidationFailures(newValidationFailures);
+    props.onChange(newModel, isValid);
+  };
+
+  const getMessage = () => {
```
helendurranttw, line 99: You could move this outside of Field too, if you want.

packages/components/src/lab/field/Field.js
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  return (
+    <div className={classNames(formGroupClasses)}>
+      {props.label && (
+        <label className="control-label" htmlFor={props.id}>
+          {props.label}
+        </label>
+      )}
+      <WithNormaliser>{React.cloneElement(props.children, childProps)}</WithNormaliser>
```
helendurranttw, line 148: 👌

packages/components/src/lab/field/Field.js
```diff
@@ -0,0 +1,182 @@
...
+  if (props.type === 'boolean') {
+    childProps.checked = model;
+  } else {
+    childProps.value = model;
+  }
+
+  return (
+    <div className={classNames(formGroupClasses)}>
+      {props.label && (
+        <label className="control-label" htmlFor={props.id}>
```
helendurranttw, line 144: I don't think this is right 🤔 Usually the label + htmlFor relationship is on sibling elements. You can also wrap the entire thing in a label, which negates the need for the id, but I'm not sure about doing the other way around. Does the label refer to the whole field? Or just the props.children? 😕 I'm not sure..

packages/components/src/lab/field/Field.spec.js
```diff
@@ -0,0 +1,264 @@
...
+  });
+
+  it('renders children', () => {
+    render(
+      <Field {...props}>
+        <input type="text" />
+      </Field>,
+    );
+
+    expect(screen.getByLabelText(props.label)).toBeTruthy();
```
helendurranttw, line 46: How does this test prove that it renders the children?

packages/components/src/lab/field/Field.spec.js
```diff
@@ -0,0 +1,264 @@
...
+        </Field>,
+      );
+      const alert = screen.getByText(props.validation.minLength.message).parentElement;
+
+      expect(alert).toBeTruthy();
+      expect(alert.className).toContain('alert-danger');
+    });
+  });
+
+  it('attaches form-group class by default', () => {
```
helendurranttw, line 163: Not sure what value this test adds

packages/docs/liveEditorCode/field.code.js
```diff
@@ -1,46 +1,22 @@
...
-        minLength: null,
-        maxLength: null,
-        minimum: 10,
-        maximum: 99,
-        validationMessages: {
-          required: 'Number is required',
-          minimum: 'Must be 10 or greater',
-          maximum: 'Must be 99 or less',
-        },
+      errors={null}
```
helendurranttw, line 4: 😌

packages/docs/next.config.js
```diff
@@ -27,6 +27,7 @@ module.exports = () =>
             transpileModules: [
               '@transferwise/dynamic-flows',
               '@transferwise/neptune-validation',
+              '@transferwise/components',
```
helendurranttw, line 30: Why is this required?

packages/docs/pages/lab/Field.mdx
helendurranttw: Good idea ☝

packages/docs/pages/lab/Field.mdx
```diff
@@ -0,0 +1,135 @@
...
+      <th>Element</th>
+      <th>type</th>
+      <th>Validation supported</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>input</td>
+      <td>string</td>
+      <td>minLenght, maxLength, pattern, required</td>
```
helendurranttw, line 27: typo

packages/docs/pages/lab/Field.mdx
```diff
@@ -0,0 +1,135 @@
...
+  <thead>
+    <tr>
+      <th>Element</th>
+      <th>type</th>
+      <th>Validation supported</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>input</td>
```
helendurranttw, line 25: You should probably specify that you mean normal HTML inputs

packages/docs/pages/lab/Field.mdx
```diff
@@ -0,0 +1,135 @@
+import { LiveEditorBlock, GeneratePropsTable } from '../../utils';
+import { Field } from '@transferwise/components/build/es/polyfill/lab';
+import { Checkbox, Alert, DateInput } from '@transferwise/components';
+import code from '../../liveEditorCode/field.code';
+
+Field is a new component that provides validation to its children. Validation messages are automatically
+attached to the validated component with styling. Currently the supported children types are:
+
+<Alert type={Alert.Type.WARNING}>
+  Please note that this component is still under an experimental phase and we do not guarantee that
```
helendurranttw, line 10: Experimental might be going too far.. how about "is still in a testing phase"

packages/docs/pages/lab/Field.mdx
```diff
@@ -0,0 +1,135 @@
...
+      <td>string</td>
+      <td>minLenght, maxLength, pattern, required</td>
+    </tr>
+    <tr>
+      <td>input</td>
+      <td>number</td>
+      <td>minimum, maximum, required</td>
+    </tr>
+    <tr>
+      <td>DateInput</td>
```
helendurranttw, line 35: Links to our components might be nice.

packages/docs/utils/sections.js
helendurranttw: I like the emoji but I'd consult with Steph

packages/components/src/lab/field/Field.spec.js
andreapiras: Because it return the field associated with label

packages/docs/next.config.js
andreapiras: because lab now exports only es version and since we are importing inside Field.mdx we need to transpile

packages/docs/utils/sections.js
andreapiras: @StephanieTW can you please tell Anton that this emojii is great?

StephanieTW: Haha. Sorry Andrea, but we tell people not to use emojis in product copy because they're hard to localise and not great for people using screen readers. So I think we should stick to those rules and not use one here too.

andreapiras: 😒😒😒😒

packages/components/src/lab/field/Field.js
andreapiras: I'm building this component to eventually wrap [BasicSchema](https://github.com/transferwise/neptune-web/blob/main/packages/dynamic-flows/src/jsonSchemaForm/basicTypeSchema/BasicTypeSchema.js#L134) , so I'm trying to keep the props aligned. This can contain multiple error in the same string maybe?

packages/components/src/lab/field/Field.js
andreapiras: I'd keep this one in since it's very related to form and not really reusable

packages/components/src/lab/field/Field.js
andreapiras: the label will refer to the element that has the same id

packages/components/src/lab/field/Field.spec.js
andreapiras: It's just checking that the class is there, is not a super valuable test, but I don't think it harms to keep it

packages/validation/src/type-validators/index.js
andreapiras: Hey @sergonius can you please give me an example?

packages/components/src/lab/field/Field.js
helendurranttw: I'm suggesting it no so much for reuse, but for performance. Moving it outside the function prevents it from being recreated every time you create a new component.

packages/components/src/lab/field/Field.js
helendurranttw: Think this still needs fixing :)

packages/components/src/lab/field/Field.js
helendurranttw: Was this resolved?

packages/components/src/lab/field/Field.js
helendurranttw: Ahh my mistake, I thought the id went on the wrapper element, not the actual field. This makes sense.

packages/components/src/lab/field/Field.spec.js
helendurranttw: 🤦‍♀️ oops I read this as getByText. Sorry!

helendurranttw (dismissed):

packages/validation/src/type-validators/index.js
helendurranttw: I presume he means like lodash or something. It's a valid point - at what point is it worth importing another library?

Side note - I don't think isFalse should go in the type-validators. False isn't a type, it's a value.

helendurranttw (commented):
Ah, I just realised you've removed the old Field. We should leave that there for now - given our lab field isn't ready for prime time yet, people who are currently using the Field will have no option if we remove the old one.

packages/validation/src/type-validators/index.js
sergonius: Yeah, could be any library we see fit. An example of [is](https://ramdajs.com/docs/#is) in the wild.

I don't see the reason for creating/exporting/importing a function for each use-case like this, especially if they are so simple and can be inlined instead.

packages/components/src/lab/field/Field.js
StevePole: I don't think this will _wrap_ BasicTypeSchema, wouldn't it be used inside of it?  I would think a single `error` string (not object) would be sufficient.  Or if we elect to support multiple, then it should probably be an array of strings.  Though I think a single string is fine.

packages/validation/src/value-validators/index.js
```diff
@@ -0,0 +1,9 @@
+import { isString, isObject, isArray } from '../type-validators';
+
+const isFalse = (value) => value === false;
```
StevePole, line 3: Really don't think we need this...

packages/components/src/lab/field/Field.js
andreapiras: yeah sorry my bad I meant it will be used inside BasicSchema to wrap FormControl.

andreapiras: Regarding errors I'm happy to call it error but I think it should be a node just in case people need to pass in a`<Message>` component

helendurranttw (approved):

PR #537, opened 2020-08-25
fix!: neptune-css clean up

helendurranttw:
## 🖼 Context

Before pushing out Neptune CSS to more product teams, we wanted to take the opportunity to tidy up some of the old CSS. 

## 🚀 Changes

Moved React specific CSS into `components`. Deprecated rarely used files, but exposed them in a `deprecated` folder.

## 🤔 Considerations

I need to write a thorough changelog / upgrade guide.

## ✅ Checklist


packages/components/src/loader/Loader.less
```diff
@@ -1,4 +1,4 @@
-@import 'bootstrap/less/variables';
+@import (reference) '@transferwise/neptune-css/src/variables/neptune-tokens.less';
```
anttio, line 1: Do we need this to be explicitly imported? There's only one colour token used in this file.

anttio (approved):

packages/components/src/loader/Loader.less
helendurranttw: Nice spotting, thanks

nabaroa (approved):
LGTM

PR #539, opened 2020-08-26
ADS-1539 currency selects

felixxiatransferwise:
## 🖼 Context

Want to add some features to DFv3 currency select to match what we have on prod atm.

## 🚀 Changes

 
For `JsonSchemaForm` only:
- Map icon names into currency flags
- Select options where title and description are less than 50 chars together will collapse to 1 line, more will separate into 2 lines
- Selects with 20 options or more has search enabled automatically

## ✅ Checklist


JulianG-TW (approved):

packages/dynamic-flows/src/jsonSchemaForm/schemaFormControl/availableCurrencyFlags.js
```diff
@@ -0,0 +1,60 @@
+export const availableCurrencyFlags = [
```
dozortsev, line 1: this should be defined only once in separate package and then reused here

I would try to add this in [currency-flags package](https://github.com/transferwise/currency-flags)

felixxiatransferwise: This is very susceptible to change right now, we are still discussing mappings as we have more icons other than currencies. This PR just aims to get something quick and working for `recipients-page` which is the only real consumer of this. Just didn't want to extract it now and introduce that overhead of involving different repos at this stage.

helendurranttw (approved):

packages/dynamic-flows/src/jsonSchemaForm/schemaFormControl/availableCurrencyFlags.js
helendurranttw: We'll merge it for now, but let's definitely revisit this at some stage.

PR #545, opened 2020-09-01
rfc: providing translations in Neptune Web

helendurranttw:
[Read the RFC here](https://github.com/transferwise/neptune-web/blob/translations-provider-rfc/rfc/0004-localisation.md).

Check out the changes to see an example of a couple of components using the context inside Neptune Web, and the docs updated to provide its own locale.

Note the [select in the docs](https://transferwise.github.io/neptune-web/branch/translations-provider-rfc/components/inputs/Select) has a search placeholder in Spanish.


rfc/0004-localisation.md
```diff
@@ -0,0 +1,143 @@
...
+## Overriding the locale
+
+To override the locale, the consumer must pass a map of strings translated to the desired language. This will be mixed in to our locale provider, and default translations will be used where none are provided:
+
+```js
+import * as React from 'react';
+import { LocaleProvider } from '@transferwise/neptune-web';
+import { MoneyInput } from '@transferwise/neptune-web';
+
+const localeOverrideEs = {
```
dozortsev, line 91: such object will need to be defined for each language, right?

so here actually we will need create a bit more then 10 of them? and then pick one based on selected lang right?

```js
const localeOverride = {
   es: {
     select: {
        placeholder: 'Seleccione una opción',
        searchPlacholder: 'buscar',
     },
  },
  en: {
     select: {
        placeholder: 'Placeholder',
        searchPlacholder: 'search placholder',
     },
  },
  ...
  ..
}

<LocaleProvider locale={getLocale(localeOverride, currentSelectedLocale)}>
    <MoneyInput />
```

something like this?

rfc/0004-localisation.md
```diff
@@ -0,0 +1,143 @@
...
+
+# Summary
+
+We have two existing RFCs attempting to solve the problem of passing translated strings to our components. There has been opposition to the way in which we pass them down - both in the naming of the prop, and how we manage passing translations down for nested props.
+
+Passing translation strings to every component is messy and error prone, and doesn't scale well for complex composite components with many sub-components.
+
+I propose we make use of React contexts / providers, which were designed to handle just such situations.
+
+I did some searching, and Uber's BaseWeb handles locales in this exact manner. The examples I've shown here are based off of the way they use context to solve this problem.
```
dozortsev, line 12: could you please send Github link?

rfc/0004-localisation.md
```diff
@@ -0,0 +1,143 @@
...
+export default enGB;
+```
+
+Each component with required translated strings defines a locale file that sits in the same folder as the component, for example:
+
+```js
+// closebutton/locale.js
+const locale = {
+  close: 'Close',
+};
```
dozortsev, line 51: from this example I don't understand where other translated messages are define?

dozortsev: or you mean that in components package we will have only default messages (English ones) and consume will need provide the rest of translated languages?

rfc/0004-localisation.md
```diff
@@ -0,0 +1,143 @@
...
+     avatarUrl="..."
+     steps={...}
+     onClose={...}
+     onGoBack={...}
+/>
+```
+
+```js
+// Suggested approach - different locale
+<LocaleProvider locale={localeOverride}>
```
dozortsev, line 135: do I understand correctly that consumers would need wrap almost each Neptune component with `LocaleProvider` in their app codebase?

rfc/0004-localisation.md
```diff
@@ -0,0 +1,143 @@
...
+export default () => (
+  <LocaleProvider locale={localeOverrideEs}>
+    <MoneyInput />
+  </LocaleProvider>
+);
+```
+
+## Nested components
+
+Because we've removed the need to pass these strings as props, there is no longer any need to pass in strings for sub-components. This produces a much cleaner interface.
```
sbrandwoo, line 107: > much cleaner interface

I disagree, I think you've chose misleading examples here. Here's how we'd need to practically use the component:

```
const Example: FC<ExampleProps> = ({}) => {
  const { translate } = useTranslations();

  return (
    <LocaleProvider
      locale={{
        avatar: {
          label: translate('mycomponent.avatar.label'),
        },
        closeButton: {
          close: translate('mycomponent.close.button'),
        },
      }}
    >
      <FlowNavigation
        avatarUrl="..."
        steps={...}
        onClose={...}
        onGoBack={...}
      />
    </LocaleProvider>
  );
};
```

sbrandwoo:
This feels similar to how [retranslate](https://github.com/Tankenstein/retranslate) works and I wonder if would could leverage that directly. For example if the component used `<Message>neptune.closeButton.label</Message>` and assumed that you provided the component inside a translation context, and a translation for `neptune.closeButton.label`.

I don't think this matches our typical use case though, because we'd typically want to vary the string used in each use of a component, rather than setting a single translation for all instances.

rfc/0004-localisation.md
danlaush: It seems like this would be the case, and pairs well with Steve's comment
> we'd typically want to vary the string used in each use of a component, rather than setting a single translation for all instances.

packages/components/src/closeButton/CloseButton.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import Types from 'prop-types';
+import Close from '@transferwise/icons/react/close';
+import { LocaleContext } from '../locale/index';
+
+const CloseButton = ({ onClose }) => {
```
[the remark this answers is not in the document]
helendurranttw: 👍

rfc/0004-localisation.md
andreapiras: @sbrandwoo Yeah but isn't that going to wrap your whole app? So you don't have to do it for each component

dozortsev: Andrea, then usually `<LocaleProvider` will be defined in some core app file like `App.js` and components in another files

`App.js`:

here we provide translations:
```js
render() {
  return (
    <LocaleProvider
      locale={{
        avatar: {
          label: translate('mycomponent.avatar.label'),
        },
        closeButton: {
          close: translate('mycomponent.close.button'),
        },
       ...
       ...
      }}
    >
      <SomeCoreComponentWhichCombineAllAppPages />
    </LocaleProvider>
  );
}
```

`SomePage.js`:
and here we provide rest of component's data:

```js
render() {
  return (
   ...
    <FlowNavigation
      avatarUrl="..."
      steps={...}
      onClose={...}
      onGoBack={...}
    />
   ...
  );
}
```

`SomeAppSubComponent.js`:

```js
render() {
  return (
   ...
    <MoneyInput />
   ...
  );
}
```

so passing data in components would split across files... which is not great

what `locale` object structure would look like if we consumer has 3+ components like FlowNavigation?

```js
    <LocaleProvider
      locale={{
        avatar: {
          label: translate('mycomponent.avatar.label'),
        },
        closeButton: {
          close: translate('mycomponent.close.button'),
        },
      }}
    >
```

please show code example?

dozortsev: @andreapiras to wrap whole app consumer would need create massive `locale` object

dozortsev: this also means splitting the way how we pass data in component across files... translations in core file (usually it will be some `App.js`) and other data in places where we define component

dozortsev: or I'm missing something? how would pass data in `local` object if you have 5+ components in you app?

helendurranttw: Andrea is correct, and I'm sorry for not posting an example that shows this. You only wrap your app once, so no need to wrap every component.

@dozortsev I think you overestimate how many translations we have in Neptune Web. Also, you only need to provide translations for the components you use.

danlaush: I might have missed some earlier discussion about this topic, but does this apply only to text in a `translation` prop? Or anywhere that text is used in a component, instead of passing down strings? Taking `<Card title="" details="" />` as an example, would those props be replaced by a translations block or are they dumb text props?

dozortsev: @danlaush 

> ...does this apply only to text in a translation prop?

no, in this approach doesn't have `translation` prop, `translation` is from another [RFC](https://github.com/transferwise/neptune-web/pull/527) as a solution to tackle same issue

> ...Taking <Card title="" details="" /> as an example, would those props be replaced by a translations block or will they remain dumb text props?

no, with Card component I guess we will have following:

[global wrapper](https://github.com/transferwise/neptune-web/pull/545#discussion_r482012114):
```
render() {
  return (
    <LocaleProvider
      locale={{
        card: {
          title: translate('mycomponent.card.title'),
          details: translate('mycomponent.card.details'),
        },
      }}
    >
      ...
    </LocaleProvider>
  );
}
```

in place where you use the component, you should have this

```js
render() {
  return (
      ..
      <Card />
      ..
   )
}
```

@helendurranttw please correct me if I'm wrong

dozortsev: how this will work if you have multiple card component I don't know

danlaush: Yeah it's the multiple instances part that I want to understand, since I'm likely to have many `<Card />`s or `<Button />`s on a page and they'll each need different strings.

helendurranttw: No these would still use normal props, and so would list items. These translated keys are only for things for which we need to provide default text within Neptune Web: things like placeholder text.

rfc/0004-localisation.md
helendurranttw: Nope - just once per app!

```
return (
      <LocaleProvider locale={localeOverrideEs}>
        <Head>
          <title>Neptune Design System — TransferWise</title>
        </Head>

        <Layout>
          <Component {...pageProps} />
        </Layout>
      </LocaleProvider>
    );
```

rfc/0004-localisation.md
sbrandwoo: Would this include placeholders on `Select`? I can see it being likely that we want to vary the value of `placeholder` and `searchPlaceholder` on each instance, and not want to have an application-wide default for it.

helendurranttw: We need translations for the default values, but we would leave the props so people could customise them if they wanted to.

helendurranttw: Thanks for your feedback and questions - I've updated the RFC to hopefully make it clearer. I'm going to resolve this conversation for now but please take a read of the updates and comment if you have more questions / feedback!

oliverviljamaa:
Thanks for sharing the RFC, the problem needs solving indeed!

I'm wondering why we couldn't just use `retranslate` inside the repo like we do in our other web libraries? The components could use the `language` from the parent `retranslate` `TranslateProvider` if one exists and if not, either a `language` prop could be passed or a simple provider could be added as a wrapper. For most of the use cases — where the app uses `retranslate` and the translated values don't need to be overridden — it would be seamless, not even a breaking change. Furthermore, overriding could always be supported via custom props, allowing the consumer to use whatever translation mechanism they choose.

If we're worried about the bundle size, `retranslate` could be a peer dependency (as most of the apps use it anyway) and the translations in all languages won't contribute that much to the size, as the keys can be minified and there's not that many values (3 from the example).

helendurranttw:
> I'm wondering why we couldn't just use `retranslate` inside the repo like we do in our other web libraries?

Good call @oliverviljamaa , I will look into this! Looks like their provider is pretty much identical. I just saw that I missed another comment from @sbrandwoo suggesting the same :)

sbrandwoo:
This doesn't solve the problem of ensuring that a consumer is passing enough translation values. If a new aria-label is added to component, requiring a new translation, consumers probably wouldn't notice and wouldn't update their locale object.

Each app's locale object would be different depending on which components they use. When an app uses a library which uses a library that uses Neptune (such as balance-flows-web using calculator), it would get very difficult to track which strings are needed.

If, instead of context, strings are provided directly to Components then we can at least use `prop-types` and TypeScript to document what's required by consumers.

--

I feel like using Context creates a disconnect between the Neptune components and the translations. I've always seen each component as being an independent item (they could be completely separate packages) but with Context now I'm configuring Neptune directly.

--

I think there's an implied decision here that we don't want to put any translated strings directly into Neptune. Doing this and giving components a `language` prop might be an all-round simpler solution.

oliverviljamaa:
> I think there's an implied decision here that we don't want to put any translated strings directly into Neptune. Doing this and giving components a `language` prop might be an all-round simpler solution.

That's actually what I was suggesting, along with a fallback of getting the `language` from the consumer app's `retranslate` provider.

dozortsev:
@oliverviljamaa wonder why you recommending retranslate? wdyt about [react-intl](https://www.npmjs.com/package/react-intl)?

I would use last one as it's based on [JS standards](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl)

oliverviljamaa:
> @oliverviljamaa wonder why you recommend retranslate? (Note: I never used that) wdyt about [react-intl](https://www.npmjs.com/package/react-intl)?
> 
> I would use last one as it's based on [JS standards](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl), many TW apps also use it, thus devs familiar with it's API..

No strong preference, I've just seen `retranslate` used more, but it may just be me!

PR #550, opened 2020-09-02
feat: Add validation to oneOfs

felixxiatransferwise:
## 🖼 Context

One of selects do not have the correct validation right now. This PR fixes it.

## 🚀 Changes

 
out

## ✅ Checklist


packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.spec.js
```diff
@@ -214,32 +215,106 @@ describe('Given a oneOfSchema component', () => {
...
     });
 
     it('should not render a generic schema', () => {
       expect(genericSchema.length).toBe(0);
     });
 
-    it('should not set a default for the SchemaFormControl value', () => {
-      expect(schemaFormControl.prop('value')).toBe(null);
+    it('should set a default for the SchemaFormControl value', () => {
+      expect(schemaFormControl.prop('value')).toBe(0);
```
StevePole, line 226: Why do you change this behaviour?

I don't believe we should because in the 'const' case we want the user to choose, otherwise some users just leave it on the default.

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.js
```diff
@@ -19,41 +42,64 @@ const OneOfSchema = (props) => {
...
+
+      // Apply validations for the change
+      setValidations(getValidationFailures(newSchema.const, props.schema, props.required));
     } else {
       props.onChange(models[index], newSchema);
     }
   };
 
-  const isConstSchema = (schema) => {
+  function isConstSchema(schema) {
```
StevePole, line 100: Why change this?

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.js
```diff
@@ -19,41 +42,64 @@ const OneOfSchema = (props) => {
...
     models[index] = model;
     setModels(models);
-
+    setChanged(true);
     props.onChange(model, schema);
   };
 
+  const onFocus = () => {
+    setFocused(true);
+    setBlurred(false);
```
StevePole, line 69: In terms of the validation, once we have blurred once, we would stay blurred.  It's maybe not the best naming (I guess you took it from the other file), but the intent is to record that the control has been interacted with at some point, so once it's set to true, it should not be reset to false.

StevePole (commented):
There are some specific questions in line, but I'm also not totally sure about this approach.  It feels like we're repeating some logic which we shouldn't be.  

There's another thing you may not be aware of, @andreapiras has been working on a `Field` component, that takes responsibility for validation away from BasicTypeSchema, and he has ported BasicTypeSchema to use the Field.  I couldn't spot the PR, but I'll try and find.  I wonder if reuse of the Field is a better approach as it would avoid reimplementing any validation logic.

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.spec.js
felixxiatransferwise: If we don't default to a value, things like `legalEntityType` won't be preselected?
Screenshot 2020-09-03 at 14 02 22

There's the other case where we don't want to default select something if we have a placeholder.
Screenshot 2020-09-03 at 14 03 53

We could introduce logic for shorter radio selects of when to default and when to not.

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.js
felixxiatransferwise: so it's hoisted. `const` does not hoist so we can't make use of it before it's declared.

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.spec.js
```diff
@@ -214,32 +215,106 @@ describe('Given a oneOfSchema component', () => {
...
+          // SchemaFormControl broadcasts the INDEX(1) of the schema
+          schemaFormControl.simulate('change', 1);
+        });
+
+        it('should broadcast the PARENT schema as the trigger, not the const', () => {
+          expect(onChange).toHaveBeenCalledWith(2, schema);
+        });
+
+        it('should only broadcast one change', () => {
+          component = shallow(<OneOfSchema {...props} required model={1} />);
```
StevePole, line 263: Not sure what this test is doing?

Why do you need to rerender the component and simulate another change?  We already rendered it up above somewhere, and we simulated a change in the beforeEach of this block.

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.spec.js
```diff
@@ -214,32 +215,106 @@ describe('Given a oneOfSchema component', () => {
...
+
+          schemaFormControl = component.find(SchemaFormControl);
+          onChange.mockReset();
+          schemaFormControl.simulate('change', 2);
+
+          expect(onChange).toHaveBeenCalledTimes(1);
+        });
+      });
+
+      it('should broadcast change to null', () => {
```
StevePole, line 273: This says what should happen but not when.  I would phrase this whole section differently, above you have a redundant `describe('change to index 1', () => {` which isn't telling us anything new, I would break out to two blocks:
```
describe('when the user selects a new schema', () => {

});
describe('when the user deselects so there is no active schema', () => {

});

File: packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.spec.js [hunk 6]
```diff
@@ -214,32 +215,106 @@ describe('Given a oneOfSchema component', () => {
...
+        schemaFormControl.simulate('change', null);
+
+        expect(onChange).toHaveBeenCalledWith(null, schema);
+      });
+    });
+
+    describe('validation errors', () => {
+      it('should not display validation errors if schema is not required and value exists', () => {
```
StevePole, line 281: Please use "given, when, then":
```
describe('when the schema is not required and the value exists', () => {
  it('should not display validation errors, () => { ... }
});

describe('when the schema is not required and no value exists', () => {
  it('should not display validation errors', () => { ... }
});
```
etc

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.js
```diff
@@ -7,8 +7,31 @@ import SchemaFormControl from '../schemaFormControl';
...
+import { getValidationFailures } from '../../common/validation/validation-failures';
+import { isEmptyObject } from '../../common/object';
 
 const OneOfSchema = (props) => {
+  const [changed, setChanged] = useState(false);
+  const [focused, setFocused] = useState(false);
+  const [blurred, setBlurred] = useState(false);
+  const [validations, setValidations] = useState(getInitialValidationFailures(props));
+
+  function getInitialValidationFailures(initialProps) {
```
StevePole, line 21: I didn't understand what this is for, when would we have 'initial' validation failures?  Doesn't validation only happen after some interaction?

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.spec.js
felixxiatransferwise: This test was already here but was failing due to the newly introduced logic around setting default values - it thinks it's broadcasted 3 times, so i'm just making it super explicit here that the assertion only pertains to the new change.

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.js
```diff
@@ -7,8 +7,31 @@ import SchemaFormControl from '../schemaFormControl';
...
+    if (
+      !initialProps.model &&
+      !initialProps.schema.placeholder &&
+      isConstSchema(initialProps.schema.oneOf[0])
+    ) {
+      // Default to first position
+      props.onChange(initialProps.schema.oneOf[0].const, props.schema);
```
StevePole, line 28: I'm not a fan of this existing in a method called `getInitialValidationFailures`.  This doesn't provide any hint that the method might change the value of the control.  If we have some initialisation work to do, we should do it explicitly, not as a by-product of performing validation.

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.js
felixxiatransferwise: In the case of a const select with placeholder where we don't default to anything and the user clicks submit straight away, we want to show for example this field is required.

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.js
felixxiatransferwise: Had some same thoughts while writing this. 👍 Will move to a separate initialisation method.

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.spec.js
```diff
@@ -214,23 +216,43 @@ describe('Given a oneOfSchema component', () => {
...
-      expect(schemaFormControl.prop('value')).toBe(null);
+    it('should set a default for the SchemaFormControl value', () => {
+      expect(schemaFormControl.prop('value')).toBe(0);
     });
 
-    it('should not broadcast a model update', () => {
-      expect(onChange).not.toHaveBeenCalled();
+    it('should broadcast first option by default', () => {
+      // enzyme shallow doesn't support useEffect right now
+      component = mount(<OneOfSchema {...props} />);
```
felixxiatransferwise, line 232: https://github.com/enzymejs/enzyme/issues/2086

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.spec.js
```diff
@@ -239,8 +261,75 @@ describe('Given a oneOfSchema component', () => {
...
+
+          schemaFormControl.simulate('change', null);
+
+          controlFeedback = component.find(ControlFeedback);
+          expect(controlFeedback.prop('validations').length).toBe(1);
+        });
+      });
+
+      describe('when no placeholder exists', () => {
+        it('should not display initial validation errors', () => {
```
StevePole, line 314: I don't think functional behaviour should differ based on the presence (or lack of) a placeholder.

felixxiatransferwise: I've introduced it for now so we can achieve the behaviour here https://github.com/transferwise/neptune-web/pull/550/files#r482963519 without introducing anything new to the spec.

We could just always default to the first option in the case of a const select but this deviates from the behaviour of what we have in production. Also I think it may be jarring if a CST sees a preselected bank code that isn't what they intend to enter. 

I'm all for improving this mechanism in the future once we have a better idea.

StevePole: How about if we actually have a default, the same as for other fields.  The only issue is that we probably need to specify the default based on `index`, so maybe it's
```
{
  defaultIndex: 0,
  oneOf: [{ ... }, { ... }]
}
```
?

packages/dynamic-flows/src/common/object.js
```diff
@@ -0,0 +1 @@
+export const isEmptyObject = (obj) => obj.constructor === Object && Object.keys(obj).length === 0;
```
dozortsev, line 1: I guess you should be able to use [`isEmpty` function from neptune-validation package](https://github.com/transferwise/neptune-web/blob/main/packages/validation/src/value-validators/index.js)

felixxiatransferwise: ah ok, thanks! will check that out.

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.spec.js
felixxiatransferwise: removed this behaviour.

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.spec.js
felixxiatransferwise: Removing this change.

StevePole (approved):
👍

PR #553, opened 2020-09-04
fix(Modal): Adjust modal with to be full width in mobile

mikknarusk:
## 🖼 Context

https://transferwise.atlassian.net/browse/SEND-1303
https://transferwise.slack.com/archives/C2XG7R3EZ/p1598961346008600

## 🚀 Changes
Makes `Modal` full width and height in modal, with 16px padding in all directions. Previously, the width was determined by the width of the content, causing the modal to change in size if the content changed and modals to look different on across pages.

Before:
Screenshot 2020-09-04 at 16 19 05

After:
Screenshot 2020-09-04 at 16 18 56

 

## 🤔 Considerations
Depending on how a consumer might be overwriting some styles to suit their needs, it's possible this will introduce unwanted visual changes.

## ✅ Checklist


fergusjordantw (approved):

helendurranttw:
Thanks for the contribution @mikknarusk! We are having an issue with our builds right now, so I'm sorry we will need to hold off on merging this for a few hours. Sorry!

packages/components/src/modal/Modal.less
```diff
@@ -1,3 +1,5 @@
+@import (reference) '@transferwise/neptune-css/src/variables/neptune-tokens.less';
```
helendurranttw, line 1: This isn't necessary - the custom properties are resolved using postcss :)

packages/components/src/modal/Modal.less
```diff
@@ -91,6 +94,19 @@
   }
 }
 
+@media (--screen-xs-max) {
+  .tw-modal-dialog {
+    padding: var(--size-16);
+    max-width: calc(
+      576px - var(--size-40) * 2 + var(--size-16) * 2
```
helendurranttw, line 101: You can just use a single number here (e.g. 528px) and explain your calculation in comments.

mikknarusk: is there a downside to doing this with calculations? In my view it explains the calculation better and provides a nice sanity check, even if the calculation is static.

helendurranttw: I can see your point, I agree it's nice having it in code. I try to avoid calc where unnecessary because it introduces an unnecessary cost (while admittedly negligible in one instance) and because IE11 has some quirks meaning it won't always work correctly in every use case, so you need to remember to check when using it.

I don't feel that strongly about it in this instance - if you prefer it you can leave it :)

helendurranttw (approved):

packages/components/src/modal/Modal.less
mikknarusk: I've checked in IE in browserstack as well just in case, looks fine.

PR #565, opened 2020-09-10
feat: InstructionsList now supports items of any renderable type

phildobsontw:
## 🖼 Context
We would like the instructions list to be able to support links, e.g:

image

## 🚀 Changes
We think the most extensible (and non breaking) way of achieving this is to allow anything that is renderable to be passed into the 'dos' and 'donts' of the InstructionsList component, by relaxing the prop types.

## 🤔 Considerations
We considered changing the contract of the component, but found it difficult to design the contract in such a way that would allow you to make just a subset of an item link.

## ✅ Checklist


packages/components/src/instructionsList/InstructionList.story.js
```diff
@@ -13,7 +13,14 @@ export const basic = () => {
     'Invite at least 3 friends',
     'Convince them to use this amazing product',
   ]);
-  const donts = array('donts', ['Paying extra hidden fees for transfers', 'Use bad exchange rate']);
+  const donts = array('donts', [
```
andreapiras, line 16: Hey @phildobsontw thanks for your contribution! Can you please also update the docs with an example? https://github.com/transferwise/neptune-web/blob/main/packages/docs/liveEditorCode/instructionList.code.js
Also can you share some mockups/screenshot of how this is going to look ? 
Thank you

phildobsontw: Sure! Have added this to the description! 🙇

StephanieTW: Hi Phil, thanks adding the screenshot of how you'd like to update the component above.

Do you have a screenshot of your particular use case? Or some a link to some designs?

Could you tell us a bit about why you need a link in this component?
Thanks!

phildobsontw: Sure - here's a screen shot of a possible design. The links were more for substantiation of marketing claims that anything else.
image

StephanieTW: Thanks, that use case makes sense to me. I can see you'd need to link off to evidence to support those claims.

andreapiras (dismissed):

andreapiras (approved):

PR #571, opened 2020-09-10
fix: typeahead visual bugs

domhorvathtw:
## 🎥 🍿 
typeaheda

## 🖼 Context

A PR related to the `<Select/>` component's styles, removed CSS rules for the `<Typeahead/>` component.
[PR link](https://github.com/transferwise/neptune-web/pull/235)

Styles for `.tw-dropdown-item` got namespaced to `.tw-select.tw-dropdown-item` and the rules were not applied anymore. 
This resulted in not having any styles for `--focused` items, even though the navigation still works with arrow up <kbd>↑</kbd> and arrow down <kbd>↓</kbd>.

The `autocomplete` value on the input element was wrongly configured. Its value was `new-password` which is not relevant for the typeahead component, [read more about autocomplete here](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete). Also it was conflicting with the dropdown items in Chrome. See screenshot:
image

## 🚀 Changes
- fix styles for Typeahead Option 
- ~~turn off autocomplete of main input element~~
- expose `inputAutoComplete` prop to allow consumers setting its value

## ✅ Checklist


packages/components/src/typeahead/typeaheadInput/TypeaheadInput.js
```diff
@@ -28,6 +28,7 @@ export default class TypeaheadInput extends Component {
     onKeyDown: Types.func.isRequired,
     onFocus: Types.func.isRequired,
     onPaste: Types.func.isRequired,
+    autoComplete: Types.string.isRequired,
```
andreapiras, line 31: Why is this required can we default to a value?

packages/components/src/typeahead/Typeahead.story.js
```diff
@@ -101,6 +101,7 @@ export const basic = () => {
       addon={<i className="input-group-text icon icon-search" />}
       onBlur={() => {}}
       options={options}
+      inputAutoComplete="off"
```
andreapiras, line 104: is this not named autoComplete?

packages/components/src/typeahead/Typeahead.js
```diff
@@ -68,6 +68,7 @@ export default class Typeahead extends Component {
     onFocus: Types.func,
     chipSeparators: Types.arrayOf(Types.string),
     size: Types.oneOf([Typeahead.Size.MEDIUM, Typeahead.Size.LARGE]),
+    inputAutoComplete: Types.string,
```
andreapiras, line 71: Can we limit the option with 'oneOf'?

domhorvathtw: There is quite a lot of options: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete
I think having it as a string is okay

andreapiras (approved):

PR #572, opened 2020-09-10
fix: move modal body class handling from Dimmer to Modal

fergusjordantw:
## 🖼 Context

https://transferwise.atlassian.net/browse/SEND-1335

We're currently removing the `modal-open` class on `document.body` in the Dimmer unmount. This is causing some problems in send-flow where unopened modals that were rendered in the React tree with `open={false}` are unmounted **after** another modal is opened, resulting in the open modal being un-scrollable.

https://transferwise.slack.com/archives/CJNBX9CP6/p1599486193015200

## 🚀 Changes

After speaking with @andreapiras - we agreed the thing that would make the most sense would be to move this handling to the `Modal` so that we're only adding/removing the class when the **modal** is actually opened.

## 🤔 Considerations

Now that the `Dimmer` is no longer adding/removing the `modal-open` class, the `Drawer` component which is also using it will be scrollable. If this is behaviour we want to avoid, we could add similar handling/styling with a `drawer-open` class or explore alternative solutions to solving this problem.

TODO: Test this with an `npm link` against send-flow to double check it fixes the original problem

## ✅ Checklist


andreapiras:
I think the drawer should have the same behaviour as the modal ie no scrolling, the drawer is position fixed and the body itself is scrollable so no need to have other confusing scrolling

fergusjordantw:
> I think the drawer should have the same behaviour as the modal ie no scrolling, the drawer is position fixed and the body itself is scrollable so no need to have other confusing scrolling

So do we think we need to add a `drawer-open` className that enables/disables scroll specifically for that component as well - or do we want to explore a fix that keeps `Dimmer` controlling the className but somehow mitigate className conflicts (i.e. unique id generated per component)

andreapiras:
> > I think the drawer should have the same behaviour as the modal ie no scrolling, the drawer is position fixed and the body itself is scrollable so no need to have other confusing scrolling
> 
> So do we think we need to add a `drawer-open` className that enables/disables scroll specifically for that component as well - or do we want to explore a fix that keeps `Dimmer` controlling the className but somehow mitigate className conflicts (i.e. unique id generated per component)

For me it makes sense for the children to control the scrolling, the Dimmer itself should be in my opinion as dumb as possible as we may use it in future also in other scenarios that are not full screen. So I'd go for drawer-open solution

fergusjordantw:
@andreapiras have updated the PR to include adding a new `drawer-open` className and am adding/removing it with the same pattern used by Modal

packages/components/src/common/DOMOperations/DOMOperations.js
```diff
@@ -1,9 +1,14 @@
 const MODAL_OPEN_BODY_CLASS = 'modal-open';
+const DRAWER_OPEN_BODY_CLASS = 'drawer-open';
```
andreapiras, line 2: Hey @fergusjordantw I don't think we need two separate classes. A `no-scroll` class would be enough. Can you please add this class in Dimmer.css. If we'll find another use case for it we can put it in neptune-css core

packages/components/src/common/DOMOperations/DOMOperations.js
```diff
@@ -1,9 +1,14 @@
...
-  document.body.classList.add(MODAL_OPEN_BODY_CLASS);
-}
+const addBodyClass = (className) => () => document.body.classList.add(className);
 
-export function removeModalOpenBodyClass() {
-  document.body.classList.remove(MODAL_OPEN_BODY_CLASS);
-}
+const removeBodyClass = (className) => () => document.body.classList.remove(className);
+
+export const addModalOpenBodyClass = addBodyClass(MODAL_OPEN_BODY_CLASS);
```
andreapiras, line 8: I think we should have a generic addNoScrollClass method

andreapiras (approved):

PR #581, opened 2020-09-14
Displaying errors after the model has changed

JulianG-TW:
# Displaying errors after the model has changed when "submitted" is true

## 🖼 Context

BasicTypeSchema/ControlFeedback components had a bug.
The errors in the errors prop were not being displayed after the model had been "changed" once.

You can see the problem in the animated gif below:

storybook-01-before

## 🚀 Changes

This PR changes the condition that makes errors visible in two places:

Once for JS: (in BasicTypeSchema.js)
```diff
- const isErrorVisible = !props.changed && props.errors;
+ const isErrorVisible = (props.submitted || !props.changed) && !!props.errors;
```

Once for CSS/Styles: (in ControlFeedback.js)
```diff
- (!changed && props.errors) ||
+ ((props.submitted || !changed) && !!props.errors) ||
```

Here's the same Storybook demo, after the changes:
storybook-02-after

## ✅ Checklist


helendurranttw (approved):

StevePole:
I'm not sure this is the right solution to the problem.  The idea was that we remove the error after the user changes the value.  Because the intended use case is to submit the form, receive an error from the server, and then remove that error once the user changes the value.  There was a bug because this only worked for the first submission.  But I don't think the new solution will remove the error after the user changes the value.  I would expect a solution to have something more like a `hasChangedSinceErrorSupplied` value that we utilise.

JulianG-TW:
Oh! I'm, sorry! I only saw this now.

I don't know how to calculate or derive a `hasChangedSinceErrorSupplied` flag.
Because the errors are passed as a prop, there isn't an event in which the error is supplied.

One of our first attempts at tackling this issue was leaving `isErrorVisible` as it was:
```js
const isErrorVisible = !props.changed && props.errors;
```
And instead we would "detect" when the errors prop changed, by running a `useEffect` (with `prop.errors` in the deps array). This effect would re-set the `changed` flag to false, enabling `isErrorVisible` to potentially be true.

But we quickly realised that the server could send the same error string twice, and that would not trigger the `useEffect`.

Any ideas on how to derive a `hasChangedSinceErrorSupplied` flag?

PR #585, opened 2020-09-15
fix: add autoFillOnBlur to typeahead component

domhorvathtw:
## 🖼 Context

We are using the `Typeahead` component in the Help Center for the new search component. It allowed us to quickly put together the new component, however, the `typeahead` component was mostly designed to other use cases and doesn't properly fit our use case. 

There is one problem:
1. user types in the input a search string
2. we show the users the search results in the dropdown menu
3. user clicks on document
4. the `handleDocumentClick` event handler updates the query in the state, triggers a new state update
5. the new state update triggers the `onChange` handler with the selected value
6. our onChange handler navigates the user to the search results page (even though it wasn't a click on an option nor pressing enter)

**Reproduce here**  https://transferwise.com/help/

In order to keep the same functionality of the component but also support the use case of the help center, I would like to introduce another prop, that can alter the default logic and not update the components internal state when users click outside of the component. 

## 🚀 Changes

 
- add `autoFillOnBlur` property to typeahead component

## 🤔 Considerations

I know adding props is not the best way of making components more generic, probably a new component will be needed but for now, this fix will be sufficient and we will rethink the anatomy of this component an either come up with something new for search or refactor this component into a compound component so it will be more customisable without introducing new props for every use case that pops up.

## ✅ Checklist


andreapiras (approved):

Michael-Terry-Transferwise (commented):
👍

PR #592, opened 2020-09-16
ADS-1582 Add alert support to dynamic flows

felixxiatransferwise:
## 🚀 Changes
This PR adds alerts to dynamic flows. Given this schema:
```json
{
  "title": "Object schema",
  "type": "object",
  "alert": {
    "context": "warning",
    "markdown": "some content\n ## How about a title?\n ```\nconsole.log('hello world')\n```"
  },
  "properties": {
    "number": {
      "type": "number",
      "title": "Number input"
    },
    "string": {
      "alert": {
        "context": "info",
        "markdown": "ah a string alert"
      },
      "type": "string",
      "title": "Text input"
    },
    "select": {
      "alert": {
        "context": "success",
        "markdown": "oh an oneOf alert"
      },
      "type": "integer",
      "title": "Select",
      "control": "radio",
      "placeholder": "Favourite number?",
      "oneOf": [
        {
          "title": "One",
          "const": 1,
          "description": "The first number"
        },
        {
          "title": "Two",
          "const": 2,
          "description": "The second number"
        },
        {
          "title": "Three",
          "const": 3,
          "description": "The third number",
          "disabled": true
        }
      ]
    }
  },
  "required": ["number", "string", "select"]
}
```

We render 
Screenshot 2020-09-16 at 15 56 33

## ✅ Checklist


packages/dynamic-flows/src/common/alert.js
```diff
@@ -0,0 +1,19 @@
+import { Alert } from '@transferwise/components';
+
+export const convertAlertContextToType = (alertContext) => {
+  if (!alertContext) {
```
helendurranttw, line 4: This is redundant - the default case will handle falsey values :)

felixxiatransferwise: Removed this since we are using `DynamicAlert` now.

helendurranttw (approved):
Discussed with Felix that we should move DynamicAlert out of layout - he will address in a separate PR.

PR #595, opened 2020-09-21
fix: IE11 double fire event

andreapiras:
## 🖼 Context

Accordion wasn't working properly in IE11 because on click element was firing the link twice. This is apparently due to the "for" attribute  which binds to its parent element and appears to trigger another click.

## 🚀 Changes
Replaced label with button. Labels should be used with inputs.An example  of accordion can be find here https://www.w3.org/TR/wai-aria-practices-1.1/examples/accordion/accordion.html

 

## ✅ Checklist


packages/components/src/accordion/AccordionItem/AccordionItem.js
```diff
@@ -2,17 +2,26 @@ import React from 'react';
...
-        <div className="media-body">{typeof title === 'string' ? <h5>{title}</h5> : title}</div>
+    <Button
+      aria-expanded={isOpen}
+      onClick={() => onClick(index)}
+      className="tw-accordion-item btn-link p-l-0 text-no-decoration"
+      type={Button.Type.LINK}
+    >
+      <div className="media p-y-2">
+        <div className="media-body text-xs-left">
+          {isString(title) ? <span className="h5">{title}</span> : title}
```
sergonius, line 18: Why not use a heading element here?

andreapiras: Hey because we can't know in advance the structure of the page, so in order to not mess up with the page SEO we do not hardcode headings

helendurranttw (approved):

PR #598, opened 2020-09-23
fix(DefinitionList): add padding for vertical layout

hendrikmolder:
## 🖼 Context

[DS-655](https://transferwise.atlassian.net/browse/DS-655)

As brought out in Slack discussions, in two-column layout we always leave some space in the bottom of the `DefinitionList` items container. However, in one-column layout, the there's no space after the last item – this means consumers would have to add these styles separately to keep the spacing consistent.

## 🚀 Changes
 
- adds padding after last item on md-sm screens

## 🤔 Considerations
Please check all layouts in both desktop and mobile vs master:
https://transferwise.github.io/neptune-web/branch/DS-655_fix-definitionlist/components/content/DefinitionList
vs
https://transferwise.github.io/neptune-web/components/content/DefinitionList

## ✅ Checklist


oliverviljamaa (approved):
👏

andreapiras (changes requested):
I don't think adding the padding here is the solution. The parent container is changing display so that's why the margin on line has not being taken into account

andreapiras (approved):

PR #600, opened 2020-09-28
chore: reintroduced well classes

andreapiras:
## 🖼 Context

Some apps still use .well bootstrap classes. In order to fix the regression we reintroduce these classes. We'll eventually replace them in future in favour of component
## ✅ Checklist


domhorvathtw (approved):

anttio:
Should this also be part of the Neptune bundle, or just a standalone CSS component?

packages/css/src/less/wells.less
```diff
@@ -0,0 +1,4 @@
+// TODO: remove bootstrap dependencies
+
+@import (reference) 'bootstrap/less/variables.less';
+@import 'bootstrap/less/wells.less';
```
anttio, line 4: Should we just refactor and clean this now? The implementation with mixins in our Bootstrap is just creating different size variants with same values as the default one. 
https://github.com/transferwise/bootstrap/blob/59d55b847efb4054173214670e8ea345176d70f1/dist/css/bootstrap.css#L10273

anttio: > Control on which breakpoints well styles are applied with some responsive classes. Works with "xs" "sm" "md" "lg" "xl"

Actually, seems that these might be needed. So nevermind this refactoring.

andreapiras: Ok do you want me to copy the generated css?

anttio (approved):

PR #601, opened 2020-09-28
feat: add logic for display order

felixxiatransferwise:
## 🚀 Changes

 
This PR adds support for `displayOrder` for object schemas.

## ✅ Checklist


packages/dynamic-flows/src/jsonSchemaForm/objectSchema/ObjectSchema.js
```diff
@@ -42,25 +60,19 @@ const ObjectSchema = (props) => {
...
-              errors={props.errors && props.errors[propertyName]}
-              locale={props.locale}
-              translations={props.translations}
-              onChange={(newModel, schema) => onChange(propertyName, newModel, schema)}
-              submitted={props.submitted}
-              required={isRequired(propertyName)}
-            />
-          </div>
-        ))}
+        {props.schema.displayOrder &&
```
StevePole, line 63: Rather than have presentational logic for both variants, let's normalise the data and have one approach to presentation e.g.
```
const orderedFields = props.schema.displayOrder || Object.keys(props.schema.properties);

orderedFields.map( ... )
```

Then you also don't need to add the WrappedGenericSchema component.

packages/dynamic-flows/src/jsonSchemaForm/objectSchema/ObjectSchema.spec.js
```diff
@@ -140,4 +140,57 @@ describe('Given a component for rendering object schemas', () => {
...
+      component = shallow(<ObjectSchema {...props} schema={schemaWithDisplayOrder} />);
+
+      genericSchemaComponents = component.find(GenericSchema);
+
+      expect(genericSchemaComponents.at(0).prop('schema')).toBe(properties.number);
+      expect(genericSchemaComponents.at(1).prop('schema')).toBe(properties.something);
+      expect(genericSchemaComponents.at(2).prop('schema')).toBe(properties.string);
+    });
+
+    it('should display any properties not specified in the display order in their natural order', () => {
```
StevePole, line 166: I'd put this in a separate 'when' as it's a different scenario.

StevePole (changes requested):

packages/dynamic-flows/src/jsonSchemaForm/objectSchema/ObjectSchema.js
felixxiatransferwise: There are more than 2 variants here:
- displayOrder contains all properties
- no displayOrder present
- **displayOrder contains only a subset of properties**

Therefore we probably can't do `const orderedFields = props.schema.displayOrder || Object.keys(props.schema.properties);`. If the concern is around the extra `WrappedGenericSchema` component, I could do 

```
const getOrderedFields = (schema) => {
  ...
  return orderedFields
}

const orderedFields = getOrderedFields(props.schema)

```
if this is preferable?

felixxiatransferwise: Updated.

packages/dynamic-flows/src/jsonSchemaForm/objectSchema/ObjectSchema.js
```diff
@@ -42,25 +54,27 @@ const ObjectSchema = (props) => {
...
-              locale={props.locale}
-              translations={props.translations}
-              onChange={(newModel, schema) => onChange(propertyName, newModel, schema)}
-              submitted={props.submitted}
-              required={isRequired(propertyName)}
-            />
-          </div>
-        ))}
+        {orderedPropertyNames.map((propertyName) => {
+          return (
```
StevePole, line 58: Why do you need the return?

packages/dynamic-flows/src/jsonSchemaForm/objectSchema/ObjectSchema.js
```diff
@@ -34,6 +34,18 @@ const ObjectSchema = (props) => {
     setModel(getValidModelParts(model, props.schema));
   }, [props.schema]);
 
+  const getOrderedPropertyNames = (schema) => {
+    if (schema.displayOrder) {
+      const unspecifiedPropertyNames = Object.keys(schema.properties).filter(
```
StevePole, line 39: I think we agreed that all properties should be in the order array, making this a bit redundant, but I guess it's fine to be safe.

felixxiatransferwise: Yeah adding this for safety.

StevePole: I was thinking about whether there's a cleaner way to do it e.g.
```
const ordered = [...new Set([...schema.displayOrder, ...Object.keys(schema.properties)];
```
But not 100% sure if that's order preserving.

StevePole (approved):

packages/dynamic-flows/src/jsonSchemaForm/objectSchema/ObjectSchema.js
StevePole: It is :-) https://stackoverflow.com/questions/55460303/why-does-js-keep-insertion-order-in-set

felixxiatransferwise: Set order should be guaranteed based on insertion order in js.

felixxiatransferwise: I was on the same post :D

StevePole (approved):
Nice

helendurranttw (approved):

PR #605, opened 2020-09-30
ADS 1617 Move alert

felixxiatransferwise:
## 🚀 Changes

 
This PR simply moves alerts for basic field types to below the field so on a "refresh on change", the user would always see the alert, whereas the alert could be out of view if it was above.

Screenshot 2020-09-30 at 10 26 55

## ✅ Checklist


helendurranttw (approved):
The code is fine, but is this really the experience you want? Would it be better to scroll up to the alert, or to show an error on the field instead of showing a separate alert? I presume you've already talked about this with the designer so if you've thought this through, all good.

PR #609, opened 2020-10-02
[ADS-1617] Add disable prop to JsonSchemaForm

JulianG-TW:
## 🖼 Context

We need a simple way to disable the entire JsonSchemaForm while refreshing requirements in recipients-page `CreateAccountForm` modal.

## 🚀 Changes

We've added an optional `disabled` prop to `<JsonSchemaForm>` and propagated it down to all schemas and form controls.

## 🤔 Considerations

There's a separate commit for disabling the `<Upload>` component: 0d1dc4cc952973a0578218e5a6158a1f603eb9f4
Because this component is different that the others, I'm not sure it's the correct approach.

Also, there may be other components/controls that I've missed? Perhaps not covered by the 4 testing schemas?

I wrote the tests using React Testing Library, as suggested in [contributing.md](https://github.com/transferwise/neptune-web/blob/main/packages/dynamic-flows/CONTRIBUTING.md). I put them on a separate file than the existing JsonSchemaForm tests because those are using Enzyme. This is not strictly necessary. I'm happy to move them upon request.

I'm not sure how to test this in IE11 before making a production build. What's the usual procedure?

## ✅ Checklist


packages/dynamic-flows/src/formControl/FormControl.js
```diff
@@ -312,6 +312,7 @@ export default class FormControl extends PureComponent {
         return (
           <Upload
             {...uploadProps}
+            usDisabled={uploadProps.usDisabled || disabled}
```
StevePole, line 315: I don't believe the upload should have a separate disabled prop, you should just use `disabled` like any other control.

packages/dynamic-flows/src/jsonSchemaForm/JsonSchemaForm.RTL.spec.js
```diff
@@ -0,0 +1,61 @@
+import React from 'react';
```
StevePole, line 1: Why a new test file for this?  We have an e2e test file that could be used to test this propagates to the DOM.  Or you could test that the disabled flag is propagated to children through the various unit tests.

StevePole (commented):
Couple comments. Mainly all good though 👍

packages/dynamic-flows/src/formControl/FormControl.js
JulianG-TW: Hmmm. I'm not sure I understand.

That's the way Upload is. We're not introducing `usDisabled` in this PR. The current version already has `usDisabled`: https://transferwise.github.io/neptune-web/components/inputs/Upload

packages/dynamic-flows/src/jsonSchemaForm/JsonSchemaForm.RTL.spec.js
JulianG-TW: hey, I completely missed the **JsonSchemaForm.e2e.spec.js** file.
I’m not sure it's a good idea to bring the new RTL tests to that file, because the new tests are testing against 4 different schemas, while the existing e2e test is only testing 1 schema.

It's technically possible, but I think the test file may become confusing.

packages/dynamic-flows/src/formControl/FormControl.js
StevePole: It shouldn't be there, as it's inconsistent, I'd rather not use it here.  @andreapiras  could this cause any problem for users of Field?

packages/dynamic-flows/src/jsonSchemaForm/JsonSchemaForm.RTL.spec.js
StevePole: I just don't think we should have the tests for one component split over two testing frameworks.  The existing e2e test would effectively test everything, because the schema it uses nests an object within a oneOf within an allOf.  So to disable successfully all of those schema types must propagate the disabled flag.  I would be tempted to add a specific disabled field to the object, to not affect any other tests.

JulianG-TW: I have deleted this file and moved the tests to the e2e file.

packages/dynamic-flows/src/formControl/FormControl.js
JulianG-TW: I don't mind leaving this untouched, but then the file upload controls won't be disabled.

StevePole (approved):

packages/dynamic-flows/src/formControl/FormControl.js
andreapiras: @StevePole Hey, that is a breaking change for people using Field and FormControl and relying on uploadProps. What we can do is:

- add `disabled` in `Upload`
-  leave also `usDisabled` 
- add a deprecation comment for `usDisabled` like we did [here](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/button/Button.js#L70). 
Line 315 will then become 
`disabled={uploadProps.usDisabled || disabled}` 
if you want to make it look like 
`disabled={disabled}` that would be a breaking change for Field and FormControl users

JulianG-TW:
Closing because we've merged https://github.com/transferwise/neptune-web/pull/618

PR #610, opened 2020-10-02
fix: fix upload validation empty mime type

andreapiras:
## 🖼 Context
Right now, mime type checking will fail on certain browser / OS combinations (such as Windows 7 + Chrome / FF / IE), because the mime type appears empty, and so fails to match the specified pattern.

The following is a way to ensure this never happens (accept all types and so impose no checking):

```
<Upload usAccept="*"  ...
```

but if you want to do something like:

```
<Upload usAccept="text/csv"
```

It can fail. We should think about the best way to solve this. Could be some / all of:

check type if present, and if not, fall back to checking extension

try and detect the failing OS / browser combinations and disable checking for them

more robust validation / error messages on incorrect extension (maybe only if type is missing)

adding a note on the docs to remind people to double check type on the server side

A start was made on this earlier in the year: https://github.com/transferwise/neptune-web/pull/158/files

## 🚀 Changes

 
Add fallback when MIME type is empty. Retrieves mimetype from data64 encoded string

## ✅ Checklist


packages/components/.storybook/config.js
```diff
@@ -24,7 +24,7 @@ if (process.env.NODE_ENV !== 'production') {
   const ReactDOM = require('react-dom');
   const axe = require('react-axe');
   const context = {
-    include: [['#story-root']],
+    include: [['#root']],
```
andreapiras, line 27: This has changed since props docs have been set to not be inline

packages/components/src/upload/utils/getFileType/getFileType.js
```diff
@@ -0,0 +1,19 @@
...
+  if (file && file.type && file.type !== '') {
+    return file.type;
+  }
+
+  if (file64) {
+    const regex = /^data:([a-z]+\/[a-z]+);/;
+    const typeFromEncoded = file64.match(regex);
+
+    if (typeFromEncoded) {
+      return typeFromEncoded[1];
```
arikanmstf-tw, line 15: is it possible if this `typeFromEncoded ` is not falsy but has 0 or 1 members ?
if `typeFromEncoded = []` or typeFromEncoded = ['foo'] , this statement returns undefined

andreapiras: From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match
```
match returns an array whose contents depend on the presence or absence of the global (g) flag, or null if no matches are found.
If the g flag is not used, only the first complete match and its related capturing groups are returned. 
```
But I can also add extra check on line 14

arikanmstf-tw (approved):

packages/components/src/upload/utils/getFileType/getFileType.js
```diff
@@ -0,0 +1,19 @@
+export const getFileType = (file, file64) => {
+  if (!file && !file64) {
+    return null;
+  }
+
+  if (file && file.type && file.type !== '') {
```
dozortsev, line 6: would be nice to add https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining to avoid such checks

(not saying we should do that in this PR)

packages/docs/pages/components/inputs/Upload.mdx
```diff
@@ -3,6 +3,8 @@ import { LiveEditorBlock, GeneratePropsTable } from '../../../utils';
 import { IMAGES } from '../../../static/assets/img/data64';
 import code from '../../../liveEditorCode/upload.code';
 
+This component uses `fetch` API which is not supported on IE11. The `fetch polyfill` is not currently provided with the library to prevent code duplication in users apps that may already have that polyfill. Please make sure you have `fetch polyfill` in your app if you intend to provide IE11 support.
```
dozortsev, line 6: what do you mean by putting "fetch polyfill" in code quotes? this looks weird for me, I would just put "fetch" in those quotes

dozortsev (approved):
don't know much about this component implement, but changes looks legit

packages/components/.storybook/preview-head.html
```diff
@@ -0,0 +1 @@
+<script src="https://polyfill.io/v3/polyfill.min.js?features=fetch"></script>
```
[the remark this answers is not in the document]
andreapiras: We don't provide fetch polyfill which is require by IE11. Probably a good idea to add it also in docs

packages/components/src/upload/utils/getFileType/getFileType.js
andreapiras: https://transferwise.atlassian.net/browse/DS-686

PR #612, opened 2020-10-02
Input accent crop

andreapiras:
## 🖼 Context

Input is currently cropping some character OÜ -> OU
image

Fixed 
image

## 🚀 Changes

 
Changed the input from height:48px to min-height:48px and max-height:48px;

## ✅ Checklist


.circleci/config.yml
```diff
@@ -168,7 +168,6 @@ workflows:
                 - main
 
       - deploy-docs-for-branch:
-          type: approval
```
andreapiras, line 171: This is not working branches are not getting deployed

packages/css/src/less/mixins/_forms.less
```diff
@@ -0,0 +1,125 @@
...
+  }
+}
+
+// Form control sizing
+//
+// Relative text size, padding, and border-radii changes for form controls. For
+// horizontal sizing, wrap controls in the predefined grid classes. `<select>`
+// element gets special love because it's special, and that's a fact!
+.input-size(@input-height; @input-padding; @font-size; @line-height; @border-radius) {
+  min-height: @input-height;
```
andreapiras, line 109: This has been changed from 
`height: @input-height;`
to 
`min-height: @input-height;
  max-height: @input-height;`

helendurranttw:
I just tested on the deploy, doesn't look like it's fixed 😞 I think height is still being set somewhere? Also, a brief look at the code turned up another place (.select.form-control) where height is being set - make sure you also test inputs used inside the Select component.

helendurranttw (changes requested):
- Take a look for the rogue height which is still being set on .form-control.
- Also test inside Select - I think it is an issue here too
- Could you update _forms.less to explain that this was copied directly from Bootstrap - and also add a comment explaining why we're using min-height / max-height instead of height.
- Lastly - have you tested across browsers? I didn't test that fix thoroughly

andreapiras:
> I just tested on the deploy, doesn't look like it's fixed 😞 I think height is still being set somewhere? Also, a brief look at the code turned up another place (.select.form-control) where height is being set - make sure you also test inputs used inside the Select component.
Locally on my docs it works fine 
image
I can't see the css on the branch docs so I think it's a cache issue

andreapiras:
> * Take a look for the rogue height which is still being set on .form-control.
> * Also test inside Select - I think it is an issue here too
image

> * Could you update _forms.less to explain that this was copied directly from Bootstrap - and also add a comment explaining why we're using min-height / max-height instead of height.
Not imported anymore replaced by overrides

> * Lastly - have you tested across browsers? I didn't test that fix thoroughly
Yes IE11,FF,Chrome e Safari

helendurranttw (approved):
Rename the mixin and you're good to go

PR #614, opened 2020-10-05
DateInput - onChange when month is not selected

JulianG-TW:
## 🖼 Context

There's an issue with Dynamic-Flows when attempting to enter a required date.

krw-dob

This PR doesn't completely fix the issue, but it's the first step.

### Before:
When month is not selected, it defaults to 0 which is January. This results in the component dispatching `onChange` with a seemingly valid data, when in reality the user hasn't finished composing the date.
before

### After:
When month is not selected, we don't construct a Date object, and the `onChange` event contains `null` instead.

after

## 🤔 Considerations

This technically breaks backwards compatibility if any user is relying on January being a default month even when no initial value is specified.

## ✅ Checklist


packages/components/src/dateInput/DateInput.spec.js
```diff
@@ -174,6 +174,20 @@ describe('Date Input Component', () => {
 
   describe('when user interacts', () => {
     describe('with day input', () => {
+      it('returns null if month is not selected', () => {
```
felixxiatransferwise, line 177: shall we add corresponding cases for day and year or do you think that's overkill?

felixxiatransferwise (approved):

packages/components/src/dateInput/DateInput.spec.js
JulianG-TW: done

packages/components/src/dateInput/DateInput.spec.js
```diff
@@ -302,6 +302,15 @@ describe('Date Input Component', () => {
...
       });
+
+      it('returns null when de-selecting month', () => {
+        component = mount(<DateInput {...props} value="2001-01-01" />);
+
+        // De-selects Month
+        simulateSelectChange(0);
+
+        expect(props.onChange).toHaveBeenLastCalledWith(null);
+      });
```
JulianG-TW, line 313: ❌  **Before:** `onChange` wasn't being called when the user de-selected the month.
edge-case--making-month-null

✅  **After:** `onChange` is called with `null` so it triggers validation errors in Dynamic Flows.
edge-case--making-month-null--fixed

packages/components/src/dateInput/DateInput.js
```diff
@@ -55,6 +55,9 @@ const DateInput = ({
   const [internalValue, setInternalValue] = useState(getDateObject);
 
   const getDateAsString = (date) => {
+    if (!isDateValid(date)) {
```
helendurranttw, line 58: Doesn't seem right to have a null return in getDateAsString - perhaps we could move this check?

packages/components/src/dateInput/DateInput.js
```diff
@@ -104,27 +107,28 @@ const DateInput = ({
...
 
   const handleInternalValue = (newDay = day, newMonth = month, newYear = year) => {
-    const dateValue = new Date(newYear, newMonth, newDay);
+    const dateValue =
+      newDay != null && newMonth != null && newYear != null
+        ? new Date(newYear, newMonth, newDay)
+        : null;
+
     const newValue = isDateValid(dateValue) ? dateValue : null;
-    // Don't broadcast already broadcasted value.
```
helendurranttw, line 109: What's the reason for removing the "Don't broadcast already broadcasted value" check?

packages/components/src/dateInput/DateInput.js
```diff
@@ -295,3 +301,14 @@ DateInput.defaultProps = {
 };
 
 export default DateInput;
+
+function useEffectAfterInitialRender(fn, deps) {
```
helendurranttw, line 305: Could you explain why this is necessary?

packages/components/src/dateInput/DateInput.js
JulianG-TW: This was removed to satisfy the following tests in **src/dateInput/DateInput.spec.js**

-› Date Input Component
--› when user interacts
---› with an empty date input
----› calls onChange with null if day is not entered [src](https://github.com/transferwise/neptune-web/blob/ADS-1617--date-input-validation-fix/packages/components/src/dateInput/DateInput.spec.js#L205)
----› calls onChange with null if month is not selected [src](https://github.com/transferwise/neptune-web/blob/ADS-1617--date-input-validation-fix/packages/components/src/dateInput/DateInput.spec.js#L219)
----› calls onChange with null if year is not entered [src](https://github.com/transferwise/neptune-web/blob/ADS-1617--date-input-validation-fix/packages/components/src/dateInput/DateInput.spec.js#L232)

We need the component to dispatch onChange(null) even if the user hasn't entered a valid date yet, to enable 

correct-warning--no-year

packages/components/src/dateInput/DateInput.js
JulianG-TW: I've added this check because:

1. `getDateAsString` implicitly expects a `Date` object (it calls `getFullYear`, `getMonth` and `getDate` 

2. `getDateAsString` is only ever called with `internalValue` as argument (from the `useEffect` callback): `getDateAsString(internalValue)`

3. `internalValue` unfortunately is **not always** a `Date` object: `setInternalValue(INITIAL_DEFAULT_STATE` where
`const INITIAL_DEFAULT_STATE = { year: null, month: null, day: null };`

Therefore `internalValue`'s inferred type is 
```ts
Date | { year: number | null, month: number | null, day: number | null }
```
and it cannot be passed as an argument to `getDateAsString` without checking.

packages/components/src/dateInput/DateInput.js
JulianG-TW: Without this the component is calling `onChange` right after mounting before any user interaction.

But I think a better solution would be to avoid calling `onChange` within a `useEffect`. I'll add 1 more commit.

JulianG-TW: here: https://github.com/transferwise/neptune-web/pull/614/commits/8fce07d93aeb087048375800b1aac981377a2124

packages/components/src/dateInput/DateInput.js
JulianG-TW: I'm happy to change the return type of `getDateAsString` from `string | null` to `string` by returning an empty string when date is not valid.

packages/components/src/dateInput/DateInput.js
JulianG-TW: Thanks for pointing this out. My initial solution was "smelly". I think the new one is better.

packages/components/src/dateInput/DateInput.js
helendurranttw: Thanks for the clear explanation!

packages/components/src/dateInput/DateInput.js
helendurranttw: I think an empty string makes most sense. If you make this change I think you're good to go 👍

packages/components/src/dateInput/DateInput.spec.js
```diff
@@ -184,14 +261,24 @@ describe('Date Input Component', () => {
...
+      it('returns empty string for invalid input', () => {
         component = mount(<DateInput {...props} value="2001-01-01" />);
 
         inputDay = component.find(DAY_SELECTOR);
 
         inputDay.simulate('change', { target: { value: 'aa' } });
 
-        expect(props.onChange).toHaveBeenCalledWith(null);
+        expect(props.onChange).toHaveBeenCalledWith('');
+      });
```
JulianG-TW, line 272: This is an existing test that had to be updated when changing `null` to empty string in 7cbc614af9ee2e48959ec4062544e0c4f374f267

packages/components/src/dateInput/DateInput.js
JulianG-TW: done! e127d71f7e6e6b144f14aa1fde5402f65e264056

PR #615, opened 2020-10-06
Not like this!

JulianG-TW:
## 🖼 Context

When using `DateInput` with Dynamic Flows and validation. We get a validation error too early, simply because the user hits **tab** to jump form **day** to **month**.

This happens because `DateInput` dispatches `onBlur` when the **day** input loses focus, even if the **month** or **year** fields receive focus immediately after.

In the GIF below we use the **tab** key to change focus from **day**, to **month**, to **year**, and back again. We're logging the `onFocus` and `onBlur` events below:

too-many-events-2

## 🚀  HORRIBLE Changes

The solution in this PR (I'm not proud of this) is to debounce dispatching `onFocus` and `onBlur` so that a sequence of `blur > focus` becomes simply `focus`:

debounced-events-2

## 🤔 Considerations

**I really hope someone can come up with a more elegant solution** because we need this for `JsonSchemaForm` (see below).

## ✅ Checklist

----

Here's the same before/after, but in `JsonSchemaForm`:

### Before:
incorrect-warning--to-early

### After:
correct-warning--


JulianG-TW:
Fortunately @helendurranttw came up with a much better solution:
2991dac5ac275633c6e2c19db902bf6be3d39d51 and 
2bf6f5fce7065d7662a689298fe6d9bb76a9e4cc

We'll kill this branch and PR and continue there.

PR #617, opened 2020-10-06
Support new currency flag schema

JulianG-TW:
## 🖼 Context

In the context of DFv3, account requirements will change the schema for currency flags in drop-downs:

currency drop-down

**Before:**
```json
        {
          "title": "GBP",
          "icon": { "name": "gbp" },
          "description": "British pound",
          "const": "GBP"
        },
```
**After:**
```json
        {
          "title": "GBP",
          "icon": { "name": "flag-gbp" },
          "description": "British pound",
          "const": "GBP"
        },
```
**Diff:**
```diff
        {
          "title": "GBP",
-         "icon": { "name": "gbp" },
+         "icon": { "name": "flag-gbp" },
          "description": "British pound",
          "const": "GBP"
        },
```

## 🚀 Changes

This PR changes `SchemaFormControl` to accept both schemas, the old and the new, to ensure backwards compatibility.

## 🤔 Considerations

We may want to remove the backwards compatibility as soon as possible?

We want to move the array of available flags to https://github.com/transferwise/currency-flags.

## ✅ Checklist


felixxiatransferwise (approved):
LGTM. Don't want to block you so approved but left comment about a small typo.

packages/dynamic-flows/src/jsonSchemaForm/JsonSchemaForm.story.js
```diff
@@ -19,9 +20,10 @@ export const basic = () => {
     oneOf: oneOfSchema,
     allOf: allOfSchema,
     'AUD Recipient': audRecipientSchema,
+    currency: currencySchema,
   };
 
-  const schema = select('schema', schemas, simpleSchema);
+  const schema = select('schema', schemas, schemas.currency);
```
StevePole, line 26: Let's not change the default

packages/dynamic-flows/src/jsonSchemaForm/schemaFormControl/SchemaFormControl.js
```diff
@@ -70,10 +70,16 @@ const SchemaFormControl = (props) => {
...
   const mapIcon = (icon) => {
     if (icon) {
-      if (icon.name && availableCurrencyFlags.includes(icon.name)) {
-        return {
-          currency: icon.name,
-        };
+      if (icon.name) {
+        let index = availableCurrencyFlags.indexOf(icon.name);
+        if (index < 0) {
+          index = availablePrefixedCurrencyFlags.indexOf(icon.name);
```
StevePole, line 76: I'd rather we only support the latest spec in the component implementation.  Can we refactor to do any mapping of legacy schemas in a single library function, and leave the rest of the implementation clean?

packages/dynamic-flows/src/jsonSchemaForm/JsonSchemaForm.story.js
JulianG-TW: oops! I meant to revert that BEFORE committing. thanks!

packages/dynamic-flows/src/jsonSchemaForm/schemaFormControl/SchemaFormControl.js
felixxiatransferwise: Not quite sure what you mean by a single library function? Where would this library function go?

If it helps, this implementation wouldn't be long lived. The backend change is already done, was thinking of releasing this PR, releasing the backend and then removing this straight after.

packages/dynamic-flows/src/jsonSchemaForm/JsonSchemaForm.story.js
JulianG-TW: done

packages/dynamic-flows/src/jsonSchemaForm/schemaFormControl/SchemaFormControl.js
StevePole: In styleguide-components we had 'RequirementsService' that mapped older formats to newer (Back then we called them 'dynamic requirements').  I would just move to a separate js file in the same folder, maybe call it legacy.js. and export a function with an appropriate name?

JulianG-TW: I've removed the support for the previous schema. Now only "flag-XXX" will be valid.

StevePole (approved):

PR #620, opened 2020-10-06
chore: update dep

andreapiras:
## 🤔 Considerations
Since this [update](https://github.com/rohanray/next-fonts/commit/960cb166af58ebc7512e23e98173a8ee436438fa#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346) next-fonts has introduced a change (with version 1.1.0) which delegates nextjs for handling css/sass/scss files. 

```
neptune-web/node_modules/@transferwise/neptune-css/dist/css/neptune.css
ModuleParseError: Module parse failed: Unexpected character '' (1:0)
```

Currently we use withCss plugin which is trying to handle the css files

```
{
    loader: 'css-loader',
    options: {
      modules: false,
      minimize: false,
      sourceMap: true,
      importLoaders: 1
    }
  },
  { loader: 'postcss-loader', options: { config: [Object] } }
```
next-css doesn't have any font loader, so we have to  pass the extra [handling](https://github.com/transferwise/neptune-web/pull/620/files#diff-586b4b25452f221c125eafec6ebfbafacab2898a8a89835ac4110391837638d1R49) 

Since we are using withCss we are not accessing the default nextJS css handler

### Can't we not use the default nextJs css handler?
  Unfortunately at the moment we can't because the way neptune provides css in component (i.e we import css per each components inside node_modules). This cause nextjs to throw an error :

   ```
- Global CSS cannot be imported from within node_modules https://github.com/vercel/next.js/issues/13282
-./node_modules/@transferwise/components/build/es/polyfill/accordion/Accordion.css
```

Following this [thread](https://github.com/vercel/next.js/issues/10059) I tried to remove custom withCss and upgrade nextjs to canary version but after upgrading I still get the same error. Because apparently this [feature](https://github.com/vercel/next.js/issues/13282. ) is being discussed to support libraries that ships css the way neptune-web does

## ✅ Checklist

### CODE REVIEW

File: packages/docs/next.config.js [hunk 1]
```diff
@@ -21,34 +20,52 @@ const assetPrefix =
...
+            const { isServer } = options;
+
+            config.module.rules.push(
+              {
                 test: [/\.code.js$/, /\.txt$/],
                 use: 'raw-loader',
-              });
+              },
+              {
+                test: /\.(woff(2)?|eot|ttf|otf|svg)(\?v=\d+\.\d+\.\d+)?$/,
```
andreapiras, line 50: WithFont has a property called issuer 
```
issuer: {
  // Next.js already handles url() in css/sass/scss files
  test: /\.\w+(?<!(s?c|sa)ss)$/i,
},
```
which is causing the build to fail and causing the following error:
ModuleParseError: Module parse failed: Unexpected character '' (1:0)

helendurranttw (approved):

PR #626, opened 2020-10-07
Date input onFocus/onBlur Grouped Events

JulianG-TW:
## 🖼 Context

When using `DateInput` with Dynamic Flows and validation. We get a validation error too early, simply because the user hits **tab** to jump form **day** to **month**.

This happens because `DateInput` dispatches `onBlur` when the **day** input loses focus, even if the **month** or **year** fields receive focus immediately after.

In the GIF below we use the **tab** key to change focus from **day**, to **month**, to **year**, and back again. We're logging the `onFocus` and `onBlur` events below:

too-many-events-2

## 🚀  Changes

Fortunately @helendurranttw came up with a much better solution than mine #615

The solution in this PR is to use [FocusEvent.relatedTarget](https://developer.mozilla.org/en-US/docs/Web/API/FocusEvent/relatedTarget) to discriminate when a blur event is followed by a focus event within the same DateInput component or not.

debounced-events-2

## 🤔 Considerations

Because this are BREAKING CHANGES, we're merging to the `beta` branch.

~~This solution doesn't completely work if we render two `DateInputs` next to each other, changing focus from one to the next would not result in `onBlur` and `onFocus` events.~~

We've solved the issue with 2 or more DateInput components rendered on the same form with a lot of help from @andreapiras. We had to jump through some hoops to make it work in IE11.

> Even though FocusEvent.relatedTarget is supported by IE
> (https://developer.mozilla.org/en-US/docs/Web/API/FocusEvent/relatedTarget)
> "IE11 sets document.activeElement to the next focused element before the blur event is called."
> (https://stackoverflow.com/a/49325196/986241)
> Therefore if the relatedTarget is null, we try the document.activeElement,
> which may contain the HTML element that is gaining focus

However this solution still leaves a difference in IE11. The `onFocus` event is triggered every time.

After spending some time trying to solve this, and considering [IE11 accounts for on only 0.13% of our users in the send flow](https://analytics.google.com/analytics/web/#/report/content-pages/a16492313w33608050p32817948/explorer-table.plotKeys=%5B%5D&explorer-table.secSegmentId=analytics.browser&_r.drilldown=analytics.pagePath:~2Fsend~2F/), we decided this was good enough.

## ✅ Checklist


packages/components/src/dateInput/DateInput.js
```diff
@@ -221,11 +235,9 @@ const DateInput = ({
                 id="date-input-day"
```
andreapiras, line 235: I don't think we should hardcode ids, they won't be unique if we have multiple field on the same page

packages/components/src/dateInput/DateInput.js
```diff
@@ -302,4 +312,9 @@ DateInput.defaultProps = {
   },
 };
 
+function shouldPropagate(e) {
```
andreapiras, line 315: Totally optional
```suggestion
const shouldPropagate = ({ relatedTarget }) =>
  !relatedTarget ||
  (!relatedTarget.className.includes('date-input-control') &&
    relatedTarget.id !== 'date-input-month');
```

andreapiras, line 315:
```suggestion
const shouldPropagate = ({ relatedTarget }) =>
  !relatedTarget ||
  (!relatedTarget.className.includes('date-input-control') &&
    relatedTarget.id !== 'date-input-month');
```

packages/components/src/dateInput/DateInput.js
```diff
@@ -302,4 +312,9 @@ DateInput.defaultProps = {
...
+  const rt = e.relatedTarget;
+  // should only propagate if the relatedTarget is null or not part of this DateInput component.
+  return !rt || (!rt.className.includes('date-input-control') && rt.id !== 'date-input-month');
+}
```
andreapiras, line 319: Totally optional 
```suggestion
const shouldPropagate = ({ relatedTarget }) =>
  !relatedTarget ||
  (!relatedTarget.className.includes('date-input-control') &&
    relatedTarget.id !== 'date-input-month');
```

packages/components/src/dateInput/DateInput.js
```diff
@@ -302,4 +312,9 @@ DateInput.defaultProps = {
   },
 };
 
+function shouldPropagate(e) {
+  const rt = e.relatedTarget;
+  // should only propagate if the relatedTarget is null or not part of this DateInput component.
+  return !rt || (!rt.className.includes('date-input-control') && rt.id !== 'date-input-month');
```
andreapiras, line 318: What happens if we have two DateInout with same ids?

packages/components/src/dateInput/DateInput.js
```diff
@@ -221,11 +235,9 @@ const DateInput = ({
                 id="date-input-day"
                 type="number"
                 name="day"
-                className="form-control"
+                className="form-control date-input-control"
```
andreapiras, line 238: I'd add a prefix here : `tw-date-input-control`

packages/components/src/dateInput/DateInput.js
```diff
@@ -208,7 +206,23 @@ const DateInput = ({
...
+            e.stopPropagation();
+          }
+        }}
+        onBlur={(e) => {
+          if (shouldPropagate(e)) {
+            onBlur();
+          } else {
+            e.stopPropagation();
+          }
+        }}
```
andreapiras, line 224:
```suggestion
        onFocus={(e) => (shouldPropagate(e) ? onFocus() : e.stopPropagation())}
        onBlur={(e) => (shouldPropagate(e) ? onBlur() : e.stopPropagation())}
```

andreapiras, line 224:
```suggestion
       onFocus={(e) => (shouldPropagate(e) ? onFocus() : e.stopPropagation())}
       onBlur={(e) => (shouldPropagate(e) ? onBlur() : e.stopPropagation())}
```

packages/components/src/dateInput/DateInput.js
JulianG-TW: There's currently no way to give them different ids, so they'll always have the same id.
If we render two DateInputs, and change focus from one to the other, it won't trigger `onBlur` and `onFocus`.

As I mentioned in the PR description, I tried avoiding this problem (but enabling different ids, but I think the solutions was too complicated for what we'd be gaining.

packages/components/src/dateInput/DateInput.js
JulianG-TW: Thank you!

I had tried
```
        onFocus={(e) => shouldPropagate(e) ? onFocus() : e.stopPropagation()}
        onBlur={(e) => shouldPropagate(e) ? onBlur() : e.stopPropagation()}
```
but I was getting linter errors.

I now realise they were just "prettier" errors.

packages/components/src/dateInput/DateInput.js
JulianG-TW: done @ 3eaecd3b3966cefae8e1bf0841839262d3d4f3a0

packages/components/src/dateInput/DateInput.js
JulianG-TW: done @ e7bdc873e945c67e054bc90993c5a766a23e17f1

packages/components/src/dateInput/DateInput.js
andreapiras: So I think the only option we have is to pass the ids as props. Since this one is a breaking change already we may as well change that. I was thinking to as something like:

packages/components/src/dateInput/DateInput.js
```diff
@@ -302,4 +299,12 @@ DateInput.defaultProps = {
   },
 };
 
+// Should only propagate if the relatedTarget is null or not part of this DateInput component.
+function shouldPropagate({ relatedTarget }) {
+  return (
+    !relatedTarget ||
+    (!relatedTarget.className.includes('tw-date-input-control') &&
+      relatedTarget.id !== 'date-input-month')
```
andreapiras, line 307: Why are we adding this check here?. Should this not be considered still part of the DateInput. While people are on the dropwnn they are still interacting with the component and that means they are still focused on it

packages/components/src/dateInput/DateInput.js
JulianG-TW: I think you have a good point. If we have unique ids for each of the 3 sub-components: day, month, year we don't need to rely on `className.includes('date-input-control')` which IIRC was only introduced in this PR.

If we do this, in the `shouldPropagate` function we can check `relatedTarget.id` against our 3 ids. **This is great!**

But I don't think we need to expose this to consumers of `DateInput` component. 
All we need to do, is ensure we have unique ids, we don't really need to make them custom.

I'll try this now. In the meantime, let me know if you have any concerns with this. I may be overlooking something.

JulianG-TW: done @ 4de3387e143aabd1dd669a7b362a653b13e3b642

packages/components/src/dateInput/DateInput.js
JulianG-TW: that was very confusing. hopefully the new implementation, which only uses ids, is more clear.

packages/dynamic-flows/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@transferwise/dynamic-flows",
   "version": "5.5.0",
-  "description": "Transferwise dynamic flows and forms",
+  "description": "Transferwise dynamic flows and forms.",
```
JulianG-TW, line 4:
```suggestion
  "description": "Transferwise dynamic flows and forms",
```

packages/components/src/dateInput/DateInput.js
```diff
@@ -75,24 +85,20 @@ const DateInput = ({
...
-          {monthLabel}
-        </label>
+      // eslint-disable-next-line
+      <label>
+        <span className="sr-only">{monthLabel}</span>
         <Select
-          id="date-input-month"
           name="month"
-          className="form-control"
+          className="form-control tw-date-input-control"
```
JulianG-TW, line 93: @andreapiras I don't think we need this?

packages/components/src/dateInput/DateInput.js
```diff
@@ -207,28 +213,29 @@ const DateInput = ({
...
           <div className="col-sm-3">
             <div className={`input-group-${size}`}>
-              <label className="sr-only" htmlFor="date-input-day">
-                {dayLabel}
+              <label>
+                <span className="sr-only">{dayLabel}</span>
+                <input
+                  type="number"
+                  name="day"
+                  className="form-control tw-date-input-control"
```
JulianG-TW, line 232: @andreapiras I don't think we need `tw-date-input-control` here?

packages/components/src/dateInput/DateInput.js
```diff
@@ -237,21 +244,18 @@ const DateInput = ({
...
         <div className="col-sm-4">
           <div className={`input-group-${size}`}>
-            <label className="sr-only" htmlFor="date-input-year">
-              {yearLabel}
+            <label>
+              <span className="sr-only">{yearLabel}</span>
+              <input
+                type="number"
+                name="year"
+                className="form-control tw-date-input-control"
```
JulianG-TW, line 252: same regarding `tw-date-input-control`

packages/components/src/dateInput/DateInput.js
JulianG-TW: (I've removed it)

packages/components/src/dateInput/DateInput.js
JulianG-TW: (I've removed it)

packages/components/src/dateInput/DateInput.js
JulianG-TW: (I've removed it)

andreapiras (approved):

PR #633, opened 2020-10-13
docs: RFC for i18n of components

dozortsev:
**Rendered version of RFC**: https://github.com/transferwise/neptune-web/blob/translation-react-intl/rfc/0004-i18n.md

See also this https://github.com/transferwise/neptune-web/pull/633#discussion_r513512287

Commit which contains most of changes for prototype: https://github.com/transferwise/neptune-web/pull/633/commits/966e79f178abf756933d3c56c6ff8ef656d2cf3b

## ✅ Checklist


.circleci/config.yml
```diff
@@ -91,6 +91,11 @@ jobs:
       - run: *test
       - *persist-workspace
 
+  push-translations-to-crowdin:
+    <<: *container
+    steps:
+      - *attach-workspace
```
dozortsev, line 97: job is not doing actually anything yet, for now it's for just for test/demo purposes

https://app.circleci.com/pipelines/github/transferwise/neptune-web/5335/workflows/59b704ea-0b0e-4a84-ac73-c5d3a9f45c91

packages/components/src/common/CloseButton/CloseButton.js
```diff
@@ -1,20 +1,24 @@
...
-    <CrossIcon size={24} />
-  </button>
-);
+export const CloseButton = ({ onClick, className }) => {
+  const intl = useIntl();
+  return (
+    <button
+      type="button"
+      className={classNames('tw-close-button', 'btn-link', 'text-no-decoration', className)}
+      aria-label={intl.formatMessage({ id: 'np.c9n.aria.lbl.close' })}
```
andreapiras, line 15: I think we can use better ids

andreapiras, line 15: We need to find a way to default this to english if for some reason we don't find the translation

packages/components/src/common/neptuneProvider/NeptuneProvider.js
```diff
@@ -0,0 +1,28 @@
...
+import Types from 'prop-types';
+import { createIntl, createIntlCache, RawIntlProvider } from 'react-intl';
+import messages from '../lang';
+
+const NeptuneProvider = ({ locale, children }) => {
+  const cache = createIntlCache();
+  const intl = createIntl(
+    {
+      locale,
+      messages: messages[locale],
```
andreapiras, line 11: We should check if this exists and if it doesn't provide a default

packages/components/src/common/CloseButton/CloseButton.js
dozortsev: for example?

packages/components/src/common/neptuneProvider/NeptuneProvider.js
dozortsev: it accepts `defaultLocale` prop next to `messages` one

dozortsev: added it here

packages/components/src/common/CloseButton/CloseButton.js
dozortsev: https://github.com/transferwise/neptune-web/pull/633#discussion_r505497856

conorcussell, line 15: I've found that co-locating a messages file with each component so you can import an object is a good way to handle these ids. 

This file will use https://formatjs.io/docs/react-intl/api/#definemessagesdefinemessage to define each mesage, then your messages file which is sent to crowdin will be generated using https://github.com/akameco/extract-react-intl-messages at build time
```
/Component.js
/Component.messages.js
```

```
import messages from './Component.messages'
...

intl.formatMessage(message.closeButtonLabel)
```

conorcussell (approved):
Seems like a good approach.

Since the list of keys is so small it won't be worth doing at the moment, but if the list of strings or supported locales grows a bit, then you could consider dynamically importing the messages file for each locale .

.circleci/config.yml
```diff
@@ -186,7 +191,16 @@ workflows:
             branches:
               only:
                 - main
-
+      - push-translations-to-crowdin:
```
oliverviljamaa, line 194: Any reason to prefer CLI in CI over the Crowdin GitHub integration? The CI approach can be flaky, has reduced visibility, requires more configuration, and doesn't allow testing languages locally. The benefit of the CI approach is that the messages are translated before they reach `main`, but I believe that could be achieved with the integration, too.

The differences are explained here:
https://transferwise.atlassian.net/wiki/spaces/EKB/pages/1054900981/Setting+up+an+integration+with+a+new+Crowdin+project

rfc/0004-i18n.md
```diff
@@ -0,0 +1,134 @@
+- Start Date: 2020-10-15
+- Status: Progress
+
+# Summary
+
+Some of Neptune Web components have default and/or constant messages which needs to be localised.
+
+This RFC describes a proposal using [react-intl](https://formatjs.io/docs/react-intl) library and Crowdin along with a custom context provider. This solution will be compatible with React or React + Angular hybrid apps/libraries.
```
oliverviljamaa, line 8: Why did you choose `react-intl` over `retranslate`? I may agree, but wondering what your rationale was.

rfc/0004-i18n.md
```diff
@@ -0,0 +1,134 @@
...
+Note: all changes in code are not a final version, they are for demo and proof of concept purposes.
+
+# Motivation
+
+Some of neptune web components required translations for 'default text' where we define 'default text':
+ - text that is not likely to be changed.
+ - text that we are able to anticipate.
+ - optional text that has a default value.
+
+Currently we don't have many occurences of such strings but we need a scalable solution to handle more in future.
```
oliverviljamaa, line 19: 👏 Love that you're doing this!

rfc/0004-i18n.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Consumers app usage
+
+Consumers needs to wrap Neptune components in the `NeptuneProvider`. Usually this should be done at root level.
+
+```js
+class App {
+  render() {
+    return (
+      <NeptuneProvider locale={'en'}>
```
oliverviljamaa, line 69: This _might_ be too magic, but could we fall back to a wrapping `react-intl` provider locale, if one exists?

rfc/0004-i18n.md
```diff
@@ -0,0 +1,134 @@
...
+```
+
+## Playground usage
+
+For example if you are using Storybook as playground it is possible to setup the env just adding a decorator that wraps your story with neptune provider.
+
+```js
+// This example with Storybook v5
+addDecorator(...);
+addDecorator((storyFn) => <NeptuneProvider locale="en">{storyFn()}</NeptuneProvider>);
```
oliverviljamaa, line 87: Could you somehow add this decorator to all stories out of the box?

rfc/0004-i18n.md
```diff
@@ -0,0 +1,134 @@
...
+```
+
+# Crowdin integration
+
+A Circle job is going to be added to the current workflow of PR branches. This job is going to take care of translations pushes to Crowdin ([see job example](https://app.circleci.com/pipelines/github/transferwise/neptune-web/5336/workflows/72e0c371-d39f-44cf-8933-20859a0c6b00)). 
+Contributors have to manually approve this job ([Holding a Workflow/Job for a Manual Approval](https://circleci.com/docs/2.0/workflows/#holding-a-workflow-for-a-manual-approval)) to trigger it they want push messages for translations.
+Having this job as part of PR branch will allow Translators start working on the translations in parallel with the developers that can keep working on their changes. Ideally all PRs should be merged once all the translations are provided, to reduce the number of merges and consequent version bumps.
```
oliverviljamaa, line 94: Could we achieve this with the Crowdin GitHub integration? An additional benefit is that engineers would see the translation progress without needing Crowdin project access.

rfc/0004-i18n.md
```diff
@@ -0,0 +1,134 @@
...
+- `[ComponentName]` - component name (e.g `CloseButton`, `Pagination` etc.);
+- `[elementName]` - element that receive the translation;
+
+Each part of the key should be separated by dot.
+
+To safe a bit of bundle file size we plan to use [`babel-plugin-react-intl-id-hash`](https://www.npmjs.com/package/babel-plugin-react-intl-id-hash) will be used to rewrite and shorten message keys.
+
+# Supported Languages
+
+Same as other TransferWise Web app/libraries.
```
oliverviljamaa, line 115: Should we just link here instead of duplicating the list?
https://transferwise.atlassian.net/wiki/spaces/EKB/pages/1054213243/Supported+languages

rfc/0004-i18n.md
andreapiras: Hey Oliver that's great, i didn't personally know that there was a Crowdin integration but it sounds good

rfc/0004-i18n.md
andreapiras: I think that people using react-intl should not use NeptuneProvider at all and just keep using their provider. Maybe we should figure it out what are the exact differences between these two and see if we provide compatibility with both type of providers
```
<RawIntlProvider value={intl}>{children}</RawIntlProvider>;```
```
<IntlProvider locale={navigator.language}>
  <App importantDate={new Date(1459913574887)} />
</IntlProvider>, 
```

File: rfc/0004-i18n.md (hunk 8 above)
Comment (andreapiras) [reply rfc/0004-i18n.md:87]: I don't think there's a way to do that if not interacting with the cosnumer `.storybook` folder of the consumer, which I don't feel very comfortable with but I might be wrong. Also config changes based on storybook version. If you have any idea please share with us :)

File: packages/components/src/common/CloseButton/CloseButton.js (hunk 2 above)
Comment (andreapiras) [reply packages/components/src/common/CloseButton/CloseButton.js:15]: That's a good idea, we'll definitely explore it. It would also reduce tech debt when you delete a component. Thank you

Comment (andreapiras) [reply packages/components/src/common/CloseButton/CloseButton.js:15]: Maybe we can also a hybrid approach where we can put shared translations in a common folder and component specific ones into the component folder. But I'm just talking out loud

**Comment by karanjthakkar** [comment]
> Since the list of keys is so small it won't be worth doing at the moment, but if the list of strings or supported locales grows a bit, then you could consider dynamically importing the messages file for each locale .

👋 super new to this project but something @conorcussell said got me wondering:

How do we plan to manage the bundlesize bloat that this will bring in the consumers? For eg: in homepage, we inject the map of current language translations into `window.config.translations` which is then consumed by the app. So at any point in time, translations of only one language are delivered over the wire to the user. If neptune bundled default text translations for all supported languages, that might cause us to deliver translations that are not needed to the user. 

for eg: in the RFC, that's 4 strings x 14 langauges, which is 56 strings. Excluding the current lang, we now have 52 strings that will not be used in the client.

Having said that, some questions to think about:

- Is there an upper limit to the number of default strings that neptune will bundle? How much is it? Do we care about the bloat added by the extra strings as and when we add support for more languages?
- Is there a way to allow consumers to provide the strings so that they can decide when to load which languages' string?

File: rfc/0004-i18n.md (hunk 10 above)
Comment (dozortsev) [reply rfc/0004-i18n.md:115]: @andreapiras @oliverviljamaa this is open source repository, we shouldn't link here things that are not accessible for people out side of TransferWise

**Comment by andreapiras** [comment]
Hey @karanjthakkar ,
> How do we plan to manage the bundlesize bloat that this will bring in the consumers? For eg: in homepage, we inject the map of current language translations into `window.config.translations` which is then consumed by the app. 

How do you guys handle this? 
Do you return just the messages from the server based on user-agent?
What happens if a user switch language? Do you run an extra request to retrieve the translations?
Before this proposal we investigated another option that would allow consumers to bundle the translations at app level not neptune level but it will force all consumers to (also) use react-intl

File: rfc/0004-i18n.md (hunk 5 above)
Comment (dozortsev) [reply rfc/0004-i18n.md:8]: don't have much experience with retranslate, but we suggest use react-intl because it's quite mature and seems well maintained library ([repo on GH](https://github.com/formatjs/formatjs/tree/main/packages)) and most likely because it's based on JS standard internationalization API, see https://formatjs.io/docs/react-intl/#runtime-requirements and [Intl](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl)

Comment (dozortsev) [reply rfc/0004-i18n.md:8]: I'm sure those who never used it but used only retranslate will be easy to understand and use it.

Comment (dozortsev) [reply rfc/0004-i18n.md:8]: would you still recommend retranslate over react-intl? if so why? 🙂

File: packages/components/src/common/CloseButton/CloseButton.js (hunk 2 above)
Comment (conorcussell) [reply packages/components/src/common/CloseButton/CloseButton.js:15]: Yep, that could also work. One of the main points of the suggestion is just that you are not writing out the strings manually so you get proper IDE support and tools like TypeScript could catch issues for you.

Generally I think its best to prefer duplication of the translated text over sharing easily between components. While two components may currently both be using the text "Close" - they are still distinct components and you can avoid coupling them by defining two keys `ComponentOne.closeButtonText` and `ComponentTwo.closeButtonText`.

The translations tools support this by suggesting similar strings to the translators

Screenshot 2020-10-20 at 14 33 00

**Comment by karanjthakkar** [comment]
> Do you return just the messages from the server based on user-agent?

I'm not sure yet. I joined the team yesterday so still unsure of the exact details here 🙈 But from what I understand, whenever you change the language from the selector in the navigation, the client sets a `localeData` cookie to the language code. Then the page is refreshed and the new language translations are injected by the server based on this cookie. Essentially the page reloads on language change which seems fair.

> Before this proposal we investigated another option that would allow consumers to bundle the translations at app level not neptune level but it will force all consumers to (also) use react-intl

Wouldn't consumers still get react-intl in their bundle anyway since neptune will come bundled with it? 

One way to make consumers independent of react-intl, and I'm just spitballing here, would be where neptune could provide a utility function that consumers could call at app startup to send translations to neptune. Then internally neptune can use those translations to provide them to react-intl using https://formatjs.io/docs/react-intl/api#definemessagesdefinemessage while allowing consumers to use translations however they want. What do you think?

**Comment by andreapiras** [comment]
> > Do you return just the messages from the server based on user-agent?
> 
> I'm not sure yet. I joined the team yesterday so still unsure of the exact details here 🙈 But from what I understand, whenever you change the language from the selector in the navigation, the client sets a `localeData` cookie to the language code. Then the page is refreshed and the new language translations are injected by the server based on this cookie. Essentially the page reloads on language change which seems fair.
> 
> > Before this proposal we investigated another option that would allow consumers to bundle the translations at app level not neptune level but it will force all consumers to (also) use react-intl
> 
> Wouldn't consumers still get react-intl in their bundle anyway since neptune will come bundled with it?
 Yes you are correct ignore my comment. Consumers will alway get.
This is the other option we were looking for 
https://formatjs.io/docs/guides/distribute-libraries that will shift the translation bundling at app level rather than library level.

File: packages/components/src/common/CloseButton/CloseButton.js (hunk 2 above)
Comment (andreapiras) [reply packages/components/src/common/CloseButton/CloseButton.js:15]: Cool that's nice, thanks for sharing this

File: rfc/0004-i18n.md (hunk 6 above)
Comment (danlaush) [reply rfc/0004-i18n.md:19]: Great initiative, definitely an important but tricky problem to solve.

I'm +1 on being pretty concerned about bundle bloat for this. Do I understand correctly that `NeptuneProvider` would include all the messages for all the components? This is a problem that grows exponentially so I would feel much more confident nipping it in the bud. One more string = 15+ more strings. One new language = all of the strings again.

For comparison (hah), this is an ES build of the `comparison-components` lib with all the translations. You can see how big it is for one widget. I think Neptune would have fewer overall than this content-heavy lib, but it's still a problem.
Screenshot 2020-10-20 at 15 17 29

I agree with @conorcussell about co-locating translations within the component, and I'd go a step further and say that components should each have their own `TranslationProvider` for just the messages in that component. This would tree-shake out messages not used by the consumer. You could still have `NeptuneProvider` for managing global language choice, but then each component provides messages for itself. Not sure how to handle shared strings but that feels like more of an edge case.

```jsx
// App.js
export default () => (
  <NeptuneProvider locale={'en'}>
    <Modal />
  </NeptuneProvider>
)

// Modal.js
import messages from './modal.messages';
export default () => {
  const intl = useIntl();
  return (
    <TranslationProvider language={intl.language} messages={messages}>
      <FormattedMessage id="random.message" />
    </TranslationProvider>
  )
}
```

rfc/0004-i18n.md
```diff
@@ -0,0 +1,134 @@
...
+ - text that we are able to anticipate.
+ - optional text that has a default value.
+
+Currently we don't have many occurences of such strings but we need a scalable solution to handle more in future.
+
+Some examples of 'default text' that need to be translated are:
+- `aria-label` in `CloseButton` ("Close");
+- `next` and `previous` button text in  `Pagination` component ("Previous" and "Next").
+- default messages in `Upload` component.
+- validation messages in `Field`.
```
danlaush, line 25: I think this will be 90% of the messages, by byte count 😂

rfc/0004-i18n.md
danlaush: I think Oliver just means all the stories in Neptune

rfc/0004-i18n.md
danlaush: Organic Growth would either need this to be a bigger list or a way to inject more translations, since we have SEO pages in languages outside the TW core (`bg, zh, cs, fi, el, ko, nb-NO, sk, sv-SE, th`) like the [currency converter](https://transferwise.com/no/currency-converter/).

rfc/0004-i18n.md
danlaush: just tricky that I wouldn't like to ship 2 different translation libs to the browser if I can avoid it 😅  don't see a good solution though unfortunately.

rfc/0004-i18n.md
danlaush: I'm also super into dynamic imports for translations, looking into trying something like that for comparison-components this quarter. The challenge with that is SSR, since nextjs has one blessed way to do it.

karanjthakkar:
> This is the other option we were looking for
https://formatjs.io/docs/guides/distribute-libraries that will shift the translation bundling at app level rather than library level.

Yes, this solves the problem that I was talking about in my initial message 🙂 Regardless of whether we choose the convention or configuration route, a standard approach prescribed by neptune would be really helpful in making sure that consuming apps can easily consolidate translations across libraries. 👌

rfc/0004-i18n.md
oliverviljamaa: @dozortsev Of course! Disregard my comment, sorry 😊

.circleci/config.yml
dozortsev: thanks, will check

rfc/0004-i18n.md
danlaush: I like retranslate for the size - react-intl is 4x larger

dozortsev: yeah, but we already do that, e.g UAS widely used component library which uses react-intl while most of their consumer apps rely on retranslate.

react-intl bundle is tree-shakable, and I already mention above it's based on native JS API which already built in browsers so it's weight should be small

rfc/0004-i18n.md
dozortsev: @andreapiras 

> ..I think that people using react-intl should not use NeptuneProvider at all and just keep using their provider

I'm afraid this is not possible as our providers knows only about our translations ([based on configs](https://github.com/transferwise/neptune-web/pull/633/files#diff-1787b0a7ee555e8de50b41f061e0517ce0ac6ec4b59ac21950426dcae905fcafR10-R12), e.g path to files with messages) and same for consumer app react-intl provider which only knows about it's own messages...

dozortsev: @oliverviljamaa 

> This might be too magic, but could we fall back to a wrapping react-intl provider locale...

I think we can try implement this by checking if there are some other react-intl providers hight or lower in tree but why would you want that? just not to define `local="jp"` two times? 🙂

rfc/0004-i18n.md
dozortsev: > ..Could you somehow add this decorator to all stories out of the box?

@oliverviljamaa but why would we do that? adding it in Storybook 5 via `addDecorator` is like adding [it in consumer application](https://github.com/transferwise/neptune-web/blob/translation-react-intl/rfc/0004-i18n.md#consumers-app-usage), I mean for same purposes.

dozortsev: @danlaush yeah, we definitely will mention about NeptuneProvider our [setup pages](https://transferwise.github.io/neptune-web/about/Setup#usage)

rfc/0004-i18n.md
dozortsev: yep, we will check it

rfc/0004-i18n.md
dozortsev: @danlaush thanks! I think we gonna add those langs as well

packages/components/src/common/CloseButton/CloseButton.js
dozortsev: @conorcussell re. https://github.com/transferwise/neptune-web/pull/633#discussion_r508413193

did I get this right that so using `.message.js` files per component we can avoid creating `.json` files (`jp.json`, `en.json` etc.), but where translations will be define? in `.message.js`? could you please show structure of the file with other langs?

rfc/0004-i18n.md
andreapiras: @danlaush but the code above does wrap all the stories or am I missing something?

rfc/0004-i18n.md
danlaush: Thanks @dozortsev, though I know this adds 10x more strings to the messages bloat 😅

rfc/0004-i18n.md
danlaush: Yeah I think we're all on the same page here

rfc/0004-i18n.md
oliverviljamaa: I'm happy with the decision, thanks for explaining 🙇

rfc/0004-i18n.md
oliverviljamaa: Yes, but the benefit might not be worth the potential fragility.

packages/components/src/common/CloseButton/CloseButton.js
conorcussell: You still store the `json` files in the repo - but they are not edited by hand. 

```
/src
  /components
    /Checkbox.tsx
    /Checkbox.message.ts # this is the file you edit
/messages/
  /default/messages.json # this is auto-generated with a pre-commit hook
  ja.json # these are committed by crowdin
  en.json
  ```

**Benefits:**
- It is clear which Components are using which messages, when you delete or move a component its messages go with it, and it encourages not sharing the keys between separate components.
- You are importing the object you can auto-complete in your editor and warnings when you try and reference something which doesn't exist.

Comment (karanjthakkar) [reply packages/components/src/common/CloseButton/CloseButton.js:15]: I'm not completely convinced colocating strings with components is a good idea for a few reasons:

- Building tooling to pull strings is much more involved because whatever custom script you write would need to look through all your nested folder hierarchies to find string files.
- For files that are going to stay fairly constant colocation doesn't provide the same benefits that you would get by colocation tests/styles. Plus, jump to definition in any editor will automatically take you to the string if that is a concern (atleast in projects that use flow/ts. not sure about this one.)
- Colocating would allow for unnecessary repetition because it's likely (a handwavy guess) that a most strings would be used in more than one place, especially in the scope of a design system. A hybrid approach where you have shared and component specific strings has very similar flaws: when a component specific string becomes used in another place, who would tell the dev to move it to shared?

Perks of a centralized file - it solves all the above: 

- Tooling to pull strings from crowdin, etc. is straightforward to build because there's only one source
- Identifying duplicates is visually more easy if there's just one source. So if a button and an input have same default text for whatever reason, now they don't need to have two copies. (I understand same string can mean different things based on the context, but the point still stands for most cases.)
- You still get autocomplete as you still import strings in the same way

Happy to hear what folks think about this? 🙂

Comment (danlaush) [reply packages/components/src/common/CloseButton/CloseButton.js:15]: Tooling shouldn't be an issue, CrowdIn handles globs pretty seamlessly ([example](https://github.com/transferwise/lienzo-frontend/blob/master/crowdin.yaml#L8-L9)) so as long there's a convention for translation file naming.

My concern is bundle size. With a centralised Neptune messages blob, could it be tree-shaken to not include unused messages in my app? If I'm not using `Field`, my users don't need all of the validation messages (times 15 languages). It adds up pretty quickly.

Comment (karanjthakkar) [reply packages/components/src/common/CloseButton/CloseButton.js:15]: > Tooling shouldn't be an issue, CrowdIn handles globs pretty seamlessly (example) so as long there's a convention for translation file naming.

ooooh fun! TIL.

> It's a fair point about re-usable strings though. I don't have a good solution for that, but I wonder how much of a problem duplication would be.

Yeah I think you reusability and bundle size kinda go hand in hand. So if you worry about the latter, it would be prudent to worry about the former as well, while things are still in early stages.

Comment (oliverviljamaa) [reply packages/components/src/common/CloseButton/CloseButton.js:15]: I feel we're talking about multiple different things at the same time here.

Maybe it'd help if we established some _requirements_? E.g.:

**Required**
- users download translations only in the language they're using
- users download translations only for the components they're seeing

**Ideal**
- using translation keys in component code is type safe
- the tooling required in the library is minimal
- unused translations are alerted of

I like the idea of colocating the messages with the component, as it aligns with how we structure the library — component code, styles, tests, and stories live together. It'd also help avoid the problem we have with "orphan" translations, as when the messages live closer to the component, they're more likely to be cleaned up. There are likely other ways to satisfy the requirements, though.

Regarding common messages and component-specific messages, I agree with @conorcussell and would opt for component-specific only and accept some repetition in tradeoff for reduced complexity, higher flexibility, and explicitness. After all, we don't know how many common messages we'd even have (the `CloseButton` would just have one message, for example, and be used in other components), and the message files compress well.

File: rfc/0004-i18n.md (hunk 10 above)
Comment (oliverviljamaa) [reply rfc/0004-i18n.md:115]: All the more reason to ensure we don't bundle all translations from the get-go.

File: packages/components/src/common/CloseButton/CloseButton.js (hunk 2 above)
Comment (karanjthakkar) [reply packages/components/src/common/CloseButton/CloseButton.js:15]: That makes sense. I don't have a strong opinion on how neptune chooses to organise strings in it's codebase. As long as consumers of the neptune have the ability to decide how and which strings are bundled in the final app shipped to the user, as mentioned [here](https://github.com/transferwise/neptune-web/pull/633#issuecomment-712826413). If distributing neptune with [this](https://github.com/transferwise/neptune-web/pull/633#issuecomment-712866074) kind of a solution works okay with the colocating strings, then I'm happy with that.

Comment (conorcussell) [reply packages/components/src/common/CloseButton/CloseButton.js:15]: > It's a fair point about re-usable strings though. I don't have a good solution for that, but I wonder how much of a problem duplication would be.

I think this exactly the sort of thing GZIP is good at, the first occurrence of a duplicate string will be included, but everything else will be references to that.

Not sure how annoying it'll be in a library like Neptune since the strings are probably going to be fairly generic, but I've previously found sharing strings quite painful. If two components are coupled by using the same key - but actually used in wildly different contexts - then when you've got translators changing them in a language you do not understand then you can end up with some weird results in your UI.

File: .circleci/config.yml (hunk 4 above)
Comment (sbrandwoo) [reply .circleci/config.yml:194]: > and doesn't allow testing languages locally

This is the main one for me - being able to test all languages locally is great.

**Comment by sbrandwoo** [comment]
If components end up with their own translations, rather than having a Neptune-wide translations, would there be any benefit to having the `NeptuneProvider` over the simpler interface of `<SomeComponent locale="en" />`?

File: packages/components/src/neptuneProvider/NeptuneProvider.js [hunk 12]
```diff
@@ -0,0 +1,44 @@
...
+
+  const intl = createIntl(
+    {
+      locale,
+      defaultLocale: locale,
+      messages,
+    },
+    cache,
+  );
+  return <RawIntlProvider value={intl}>{children}</RawIntlProvider>;
```
dozortsev, line 31: Hey

would like address some changes to the RFC which were added based on your feedback

**re. not bundling all translations in consumer application bundle**

Main concept remains the same, we use global context provider (`NeptuneProvider`) but it got changed inside.

We implemented solution which bundle only one default language (English) in the bundle of consumer app and loads necessary translations on demand, this been done via dynamic imports ([`import()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#Dynamic_Imports)) (@conorcussell thanks for [hint](https://github.com/transferwise/neptune-web/pull/633#pullrequestreview-512600566)).

Please see [updated `NeptuneProvider`](https://github.com/transferwise/neptune-web/pull/633/files#diff-fa11d57d55ae24aab6f002e1e3f49b34a62b3f74b065c49f477f457ae9199bf7R4-R21) which loads necessary translations file once `locale` gets updated. This is possible also due to code-splitting, when build tools see syntax like `import()` they start split code on smaller bundlers which can later be loaded on demand or in parallel.

Following example tries to demo this, you can see that translation bundles gets downloaded only when we change `locale` for `NeptuneProvider`:

Screen Shot 2020-10-28 at 2 26 53 PM

Learn more about it e.g in [Rollup](https://rollupjs.org/guide/en/#dynamic-import) and [Webpack](https://webpack.js.org/guides/code-splitting/#dynamic-imports) docs.

This works as for UMD as for ES bundles of `@transferwise/components` package.

This also been tested and proven to works in apps which do SSR.

**re. extra languages**

we will add support for those extra languages which Dan pointed out https://github.com/transferwise/neptune-web/pull/633#discussion_r508564238

**re. co-locating translations within the component**

see no harm rather benefits if we follow this approach, agree to add it 👍 🙂

dozortsev (commented):
thanks for feedback! @conorcussell @oliverviljamaa @danlaush @karanjthakkar @sbrandwoo

danlaush:
@sbrandwoo I guess it could support both? I like the idea of Neptune providing a LocaleProvider context instead of having to make my own. I wouldn't want to prop drill locale everywhere, so if enough people need a LocaleProvider why not make one available.

karanjthakkar (approved):
Looks like a great first step. Thanks!

packages/components/src/neptuneProvider/NeptuneProvider.js
danlaush: With co-locating, you'd do this as part of the MVP? So each component has its own messages and those messages are broken up into languages.

danlaush: With co-locating, would you see each component having its own Intl Provider? Or were you thinking that the translation file lives next to the component but all messages end up managed by one Neptune Provider? I'd much prefer the former - like splitting out the language into multiple modules it would make a big difference to bundle size.

dozortsev: I mean each component has `.message.js` file with default lang, rest of langs are located in some `lang/` folder, similar to what is showed in this https://github.com/transferwise/neptune-web/pull/633#discussion_r509366873

danlaush (approved):
👏

andreapiras (approved):

rfc/0004-i18n.md
oliverviljamaa: Why did you decide against the GitHub integration?

dozortsev: No, we didn't against it at all, in fact we would like receive PR from Crowdin after we manually push translations and once some of translations ready, so we kinda meed mix of both.

Most likely we will use integration approach for `main` branch too because I'm sure sometimes people won't want/need to wait until translations are ready and they will merge PR quicker so we need cover that case too.

sorry I didn't describe this thing  last comment because this needs a bit of investigation and testing a bit while at the same time this doesn't impact the translations implementation inside neptune-web components.

oliverviljamaa: Cool, it's just that this paragraph indicates we'll have the CI approach only. Also, why is the manual push necessary? Could we rely on just the branch naming regex with the GitHub integration?

oliverviljamaa (approved):
Nice job 👏

rfc/0004-i18n.md
dozortsev: > ..Cool, it's just that this paragraph indicates we'll have the CI approach only.

will update as implement it 🙂

> Also, why is the manual push necessary?

I guess because we should push translations into Crowdin only when we sure that they are final, you know what I mean? sometimes devs need answer from Copy team and while they sorting that out they usually put some temp message there which I guess we don't want translate... plus I'm not sure but I guess we should push translations only once, right? I mean if one commits and pushes many times in the PR after pushing translations will this somehow affect Crowdin or it will just ignore that as message already there?

> Could we rely on just the branch naming regex with the GitHub integration?

I guess we should not add some rules for people to keep contribution process as regular as possible, perhaps if only we automate this somehow?

oliverviljamaa: > I guess because we should push translations into Crowdin only when we sure that they are final, you know what I mean? sometimes devs need answer from Copy team and while they sorting that out they usually put some temp message there which I guess we don't want translate...
Yup, agreed! We could also just use plain text while the copy is not final 👍 

> I mean if one commits and pushes many times in the PR after pushing translations will this somehow affect Crowdin or it will just ignore that as message already there?
That should not be a problem — we've done pushing on every branch for quite some time, as it's one of the solutions outlined here:
https://transferwise.atlassian.net/wiki/spaces/EKB/pages/1054900981/Setting+up+an+integration+with+a+new+Crowdin+project

PR #634, opened 2020-10-13
Dynamic Flows - Optional invalid fields should disallow submission [ADS-1640]

JulianG-TW:
## 🖼 Context

In recipients page the KRW currency schema contains an optional phone number field. In the video below the user enters an invalid phone number but the form is submitted regardless.

The same issue can happen with optional email addresses in other currencies.
Screen Recording 2020-10-12 at 16 46 29

This is caused because currently, the onChange event is called with valid = true because the phone number is removed from the model.

Consider the following schema:

```json
{
  "allOf": [
    {
      "title": "Phone schema",
      "type": "object",
      "properties": {
        "name": {
          "type": "string",
          "title": "Your name (required)",
          "minLength": 2,
          "maxLength": 45,
          "default": ""
        },
        "phone": {
          "type": "string",
          "format": "phone",
          "title": "Phone number (optional)",
          "minLength": 10,
          "maxLength": 12
        }
      },
      "required": ["name"]
    }
  ]
}
```

before

### Proposed Fix

We want to avoid removing invalid values from the model, so that onChange is called with valid = false when an optional field is invalid.

after

## 🤔 Considerations

Also tested with the following schema:

```json
{
  "allOf": [
    {
      "title": "Optional Email Schema",
      "type": "object",
      "properties": {
        "name": {
          "type": "string",
          "title": "Your name (required)",
          "minLength": 2,
          "maxLength": 45,
          "default": ""
        },
        "email": {
          "title": "Recipient Email",
          "type": "string",
          "placeholder": "example@example.ex",
          "validationMessages": {
            "pattern": "The email address is invalid."
          },
          "refreshFormOnChange": false,
          "pattern": "^[^\\s]+@[^\\s]+\\.[^\\s]+$",
          "maxLength": 255
        }
      },
      "required": ["name"]
    }
  ]
}
```

optional-email

## ✅ Checklist


StevePole:
I think the intent is that the phone control should have a minLength, then a partially entered value would be invalid.

packages/dynamic-flows/src/jsonSchemaForm/JsonSchemaForm.e2e.spec.js
```diff
@@ -137,17 +137,25 @@ describe('E2E: Given a component for rendering a JSON schema form', () => {
...
+      expect(onChange).toHaveBeenLastCalledWith(
+        { const: 'abcd', string: 'x' },
+        false,
+        stringSchema,
+      );
+    });
+
+    it('should call onChange with valid = false', () => {
+      expect(onChange.mock.calls[onChange.mock.calls.length - 1][1]).toBe(false);
     });
```
JulianG-TW, line 150: new test!

felixxiatransferwise (approved):

andreapiras (approved):

PR #637, opened 2020-10-14
feat: add a min-width and the ability to right align the date lookup calendar

helendurranttw:
## 🖼 Context

When the input width is limited, the calendar will only be as wide as the input. This forces the user to scroll horizontally within the calendar, which is a poor experience.

Screenshot 2020-10-14 at 13 32 37

To fix this we can increase the minimum width of the calendar. This, however, can create further issues if the input is positioned to the very right of the screen - the calendar may open in a position where it is cut off.

Screenshot 2020-10-14 at 13 31 56

## 🚀 Changes

Added a minimum width to the dropdown, to ensure it's always visible when the input is narrower than it. Added a check on open to see if the calendar is cut off on the right. If so, I align to the right.

## 🤔 Considerations

I didn't bother adding handling on resize, as it seems like the edge case of an edge case. I can easily do this though, if it seems necessary.

I'm concerned that some people may be relying on the width of the calendar limiting itself to the width of the input (i.e. it not having a min-width). I could consider adding a prop to trigger this behaviour, instead of applying it across the board.

## ✅ Checklist


sergonius:
> This, however, can create further issues if the input is positioned to the very right of the screen - the calendar may open in a position where it is cut off.

Couldn't we just check whether it's not visible and then programmatically change the positioning?

helendurranttw:
> Couldn't we just check whether it's visible and then programmatically change the positioning if needed?

@sergonius Good point! Will give that a shot :)

packages/components/src/dateLookup/DateLookup.js
```diff
@@ -68,24 +68,45 @@ class DateLookup extends PureComponent {
...
+    if (this.open && this.dropdown && this.dropdown.current) {
+      const dropdown = this.dropdown.current;
+      const bounding = dropdown.getBoundingClientRect();
+
+      const rightSideOffscreen =
+        bounding.right > (window.innerWidth || document.documentElement.clientWidth);
+      const leftSideOffscreen = bounding.left < 0;
+
+      if (rightSideOffscreen) {
+        // If both sides are offscreen, 🤷
```
sergonius, line 97: So what actually happens if both side appear to be off screen?
Since the min-width is 320px and that's the size of the smallest screens used these days, it could perhaps be displayed full screen, like a modal or something.

andreapiras (approved):

packages/components/src/dateLookup/DateLookup.js
helendurranttw: Yeah we had the same idea, to open in a bottom sheet like the select. But we will tackle this in a separate ticket :)

PR #640, opened 2020-10-19
fix: add font-display to fonts

nabaroa:
## 🖼 Context

Ticket: https://transferwise.atlassian.net/browse/DS-685

Adding  font-display to fonts we ensure the text remains visible during webfont load, improving performance and avoiding flash of invisible text.

https://web.dev/font-display/

imagen

## 🚀 Changes

Adds the font-display property to avoid the flash of invisible text.

## ✅ Checklist


gusso:
Nice! Did you consider using `fallback` instead of `swap`? I think `fallback` gives a chance for fast connections to fetch the font and avoid the flash of unstyled content, so it caters to both slow and fast.

gusso (approved):

anttio:
Late to comment on this but how I understand `fallback` works is that it has a short `block` period but also a short `swap` period. And reading from the [specs](https://drafts.csswg.org/css-fonts-4/#font-display-desc) this means that if the font doesn't load in the short swap period, usually 3 sec, then fallback font will be used. Whereas `swap` has infinite loading period for the custom font.

> In other words, the font face is rendered with a fallback at first if it’s not loaded, but it’s swapped in as soon as it loads. However, if too much time passes, the fallback will be used for the rest of the page’s lifetime instead.

gusso:
@anttio Oh I see, you're right. So if a connection is not fast enough, on `fallback` the app has the chance of staying in the fallback font for entire duration of the session, while `swap` will swap to the correct font whenever it finally becomes available.

I guess `swap` would indeed be more appropriate, sorry for nudging you in the wrong direction @nabaroa 🤦‍♂️

nabaroa:
I'll update it :)

PR #646, opened 2020-10-26
fix: improve positioning logic to the date lookup cal

helendurranttw:
## 🖼 Context

I added some logic last week to right align the calendar if the edge is cut off on the right. I didn't account for there not being sufficient room to display it to the left once aligned.

## 🚀 Changes

I've added logic to only align right if there is sufficient room to display it right aligned: if not, it will just align itself to the leftmost edge. The calendar width is 320px, so the chance of anyone using a screen smaller than this is minimal.

## 🤔 Considerations

We will be adding a better fix for this soon; opening the calendar in a bottom drawer style similar to how we render select on mobile web. I'm adding this as a quick fix as the bug still exists in production.

## ✅ Checklist


helendurranttw:
We came up with a better way of doing it - coming soon in a new PR.

PR #647, opened 2020-10-26
fix: change tw-btn class so as not to clash with ad-blockers

helendurranttw:
## 🖼 Context

Ad blockers have taken to blocking buttons with the class `tw-btn` - a class applied to twitter sharing buttons. We need to change the class used by our buttons so we're not affected by this.

## 🚀 Changes

Changed the `tw-btn` classes to `np-btn`.

## 🤔 Considerations

Should we leave `tw-btn` on there for a few releases to be safe? Generally I’m in favour of declaring internal classes implementation detail, but I’m aware this is an extremely widely used component so I’m willing to compromise here.

## ✅ Checklist


andreapiras (approved):

helendurranttw:
As per @StevePole's suggestion we could move this CSS into Neptune. I think this is probably a good idea, but I'll do it separately in our next major Neptune CSS release, to ensure the change gets sufficient testing when we roll it out.

PR #653, opened 2020-11-04
fix!: invoke onAmountChange if value is empty in MoneyInput

dozortsev:
## 🖼 Context

`onAmountChange` prop function doesn't get triggered when input value is empty.

## 🚀 Changes

- Fix for money input.

## 🤔 Considerations

This is breaking change, because now `onAmountChange` will be invoked in this case (with `null` value passed in).

Commit message:

```
fix!: buggy behaviour in MoneyInput

BREAKING CHANGE: invoke onAmountChange if input value is empty and pass null in there
```

## ✅ Checklist


packages/components/src/moneyInput/MoneyInput.js
```diff
@@ -52,7 +53,9 @@ export default class MoneyInput extends Component {
     this.setState({
       formattedAmount: value,
     });
-    const parsed = parseAmount(value, this.props.selectedCurrency.currency, this.props.locale);
+    const parsed = isEmpty(value)
+      ? null
+      : parseAmount(value, this.props.selectedCurrency.currency, this.props.locale);
     if (!Number.isNaN(parsed)) {
```
dozortsev, line 59: note 

```js
!Number.isNaN(null)
// => true
```

mikknarusk (approved):

fergusjordantw:
Any chance this could break things if consumers aren't handling it and are always expecting a number?

packages/components/src/moneyInput/MoneyInput.js
helendurranttw: What about undefined?

dozortsev: I guess we better pass `null` value rather than `undefined` in `onAmountChange`

dozortsev:
@mikknarusk @fergusjordantw well yeah it will break things if they not handling `null` but we need fix thing wrong behavior 🙂 we plan to release it as breaking change.

fergusjordantw:
@dozortsev yep - was just making sure that we considered it a breaking change

packages/components/src/moneyInput/MoneyInput.js
dozortsev: re. `value` being `undefined`

I think it's can't be `undefined` in this case because it's coming from onchange event object so it always will be a string value

dozortsev: in some cases the value based on output of `formatAmountIfSet` ([example](https://github.com/transferwise/neptune-web/blob/c482de02912267b6471da379639cca46682ee083/packages/components/src/moneyInput/MoneyInput.js#L96)) but I guess [this function](https://github.com/transferwise/neptune-web/blob/c482de02912267b6471da379639cca46682ee083/packages/components/src/moneyInput/MoneyInput.js#L22) doesn't return `undefined`

dozortsev: here is implementation of formatAmount just incase 🙂

https://github.com/transferwise/formatting/blob/947cc15b27fc0a3381607300ecd9f42a2e073869/src/currency/currency.ts#L57-L69

helendurranttw: Ok, thanks for that!

helendurranttw (approved):

PR #654, opened 2020-11-04
chore: drawer & switch updates

helendurranttw:
## 🖼 Context

We had done some a11y refactoring on these components, but in the process we'd made some new props required. In order to prevent many breaking changes, we've decided to just add a warning for now. We will be releasing a breaking change next year where we enforce these "suggestions".

## 🚀 Changes

Un-broke the breaking changes we had previously committed to this branch, by making the newly required props not required. Added warnings instead. Also added our new np prefix to the classes.

## ✅ Checklist


packages/components/src/moneyInput/MoneyInput.js
```diff
@@ -4,6 +4,7 @@ import classNames from 'classnames';
 import Select from '../select';
 import './MoneyInput.css';
 import { Size } from '../common/propsValues/size';
+import { isEmpty } from '@transferwise/neptune-validation';
```
andreapiras, line 7: Is this supposed to be here?

helendurranttw: 🙈

packages/components/src/drawer/Drawer.js
```diff
@@ -11,6 +11,14 @@ import { Position } from '../common/propsValues/position';
 import KEY_CODES from '../common/keyCodes';
 
 const Drawer = ({ children, footerContent, headerTitle, onClose, open, position }) => {
+  if (!onClose && process.env.NODE_ENV === 'development') {
```
andreapiras, line 14: process doesn't exists on clientside. Please check how is this bundled in UMD

packages/components/src/drawer/Drawer.js
```diff
@@ -41,35 +49,35 @@ const Drawer = ({ children, footerContent, headerTitle, onClose, open, position
...
-              onKeyDown={handleOnKeyDown}
+              onKeyDown={onClose && handleOnKeyDown}
               aria-label="Close"
             >
               <CrossIcon size={24} />
             </div>
             {headerTitle && (
               <div className="align-heading m-l-2">
-                <h3 className="drawer-header--title">{headerTitle}</h3>
+                <h3 className="np-drawer-header--title">{headerTitle}</h3>
```
andreapiras, line 70: we should change this to class .h3

packages/components/src/button/Button.js
```diff
@@ -33,14 +34,10 @@ const Button = ({
...
-  /** @DEPRECATED Size.EXTRA_SMALL */
-  if (size === Size.EXTRA_SMALL && process.env.NODE_ENV === 'development') {
-    // eslint-disable-next-line
-    console.warn(
-      '@transferwise/components: The value Button.Size.EXTRA_SMALL for the prop `size` in the Button component has been deprecated, and will be removed in the next major release. Please use Button.Size.SMALL instead.',
-    );
-  }
-
+  logActionRequiredIf(
+    `Button has deprecated the \`Button.Size.EXTRA_SMALL\` value for the \`size\` prop. Please use Button.Size.SMALL instead.`,
```
andreapiras, line 38: You can use single quotes like [here](https://github.com/transferwise/neptune-web/pull/654/files#diff-3778ef06b1cc70d1c85dd8d07e2cd5e6e2da9fc4c365ab8176a045c03553ab09R16)
) to avoid backslash

andreapiras (approved):

packages/components/src/button/Button.js
helendurranttw: Ah yes, I originally had a ${} in there but removed it - thanks!

helendurranttw, line 38:
```suggestion
    'Button has deprecated the `Button.Size.EXTRA_SMALL` value for the `size` prop. Please use Button.Size.SMALL instead.',
```

PR #661, opened 2020-11-06
ADS-1664 Broadcast model on schema change

felixxiatransferwise:
## 🖼 Context

**For video and explanation see "Notes on Neptune dynamic flows" section of** https://transferwise.atlassian.net/browse/ADS-1664

Essentially we need to broadcast valid parts of the model on schema change because if we don't, the user could submit something that is out of sync with the current schema.

## 🚀 Changes

 
- ObjectSchema will now broadcast valid parts of the model on schema change.

## ✅ Checklist


packages/dynamic-flows/package.json
```diff
@@ -66,7 +66,8 @@
     "@transferwise/icons": "^2.4.0",
     "@transferwise/neptune-css": "^4.0.6",
     "classnames": "^2.2.6",
-    "core-js": "^3.6.5"
+    "core-js": "^3.6.5",
+    "lodash": "^4.17.20"
```
andreapiras, line 70: Hey @felixxiatransferwise do you need the whole lodash? Can you not install [lodash.isEqual](https://www.npmjs.com/package/lodash.isequal)

packages/dynamic-flows/package.json
```diff
@@ -57,7 +57,7 @@
   },
   "peerDependencies": {
     "prop-types": "^15.7.2",
-    "react": "16.13.1",
+    "react": "16.14.0",
```
andreapiras, line 60: Why is this update necessary at this point?

packages/dynamic-flows/src/jsonSchemaForm/objectSchema/ObjectSchema.js
```diff
@@ -31,7 +32,11 @@ const ObjectSchema = (props) => {
 
   useEffect(() => {
     // When the schema changes, only retain valid parts of the model
-    setModel(getValidModelParts(model, props.schema));
+    const newModel = getValidModelParts(model, props.schema);
+    setModel(newModel);
```
andreapiras, line 36: Should you reset the model only if not equal to model? Otherwise you are resetting an already set value 🤔

andreapiras: Also I don't think it's a good idea to set a state inside a hook. You can cause unexpected behaviors or worst loops. I don't understand why you need a model state at all. I would not get into this area of state vs props - controlled vs uncontrolled. What are you doing here is considered an anti pattern '**Anti-pattern: Erasing state when props change'** Maybe have a look at this [doc](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#when-to-use-derived-state)

andreapiras: If you are trying to limit the broadcast values maybe you should add this check in the parent that broadcast the model

felixxiatransferwise: It was setting the model before `setModel(getValidModelParts(model, props.schema));`. My change is only to broadcast the model when the schema changes.

packages/dynamic-flows/package.json
felixxiatransferwise: Wasn't playing nicely with lodash but is fine with lodash.isequal. No harm in minor version update right?

andreapiras: I'd rather keep the versions of react in sync in the monorepo if possible so if is not required I'd put back to what it was please

packages/dynamic-flows/src/jsonSchemaForm/objectSchema/ObjectSchema.js
andreapiras: I saw that but I have to also point out some code that could be improved or that can cause issue :)

andreapiras (approved):
Approved. @felixxiatransferwise to address my concerns in a separate ticket :)

PR #665, opened 2020-11-09
[WIP] Deprecation strategy update

andreapiras:
## 🖼 Context

We need to introduce a simpler deprecation strategy that allows us to move faster and create the smallest amount of tech debt as possible. Our aim is to deprecate a prop and still leave a window of time to allow consumer to upgrade the library version to adhere to the new API.

## 🚀 Changes

 
- Update RFC
- Introduced the idea of args defaultValue assignments
- Introduced the `deprecated` function

## ✅ Checklist

### DIFF
--- a/rfc/0003-deprecation-strategy-and-higher-order-component.md
+++ b/rfc/0003-deprecation-strategy-and-higher-order-component.md
@@ -52,7 +52,7 @@ const Button = ({ label }) => <button>{label}</button>
 - add the new prop `children` and default its value to old one `label`.
 
 ```js
-const Button = ({ children = label, label  }) => <button>{children}</button>
+const Button = ({ label, children = label }) => <button>{children}</button>
 
 ```
 This solution is back-compatible and if `children` is not provided it will default to `label`.
@@ -61,7 +61,7 @@ To achieve #2, we will:
  - add a customProp deprecate function that will throw a warn message for the deprecated prop, including the `expire date`
 
 ```js
-const Button = ({ children = label, label  }) => <button>{children}</button>
+const Button = ({ label, children = label }) => <button>{children}</button>
 
 Button.propTypes = {
   // eslint-disable-next-line react/require-default-props
--- a/rfc/0003-deprecation-strategy-and-higher-order-component.md
+++ b/rfc/0003-deprecation-strategy-and-higher-order-component.md
@@ -1,15 +1,11 @@
 - Start Date: 2019-11-15
-- Status: Agreed
+- Status: Progress
 
 # Summary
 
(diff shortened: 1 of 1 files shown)


rfc/0003-deprecation-strategy-and-higher-order-component.md
```diff
@@ -19,201 +15,92 @@ While there are many facets to this, we have agreed that breaking changes should
...
-
-1. Deciding which version of a component to serve
-2. Warning the consumer of API changes, when using the old API.
-
-This could be controlled by a `deprecationConfiguration` capable of generating useful warnings for the consumer.
+As part of the deprecation process we are proposing a light approach that introduces breaking changes progressively. The deprecation's steps will be:
+ 
+- Default the new prop value to old prop (valid only for prop renaming)
+- Introduction of a deprecation message with final deprecation date.
+- Remove the deprecated prop/s.
```
dozortsev, line 22: why is this only about new/old props? what if there is no new prop? what if it's about some CSS breaking changes? or component behaviour breaking changes? (e.g https://github.com/transferwise/neptune-web/pull/653 started invoking prop in new cases passing there new value in some cases (`null`))

so basically how about other use cases of introducing breaking changes?

rfc/0003-deprecation-strategy-and-higher-order-component.md
```diff
@@ -19,201 +15,92 @@ While there are many facets to this, we have agreed that breaking changes should
...
 
 ```js
-function decider(props) {
-  if (props.label) {
-    console.warn("Button's `label` prop is deprecated, please use `children` instead");
-  }
-
-  return true;
-}
+const Button = ({ children = label, label  }) => <button>{children}</button>
```
dozortsev, line 55: I guess `label` prop should be defined before `children` otherwise this won't work, right?

rfc/0003-deprecation-strategy-and-higher-order-component.md
dozortsev: IMHO I would better invest nice changelog (nice example from [Reakit](https://github.com/reakit/reakit/blob/master/CHANGELOG.md#breaking-changes)) and in education of our consumers to look into it and be mindful when upgrading to latest major version

andreapiras: If there's no new prop the mechanism will be the same except for the fact that there won't be a default assignment parameter value. The whole idea here is to give people enough time to update their codes without have to immediately follow up. Css breaking changes were not in scope in the original RFC neither but we can think about a strategy to cover them. A nice changelog would be nice but I think it could be an enhancement to the suggested strategy rather than a replacement

PR #666, opened 2020-11-10
fix: make moneyInput button width consistent

mikknarusk:
## 🖼 Context
https://transferwise.atlassian.net/browse/SEND-1671

Example in product:
54e4e7df-18dd-4eec-97de-016635e32db2

Extreme example in dev:
Screenshot 2020-11-10 at 10 42 12
After (desktop):
Screenshot 2020-11-10 at 10 49 02
After (mobile):
Screenshot 2020-11-10 at 10 49 06

## 🚀 Changes
Adds a consistent `min-width` to the `MoneyInput` based on some tests.
 

## 🤔 Considerations
This is obviously not future proof, I'd love to hear some other ideas.

## ✅ Checklist


packages/components/src/moneyInput/MoneyInput.story.js
```diff
@@ -9,7 +9,7 @@ export default {
 };
 
 export const basic = () => {
-  const size = select('size', Object.values(MoneyInput.Size), MoneyInput.Size.MEDIUM);
+  const size = select('size', Object.values(MoneyInput.Size), MoneyInput.Size.LARGE);
```
mikknarusk, line 12: Changed the default size here as that's what we use in the product, I haven't seen any other sizes in use anywhere.

packages/components/src/moneyInput/MoneyInput.less
```diff
@@ -32,6 +32,15 @@
 .tw-money-input .amount-currency-select-btn > .tw-select > .btn {
   display: flex;
   align-items: center;
+  justify-content: space-between;
+
```
oliverviljamaa, line 36:
```suggestion
  justify-content: space-between;
```

mikknarusk:
If I use `rem` instead, it works out of the box for all sizes, but it will looks increasingly bad in smaller sizes. But compared to the starting point I'd say it's fine:

LG:
Screenshot 2020-11-10 at 11 08 33

MD:
Screenshot 2020-11-10 at 11 08 36

SM:
Screenshot 2020-11-10 at 11 08 43

SM without my changes for reference:
Screenshot 2020-11-10 at 11 08 54

oliverviljamaa (approved):

packages/components/src/moneyInput/MoneyInput.story.js
danlaush: It's the storybook entry so happy for this to change, but if you're curious I'm using Medium ([code](https://github.com/transferwise/lienzo-frontend/blob/master/components/PaypalCalculator/PaypalCalculator.js#L187), [url](https://transferwise.com/gb/paypal-fee-calculator/)). Also now realise I should be using an enum for the size 😅

fergusjordantw (approved):

packages/components/src/moneyInput/MoneyInput.story.js
mikknarusk: Cool, happy to undo

danlaush:
@mikknarusk did you consider the `ch` unit? https://meyerweb.com/eric/thoughts/2018/06/28/what-is-the-css-ch-unit/

mikknarusk:
> @mikknarusk did you consider the `ch` unit? https://meyerweb.com/eric/thoughts/2018/06/28/what-is-the-css-ch-unit/
> I would guess that setting it to `3ch` wide would make it always big enough to fit `000` for the given font, so it should always have enough space for any 3-letter currency.

I didn't, interesting read. 
Looking at my change, I need to target the entire button, not just the three letters. Not because I want to, but because there isn't anything to target for just the text. The fact that the text itself is not a separate element + the fact that the span doesn't have a class + the fact that there's a second span makes it a bit inconvenient and flaky to target

packages/components/src/moneyInput/MoneyInput.less
```diff
@@ -32,6 +32,13 @@
 .tw-money-input .amount-currency-select-btn > .tw-select > .btn {
   display: flex;
   align-items: center;
+
+  span:not(.tw-icon) {
+    min-width: 2.6em;
+    @media (--screen-sm) {
+      min-width: calc(~"2.6em +" + (var(--size-8) * 4));
```
helendurranttw, line 39: Unfortunately this won't work in IE11 - the custom props polyfill has trouble working with custom props inside calc statements. You'll need to remove the custom prop and swap it out for a normal value :)

danlaush (approved):

packages/components/src/moneyInput/MoneyInput.less
mikknarusk: it seems to work fine locally, is it different in the prod build? 
Screenshot 2020-11-10 at 12 17 56

andreapiras: Why not using `--size-32` if it works anyway?

mikknarusk: @andreapiras I don't know, I like to keep things complicated 😂

helendurranttw: Oh! I'm happily mistaken. Though the build seems to be failing - did you notice? Maybe it's a different issue.

packages/components/src/moneyInput/MoneyInput.less
```diff
@@ -32,6 +32,13 @@
 .tw-money-input .amount-currency-select-btn > .tw-select > .btn {
   display: flex;
   align-items: center;
+
+  span:not(.tw-icon) {
+    min-width: 2.6em; // fits a 3 letter string, ZMW being the longest one
+    @media (--screen-sm) {
+      min-width: calc(~"2.6em" + var(--size-32));
```
andreapiras, line 39: 😌

packages/components/src/moneyInput/MoneyInput.less
mikknarusk: @helendurranttw it was this one https://github.com/transferwise/neptune-web/pull/666/commits/7e30eb9367fb8999250fce2004d8e9c13d447fbb

helendurranttw: Ah cool, thanks for that. And thanks for the contribution! 😁

helendurranttw (approved):

PR #667, opened 2020-11-10
feat: Add support for extra small chevron

twrdjpalmer:
## 🖼 Context

[Per thread](https://transferwise.slack.com/archives/C010NMC34AE/p1601563072045000), this introduces an XS chevron icon for usage in the [public navigation](https://github.com/transferwise/public-navigation) and other marketing components.

## 🚀 Changes

Returns a bespoke chevron SVG when providing `size={Size.EXTRA_SMALL}`.

## 🤔 Considerations

N/A

## ✅ Checklist


packages/components/src/chevron/Chevron.js
```diff
@@ -6,21 +6,35 @@ import { ChevronUp } from '@transferwise/icons';
...
   [Size.MEDIUM]: 24,
 };
 
+const ExtraSmallChevron = (props) => (
+  <svg {...props} width="10" height="10" viewBox="0 0 10 10" fill="currentColor">
+    <path
+      fillRule="evenodd"
+      d="M4.9995 2.0865L0.0635004 6.7645L1.2745 7.9125L4.9995 4.3835L8.7245 7.9135L9.9355 6.7645L4.9995 2.0865Z"
+    />
+  </svg>
```
dozortsev, line 20: DOM for this custom icon should be same as for other icons, `svg` wrapped in `span` etc.

https://github.com/transferwise/icons/blob/main/src/build-scripts/utils/react-components.ts#L30-L35

dozortsev:
```jsx
<span className={classname('tw-icon', 'tw-icon-up', className)} 
```

packages/components/src/chevron/Chevron.js
```diff
@@ -6,21 +6,50 @@ import { ChevronUp } from '@transferwise/icons';
...
+    </svg>
+  </span>
+);
+
+ExtraSmallChevron.propTypes = {
+  className: Types.string,
+};
+
+ExtraSmallChevron.defaultProps = {
+  className: '',
```
dozortsev, line 35: we use `null` as default value in all other component, please change it

dozortsev (approved):

packages/components/src/chevron/Chevron.js
dozortsev:
```suggestion
  className: null,
```

twrdjpalmer: Out of curiosity, why would you use `null` for a value which you've actively typed as `string`? In a strongly typed env, your default value would fail type checks.

Especially, strange given you then have to use the or operator to product a string in the way it's been used

dozortsev: this more comes from general idea of avoiding give props empty string values because you might end up in case where attr is in DOM but it's empty (`<div id >...</div>`) which is not nice, in this case `className` won't be empty attribute but in some other it could be that's why we prefer `null` or `undefined`.

twrdjpalmer: `undefined` would make more sense, as the prop is not required (per typing). Perhaps it would be better to turn off the eslint rule which requires you to defined a default props for optional props, given `undefined` is a valid value

packages/components/src/chevron/Chevron.js
```diff
@@ -6,21 +6,50 @@ import { ChevronUp } from '@transferwise/icons';
...
 
 const CONTAINER_SIZE = {
+  [Size.EXTRA_SMALL]: 10,
   [Size.SMALL]: 16,
   [Size.MEDIUM]: 24,
 };
 
+const ExtraSmallChevron = ({ className, ...restProps }) => (
+  <span
+    {...restProps}
```
andreapiras, line 16: Hey what would be the use case for the rest prop?

twrdjpalmer, line 16:
```suggestion
```

packages/components/src/chevron/Chevron.js
```diff
@@ -6,21 +6,50 @@ import { ChevronUp } from '@transferwise/icons';
 import { Size, Position } from '../common';
 
 const CONTAINER_SIZE = {
+  [Size.EXTRA_SMALL]: 10,
   [Size.SMALL]: 16,
   [Size.MEDIUM]: 24,
 };
 
+const ExtraSmallChevron = ({ className, ...restProps }) => (
```
twrdjpalmer, line 14:
```suggestion
const ExtraSmallChevron = ({ className }) => (
```

packages/components/src/chevron/Chevron.js
```diff
@@ -6,21 +6,45 @@ import { ChevronUp } from '@transferwise/icons';
 import { Size, Position } from '../common';
 
 const CONTAINER_SIZE = {
+  [Size.EXTRA_SMALL]: 10,
   [Size.SMALL]: 16,
   [Size.MEDIUM]: 24,
 };
 
+const ExtraSmallChevron = ({ className }) => (
+  <span className={`tw-icon tw-icon-chevron ${className || ''}`} role="presentation" aria-hidden>
```
KarlEST, line 15: Let's use `classNames()` here. It's already imported as well

KarlEST (approved):

packages/components/src/chevron/Chevron.js
```diff
@@ -6,21 +6,49 @@ import { ChevronUp } from '@transferwise/icons';
...
+    className,
+  );
+
+  const sizeValue = CONTAINER_SIZE[size];
+
+  if (size === Size.EXTRA_SMALL) {
+    return <ExtraSmallChevron className={classNameValue} />;
+  }
+
+  return <ChevronUp className={classNameValue} size={sizeValue} />;
```
andreapiras, line 51:
```suggestion
return size === Size.EXTRA_SMALL ? <ExtraSmallChevron className={classNameValue} />: <ChevronUp className={classNameValue} size={sizeValue} />;
  }
```

andreapiras: Optional :)

packages/components/src/chevron/Chevron.js
```diff
@@ -6,21 +6,49 @@ import { ChevronUp } from '@transferwise/icons';
...
-  };
-  return <ChevronUp {...props} />;
+  const classNameValue = classNames(
+    'tw-chevron',
+    { 'chevron-color': !disabled },
+    `${[orientation]}`.toLowerCase(),
+    className,
+  );
+
+  const sizeValue = CONTAINER_SIZE[size];
```
andreapiras, line 45: This can be defined at line 50

PR #671, opened 2020-11-12
feat: adds the ability to provide an icon and use markdown inside an alert

helendurranttw:
## 🖼 Context

We're bringing alerts on web in line with the way they work on mobile.

✨ [Live demo](https://transferwise.github.io/neptune-web/branch/alert-refactor/components/alerts/Alert) ✨

Screenshot 2020-11-12 at 17 07 21

## 🚀 Changes

* Deprecated the use of `children`, replacing it with the `message` prop
* Added the ability to use bold inside alert messages using **markdown notation**
* Added an `icon` prop
* Added an `action` prop that allows the user to specify an action

## 🤔 Considerations

There are no breaking changes to the API, but users will get warnings telling them to update their use of deprecated props.

## ✅ Checklist


packages/components/src/alert/Alert.js
```diff
@@ -1,121 +1,132 @@
...
+
+const iconTypeMap = {
+  [MessageType.POSITIVE]: CheckCircle,
+  [MessageType.NEUTRAL]: InfoCircle,
+  [MessageType.WARNING]: AlertIcon,
+  [MessageType.NEGATIVE]: CrossCircle,
 };
 
-function arrowClasses(arrow) {
+const Alert = (props) => {
```
andreapiras, line 25: Why not deconstruct here?

packages/components/src/alert/Alert.js
```diff
@@ -1,121 +1,132 @@
...
-      case TOP:
-        return classes.concat('arrow-center');
-      case TOP_RIGHT:
-        return classes.concat('arrow-right');
-      case TOP_LEFT:
-      default:
-        return classes;
-    }
+    const AlertWithArrow = withArrow(Alert, arrow);
+    return <AlertWithArrow {...props} arrow={undefined} />;
```
andreapiras, line 32: Can you please explain undefined here? would it not be the same not passing anything?

packages/components/src/alert/Alert.js
```diff
@@ -1,121 +1,132 @@
...
+const deprecatedTypeMapMessage = {
+  [MessageType.SUCCESS]: 'Alert.Type.POSITIVE',
+  [MessageType.INFO]: 'Alert.Type.NEUTRAL',
+  [MessageType.ERROR]: 'Alert.Type.NEGATIVE',
+};
+
+const deprecatedTypes = Object.keys(deprecatedTypeMap);
+
+function logDeprecationNotices({ arrow, children, dismissible, size, type }) {
+  logActionRequiredIf('Alert has deprecated the `arrow` prop. Please remove it.', !!arrow);
```
andreapiras, line 69: How come we are not using `deprecated` prop instead?

packages/components/src/alert/Alert.js
```diff
@@ -1,121 +1,132 @@
...
+    ArrowPosition.TOP_RIGHT,
+    ArrowPosition.BOTTOM_LEFT,
+    ArrowPosition.BOTTOM,
+    ArrowPosition.BOTTOM_RIGHT,
   ]),
+  /** @DEPRECATED Use message instead */
+  children: requiredIf(Types.node, ({ message }) => !message),
   className: Types.string,
+  /** @DEPRECATED See onDismiss */
+  dismissible: Types.bool,
```
andreapiras, line 109: We can use deprecate here

packages/components/src/alert/Alert.js
```diff
@@ -1,121 +1,132 @@
...
+  const mappedType = deprecatedTypeMap[type] || type;
+
+  const Icon = iconTypeMap[mappedType];
+  const iconEl = icon ? cloneElement(icon, { size: 24 }) : <Icon size={24} />;
+
+  const alert = (
+    <div role="alert" className={classNames('alert d-flex', `alert-${mappedType}`, className)}>
+      {iconEl}
+      <div className="alert__message p-l-2 flex-grow-1">
+        <div>{children || <InlineMarkdown>{message}</InlineMarkdown>}</div>
```
andreapiras, line 46: Do we need this div?

packages/components/src/alert/Alert.spec.js
```diff
@@ -1,144 +1,245 @@
...
+    [Alert.Type.NEGATIVE]: renderIcon(CrossCircle),
+  };
+
+  const message = 'Your card is on its way.';
+
+  const origWarn = console.warn;
+  let mockedWarn;
+
+  beforeEach(() => {
+    mockedWarn = jest.fn();
```
andreapiras, line 37: does this need to be done BeforeEach? Would it not be better to do it BeforeAll/AfterAll?

packages/components/src/alert/Alert.js
helendurranttw: Because I want to be able to use `props` on line 32 :)

packages/components/src/alert/Alert.js
helendurranttw: It's overriding the prop. If we pass `props`, it will still have `arrow`, and we'll get into an infinite loop :)

packages/components/src/alert/Alert.js
helendurranttw: Yeah we do, because children may be inline (and so is `InlineMarkdown`).

packages/components/src/alert/Alert.js
helendurranttw: I will update!

packages/components/src/alert/Alert.spec.js
```diff
@@ -1,144 +1,245 @@
...
-      expect(component.hasClass('alert-detach')).toBe(true);
-      expect(component.hasClass('alert-info')).toBe(true);
+    it('the message is rendered', () => {
+      getByText(message);
     });
 
-    it('checks that by default its not dismissible', () => {
-      expect(component.find('button')).toHaveLength(0);
+    it('will be of type neutral', () => {
+      expect(component.className).toContain('alert-neutral');
```
andreapiras, line 58: it would be nice to have these classes defined as constants to make our tests more resilient

packages/components/src/alert/Alert.spec.js
```diff
@@ -1,144 +1,245 @@
...
     });
   });
 
-  describe('check arrow positioning', () => {
-    it('checks left top arrow', () => {
-      component.setProps({ arrow: Alert.ArrowPosition.TOP_LEFT });
+  describe('action', () => {
+    const action = {
+      'aria-label': 'Learn more about fluffy kittens',
+      href: 'fluffykittens.com',
```
andreapiras, line 137: This actually exists and it's beautiful

helendurranttw: I had no idea 😂😭 shame there are no kittens!

packages/components/src/alert/Alert.story.js
```diff
@@ -8,28 +10,125 @@ export default {
...
+  );
+
+  const message = text(
+    'message',
+    'Payments sent to your bank details **today** might not arrive in time for the holidays.',
+  );
+  const [dismissed, setDismissed] = useState(false);
+
+  return dismissed ? (
+    <></>
```
andreapiras, line 29: null?

packages/components/src/alert/Alert.story.js
```diff
@@ -8,28 +10,125 @@ export default {
...
+  const icon = boolean('custom icon', false);
+  const type = select(
+    'type',
+    [Alert.Type.POSITIVE, Alert.Type.NEGATIVE, Alert.Type.NEUTRAL, Alert.Type.WARNING],
+    Alert.Type.POSITIVE,
+  );
+
+  const message = text(
+    'message',
+    'Payments sent to your bank details **today** might not arrive in time for the holidays.',
```
andreapiras, line 24: 😩

packages/components/src/alert/Alert.story.js
```diff
@@ -8,28 +10,125 @@ export default {
...
+  const type = select(
+    'type',
+    [Alert.Type.SUCCESS, Alert.Type.ERROR, Alert.Type.INFO, Alert.Type.WARNING],
+    Alert.Type.SUCCESS,
+  );
+
+  const [dismissed, setDismissed] = useState(false);
+
+  return dismissed ? (
+    <></>
```
andreapiras, line 83: null?

packages/components/src/alert/Alert.story.js
```diff
@@ -8,28 +10,125 @@ export default {
...
+  );
+  return (
+    <>
+      <h1>Bootstrap (deprecated)</h1>
+      <h2>Simple</h2>
+      <div className="alert alert-success" role="alert">
+        <p>This is a paragraph</p>
+        <p>This is a paragraph</p>
+      </div>
+      <div className="alert alert-info" role="alert">
```
andreapiras, line 108: Nice!, since you put here shall we remove it from neptuneCssTest/alerts.html?

packages/components/src/alert/closeButton/CloseButton.js
```diff
@@ -0,0 +1,35 @@
...
+    className={classNames('np-close-button', className)}
+    aria-label="Close"
+    onClick={onClick}
+  >
+    <CrossIcon size={size} />
+  </button>
+);
+
+CloseButton.Size = {
+  SMALL: 16,
```
andreapiras, line 20: Should we actually pass Small and Large here rather than a number just in case we want to change the number in the future?

packages/components/src/alert/Alert.story.js
helendurranttw: 😔

packages/components/src/alert/Alert.story.js
helendurranttw: I think it's ok to have both, at least while we still have some people using plain css alerts

packages/components/src/alert/closeButton/CloseButton.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
```
andreapiras, line 1: should this be in a shared folder?

packages/components/src/alert/closeButton/CloseButton.less
```diff
@@ -0,0 +1,16 @@
+.np-close-button {
+  transition: color 0.15s ease-in-out;
+  color: var(--color-control-accent);
+  border: none;
```
andreapiras, line 4: we can use `btn-unstyled`and remove close

packages/components/src/alert/closeButton/CloseButton.js
helendurranttw: Let's leave it private to alert until you’ve untangled the existing closeButton as part of the flow nav refactoring - then we can merge them.

packages/components/src/alert/Alert.js
```diff
@@ -1,121 +1,142 @@
...
+        {action && (
+          <a href={action.href} className="m-t-1 d-inline-block" aria-label={action['aria-label']}>
+            {action.text}
+          </a>
+        )}
+      </div>
+      {onDismiss && <CloseButton onClick={onDismiss} size={16} className="m-l-2 close" />}
```
andreapiras, line 53: `close` contains a bunch of css that seems related to the old closebutton where the x was font
image
I'd unse `btn-unstyled` instead embedded into CloseButton

packages/components/src/alert/withArrow/withArrow.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import Types from 'prop-types';
+import classNames from 'classnames';
+
+const ArrowPosition = {
+  TOP_LEFT: 'up-left',
```
andreapiras, line 6: 🤷

packages/components/src/inlineAlert/InlineAlert.js
```diff
@@ -0,0 +1,47 @@
...
+    );
+  }, ArrowPosition.TOP_LEFT);
+
+  return <AlertWithArrow {...props} />;
+};
+
+InlineAlert.ArrowPosition = ArrowPosition;
+
+InlineAlert.Type = {
+  SUCCESS: MessageType.SUCCESS,
```
andreapiras, line 30: Should we not deprecate these too in favour of positive,neutral etc?

packages/components/src/alert/closeButton/CloseButton.less
helendurranttw: Actually we need close, it has special handling for the various alert colours.

packages/components/src/inlineAlert/InlineAlert.js
helendurranttw: I did start to, but decided to leave it since this is really a part of field, and will require some changes in dynamic forms.

packages/components/src/alert/Alert.js
```diff
@@ -1,121 +1,144 @@
...
+};
+
+const expiryDate = new Date('03-01-2021');
+
+const Alert = (props) => {
+  const { arrow, action, children, className, icon, onDismiss, message, type } = props;
+
+  if (arrow) {
+    const AlertWithArrow = withArrow(Alert, arrow);
+    return <AlertWithArrow {...props} />;
```
andreapiras, line 32: Should we not default to InlineAlert here?

helendurranttw: Hmm no? InlineAlert is for use inline in fields, and doesn't have an icon.

packages/components/src/markdown/Markdown.js
```diff
@@ -1,42 +1,81 @@
...
+const Markdown = ({ as: Element, children, className, allowList, blockList }) => {
   if (!children) {
     return null;
   }
 
-  const createMarkup = () => {
-    const parsed = reader.parse(children);
+  if (allowList?.length && blockList?.length) {
+    // eslint-disable-next-line
+    console.warn(
```
andreapiras, line 19: Do we want this in Production?

helendurranttw: I'll use logActionRequired :)

packages/css/src/less/alerts.less
```diff
@@ -1,15 +1,136 @@
...
+  }
 
-.alert {
   .close {
     margin-top: var(--size-4);
   }
+
+  p,
+  ul {
+    margin-bottom: 0;
```
andreapiras, line 23: m-b-0?

packages/css/src/less/alerts.less
```diff
@@ -1,15 +1,136 @@
...
+  }
+
+  p + p {
+    margin-top: 8px;
+  }
+}
+
```
andreapiras, line 30: m-t-1

packages/components/src/alert/Alert.js
```diff
@@ -1,121 +1,144 @@
...
+};
+
+const iconTypeMap = {
+  [MessageType.POSITIVE]: CheckCircle,
+  [MessageType.NEUTRAL]: InfoCircle,
+  [MessageType.WARNING]: AlertIcon,
+  [MessageType.NEGATIVE]: CrossCircle,
+};
+
+const expiryDate = new Date('03-01-2021');
```
anttio, line 25: I have seen this on other components as well. When I'm reviewing the code and see this my first reaction is that will this component be deprecated on that date? Or what happens.

helendurranttw: Good point. I'll name the variable with a more descriptive name :)

packages/css/src/less/alerts.less
```diff
@@ -1,15 +1,136 @@
-// TODO: remove bootstrap dependencies
+@import (reference) '../variables/legacy-variables.less';
+@import (reference) '../variables/neptune-tokens.less';
+@import (reference) './mixins/_alerts.less';
 
-// variables
-@import (reference) 'bootstrap/less/variables.less';
+.alert {
+  padding: var(--size-16);
```
andreapiras, line 6: Optional: p-a-2

packages/css/src/less/alerts.less
```diff
@@ -1,15 +1,136 @@
...
+
+    .arrow(@alert-arrow-width);
+  }
+
+  .radio ~ .alert,
+  .checkbox ~ .alert,
+  .alert-detach {
+    border-width: 0;
+    margin-top: 12px;
+    border-radius: 3px;
```
andreapiras, line 57: I think we have a custom prop for this. Feel free to skip if out of scope

packages/css/src/less/alerts.less
```diff
@@ -1,15 +1,136 @@
...
+  .alert {
+    position: relative; // Elevate over input
+    padding: 4px 16px 3px;
+    margin-top: -2px;
+    border-style: solid;
+    border-color: @input-border;
+    border-width: 0 1px 1px 1px;
+    border-top-left-radius: 0;
+    border-top-right-radius: 0;
+    margin-bottom: 0;
```
andreapiras, line 46: .m-b-0

packages/components/src/alert/Alert.js
andreapiras: So will we still have alerts with arrows outside forms?

packages/components/src/alert/Alert.spec.js
```diff
@@ -1,144 +1,250 @@
...
+    };
+
+    it('is rendered with aria-label and href set', () => {
+      ({ container, getByText } = render(<Alert action={action} message={message} />));
+      component = getByRole(container, 'alert');
 
-      expect(component.hasClass('arrow')).toBe(true);
-      expect(component.hasClass('arrow-left')).toBe(false);
+      const el = getByText(action.text);
+      expect(el.attributes.getNamedItem('aria-label').value).toBe(action['aria-label']);
```
andreapiras, line 151: expect(el).toHaveAttribute('aria-label', action['aria-label']);

packages/components/src/alert/Alert.spec.js
```diff
@@ -1,144 +1,250 @@
...
+      expect(el.attributes.getNamedItem('href').value).toBe(action.href);
```
andreapiras, line 152:
```suggestion
      expect(el).toHaveAttribute('aria-label', action['aria-label']);
      expect(el).toHaveAttribute('href', action.href);
```

packages/components/src/alert/Alert.spec.js
```diff
@@ -1,144 +1,250 @@
...
     });
+  });
+
+  describe('onDismiss', () => {
+    it('renders the close button if onDismiss is provided', () => {
+      const onDismiss = jest.fn();
 
-    it('checks center bottom arrow', () => {
-      component.setProps({ arrow: Alert.ArrowPosition.BOTTOM });
+      ({ container } = render(<Alert onDismiss={onDismiss} message={message} />));
```
andreapiras, line 201: import { screen } from '@testing-library/dom';

```suggestion
      render(<Alert onDismiss={onDismiss} message={message} />);

      expect(screen.getByRole('button')).toBeTruthy();
```

packages/components/src/alert/Alert.spec.js
```diff
@@ -1,144 +1,250 @@
...
+      component = getByRole(container, 'alert');
```
andreapiras, line 202:
```suggestion
      render(<Alert onDismiss={onDismiss} message={message} />);
```

packages/components/src/alert/Alert.spec.js
```diff
@@ -1,144 +1,250 @@
...
 
-      expect(component.hasClass('arrow')).toBe(true);
-      expect(component.hasClass('arrow-center')).toBe(true);
-      expect(component.hasClass('arrow-bottom')).toBe(true);
+      expect(container.querySelector('button')).toBeTruthy();
```
andreapiras, line 204:
```suggestion
      expect(screen.getByRole('button')).toBeTruthy();
```

packages/components/src/alert/Alert.spec.js
```diff
@@ -1,144 +1,250 @@
+/* eslint-disable no-console */
 import React from 'react';
-import { shallow } from 'enzyme';
-
+import '@testing-library/jest-dom';
+import userEvent from '@testing-library/user-event';
+import { render, getByRole, cleanup } from '@testing-library/react';
```
andreapiras, line 5:
```suggestion
import { render, getByRole, cleanup } from '@testing-library/react';
import { screen } from '@testing-library/dom';
```

packages/components/src/alert/Alert.js
helendurranttw: Not in Neptune, but it's possible someone outside is using them

mikknarusk:
Cool!
Couple of questions:
> Deprecated the use of children, replacing it with the message prop
Why is that? In most cases I'd love to still keep using `children` to be able to use jsx for the content. Seems weird to switch to markdown for this one component. Alerts are used widely in the product (granted, more than they should be), so it will probably be a bumpy ride to upgrade for larger repos.

Should the icon, text & close button be aligned somewhat? At least the gap between the icon and text top alignment stands out to me
Screenshot 2020-11-23 at 12 54 07

Just to be double check, this will automatically add an icon to every `Alert` with no opt out mechanism? I think that's fine, but seeing as the prop is optional, it's maybe a bit unexpected.
Btw is there a use case to support overriding the default icon? I think we should encourage consistent icons here across the product

packages/components/src/alert/Alert.js
```diff
@@ -1,121 +1,142 @@
...
-  onDismiss: requiredIf(Types.func, ({ dismissible }) => dismissible),
-  arrow: Types.oneOf([
-    Alert.ArrowPosition.TOP_LEFT,
-    Alert.ArrowPosition.TOP,
-    Alert.ArrowPosition.TOP_RIGHT,
-    Alert.ArrowPosition.BOTTOM_LEFT,
-    Alert.ArrowPosition.BOTTOM,
-    Alert.ArrowPosition.BOTTOM_RIGHT,
-  ]),
+  /** An optional call to action to sit under the main body of the alert. If your label is short, use aria-label to provide more context */
```
oliverviljamaa, line 80: What's the use case for providing extra information for assistive technologies only? Is the reason that with assistive technologies, the message that normally adds the context is more disconnected from the label?

packages/components/src/alert/Alert.js
```diff
@@ -1,121 +1,142 @@
...
   className: Types.string,
+  /** An optional icon. If not provided, we will default the icon to something appropriate for the type */
+  icon: Types.element,
+  /** The main body of the alert. Accepts plain text and bold words specified with **double stars** */
+  message: requiredIf(Types.node, ({ children }) => !children),
+  /** The presence of the onDismiss handler will trigger the visibility of the close button */
+  onDismiss: Types.func,
+  /** The type dictates which icon and colour will be used */
+  type: Types.oneOf(Object.values(Alert.Type)),
+
```
oliverviljamaa, line 95: Absolutely beautiful interface 👏

packages/components/src/common/markdownNodeTypes.js
```diff
@@ -0,0 +1,18 @@
+export const MARKDOWN_NODE_TYPES = {
```
oliverviljamaa, line 1: Could we use the enum style here as well?
```suggestion
export const MarkdownNodeType = {
```

packages/components/src/alert/inlineMarkdown/InlineMarkdown.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import Types from 'prop-types';
+import Markdown from '../../markdown';
+import { MARKDOWN_NODE_TYPES } from '../../common/markdownNodeTypes';
+
+const allowList = [MARKDOWN_NODE_TYPES.EMPH, MARKDOWN_NODE_TYPES.STRONG];
```
oliverviljamaa, line 6: Does this mean alerts can have *emphasis* too, albeit undocumented?

packages/components/src/markdown/Markdown.js
```diff
@@ -1,42 +1,81 @@
...
- *
- * @param {object} [as] - html wrapper for the parsed markdown.
- * @param {string} [children] - the string markdown to be converted.
- *
- * @usage `<Markdown>'# Some valid markdown'</Markdown>`
- * */
+const NODE_TYPE_LIST = Object.values(MARKDOWN_NODE_TYPES);
 
-const Markdown = ({ as: Element, children, className }) => {
+const Markdown = ({ as: Element, children, className, allowList, blockList }) => {
```
oliverviljamaa, line 13: Love the list names ❤️

packages/components/src/markdown/Markdown.js
```diff
@@ -1,42 +1,81 @@
...
+  if (!parsed) {
+    return parsed;
+  }
+
+  const walker = parsed.walker();
+  let event = walker.next();
+
+  while (event) {
+    const { node } = event;
+    if (blockList.indexOf(node.type) >= 0) {
```
oliverviljamaa, line 53: We're using `Object.values`, so could we also use `Array#includes`?

packages/components/src/markdown/Markdown.js
oliverviljamaa: Does this component power dynamic forms Markdown rendering, too?

packages/components/src/markdown/Markdown.spec.js
```diff
@@ -1,52 +1,104 @@
...
+      expect(container).not.toContainHTML(output.sublist);
+      expect(container).toContainHTML('sublistsublist');
     });
   });
 
-  it('strips unsafe values', () => {
-    const testValue = `<script type="text/javascript"></script>`;
-    const { container } = render(<Markdown>{testValue}</Markdown>);
-    expect(container).toContainHTML('<!-- raw HTML omitted -->');
+  describe('blockList', () => {
```
oliverviljamaa, line 77: I think these tests' readability suffers a bit from being overly DRY.

oliverviljamaa: I found myself scrolling back up to `input` and `output`.

packages/components/src/utilities/deprecatedProp/deprecatedProp.js
```diff
@@ -1,19 +1,36 @@
...
+import { logActionRequired } from '..';
+
+const deprecatedMessage = ({ component, propName, message, expiryDate }) => {
+  const messages = [`${component} has deprecated the use of ${propName}.`];
+
+  if (message) {
+    messages.push(message);
+  }
+
+  if (expiryDate) {
```
oliverviljamaa, line 10: 👏

oliverviljamaa (commented):
It looks beautiful 😍

packages/components/src/alert/closeButton/CloseButton.js
```diff
@@ -0,0 +1,35 @@
...
+import Types from 'prop-types';
+import classNames from 'classnames';
+import { Cross as CrossIcon } from '@transferwise/icons';
+
+import './CloseButton.css';
+
+export const CloseButton = ({ onClick, className, size }) => (
```
dozortsev, line 8: why not use [common `CloseButton`](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/common/CloseButton/CloseButton.js) it meant to be eventually used for such cases like in Alert or Modal or Drawer etc.

packages/components/src/alert/Alert.js
```diff
@@ -1,121 +1,142 @@
...
 };
 
 Alert.defaultProps = {
-  type: Alert.Type.INFO,
-  size: Alert.Size.LARGE,
-  dismissible: false,
-  arrow: null,
+  action: undefined,
+  arrow: undefined,
   className: '',
```
dozortsev, line 135: would set `undefined` here

packages/components/src/alert/Alert.spec.js
```diff
@@ -1,144 +1,245 @@
...
+    const action = {
+      'aria-label': 'Learn more about fluffy kittens',
+      href: 'fluffykittens.com',
+      text: 'Learn more',
+    };
 
-    it('checks right top arrow', () => {
-      component.setProps({ arrow: Alert.ArrowPosition.TOP_RIGHT });
+    it('is rendered with aria-label and href set', () => {
+      render(<Alert action={action} message={message} />);
```
dozortsev, line 145: minor thing: you could use here explicit spread 

```jsx
<Alert {...{action, message}} />
```

packages/components/src/alert/inlineMarkdown/InlineMarkdown.js
```diff
@@ -0,0 +1,21 @@
...
+
+const InlineMarkdown = (props) => {
+  return <Markdown {...props} as="span" allowList={allowList} blockList={undefined} />;
```
dozortsev, line 9: wait, aren't `children` should be passed only as following?

```jsx
<Component>{children}</Component>
```

can it be passed like this? 

```jsx
<Component children={children} .. />
```

packages/components/src/alert/withArrow/withArrow.js
dozortsev: we already have such positions defined https://github.com/transferwise/neptune-web/blob/2d917e839baf17f9e14a910aece6436db59916ba/packages/components/src/common/propsValues/position.js#L1-L10, could you please re use them?

packages/components/src/alert/withArrow/withArrow.js
```diff
@@ -0,0 +1,59 @@
...
+  TOP: 'up-center',
+  TOP_RIGHT: 'up-right',
+  BOTTOM_LEFT: 'down-left',
+  BOTTOM: 'down-center',
+  BOTTOM_RIGHT: 'down-right',
+};
+
+function withArrow(Alert, arrow) {
```
dozortsev, line 14: why this needs to be wrapped in function?

packages/components/src/common/propsValues/messageType.js
```diff
@@ -1,6 +1,11 @@
 export const MessageType = {
-  INFO: 'info',
+  NEGATIVE: 'negative',
+  NEUTRAL: 'neutral',
+  POSITIVE: 'positive',
```
dozortsev, line 4: I believe these types could be used not only for messages, out Button could be negative (red) and positive etc., so I would name this const and the file

'contextual types' maybe? name inspired by BT naming https://bootstrap.transferwise.com/css/#tables-contextual-classes

packages/components/src/inlineAlert/InlineAlert.js
```diff
@@ -0,0 +1,47 @@
...
+
+const InlineAlert = (props) => {
+  const AlertWithArrow = withArrow(({ children, className, type }) => {
+    const typeClass = `alert-${typeClassMap[type] || type}`;
+    return (
+      <div
+        role="alert"
+        className={classNames('alert alert-detach p-x-2 p-y-1', typeClass, className)}
+      >
+        {children}
```
dozortsev, line 19: why this prop is not called 'message' as in `Alert` component?

packages/components/src/inlineAlert/InlineAlert.js
```diff
@@ -0,0 +1,47 @@
...
+};
+
+InlineAlert.propTypes = {
+  children: Types.node.isRequired,
+  className: Types.string,
+  type: Types.oneOf(Object.values(InlineAlert.Type)),
+};
+
+InlineAlert.defaultProps = {
+  className: '',
```
dozortsev, line 43: `undefined`

packages/components/src/alert/withArrow/withArrow.js
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
```
dozortsev, line 1: I guess this code should sit next to inline alert rather than in alert, or I'm missing something?

packages/components/src/alert/Alert.js
```diff
@@ -1,121 +1,142 @@
...
+    },
+  ),
+  /** @DEPRECATED */
+  dismissible: deprecated(Types.bool, {
+    component: 'Alert',
+    message: 'The Alert will now be considered dismissible if an `onDismiss` hander is present.',
+    expiryDate: new Date('03-01-2021'),
+  }),
+  /** @DEPRECATED */
+  size: deprecated(Types.oneOf([Alert.Size.SMALL, Alert.Size.LARGE]), {
```
dozortsev, line 125: but according to chosen deprecation strategy we should support this feature until `expiryDate: new Date('03-01-2021')`

packages/components/src/alert/Alert.js
helendurranttw: That's exactly right Oliver - there is a WCAG rule that says the link should provide all the context required for the user to understand it, unless it is inside a descriptive block of text or next to a label. Because ours is not, it could sometimes use more context (for example if you use the link "Read more" or "Learn more").

helendurranttw:
Hey @mikknarusk :)

> > Deprecated the use of children, replacing it with the message prop
> 
> Why is that? In most cases I'd love to still keep using `children` to be able to use jsx for the content. Seems weird to switch to markdown for this one component. Alerts are used widely in the product (granted, more than they should be), so it will probably be a bumpy ride to upgrade for larger repos.

I guess that's somewhat the point - we don't want people using arbitrary content inside Alerts, and by not using children we encourage the best behaviour. Also - for now we are supporting both ways of using it, so people can continue using children for a while, and people who are ready can opt in to the new behaviour by using message.

It does seem a little strange to use markdown here, but we imagine InlineMarkdown will eventually be supported in other components soon (mobile libraries can already do this using text styles). This also supports the dynamic forms use case, and gives us a way of allowing some text markup without allowing arbitrary HTML content.

> Should the icon, text & close button be aligned somewhat? At least the gap between the icon and text top alignment stands out to me

I've checked with the designers - it may look a tiny bit off here but if you change the alignment it will look misaligned when there's only one line of text.

> Just to be double check, this will automatically add an icon to every `Alert` with no opt out mechanism? I think that's fine, but seeing as the prop is optional, it's maybe a bit unexpected.
> Btw is there a use case to support overriding the default icon? I think we should encourage consistent icons here across the product

Yes - it will always add an icon, and there is no opt out. We agree that where possible people should use the default icons, but we decided that if people felt they could give the user more context through use of a custom icon, then there was no harm in that. We also knew of at least one existing case on mobile, so wanted to eventually be able to support this on web,

packages/components/src/alert/Alert.js
oliverviljamaa: Makes sense, thanks!

packages/components/src/alert/inlineMarkdown/InlineMarkdown.js
helendurranttw: Ah good catch, I don't think this should be here. At first I was under the impression that EMPH is converted to STRONG because we don't support italics, but that hasn't been implemented. Will remove!

packages/components/src/markdown/Markdown.js
helendurranttw: Yep!

packages/components/src/markdown/Markdown.spec.js
helendurranttw: Gotcha, will tidy up. Thanks 😊

packages/components/src/alert/closeButton/CloseButton.js
helendurranttw: That's the plan! The existing CloseButton needs some refactoring - it has an embedded line, which is used by the OverlayHeader and FlowNav components. Andrea is in the middle of refactoring these, so we decided to leave this one internal for the moment, and he'll merge them when he's done. I didn't want to include that refactoring in this PR.

packages/components/src/alert/inlineMarkdown/InlineMarkdown.js
helendurranttw: It can!

packages/components/src/alert/withArrow/withArrow.js
helendurranttw: It's a higher order function. It produces a new component - an ArrowAlert :)

helendurranttw: About the position values - these aren't new - changing them now would be a breaking change.

packages/components/src/common/propsValues/messageType.js
helendurranttw: Nice idea. Will do 👍

packages/components/src/inlineAlert/InlineAlert.js
helendurranttw: InlineAlert still uses children - it doesn't support markdown yet. This component is what we use in the validation of Fields. It may not live past when we do Field validation alignment 😄

packages/components/src/alert/withArrow/withArrow.js
helendurranttw: It's still used inside Alert, so it probably makes sense both ways. Or InlineAlert could live inside Alert? 🤔

packages/components/src/alert/Alert.js
helendurranttw: The reason for handling this differently is that this is a style update, which doesn't impact the functionality of the alert. We could just remove the prop, but leaving in the warning will encourage people to update their usage.

packages/components/src/common/propsValues/contextualType.js
```diff
@@ -0,0 +1,11 @@
+export const ContextualType = {
```
oliverviljamaa, line 1: I like this better than the previous name, but it feels a bit clunky. Would `Sentiment` be more natural?

oliverviljamaa (approved):
Amazing work 👏

packages/components/src/common/propsValues/contextualType.js
helendurranttw: We love `Sentiment`! Will change, thanks for the suggestion :)

andreapiras (approved):

PR #674, opened 2020-11-13
feat!: internationalization support for @transferwise/components

dozortsev:
## Demo

Updated Setup page: https://transferwise.github.io/neptune-web/branch/i18y/about/Setup#usage
Provider component page: https://transferwise.github.io/neptune-web/branch/i18y/components/context/Provider

## 🚀 Changes

List of key changes:

- setup Crowdin integration
- setup react-intl for managing translations plus tooling around it (babel configs and message extraction)
- added i18n support for a few components
- created new component - `Provider`
- adjusted Storybook config and RTL tests because of the new changes
- updated loads of docs

See PR comments to code, they should explain most of details.

## ✅ Checklist


packages/components/lang/en.json
```diff
@@ -0,0 +1,6 @@
+{
+  "0SKBlf": "Go to page {pageNumber}",
+  "2xdm3a": "Close",
```
andreapiras, line 3: is this autogenerated?

packages/components/package.json
```diff
@@ -59,18 +61,24 @@
...
+    "npm-watch": "^0.7.0"
   },
   "peerDependencies": {
     "currency-flags": "^2.1.1",
     "prop-types": "^15.7.2",
     "react": "16.13.1",
-    "react-dom": "16.13.1"
+    "react-dom": "16.13.1",
+    "react-intl": "^5.10.0",
+    "babel-plugin-react-intl": "^8.2.16"
```
andreapiras, line 81: Do consumers need this?

packages/components/lang/en.json
```diff
@@ -0,0 +1,6 @@
...
+  "OesMJh": "Current page, page {pageNumber}",
+  "nnpz2g": "Pagination navigation"
+}
```
dozortsev, line 6: this is main source file for Crowdin, which is generated by `@formatjs/cli` out of all `*.message.js` files

see docs for more details: https://formatjs.io/docs/getting-started/message-extraction

packages/components/package.json
```diff
@@ -20,16 +20,18 @@
   "scripts": {
     "dev": "npm-run-all --parallel dev:*",
     "dev:watch-less": "gulp watchLess --dest=src",
-    "dev:storybook": "start-storybook -p 3001",
+    "dev:storybook": "npx npm-watch & start-storybook -p 3001",
     "build": "npm-run-all build:*",
     "build:clean": "rm -rf lib build",
+    "build:default-locale-source": "formatjs extract 'src/**/*.messages.js' --out-file lang/en.json --id-interpolation-pattern '[sha512:contenthash:base64:6]' --format lang/formatter.js",
```
dozortsev, line 26: generate main source file out of all `*.messages.js` files

packages/components/package.json
```diff
@@ -20,16 +20,18 @@
...
     "build:compile-css": "gulp compileLess --dest=src",
     "build:umd": "NODE_ENV=umd rollup -c",
     "build:umd-nopolyfill": "NODE_ENV=umd-nopolyfill rollup -c",
-    "build:es": "NODE_ENV=es babel src -d build/es/polyfill --ignore '**/*.spec.js','**/*.story.js','**/__mocks__'",
-    "build:es-nopolyfill": "NODE_ENV=es-nopolyfill babel src -d build/es/no-polyfill --ignore '**/*.spec.js','**/*.story.js','**/__mocks__'",
+    "build:es": "NODE_ENV=es babel src -d build/es/polyfill --ignore '**/*.spec.js','**/*.story.js','**/__mocks__','**/test-utils'",
+    "build:es-nopolyfill": "NODE_ENV=es-nopolyfill babel src -d build/es/no-polyfill --ignore '**/*.spec.js','**/*.story.js','**/__mocks__','**/test-utils'",
     "build:copy-files": "cpx 'src/**/!(db)/*.{css,json,svg}' build/es/polyfill",
     "build:copy-files2": "cpx 'src/**/!(db)/*.{css,json,svg}' build/es/no-polyfill",
+    "build:copy-lang": "cpx 'lang/*.json' build/es/lang & cpx 'lang/*.json' build/lang",
```
dozortsev, line 34: copy translation files next to ES and UMD bundles

packages/components/package.json
```diff
@@ -20,16 +20,18 @@
   "scripts": {
     "dev": "npm-run-all --parallel dev:*",
     "dev:watch-less": "gulp watchLess --dest=src",
-    "dev:storybook": "start-storybook -p 3001",
+    "dev:storybook": "npx npm-watch & start-storybook -p 3001",
```
dozortsev, line 23: it's useful to watch and re generate main `en.json` file while developing against Storybook (see `"watch"` prop lower in this file)

packages/components/src/avatar/Avatar.spec.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { render } from '@testing-library/react';
+import { render } from '../test-utils';
```
dozortsev, line 2: to avoid adding `<Provider >` component in every render in tests I had to "hack" render method of RTL, see `test-utils.js` file

this seem as official way todo this https://testing-library.com/docs/example-react-intl/#creating-a-custom-render-function

packages/components/src/common/CloseButton/CloseButton.messages.js
```diff
@@ -0,0 +1,8 @@
+import { defineMessages } from 'react-intl';
+
+export default defineMessages({
+  ariaLabel: {
+    defaultMessage: 'Close',
+    description: 'For `aria-label` of root button element of the component',
```
dozortsev, line 6: adding `description` is optional and it will be deleted by babel react-intl plugin while transpiling so it won't affect bundlesize

packages/components/src/provider/Provider.js
```diff
@@ -0,0 +1,52 @@
...
+          const newMessages = await import(`../../lang/${locale}.json`);
+          setIntl(createIntlInstance(locale, newMessages));
+        } catch (e) {
+          setIntl(createIntlInstance(locale, en));
+        }
+      })();
+    }
+  }, [locale]);
+
+  return <RawIntlProvider value={intl}>{children}</RawIntlProvider>;
```
dozortsev, line 39: for more into see https://formatjs.io/docs/react-intl/components#rawintlprovider

packages/components/src/test-utils/index.js
```diff
@@ -0,0 +1,17 @@
...
+function render(ui, { locale = 'en', ...renderOptions } = {}) {
+  // eslint-disable-next-line react/prop-types
+  function Wrapper({ children }) {
+    return <Provider {...{ locale }}>{children}</Provider>;
+  }
+  return rtlRender(ui, { wrapper: Wrapper, ...renderOptions });
+}
+
+export * from '@testing-library/react';
+export { render };
```
dozortsev, line 17: this seem as an official way todo this https://testing-library.com/docs/example-react-intl/#creating-a-custom-render-function

packages/docs/liveEditorCode/provider.code.js
```diff
@@ -0,0 +1,26 @@
...
+          { value: 'en', label: 'English (en)' },
+          { value: 'fr', label: 'French (fr)' },
+          { value: 'pl', label: 'Poland (pl)' },
+          { value: 'ja', label: 'Japan (ja)' },
+          { value: 'ru', label: 'Russian (ru)' },
+        ]}
+      />
+    </Provider>
+  );
+}
```
dozortsev, line 26: demo: https://transferwise.github.io/neptune-web/branch/i18y/components/context/Provider

rfc/0004-i18n.md
```diff
@@ -1,5 +1,5 @@
 - Start Date: 2020-10-15
-- Status: Progress
+- Status: Done
```
dozortsev, line 2: Update RFC files since some little things changed a slightly

.scaffold-templates/template.4.js
```diff
@@ -10,8 +10,9 @@ module.exports = {
   name: (name) => capitalizeFirstLetter(name),
   template: (name) => `import React from 'react';
 import '@testing-library/jest-dom';
-import { render } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
+import { render } from '../common/test';
```
dozortsev, line 14: just ignore this file 😄

packages/docs/pages/about/Setup.mdx
```diff
@@ -31,13 +31,16 @@ import '@transferwise/icons/lib/styles/main.min.css';
...
   return (
-    <Button size={Button.Size.SMALL} block={true}>
-      Hello Neptune
-    </Button>
+    // Components should be wrapped by Provider component which init and provides context data for them
+    // usually this should be done once at the top level component
+    <Provider locale={currentLocale}>
+      ...
+      <Button size={Button.Size.SMALL} block>...</Button>
+    </Provider>
```
dozortsev, line 43: worth emphasize, even if none of Neptune components which you currently use in your app/library rely on translations you still should add it in you consumer app, because:

- tomorrow you might would need start use one of such components
- sometimes internal components of those that you use may rely on translations (basically to avoid any unexpected error)
- tomorrow we might add support.. let's say for theming `<Provider locale={'en'} theme={..}>` and for fast migration we want you to be already "onboard" now 🙂

packages/components/.storybook/config.js
```diff
@@ -47,6 +48,7 @@ addDecorator(
 );
 addDecorator(CenterDecorator);
 addDecorator(StrictModeDecorator);
+addDecorator((storyFn) => <Provider locale="en">{storyFn()}</Provider>);
```
sbrandwoo, line 51: It's quite useful to be able to test components with real translations. Unfortunately I think that's hard to do with the decorator at the top level - you need to specify the decorator in the stories themselves and use a knob/control to set the locale.

packages/components/lang/de.json
```diff
@@ -0,0 +1,6 @@
+{
+  "0SKBlf": "Go to page {pageNumber}",
```
sbrandwoo, line 2: Why are you obfuscating the keys? They are very useful for translators in crowdin to understand context, and useful for developers to see where copy is being used.

packages/components/src/provider/Provider.js
```diff
@@ -0,0 +1,52 @@
+import React, { useEffect, useState } from 'react';
+import Types from 'prop-types';
+import { createIntl, createIntlCache, RawIntlProvider } from 'react-intl';
+import en from '../../lang/en.json';
+
+const defaultLocale = 'en';
+const cache = createIntlCache();
```
sbrandwoo, line 7: Do you have any documentation on this function?

Usually we can't use global caches, because they will be shared between customers - this risks leaking data between customers.

packages/components/src/provider/Provider.js
```diff
@@ -0,0 +1,52 @@
...
+  );
+}
+
+let currentLocale = defaultLocale;
+
+function Provider({ locale, children }) {
+  const [intl, setIntl] = useState(createIntlInstance(defaultLocale, en));
+
+  useEffect(() => {
+    if (currentLocale !== locale) {
```
sbrandwoo, line 26: Regarding React hydration warnings - you won't get them because the page is rendering in English, and then re-rendering in a different language afterwards. This does mean that any text visible on the screen will flash though as it'll change in front of the user.

packages/components/src/provider/Provider.js
```diff
@@ -0,0 +1,52 @@
...
+      defaultLocale,
+      messages,
+    },
+    cache,
+  );
+}
+
+let currentLocale = defaultLocale;
+
+function Provider({ locale, children }) {
```
sbrandwoo, line 22: I believe this `locale` is coming from the user - in which case you will need to validate it, and you'll need to map locales to the specific languages that Neptune supports. For example a locale of `es-ES` needs to load the `es.json` file.

packages/components/lang/hk.json
```diff
@@ -0,0 +1,6 @@
+{
```
sbrandwoo, line 1: `hk` isn't a valid locale. It doesn't matter than the file is called `hk.json`, but be aware that a valid value of `locale` will never be `hk`.

packages/components/lang/es.json
```diff
@@ -0,0 +1,6 @@
+{
```
danlaush, line 1: Having a single Provider wrapping the whole app seems `fine` when there's only these 4 strings, but I'm expecting that eventually there will be a bunch of form input validation messages. If I don't use a form element, that's dead code I'd rather not ship.

Did you consider giving each component its own messages Provider? I see the value in having one wrapping Provider so all the components I use can access a shared language value, but I think both can be used in tandem. The main Neptune Provider can just manage the locale, and each component can manage its own messages.

```jsx
// Render tree
<NeptuneProvider locale="es">
  <Modal>
    <InternalMessagesProvider component="modal"> // uses the NeptuneProvider context for language, but fetches messages just for the modal component
      <Message>modal.close</Message> // sorry I don't know how react-intl works
    </InternalMessagesProvider>
  </Modal>
</NeptuneProvider>
```

packages/components/lang/hk.json
danlaush: Yeah we'll probably want this to be `zh-HK`.

packages/components/CONTRIBUTING.md
```diff
@@ -46,6 +46,22 @@ Please use [react-testing-library](https://github.com/testing-library/react-test
...
+
+Later once you define all message(s) and agreed on copies, send them to translations vendor (Crowdin):
+ 
+1. Run `yarn build` (it will regenerate main source file (`en.json`) out of all `*.message.js` files)
+2. Push changes (or only `en.json`) to your remote branch (this will make Crowdin notice new messages)
+3. Right after that Crowdin will create a PR (where base branch will be your feature branch) with placeholders (English messages) in other translation files
+and will commit into same PR (or new PR in case you merged the first one) with translated messages as soon as translators translate them.
+
+If you merge your feature PR still not having all translations (which is okay-ish), Crowdin should do same just for `main` branch.
+
```
helendurranttw, line 64:
```suggestion
# Internationalisation

This package has support for internationalisation, and all messages inside components should be translated if they are not required to be provided by the consumer.

Define messages using the [`defineMessages`](https://formatjs.io/docs/react-intl/api/#definemessagesdefinemessage) function in a file named `[ComponentName].message.js` next to the component file. You can then import and use them in the component.

Once you have defined the messages and agreed on copy, they must be sent to the translations vendor (Crowdin). To do this:
 
1. Run `yarn build` (it will regenerate a main source file (`en.json`) out of all `*.message.js` files)
2. Push changes (or only `en.json`) to your remote branch (this will make Crowdin notice new messages)
3. Right after that Crowdin will create a PR (where the base branch is your feature branch) with placeholders (English messages) in other translation files. It will commit into same PR (or a new PR in case you have merged the first one) with translated messages, as soon as translators translate them.

If you merge your feature PR without all the translations (which is okay-ish), Crowdin should do the same, just on the `main` branch.

```

helendurranttw: Terrible diff.. it's just a few tweaks here and there. Nice docs!

packages/components/lang/de.json
helendurranttw: I agree - I guess you might be trying to save a few bytes? But as Stephen points out, they're probably worth leaving in.

packages/components/babel.config.js
```diff
@@ -80,6 +80,13 @@ module.exports = {
     '@babel/plugin-proposal-optional-chaining',
     '@babel/plugin-transform-react-jsx',
     '@babel/plugin-transform-runtime',
+    [
+      'react-intl',
+      {
+        idInterpolationPattern: '[sha512:contenthash:base64:6]',
+        removeDefaultMessage: true,
+      },
+    ],
```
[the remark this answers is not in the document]
helendurranttw: I guess you won't need idInterpolationPattern if you remove the key hashing

packages/components/package.json
helendurranttw: You could probably just give it its own task (`dev:translations`) - all the `dev:` tasks will get run on `dev`.

packages/components/package.json
```diff
@@ -62,20 +64,26 @@
...
   "peerDependencies": {
     "@transferwise/icons": "^2.5.0",
     "@transferwise/neptune-css": "^4.0.8",
     "currency-flags": "^2.1.1",
     "prop-types": "^15.7.2",
     "react": "16.13.1",
-    "react-dom": "16.13.1"
+    "react-dom": "16.13.1",
+    "react-intl": "^5.10.0",
+    "babel-plugin-react-intl": "^8.2.16"
```
helendurranttw, line 86: Is this supposed to be a peer? Seems like a dev

packages/components/src/lab/pagination/paginationLink/PaginationLink.js
```diff
@@ -16,8 +21,10 @@ const PaginationLink = ({ disabled, active, pageNumber, onClick, children }) =>
     }
   };
 
-  const ariaLabel = active ? `Go to page ${pageNumber}` : `Current page, page ${pageNumber}`;
-
+  const ariaLabel = intl.formatMessage(
```
helendurranttw, line 24: I had a peek at the deploy - it seems like the labels aren't quite right. I think we'll need more than this, too - the '...' will need its own label (maybe 'Click to access pages x - y' or something?).

Screenshot 2020-11-30 at 20 40 38

packages/components/src/provider/Provider.js
```diff
@@ -0,0 +1,52 @@
...
+
+  useEffect(() => {
+    if (currentLocale !== locale) {
+      currentLocale = locale;
+      (async function changeLanguage() {
+        try {
+          const newMessages = await import(`../../lang/${locale}.json`);
+          setIntl(createIntlInstance(locale, newMessages));
+        } catch (e) {
+          setIntl(createIntlInstance(locale, en));
```
helendurranttw, line 33: Under what circumstances would this fail? Maybe we should do more than just swallow the error?

Also - maybe we should revert back to `currentLocale` rather than English?

packages/docs/pages/about/Setup.mdx
helendurranttw: You can safely say "must" - because even if they don't need them, we will need them for our internal components :)

I think this deserves more explanation. If I haven't done this before, what locales are supported? What format do I pass them in in?

You could a further section below the code example, perhaps - with a link to the demo.

packages/docs/liveEditorCode/provider.code.js
helendurranttw: The demo isn't working :(

I think this could use a bit of an explanation as to what it is, too - for those who haven't come across it.

helendurranttw (commented):
Great work Anton - lots of moving pieces and it's coming together nicely 😄

packages/components/lang/de.json
conorcussell: Yep, ideally we would have some docs on key naming from the translators perspective to tell us this, but agree with Stephen, they'll at least want the key name for context

packages/components/lang/es.json
conorcussell: I think this seems like a good idea - you could still use `IntlProvider` in the top-level Provider - but expose a function called something like `setMessages` which updates the messages passed in to that component as props. Instead of an additional Provider per component, you could maybe just use a hook

(sorry untested-code ahead)

```TypeScript
const useMessages = (componentSpecificMessages) => {
  const intl = useIntl()
  const { setMessages }  = useContext(NeptuneContext)

  useEffect(() => {
    setMessages({ ...intl.messages, ...componentSpecificMessages })
  }, [intl.locale])
}


// Button.jsx

const Button = () => {
  useMessages(componentSpecificMessages);
}
```

conorcussell (changes requested):
The translation key names being obfuscated is the only blocker for me

packages/components/package.json
dozortsev: 👌

packages/docs/liveEditorCode/provider.code.js
dozortsev: it works, it's just all translations files have English version by default, we haven't received translations yet

Screen Shot 2020-11-30 at 2 17 26 PM

packages/docs/pages/about/Setup.mdx
dozortsev: 👌

danlaush (approved):
- Agree with Conor on the string obfuscation but he's already marked this as Changes Requested
- Not a blocker for now but I'd encourage looking into only loading messages for the components I'm using

packages/components/.storybook/config.js
dozortsev: good point, I'm adding global know for this 

Screen Shot 2020-11-30 at 3 34 28 PM

packages/components/src/provider/Provider.js
dozortsev: partially I configured react-intl like in this example: https://formatjs.io/docs/getting-started/installation/#minimal-application where it only says that it's optional but kinda important

there is seems really nothing particularly about this function in the docs 🤔

dozortsev: Looking at [it's source code](https://github.com/formatjs/formatjs/blob/cca3ae4878aaf286c12bc7400e14e9af4e0f26ad/packages/intl/src/utils.ts#L58-L67) it seems like it only necessary for preventing of creating of some internal it's objects and don't think that it's global one.

packages/components/src/provider/Provider.js
dozortsev: right, is there a way to fetch data based on prop value and only then do render?

packages/components/package.json
dozortsev: yep, correct, it's only dev dep

packages/components/src/provider/Provider.js
dozortsev: fixed by aligning names of files same was as country codes in Crowdin:

```
  'en',
  'ru',
  'zh-HK',
  'cs',
  'fr',
  'de',
  'hu',
  'id',
  'it',
  'ja',
  'pl',
  'pt-BR',
  'ro',
  'es-ES',
  'tr',
```

packages/components/lang/hk.json
dozortsev: https://github.com/transferwise/neptune-web/pull/674#discussion_r532816479

packages/components/src/provider/Provider.js
dozortsev: it could fail when some unknown values was passed,

good shout, I updated it to handle such cases with logs for errors https://github.com/transferwise/neptune-web/blob/i18y/packages/components/src/provider/Provider.js

packages/components/src/lab/pagination/paginationLink/PaginationLink.js
dozortsev: I do agree it has wrong messages aria attrs, I did review the code and seems like not quick fix, it will take quite a bit of changes to improve that which I think getting out of scope of this PR.

My changes do make thing worse so I would suggest we do that later separately. 🙂

packages/docs/pages/about/Setup.mdx
dozortsev: I update sentence 👍 for rest stuff I think consumers can find on providers page https://transferwise.github.io/neptune-web/branch/i18y/components/context/Provider#props

packages/components/lang/de.json
oliverviljamaa: The namespacing in translation keys compresses well, too!

packages/components/src/provider/Provider.js
oliverviljamaa: Shouldn't they align with the TransferWise values from the internal wiki instead?
https://transferwise.atlassian.net/wiki/spaces/EKB/pages/1054213243/Supported+languages

crowdin.yml
```diff
@@ -0,0 +1,25 @@
...
+  - source: packages/components/lang/en.json
+    translation: /packages/components/lang/%two_letters_code%.%file_extension%
+    languages_mapping:
+      two_letters_code:
+        fr: fr
+        hu: hu
+        es-ES: es-ES
+        tr: tr
+        en-GB: en
+        pt-BR: pt-BR
```
oliverviljamaa, line 15: Is it intentional that they're different from our regular, underscored (sadly) and mostly 2-letter, codes?

crowdin.yml
```diff
@@ -0,0 +1,25 @@
+api_key_env: CROWDIN_KEY
```
oliverviljamaa, line 1: Why do we need the Crowdin key? If we're using the GitHub integration, it should not be necessary.

crowdin.yml
```diff
@@ -0,0 +1,25 @@
...
+project_identifier: neptune-web-components
+commit_message: 'chore: new translations ✨'
+append_commit_message: false
```
oliverviljamaa, line 4: Do we need all of these and if yes, why? Could any of them be configured from the Crowdin GUI?

crowdin.yml
```diff
@@ -0,0 +1,25 @@
...
+files:
+  - source: packages/components/lang/en.json
```
oliverviljamaa, line 6: For CRAB, the standard directory naming is `i18n/${code}.json`, any reason to not follow suit here?

crowdin.yml
oliverviljamaa: That's all we have for the `BankDetails` component, for example:
https://github.com/transferwise/borderless-account-components/blob/master/crowdin.yaml

packages/components/CONTRIBUTING.md
```diff
@@ -46,6 +46,20 @@ Please use [react-testing-library](https://github.com/testing-library/react-test
...
+Define messages using the [`defineMessages`](https://formatjs.io/docs/react-intl/api/#definemessagesdefinemessage) function in a file named `[ComponentName].message.js` next to the component file. You can then import and use them in the component.
+
+Once you have defined the messages and agreed on copy, they must be sent to the translations vendor (Crowdin). To do this:
+ 
+1. Run `yarn build` (it will regenerate a main source file (`en.json`) out of all `*.message.js` files)
+2. Push changes (or only `en.json`) to your remote branch (this will make Crowdin notice new messages)
+3. Right after that Crowdin will create a PR (where the base branch is your feature branch) with placeholders (English messages) in other translation files. It will commit into same PR (or a new PR in case you have merged the first one) with translated messages, as soon as translators translate them.
+
+If you merge your feature PR without all the translations (which is okay-ish), Crowdin should do the same, just on the `main` branch.
```
oliverviljamaa, line 61: Love the docs.

packages/components/CONTRIBUTING.md
```diff
@@ -46,6 +46,20 @@ Please use [react-testing-library](https://github.com/testing-library/react-test
...
 
+# Internationalisation
+
+This package has support for internationalisation, and all messages inside components should be translated if they are not required to be provided by the consumer.
+
+Define messages using the [`defineMessages`](https://formatjs.io/docs/react-intl/api/#definemessagesdefinemessage) function in a file named `[ComponentName].message.js` next to the component file. You can then import and use them in the component.
+
+Once you have defined the messages and agreed on copy, they must be sent to the translations vendor (Crowdin). To do this:
+ 
+1. Run `yarn build` (it will regenerate a main source file (`en.json`) out of all `*.message.js` files)
```
oliverviljamaa, line 57: Is there any way to avoid this? E.g., make Crowdin understand the `js` files somehow (unlikely) or utilise multiple JSON paths, like @karanjthakkar did?

packages/components/lang/en.json
oliverviljamaa: To me, `source.json` or `messages.json` would seem cleaner for source files — especially considering all the translations are checked in here. We can still assign it under `en` in the translation map, of course.

packages/components/src/lab/pagination/paginationLink/PaginationLink.js
oliverviljamaa: I usually like to add contextual `TODO`s for cases like this, might be nice here?

packages/components/src/provider/Provider.js
```diff
@@ -0,0 +1,81 @@
...
+      messages,
+    },
+    cache,
+  );
+}
+
+let currentLocale = defaultLocale;
+
+function Provider({ locale, children }) {
+  const [intl, setIntl] = useState(createIntlInstance(defaultLocale, en));
```
oliverviljamaa, line 42: Why can't we create it with the `locale` from the prop?

packages/components/src/provider/Provider.spec.js
```diff
@@ -0,0 +1,35 @@
...
+import { FormattedMessage } from 'react-intl';
+import { render, waitFor, cleanup, screen } from '../test-utils';
+
+import Provider from '.';
+import closeButtonMessages from '../common/CloseButton/CloseButton.messages';
+
+describe('Provider', () => {
+  beforeEach(cleanup);
+
+  it('tests that provider does not add any elements in the DOM', () => {
```
oliverviljamaa, line 12: _Provider tests..._ sounds a bit weird, why not _Provider does not add..._?

packages/components/src/provider/Provider.spec.js
```diff
@@ -0,0 +1,35 @@
...
+
+  test.each([
+    ['pl', 'Blisko'],
+    ['fr', 'Proche'],
+    ['en', 'Close'],
+  ])('tests switching locale (%s)', async (locale, expectedMessage) => {
+    expect(document.body).toBeEmptyDOMElement();
+
+    render(
+      <Provider {...{ locale }}>
```
oliverviljamaa, line 26: Can't we?
```suggestion
      <Provider locale={locale}>
```

packages/components/src/provider/Provider.spec.js
```diff
@@ -0,0 +1,35 @@
...
+describe('Provider', () => {
+  beforeEach(cleanup);
+
+  it('tests that provider does not add any elements in the DOM', () => {
+    const { container } = render(<Provider />);
+
+    expect(container).toBeEmptyDOMElement();
+  });
+
+  test.each([
```
oliverviljamaa, line 18: Should we test a missing locale, too?

packages/components/src/test-utils/index.js
oliverviljamaa: Nice, perhaps I'd even link this in a comment in the code.

packages/docs/liveEditorCode/provider.code.js
oliverviljamaa: Neat!

oliverviljamaa (commented):
Amazing work, @dozortsev 🚀 🌐 💪 I left some feedback, happy to discuss!

packages/components/src/provider/Provider.js
danlaush: I'm on board with migrating to IETF `zh-HK` over `zh_HK`, but yeah everywhere else just uses `pt` and `es`.

crowdin.yml
danlaush: Maybe you have a tip for this @oliverviljamaa but I've noticed that if you don't specify the two-letter code, crowdin downloads the file as a 5-char string. Here's public-nav's dist folder, you can see the languages that don't have a 2-letter mapping 

image
That's even when crowdin's UI shows it as a 2-char string, e.g. https://crowdin.com/project/public-navigation/bg
Screenshot 2020-12-01 at 09 44 16

packages/components/src/provider/Provider.js
sbrandwoo: The param is `locale` and this is a public package, so any values here should be IETF.

sbrandwoo:
I think I'd prefer this PR to go live without the dynamic loading of language files - I'm not convinced it's the right solution.

I'd prefer if we could load translation files on the server-side somehow, and pass these translations into the page, rather than relying on the page loading the translations for themselves later.

This would overall be better for performance in terms of download time, and would prevent the page from being rendered twice, with a lag between the two renders.

packages/components/src/provider/Provider.js
oliverviljamaa: Happy with either, as long as it doesn't add extra mapping burden to all consumers.

twrdjpalmer:
Have to agree with @sbrandwoo here. Lazily loading the correct language seems far less than ideal. I've been thinking about this problem recently, after discussions with @danlaush about how Crab Lienzo is going to handle it.

Rather than put all the load on neptune itself, we could provide the tools for the consumer to enable server-side translations.

This could be a middleware, which reads the translations from `node_modules` and includes them in crab's `res.locals` for the consumer to provide to the `NeptuneProvider` component. Much in the same way the current translation's middleware works.

This is quite common amongst other libs, such as `emotion` which requires you to call something in `getInitialProps` to avoid memory leaks.

Of course, this won't work for all consumers. They can use said middleware as a foundation to roll their own solution, which could be contributed back to neptune as necessary.

danlaush:
I definitely pushed Anton toward this solution, which I'm sure was influenced by the fact that I came up with it working on a client-side-only app [rate-alerts](https://transferwise.com/tools/exchange-rate-alerts/). For that use case it makes for a really smooth language change but I wonder if I would have come at the problem differently if I'd worked on a Crab app first.

packages/components/lang/formatter.js
```diff
@@ -0,0 +1,7 @@
+exports.format = (messages) => {
+  const results = {};
+  Object.entries(messages).forEach(([id, msg]) => {
```
andreapiras, line 3: This is [not supported on IE](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries). So it will probably results in polyfills to be added to the final bundle, better to use Object.keys

packages/components/src/chevron/Chevron.spec.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import '@testing-library/jest-dom';
-import { render } from '@testing-library/react';
+import { render } from '../test-utils';
```
andreapiras, line 3: Do we need to add it even in files where we don't use translations?

packages/components/src/lab/pagination/paginationLink/PaginationLink.js
```diff
@@ -16,8 +21,10 @@ const PaginationLink = ({ disabled, active, pageNumber, onClick, children }) =>
...
+    active ? messages.ariaLabelActive : messages.ariaLabelPassive,
```
andreapiras, line 25: maybe you meant `ariaLabelInactive` ? 😅

packages/components/src/provider/Provider.js
```diff
@@ -0,0 +1,81 @@
...
+      defaultLocale,
+      messages,
+    },
+    cache,
+  );
+}
+
+let currentLocale = defaultLocale;
+
+function Provider({ locale, children }) {
```
andreapiras, line 41: I think i would give it a more recognizable name `NeptuneProvider`

packages/components/src/provider/Provider.js
```diff
@@ -0,0 +1,81 @@
...
+        // eslint-disable-next-line no-console
+        console.error(`Could not load translations for ${locale}`, e);
+        messages = await import(`../../lang/${currentLocale}.json`);
+      }
+      setIntl(createIntlInstance(locale, messages));
+      currentLocale = locale;
+    };
+
+    if (currentLocale !== locale) {
+      if (supportedLocale.includes(locale)) {
```
andreapiras, line 59: includes is not supported on Ie11

helendurranttw: We have a polyfill though :)

dozortsev: yep, babel provides polyfills for JS native API

andreapiras: Yes I know, but that means polyfills will get added when really we can avoid by using indexOf

packages/components/src/chevron/Chevron.spec.js
dozortsev: I would say yes, https://github.com/transferwise/neptune-web/pull/674#discussion_r530583387

crowdin.yml
oliverviljamaa: Public navigation has `%locale%`, not `%two_letter_code%` — if you have the latter, it'll have 2-letter code for all the locales, and the only mapping we'll have to do is `zh` to `zh_HK`.

packages/components/src/provider/Provider.js
dozortsev: I know, Provider is not quite regular component e.g like Button one but for me NeptuneProvider feels same like we would name Button component a NeptuneButton, I pick generic name and didn’t name it e.g TranslationProvider because in future we might extend it’s functionality to provide context for other stuff.

packages/components/lang/en.json
dozortsev: I see but as was suggested https://github.com/transferwise/neptune-web/pull/674#issuecomment-736385958 with new interface consumers will need load file for by themselves, and I think it would be convenient if they would be able todo that just via locale value without any extra mapping (locale to file name) when they would do dynamic import

```js
const file = import(`@transferwise/components/build/lang/${locale}.json`)
```

packages/components/src/test-utils/index.js
dozortsev: 👌

packages/components/src/provider/Provider.spec.js
dozortsev: test will fail because of warning of unrecognized value passed into prop ([test setup](https://github.com/transferwise/neptune-web/blob/a9da4cd0bba2535bc7bb8222ddb1cde9a27e6cba/packages/test-config/config/setupTests.js#L12-L13)).

packages/components/src/provider/Provider.spec.js
dozortsev: fixed

crowdin.yml
dozortsev: renamed to `i18n`

packages/components/src/provider/Provider.js
```diff
@@ -0,0 +1,51 @@
...
+  'zh_HK',
+];
+
+function Provider({ i18n, children }) {
+  const cache = createIntlCache();
+  const { locale, messages } = i18n;
+  let intlConfig;
+  if (Locales.includes(locale)) {
+    intlConfig = {
+      locale: locale.replace('_', '-'),
```
dozortsev, line 29: `createIntl` does support code with underscore

packages/components/i18n/en.json
```diff
@@ -0,0 +1,6 @@
+{
+  "np.CloseButton.aria-label": "Close",
```
conorcussell, line 2: obviously this is about as minor as it gets, but, whats with the mix of kebab and camel case in these key names?

conorcussell (approved):

packages/docs/pages/_app.js
```diff
@@ -21,6 +21,12 @@ if (process.env.NODE_ENV !== 'production' && typeof window !== 'undefined') {
 }
 
 class MyApp extends App {
+  static async getInitialProps() {
+    const defaultLocale = 'en';
+    const messages = await import(`@transferwise/components/i18n/${defaultLocale}`);
+    return { locale: defaultLocale, messages };
+  }
```
twrdjpalmer, line 28: Can you provide documentation for the recommended way of setting up the use of neptune's translations in the consumer?

packages/components/i18n/en.json
dozortsev: 👌

packages/docs/pages/_app.js
dozortsev: good point, updated notes in [Setup.mdx](https://github.com/transferwise/neptune-web/pull/674/files#diff-64a81e19012827758c14ce4ed4f10be52fa187b4471a7c4955a599caa57055abR51-R65) ([rendered](https://transferwise.github.io/neptune-web/branch/i18y/about/Setup))

packages/components/src/common/CloseButton/CloseButton.messages.js
```diff
@@ -0,0 +1,9 @@
+import { defineMessages } from 'react-intl';
+
+export default defineMessages({
+  ariaLabel: {
```
helendurranttw, line 4: Why do the generated translation files have `np.CloseButton.aria-label` instead of `np.CloseButton.ariaLabel`?

packages/docs/pages/components/context/Provider.mdx
```diff
@@ -0,0 +1,16 @@
...
+import { Provider, Select } from '@transferwise/components';
+import * as supportedLanguages from '@transferwise/components/build/i18n';
+
+import messages from '@transferwise/components/build/es/no-polyfill/common/CloseButton/CloseButton.messages';
+import code from '../../../liveEditorCode/provider.code';
+
+export const defaultLocale = 'en';
+
+<LiveEditorBlock code={code} scope={{ Provider, Select, FormattedMessage, messages, supportedLanguages, defaultLocale }} />
+<GeneratePropsTable componentName="Provider" />
```
helendurranttw, line 12:
```suggestion

export const defaultLocale = 'en';

`Provider` provides contextual data such as translations to components. See the [usage guide](https://transferwise.github.io/neptune-web/branch/i18y/about/Setup#usage) for more.

<LiveEditorBlock code={code} scope={{ Provider, Select, FormattedMessage, messages, supportedLanguages, defaultLocale }} />
<GeneratePropsTable componentName="Provider" />
```

helendurranttw:
This looks great Anton. I did notice the example you have on the docs at the moment isn't working - potentially related to the comment I made about `np.CloseButton.aria-label` vs `np.CloseButton.ariaLabel`? Also the tests are failing at the moment (maybe also related?).

Once you've fixed the docs example and the tests and addressed the comments - I think it's good to go.

packages/components/src/provider/Provider.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import Types from 'prop-types';
+import { createIntl, createIntlCache, RawIntlProvider } from 'react-intl';
+import en from '../../i18n/en.json';
+
+const Locales = [
+  'de',
+  'en',
```
twrdjpalmer, line 8: Good chance we also pass `en_US`—do we need to support that here too?

crowdin.yml
```diff
@@ -0,0 +1,22 @@
...
+        pt-BR: pt
+        it: it
+        de: de
+        ru: ru
+        pl: pl
+        ro: ro
+        ja: ja
+        zh-HK: zh_HK
+        cs: cs
+        id: id
```
oliverviljamaa, line 21: Why would this not work? One less thing to change when supporting a new language.
```suggestion
      two_letters_code:
        zh-HK: zh_HK
```

packages/components/i18n/en.json
```diff
@@ -0,0 +1,6 @@
+{
+  "np.CloseButton.ariaLabel": "Close",
```
oliverviljamaa, line 2: Why do we need the namespacing? If we do, why not write _neptune_ in full?

packages/components/i18n/formatter.js
```diff
@@ -0,0 +1,7 @@
+exports.format = (messages) => {
+  const results = {};
+  Object.entries(messages).forEach(([id, msg]) => {
+    results[id] = msg.defaultMessage;
+  });
+  return results;
```
oliverviljamaa, line 6: Is this the same as:
```suggestion
  return Object.fromEntries(
    Object.entries(messages).map(([id, { defaultMessage }]) => ([id, defaultMessage])),
  );
```

packages/components/i18n/zh_HK.json
```diff
@@ -0,0 +1,6 @@
+{
+  "np.CloseButton.aria-label": "Close",
+  "np.Pagination.aria-label": "Pagination navigation",
+  "np.PaginationLink.ariaLabel.active": "Go to page {pageNumber}",
+  "np.PaginationLink.ariaLabel.inactive": "Current page, page {pageNumber}"
+}
```
oliverviljamaa, line 6: Is this all the copy we want to add for now or will there be separate PRs for more? (general Q, not related to this this language)

packages/components/src/provider/Provider.js
oliverviljamaa: Do I understand correctly that although we don't support `en_US` as a separate language in our systems (Crowdin, etc.), but want to keep showing it as a language customers can select? If yes, could we keep the knowledge of `en_US` higher up and make the conversion in CRAB, etc.?

packages/components/src/provider/Provider.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import Types from 'prop-types';
+import { createIntl, createIntlCache, RawIntlProvider } from 'react-intl';
+import en from '../../i18n/en.json';
+
+const Locales = [
```
oliverviljamaa, line 6: `LOCALES` for constant naming? `PascalCase` is usually for enums, components, types, and classes.

packages/docs/pages/about/Setup.mdx
```diff
@@ -46,6 +52,28 @@ For information on what classes are available, see the <DocLink href="/styles/co
...
 By default, apps come polyfilled. If you would like to use an unpolyfilled version because your app provides its own polyfills, see [Advanced Components Usage](#advanced-components-usage) below.
 
+## Translations and the Provider
+
+You need to pass a translations object to the `Provider` component. This can be done in a few different ways, depending on the setup of your library / application.
+
+Translation files can be accessed directly, which gives you the opportunity to load them dynamically:
+
+```js
+import en from '@transferwise/components/build/i18n/de.json';
```
oliverviljamaa, line 62: Nice!

packages/docs/pages/about/Setup.mdx
```diff
@@ -46,6 +52,28 @@ For information on what classes are available, see the <DocLink href="/styles/co
...
+import(`@transferwise/components/build/i18n/${locale}.json`);
+```
+
+Or you can load them all at once:
+
+```js
+import * as translations from '@transferwise/components/build/i18n';
+```
+
+**Note:** `Provider` must be used in repositories which bundle components (i.e. lists `@transferwise/components` as a `dependency`). For repositories which don't bundle components (typically those that list it as a `peerDependency`) it is unnecessary (often it's similar component libraries).
```
oliverviljamaa, line 73: 👏

packages/components/src/provider/Provider.js
twrdjpalmer: We use `en_US` bit in the money transfer homepage, so want to keep that value around. For as long as the code is in our [supported language matrix](https://transferwise.atlassian.net/wiki/spaces/EKB/pages/1054213243/Supported+languages), we should support it here too, right? I'm not sure, just making sure we've thought about it :)

oliverviljamaa: I'm not sure either — at some point, translators recommended us to stop differentiating between `en` and `en_US` (hence the _use source directly_ on the languages page), but it's still present in language selectors.

packages/components/src/provider/Provider.js
```diff
@@ -0,0 +1,51 @@
...
+  'hu',
+  'id',
+  'it',
+  'ja',
+  'pl',
+  'pt',
+  'ro',
+  'ru',
+  'tr',
+  'zh_HK',
```
sbrandwoo, line 20: `zh_HK`, with an underscore, is not a valid "locale" and should never be mentioned in a variable called `locale`.

packages/components/src/provider/Provider.js
sbrandwoo: > but want to keep showing it as a language customers can select?

This is my understanding too. In still affects how dates are formatted. It's also a good example of where TW language doesn't match locale, as our language of `en` is British English, whereas the international convention is to have `en` as US English.

Crab will use the following:
- Language US English `en_US`, locale `en`
- Language British English `en`, locale `en-GB`

packages/components/src/provider/Provider.js
sbrandwoo: For reference, locales should be [BCP 47 / IETF language tags](https://en.wikipedia.org/wiki/IETF_language_tag).

An easy way to test is to see if it works with the browser's Intl library:

```
new Date().toLocaleDateString("zh_HK")
```

sbrandwoo:
As this is a public package I believe we should be as correct as possible when it comes to locales. Values such as `en_US` and `zh_HK`, with underscores, exist because of TW tech debt (Webapp won't accept hyphens) and we should avoid them where possible.

packages/components/src/provider/Provider.js
oliverviljamaa: I agree with @sbrandwoo — even if this wasn't a public package, I'd aim to hide our legacy-related constraints as much as possible, rather than spread it further.

Perhaps we could accept standard locales and map to the best option?

packages/components/src/common/CloseButton/CloseButton.messages.js
dozortsev: this is now fixed, 

as a developer you don't have to deal with translation json files, you only make changes in `*.messages.js` and update Crowdin source file (`en.json`) by running `yarn build` then push and wait until Crowdin update other files for you https://github.com/transferwise/neptune-web/pull/728

packages/components/i18n/en.json
dozortsev: okey, fixed

packages/components/i18n/zh_HK.json
dozortsev: yeah, we there will be follow up PR

don't want do that in this PR as it's already pretty big

packages/components/src/provider/Provider.js
dozortsev: 👌

dozortsev:
Made a change to support only IETF language tag, I agree we should only them and don't tight our Neptune library to TW specifics.

https://transferwise.github.io/neptune-web/branch/i18y/components/context/Provider here in Select you can see list of supported locale codes, if unrecognized code supplied the component will fallback to `en`.

cc @oliverviljamaa @sbrandwoo @twrdjpalmer

packages/components/src/provider/Provider.js
```diff
@@ -0,0 +1,50 @@
...
+  'ru',
+  'tr',
+  'zh-HK',
+];
+
+function Provider({ i18n, children }) {
+  const cache = createIntlCache();
+  const { locale, messages } = i18n;
+  let intlConfig;
+  if (LOCALES.includes(locale)) {
```
oliverviljamaa, line 27: Could we even be more flexible and support _language + region_ locales, especially considering we support `zh_HK`? We could just check if we have `locale.slice(0,2)` and fall back to that if we do.

oliverviljamaa (approved):
Awesome job on this 🚀

packages/components/src/provider/Provider.js
sbrandwoo: Crab will give `en-GB` and `en-US` locales, so just checking the first two characters sounds good as a fallback.

@danlaush might have further thoughts about how this will work with lienzo languages.

packages/docs/pages/about/Setup.mdx
```diff
@@ -46,6 +52,29 @@ For information on what classes are available, see the <DocLink href="/styles/co
...
+
+You need to pass a translations object to the `Provider` component. This can be done in a few different ways, depending on the setup of your library / application.
+
+Translation files can be accessed directly, which gives you the opportunity to load them dynamically:
+
+```js
+import en from '@transferwise/components/build/i18n/de.json';
+
+const locale = 'zh-HK';
+import(`@transferwise/components/build/i18n/${locale}.json`);
```
sbrandwoo, line 65: As mentioned in `Provider`, we might need to match IETF locales into locales that Neptune supports. That would also be useful here when selecting which translation to load.

Perhaps Neptune could expose a function such as `toNeptuneLocale(locale: string): string` that consumers can use to map IETF locales to supported Neptune locales?

packages/components/src/provider/Provider.js
danlaush: Thanks for thinking of Lienzo <3 We'll normalise the language from Lienzo before we use it anywhere—it's a free text field so there are >70 uniques in use. https://looker.tw.ee/sql/cn2v3hfgpthxwc

I think this could be fine for now, with one thing to think about. When (not if—only a matter of time) we start supporting mainland Chinese (`zh`, `zh-Hans`, `zh-CN`), how could we handle that separately from `zh-HK`?

dozortsev: @danlaush we will add support for them `zh`, `zh-Hans`, `zh-CN` (meaning add it in [list of supported locales](https://github.com/transferwise/neptune-web/pull/674/files#diff-2e5996335d2ef854008793db05763e311536f765f40dd7a340f5c8ed2eb86914R47) and provide you a JSON file of it)

danlaush: I think in that case we couldn't safely use the two-letter code as a fallback. If `zh` existed and someone passed in `zh_HK`, it would clip it to `zh` and use simplified Chinese, which wouldn't be readable by traditional Chinese readers. Do I understand it right?

sbrandwoo (approved):

packages/docs/pages/about/Setup.mdx
dozortsev: add mapping API https://github.com/transferwise/neptune-web/pull/674/commits/1e2d6222f2304f6a4188e70b7c7732b8aa66eafc and updated docs https://github.com/transferwise/neptune-web/pull/674/commits/f6747be1bb5efa90fd2f3e8f24d4e367ce819ffb

packages/components/src/provider/Provider.js
```diff
@@ -0,0 +1,36 @@
...
+
+function Provider({ i18n, children }) {
+  const cache = createIntlCache();
+  const { locale, messages } = i18n;
+  const localeCode = mapLocale(locale);
+
+  let intlConfig;
+  if (localeCode === null) {
+    // eslint-disable-next-line no-console
+    console.warn(`Unsupported locale value was provided: '${locale}', use English by default`);
```
twrdjpalmer, line 15:
```suggestion
    console.warn(`Unsupported locale value was provided: '${locale}', defaulting to 'en'`);
```

packages/docs/pages/about/Setup.mdx
```diff
@@ -46,6 +52,29 @@ For information on what classes are available, see the <DocLink href="/styles/co
...
 By default, apps come polyfilled. If you would like to use an unpolyfilled version because your app provides its own polyfills, see [Advanced Components Usage](#advanced-components-usage) below.
 
+## Translations and the Provider
+
+You need to pass a translations object to the `Provider` component. This can be done in a few different ways, depending on the setup of your library / application.
+
+Translation files can be accessed directly, which gives you the opportunity to load them dynamically:
+
+```js
+import en from '@transferwise/components/build/i18n/de.json';
```
twrdjpalmer, line 62:
```suggestion
import de from '@transferwise/components/build/i18n/de.json';
```

packages/docs/pages/about/Setup.mdx
```diff
@@ -31,13 +31,19 @@ import '@transferwise/icons/lib/styles/main.min.css';
...
+  const currentLocale = ...
 
-export default function Hello() {
   return (
-    <Button size={Button.Size.SMALL} block={true}>
-      Hello Neptune
-    </Button>
+    // Components must be wrapped by the Provider component, which initialises and provides
+    // context data for them. This should usually be done once at the top app level component
+    <Provider i18n={{ locale: currentLocale, messages: translations[currentLocale] }}>
```
twrdjpalmer, line 43: Might be OTT, but might be worth adding a note that this is the spec locale & not the TW internal one, so we remember to do the conversion

twrdjpalmer (approved):

packages/components/src/test-utils/index.js
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Custom `render` function which wraps passed elements in Provider component
+ * Source: https://testing-library.com/docs/example-react-intl/#creating-a-custom-render-function
+ */
+function render(ui, { locale = 'en', messages = en, ...renderOptions } = {}) {
+  // eslint-disable-next-line react/prop-types
+  function Wrapper({ children }) {
+    return <Provider i18n={{ locale, messages }}>{children}</Provider>;
+  }
+  return rtlRender(ui, { wrapper: Wrapper, ...renderOptions });
```
helendurranttw, line 16: Abbreviating React Testing Library is going to get confusing once we start talking more about Right To Left 😀

helendurranttw (approved):
👏.  🎉.  💪

andreapiras (approved):

packages/components/src/test-utils/index.js
dozortsev: fixed

PR #676, opened 2020-11-16
feat: add callback to react to search changes in money input

mikknarusk:
## 🖼 Context
We wish to customise the `customActionLabel` based on what the user has searched for, in order to be more helpful and clear about which currency routes we support. Designs in the ticket:
https://transferwise.atlassian.net/browse/SEND-1171

## 🚀 Changes
Adds optional `onSearchChange` callback that exposes the searched string and resulting filtered options to the consumer, both are currently isolated inside the component.

Quick mockup in storybook:
ezgif com-gif-maker

 

## 🤔 Considerations
The search API now looks simiar to that of [Select](https://transferwise.github.io/neptune-web/components/inputs/Select), but the behavior is actually different:
- If you pass `onSearchChange` to Select, it will stop searching internally and over control to the consumer
- If you pass `onSearchChange` to MoneyInput, it will keep searching internally, making the callback essentially readonly. I've tried to address this in the docs, let me know if it's (not) clear or if we should change the prop name. 

## ✅ Checklist


fergusjordantw (approved):

packages/components/src/moneyInput/MoneyInput.js
```diff
@@ -296,6 +308,7 @@ MoneyInput.defaultProps = {
   locale: 'en-GB',
   addon: null,
   searchPlaceholder: '',
+  onSearchChange: () => {},
```
dozortsev, line 311: this is intentional? I would pass here `undefined`

mikknarusk: Yes, it's intentional. It's a choice between either this default empty function or checking `if (typeof this.props. onSearchChange === 'function')` in the logic above. I feel like this is cleaner but no strong preference

packages/components/src/moneyInput/MoneyInput.js
```diff
@@ -109,6 +109,14 @@ export default class MoneyInput extends Component {
     }
   };
 
+  handleSearchChange = (searchQuery) => {
+    this.setState({ searchQuery });
+    this.props.onSearchChange({
+      searchQuery,
+      filteredOptions: filterOptionsForQuery(this.props.currencies, searchQuery),
```
dozortsev, line 116: maybe here would be better invoke `getSelectOptions()` instead of `filterOptionsForQuery(this.props.currencies, searchQuery),`?

mikknarusk: `getSelectOptions` appends the `customActionLabel` to the list, which is useless for the client and redundant since the client is already providing the label

packages/components/src/moneyInput/MoneyInput.js
dozortsev: yeah, I would prefer checking if it's assigned, and Select already handle this case (e.g [here](https://github.com/transferwise/neptune-web/blob/i18y/packages/components/src/select/Select.js/#L151))

packages/components/src/moneyInput/MoneyInput.js
dozortsev: alright, okey, just wanted to confirm this.

packages/components/src/moneyInput/MoneyInput.js
mikknarusk: Will make the change. Note this is not directly related to the callback check in Select though

dozortsev (approved):

PR #679, opened 2020-11-18
ADS-1698 feat: Add tabs for oneOf of objects

felixxiatransferwise:
## 🖼 Context

We would like to introduce tabs so that customers are able to see all available account type options without expanding the dropdown.

## 🚀 Changes

 
- Use tabs when the server specifies it as the control
Screenshot 2020-11-18 at 11 13 31

## ✅ Checklist


packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.js
```diff
@@ -102,7 +102,14 @@ const OneOfSchema = (props) => {
   };
 
   const mapSchemas = (schema) => {
-    return { ...schema, oneOf: schema.oneOf.map(mapOneOfToConst) };
+    return {
+      ...schema,
+      originalType:
+        schema.oneOf[0].type === OriginalOneOfTypes.OBJECT
+          ? OriginalOneOfTypes.OBJECT
+          : OriginalOneOfTypes.CONST,
```
felixxiatransferwise, line 110: This is introduced because we only want to render tabs for oneOf objects and not oneOf consts. The way this was mapped before means we lose this information once we end up inside `SchemaFormControl` where we decide the type of component to render.

StevePole: Is this really necessary?  Sure, a tab for a const might be slightly odd UX, but is that something we can prevent in the library rather than the client?  Just seems to add needless complexity here.

felixxiatransferwise: Was thinking `tab` was a client decision of how to display a `select` but sure we can move this to the server side.

StevePole: We can't assume web has more screen width... 

We're using 'tabs' (or something that looks very similar) on mobile already, we use it for the account type choice (e.g. IBAN vs Sort code).  How does mobile know to use it for that choice but not for other oneOfs?  I don't think we can assume that any oneOf with a small number of options should be rendered as tabs.  There's also the problem of 'title'.  When we render a oneOf as a select (or radio), we use the title to explain the choice.  But when we render tabs we don't have that title.  How do we know if the title is necessary if the backend doesn't tell us?

Based on that, I feel it should come from the server.  Though in either case (mobile or web), the device may have to choose to ignore it based on the available space.

felixxiatransferwise: > How does mobile know to use it for that choice but not for other oneOfs?
Couldn't it do it from a combination of the view size, the sum of the tab title lengths and the number of tabs? I.e if we can fit all of the options in a tab format, do it, otherwise render it as some other component?

@JulianG-TW also floated the idea that on web we may want to dynamically change the tab to a dropdown as the viewport shrinks.

> There's also the problem of 'title'. When we render a oneOf as a select (or radio), we use the title to explain the choice. But when we render tabs we don't have that title.
I might be misunderstanding but why don't we have the title here? In `OneOfSchema`, the title would still get mapped to title on the option 
```
{
      title: schema.title,
      description: schema.description,
      const: index,
      disabled: schema.disabled,
      icon: schema.icon,
    }
```
so we can still render it in the tab. 

> Based on that, I feel it should come from the server. Though in either case (mobile or web), the device may have to choose to ignore it based on the available space.
👍  Sure, let's do this for now since it is probably the most straight forward option.

StevePole: I didn't mean the title of the choices, but the title that describes what choice you're making:
```
{
  title: "What's your favourite colour?",
  oneOf: [{ const: "red" }, {const: "blue" }]
}
```
If we render this as tabs it will make no sense, because you need the context of the title.

StevePole: (Yes, these are consts, but you could construct a choice between object schemas with the same problem)

StevePole: We can't infer that tabs are ok because they'd fit.  We need to know that the title is not necessary.

felixxiatransferwise:
Closing because we are moving away from tabs.

PR #688, opened 2020-11-20
feat: Remove oneOf schemas defaulting logic

felixxiatransferwise:
## 🖼 Context

We want to remove oneOf defaulting logic such that:

- model matches -> model value
- model doesn’t match -> default value exists -> default value
- model doesn’t match -> default value doesn’t exist -> no value selected

**Since `default` is not in the dynamic interfaces lib yet, I've excluded the "defaulting logic" from this PR (will be added once it's in the lib).** 

## ✅ Checklist


packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.js
```diff
@@ -35,10 +35,12 @@ const OneOfSchema = (props) => {
     if (index >= 0) {
       return index;
     }
-    // Otherwise, default to the first schema if it's not a const
-    if (schema.oneOf[0] && !isConstSchema(schema.oneOf[0])) {
+
+    // Otherwise, render the first oneOf non-const schemas if there's only 1 (otherwise it won't be displayed)
```
StevePole, line 39: I'm not sure this is right.  After this change, we don't automatically default const or non-const oneOfs, so I think we just need:
```
// Otherwise, if there's only one schema, default to it
if (schema.oneOf.length === 1) {
  return 0;
}
```
?

packages/dynamic-flows/src/jsonSchemaForm/schemas/oneOf.json
```diff
@@ -6,6 +6,86 @@
       "type": "string",
       "title": "Full name"
     },
+    "justOne": {
```
StevePole, line 9: Can we make an example that feels more realistic?

felixxiatransferwise: Removed this one because it's pretty unrealistic to have just 1 object schema inside an oneOf anyway and it's covered by tests. I forgot to remove it after manual testing.

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.js
felixxiatransferwise: We weren't defaulting on oneOf consts before because we want the customer to make the conscious choice unless a default is available (which I'll add once oneOfs have defaults).

StevePole (approved):
👍

PR #694, opened 2020-11-24
feat: ADS-1721 Client side persist async validation

felixxiatransferwise:
## 🖼 Context

Currently, there are a couple of issues with persist async validation:
- we trigger a persist async call even when the persist async model is `null`, we should prevent this.
- we don't show validation errors when a persist async field is required upon submission

## 🚀 Changes

 
- Check the persist async model value is truthy before triggering persist async call on blur
- Pass down `required` prop to `BasicTypeSchema`

## ✅ Checklist


packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/PersistAsyncSchema.js
```diff
@@ -72,7 +72,9 @@ const PersistAsyncSchema = (props) => {
   const getErrorFromResponse = (errorProperty, response) => response.validation?.[errorProperty];
 
   const onBlur = () => {
-    getPersistAsyncResponse(persistAsyncModel, props.schema.persistAsync);
+    if (persistAsyncModel) {
```
StevePole, line 75: Not sure this is specific enough.  While unlikely, we might have:
```
{
  type: "boolean",
  persistAsync: { ... }
}
```
Perhaps use isNull from our library?

felixxiatransferwise: ah good point, will do

StevePole (approved):
👍

PR #696, opened 2020-11-25
feat: add SwitchOption component

fergusjordantw:
## 🖼 Context

https://transferwise.atlassian.net/browse/SEND-1856 adding new stuff to settings which requires a list of switches
Designs here: https://www.figma.com/file/r0bTJPZ16amp33hNefBnEZ/Discoverability-opt-in?node-id=1575%3A37795

## 🚀 Changes

- Added a new SwitchOption component
- Adds a SwitchOptionList component based on the NavigationOptionList

## 🤔 Considerations

I've basically just copied the NavigationOptionList component, but it's actually very generic and just rendering whatever children are passed in with a little bit of styling. I think it would actually be best to make a generic OptionList component that isn't limited (although it's only constricted by naming at the moment) to the NavigationOption. Doing this would be a breaking change though so figured I'd raise it here first.

Option:

image

List:

image

## ✅ Checklist


packages/components/src/switchOption/SwitchOption.js
```diff
@@ -0,0 +1,52 @@
...
+import Switch from '../switch';
+
+const SwitchOption = ({ media, title, content, id, checked, onChange, complex }) => {
+  const sharedProps = { media, title, content, complex };
+
+  const stopPropagation = (event) => {
+    if (event) {
+      event.stopPropagation();
+      event.preventDefault();
+      if (event.nativeEvent && event.nativeEvent.stopImmediatePropagation) {
```
sergonius, line 14: `event.nativeEvent?.stopImmediatePropagation`?
Could even go further and get rid of the `if`.

PR #703, opened 2020-11-26
fix: sizeswapper fix

andreapiras:
## 🚀 Changes

SizeSwapper is using an adaptive technique which serves markup based on device size. This doesn't works perfectly with SSR where device size is unknown. We still need a solution to monitor parent containers on resize vs windows resize so we can't replace this solution with traditional responsive content that won't be take consideration parent width. 

## 🤔 Considerations
This function should be used sparingly and also with an eye to SEO. For now the resolution would be render all elements on SSR env (like a responsive technique would do) and replace this content on first hydration

## ✅ Checklist

- [] Changes are tested and all tests pass
- [] Changes meet [accessibility standards](https://github.com/transferwise/neptune-web/blob/main/ACCESSIBILITY.md) and there are no violations in the console
- [] Changes work in all supported browsers (don't forget IE11)
- [] You've updated the documentation if necessary


packages/components/src/sizeSwapper/SizeSwapper.spec.js
```diff
@@ -31,14 +31,24 @@ describe('SizeSwapper', () => {
...
+    expect(queryByText(`element-2-${breakpoints[1]}`)).toBeInTheDocument();
+  });
+
   it('switches elements according to breakpoints', () => {
     const { getByText, queryByText } = render(<SizeSwapper items={items} />);
 
     expect(getByText('element-1-0')).toBeInTheDocument();
     expect(queryByText('element-2-0')).toBeInTheDocument();
 
-    expect(queryByText(`element-1-${breakpoints[1]}`)).not.toBeInTheDocument();
```
helendurranttw, line 40: Now that everything is in the document, I don't think this really proves that it's switching correctly. Could we also check if it's visible or not visible?

packages/components/src/sizeSwapper/SizeSwapper.js
```diff
@@ -16,9 +16,16 @@ const SizeSwapper = ({ items }) => {
...
-  const itemsToRender = items
-    .filter(({ breakpoint = 0 }) => clientWidth >= breakpoint)
-    .slice(-1)[0];
+  let itemsToRender = [];
+
+  if (clientWidth) {
+    // eslint-disable-next-line
+    itemsToRender = items.filter(({ breakpoint = 0 }) => clientWidth >= breakpoint).slice(-1)[0];
+  } else {
+    // If clientWidth is not defined renders all elements like a responsive technique would do.
```
helendurranttw, line 25: Might be worth calling out that this will happen if it's SSR

packages/components/src/sizeSwapper/SizeSwapper.js
```diff
@@ -16,9 +16,16 @@ const SizeSwapper = ({ items }) => {
...
   // If all breakpoints are specified and clientWidth never > breakpoint itemsToRender can be undefined.
   // Do not use deconstruct here to get items and layout.
-  const itemsToRender = items
-    .filter(({ breakpoint = 0 }) => clientWidth >= breakpoint)
-    .slice(-1)[0];
+  let itemsToRender = [];
+
+  if (clientWidth) {
+    // eslint-disable-next-line
+    itemsToRender = items.filter(({ breakpoint = 0 }) => clientWidth >= breakpoint).slice(-1)[0];
```
helendurranttw, line 23: You could use a pop instead of slice etc

packages/components/src/sizeSwapper/SizeSwapper.spec.js
andreapiras: All Element s are going to be in the document only if clientWidth is not defined. After that , on hydration only the right set of elements will be in the document

packages/components/src/decision/Decision.js
```diff
@@ -21,7 +21,7 @@ const Decision = ({ options, presentation, type, size }) => {
         },
         {
           items: [],
-          breakpoint: SizeSwapper.Breakpoint.MEDIUM,
+          breakpoint: SizeSwapper.Breakpoint.EXTRA_SMALL,
```
andreapiras, line 24: We discussed this change with @mikknarusk and @gusso and we decided to go for this breakpoint

packages/components/src/sizeSwapper/SizeSwapper.js
mikknarusk: random driveby: I'm vary of using pop since alters the original value, which is fine now since it's mitigated by the `.filter` but might cause a bug in the future

packages/components/src/sizeSwapper/SizeSwapper.spec.js
helendurranttw: Right, I get that, but my point is - the point of sizeswapper is that it shows some elements at one width, and other elements at another, right?

Now that all elements are shown at the first breakpoint (on SSR), how do we know they're not all displayed to the user?

andreapiras: Got it! updated

packages/components/src/sizeSwapper/SizeSwapper.js
```diff
@@ -12,13 +12,23 @@ const SizeSwapper = ({ items }) => {
...
-    .slice(-1)[0];
+  let itemsToRender = [];
+
+  if (clientWidth) {
+    // eslint-disable-next-line
+    itemsToRender = items.filter(({ breakpoint = 0 }) => clientWidth >= breakpoint).pop();
+  } else {
+    // On SSR environments clientWidth is undefined because window is undefined so we render
+    // all elements by default. If there's no SSR or on first Hydration only the right elements are going to be in the markup.
+    // If clientWidth is not defined renders all elements like a responsive technique would do.
```
andreapiras, line 28:
```suggestion
    // On SSR environments clientWidth returns null because window is undefined so we render
    // all elements by default.
    // If there's no SSR or on first Hydration only the right elements are going to be in the markup.
    // If clientWidth is null or zero all elements render like a responsive technique would do.
```

helendurranttw (approved):

PR #707, opened 2020-11-30
feat: add aria-label for Info button

andreapiras:
## 🖼 Context

This PR addresses some improvements in Summary:
- Adds aria-label to info button
- Update docgen to handle optional chaining

## ✅ Checklist


packages/components/src/summary/Summary.less
```diff
@@ -22,6 +22,8 @@
     }
     // The second icon is the badge
     :nth-child(2) {
+      background-color: --color-background-screen;
```
andreapiras, line 25: Add background to transparent badge to prevent the content underneath to show through

packages/components/src/popover/Popover.js
```diff
@@ -55,7 +55,17 @@ export default class Popover extends Component {
     document.removeEventListener('click', this.closePopoverOnOutsideClick, true);
   };
 
-  toggle = () => (this.state.isOpen ? this.close() : this.open());
+  toggle = () => {
+    const { onClick } = this.props;
+    if (onClick) {
+      onClick();
```
helendurranttw, line 61: Should we call this onOpen? 🤔

packages/components/src/popover/Popover.js
```diff
@@ -140,6 +153,8 @@ Popover.propTypes = {
 };
 
 Popover.defaultProps = {
+  'aria-label': null,
+  onClick: null,
   title: null,
```
helendurranttw, line 158: All of these should be undefined.

packages/components/src/summary/info/Info.js
```diff
@@ -58,8 +77,10 @@ Info.propTypes = {
 };
 
 Info.defaultProps = {
+  'aria-label': null,
   className: null,
   content: null,
+  onClick: null,
```
helendurranttw, line 83: Undefined everywhere!

helendurranttw:
You added features, so perhaps this is a feat?

packages/components/src/summary/Summary.less
helendurranttw: Don't think this is doing what you think it is! You're missing the var() so this is probably not doing anything.

packages/components/src/summary/Summary.less
```diff
@@ -22,6 +22,8 @@
...
+      border: 50%;
```
helendurranttw, line 26: Is this supposed to be border-radius?

packages/components/src/popover/Popover.js
andreapiras: This run on both open and close,maybe I should pass outside this.state.isOpen instead?

packages/components/src/popover/Popover.js
```diff
@@ -140,6 +153,8 @@ Popover.propTypes = {
 };
 
 Popover.defaultProps = {
+  'aria-label': undefined,
+  onClick: undefined,
   title: null,
```
helendurranttw, line 158: Missed one 👀

packages/components/src/popover/Popover.js
```diff
@@ -122,6 +133,8 @@ export default class Popover extends Component {
 }
 
 Popover.propTypes = {
+  'aria-label': Types.node,
+  onClick: Types.func,
```
helendurranttw, line 137: Why does the Popover need an onClick? 🤔

helendurranttw: Oh I see. I think this is better named onOpen :)

helendurranttw: Or maybe onToggle? Since you're calling it on open and close.

packages/components/src/summary/info/Info.js
```diff
@@ -23,13 +24,29 @@ const Info = ({ className, content, presentation, size, title }) => {
...
         <>
-          <button type="button" onClick={() => setOpen(true)}>
+          <button
+            aria-label={props['aria-label']}
+            className="btn-unstyled"
+            type="button"
+            onClick={() => {
+              setOpen(true);
+              if (onClick) {
+                onClick(!open);
```
helendurranttw, line 34: I think an `onClick` function makes more sense in `Info`, but if you're passing the state of the popover / modal I think you'd better also call it `onToggle` or similar

packages/components/src/summary/info/Info.spec.js
```diff
@@ -48,9 +50,31 @@ describe('Info', () => {
...
   });
   describe(`when in ${Info.Presentation.MODAL} mode`, () => {
     it('renders modal component', () => {
+      const onClick = jest.fn();
+      const { getByLabelText } = render(
+        <Info {...props} onClick={onClick} presentation={Info.Presentation.MODAL} />,
+      );
+      const helpButton = getByLabelText('aria-label');
+
+      fireEvent.click(helpButton);
```
helendurranttw, line 73: This test doesn't seem to test what it says it's testing for

packages/components/src/summary/info/Info.spec.js
```diff
@@ -59,6 +83,21 @@ describe('Info', () => {
...
     });
+
+    it('fires onclick with expected values', () => {
+      const onClick = jest.fn();
+      const { container } = render(
+        <Info {...props} presentation={Info.Presentation.MODAL} onClick={onClick} />,
+      );
+      const helpIcon = getHelpIcon({ container });
+
+      expect(helpIcon).toBeInTheDocument();
```
helendurranttw, line 94: Do you need this check in this test?

packages/components/src/summary/info/Info.js
```diff
@@ -58,6 +64,7 @@ Info.propTypes = {
 };
 
 Info.defaultProps = {
+  'aria-label': null,
   className: null,
   content: null,
```
helendurranttw, line 69: undefined 👩‍🌾

helendurranttw (approved):
Change those last nulls to undefined and you're good to go

PR #709, opened 2020-12-01
feat: change container of Modal component to allow to be 'adaptive' / 'scrollable'

dozortsev:
## 🖼 Context

Current scrollable body of Modal sometimes makes it inconvenient UX if the body contains quite big amount of copy and/or some components (which have ability to expand, popup etc).

<details>
<summary>See example with Select:</summary>

occluded-dropdown

</details>

## 🚀 Changes

Adding ability to make body of Modal component adaptive to size of their content and allow any kind of dropdowns appear over Modal's footer, this behaviour is not default, it needs to be enabled via new prop:

```jsx
<Modal container={'adaptive'} />
```

Because of that change we also had to make Dimmer auto scrollable if necessary, this could be enabled/disabled via new prop `scrollable`.

Modal's current behavious (which you currently can see on [prod instance](https://transferwise.github.io/neptune-web/components/overlays/Modal)) will be still available by default:

```jsx
<Modal />

// or
<Modal container={'scrollable'} />
```

---

We also added `position` prop to position the component at the top or at the center. 

<details>
<summary> `position={TOP}` is useful when you use e.g Tabs component where each tab has diff amount of content and you want to avoid Modal jumping when switching between these tabs:</summary>

skipipng

</details>

---

Demo of the changes: https://transferwise.github.io/neptune-web/branch/fix-select-inside-modal/components/overlays/Modal

## ✅ Checklist


andreapiras (changes requested):
We need to implement the select dropdown reposition

packages/components/src/modal/Modal.js
```diff
@@ -5,14 +5,19 @@ import CSSTransition from 'react-transition-group/CSSTransition';
...
-import { Size, addNoScrollBodyClass, removeNoScrollBodyClass } from '../common';
+import { Size, Container, addNoScrollBodyClass, removeNoScrollBodyClass } from '../common';
 import Dimmer from '../dimmer';
 
 const TRANSITION_DURATION_IN_MILLISECONDS = 150;
 
 class Modal extends Component {
   static Size = Size;
 
+  static Container = {
```
andreapiras, line 16: I think having a prop called 'fitHeight' = {true|false} would be enough

packages/components/src/modal/Modal.js
```diff
@@ -85,7 +91,12 @@ class Modal extends Component {
           unmountOnExit
         >
           <div
-            className={`tw-modal fade ${className}`}
+            className={classNames(
+              'tw-modal',
+              { 'tw-modal--adaptive': container === Container.ADAPTIVE },
```
andreapiras, line 96: This would become
'tw-modal-fit-height' : fitHeight

packages/components/src/modal/Modal.js
dozortsev: I prefer we use enum approach here (please check [bool vs enum](https://material-ui.com/guides/api/#boolean-vs-enum)) as it's more scalable approach, we might add third type of container in Modal, and/or we might add a few more types of containers [there](https://github.com/transferwise/neptune-web/pull/709/files#diff-9b93bb4e4e56b73b6c4bfc3a638d6ea371472568c45d7a66cdb1dd63e7c1bd3aR1) for other components which won't be used by Modal, so it's not only for Modal, tomorrow it could be about Popover, Alert etc.

dozortsev: I'm fine to consider to change API naming, 

wdyt?
```ts
container: 'fitContent' | 'fitMaxHeight'
```

packages/components/src/modal/Modal.js
dozortsev: https://github.com/transferwise/neptune-web/pull/709#discussion_r535359292

packages/components/src/modal/Modal.js
helendurranttw: I think container is a little bit too generic - how about `contentFit`? I like the idea of a single boolean prop, but am also ok with the enum - if you rename it to be a little more descriptive. How about: `contentFit: 'scroll' | 'expand'`? Or for a boolean: `contentScroll: true | false`?

Also - one quick note on the enum - let's not put it in common just yet - leave it inside Modal, and we can pull it out when we want to reuse it. I can't see this being a frequently used prop.

helendurranttw (commented):
Looks great Anton, thank you.

helendurranttw:
@dozortsev - looks like there's an issue in IE11 - the modal isn't rendering correctly.

packages/components/src/modal/Modal.js
dozortsev: @helendurranttw thanks for feedback

re. `contentFit: 'scroll' | 'expand'` I would not use word "expand" because it **adapting** to content size which mean that sometimes it shrinks.

<details>
<summary>Example of adaptive modal container (shrinked, expanded and then again shrinked):</summary>
Screen Shot 2020-12-07 at 11 22 40 AM
Screen Shot 2020-12-07 at 11 22 25 AM
Screen Shot 2020-12-07 at 11 22 16 AM
</details>

I think "adaptive" is perfect word here 🙂

---

re. following:

> ...Also - one quick note on the enum - let's not put it in common just yet - leave it inside Modal, and we can pull it out when we want to reuse it. I can't see this being a frequently used prop.

I'm afraid that we might forget about it, and it won't be discoverable for contributors either if we only store in inside this component. I see no harm to define it in common, this way it will find it's second usage faster, defining something in common tell contributors that this meant to be reused so we will avoid people inventing 🙂

oliverviljamaa:
Having the options behind the footer feels like a bug to me. Did you consider showing them above the footer and above the modal for modals with the current, `scrollable`, container, too?
image

packages/components/src/modal/Modal.js
oliverviljamaa: > we might add third type of container in Modal
What could that be? Just challenging and wondering whether we're sacrificing a bit of readability for the unknown.

Regarding naming — I feel the option defines how the _height_ of the modal is determined, either by _content_ or _screen/viewport (height)_, so would recommend using something along those lines in the lingo.

oliverviljamaa: `height*: 'content' | 'viewport'`, where `*` is a word I don't have yet. `heightBy`? `heightDeterminedBy`? 🤔

andreapiras:
> Having the options behind the footer feels like a bug to me. Did you consider showing them above the footer and above the modal for modals with the current, `scrollable`, container, too?
> 

Hey @oliverviljamaa, at the moment the body of modal has an overflow-y:auto, so the dropdown can't exceed tw-modal-body boundaries. The only solution we could see here is too render the select modal into a portal which would then cause many headaches  on attaching to the triggering field (also imagine scrolling etc). So for now if you want to use select in modal you have to use adaptive modals that will adapt to the content. If you have any other idea please share with us :)

packages/components/src/modal/Modal.js
dozortsev: @oliverviljamaa 

> ..What could that be? Just challenging and wondering whether we're sacrificing a bit of readability for the unknown.

I don't know, but it can be third type of container for Modal or for another component

packages/docs/liveEditorCode/modal.code.js
```diff
@@ -17,6 +17,7 @@
           </>
         }
         open={open}
+        container={Modal.Container.ADAPTIVE}
```
andreapiras, line 20: should we document position too here?

packages/components/src/modal/Modal.js
andreapiras: I think container is way too generic and my preference would definitely go for boolean as I don't see any other use cases. If we use tight the prop name with the height (fitHeight true or false) if tomorrow we'll need another prop we'll be able to introduce it in the same way -> fitWidth={true|false} etc

helendurranttw: scroll: content | viewport

packages/docs/liveEditorCode/modal.code.js
dozortsev: added

packages/components/src/dimmer/Dimmer.spec.js
```diff
@@ -45,7 +45,16 @@ describe('Dimmer', () => {
...
+    expect(component.find(Dimmer).props()).toEqual({ ...props, scrollable: false, children: null });
+  });
+
+  it('do not make Dimmer auto scrollable if scrollable=false', () => {
+    expect(component.find('.dimmer').hasClass('dimmer--scrollable')).toBe(false);
+  });
+
+  it('makes Dimmer auto scrollable if scrollable=true', () => {
+    component.setProps({ scrollable: true });
+    expect(component.find('.dimmer').hasClass('dimmer--scrollable')).toBe(true);
```
andreapiras, line 57:
```suggestion
  it('is not scrollable when scrollable=false', () => {
    expect(component.find('.dimmer--scrollable')).toHaveLength(0);
  });

  it('is scrollable when scrollable=true', () => {
    component.setProps({ scrollable: true });
    expect(component.find('.dimmer--scrollable')).toHaveLength(1);
```

andreapiras (approved):

oliverviljamaa:
> Hey @oliverviljamaa, at the moment the body of modal has an overflow-y:auto, so the dropdown can't exceed tw-modal-body boundaries. The only solution we could see here is too render the select modal into a portal which would then cause many headaches on attaching to the triggering field (also imagine scrolling etc). So for now if you want to use select in modal you have to use adaptive modals that will adapt to the content. If you have any other idea please share with us :)

Thanks, @andreapiras! The workarounds don't seem great indeed. I'm wondering why we need the `overflow-y:auto`?

PR #710, opened 2020-12-01
[ADS-1747] refreshFormOnChange is now conditional

JulianG-TW:
## 🖼 Context

[ADS-1747](https://transferwise.atlassian.net/browse/ADS-1747)

In the create account form, when the user types in an email address, we're debouncing refresh by 1 second, but it's not enough:

If the users makes a pause longer than 1 second while typing the email address, we refresh the requirements, even when the email is not valid.

too-many-refreshes

The fact that focus is lost after refreshing only makes things worse.

## 🚀 Changes

In `BasicTypeSchema`, `refreshFormOnChange` will only be broadcasted as true if the schema is valid.

only-one-refresh

## ✅ Checklist


packages/dynamic-flows/src/jsonSchemaForm/basicTypeSchema/BasicTypeSchema.js
```diff
@@ -28,7 +28,9 @@ const BasicTypeSchema = (props) => {
     setLastModel(broadcastModel);
 
     if (broadcastModel !== lastModel) {
-      props.onChange(broadcastModel, props.schema);
+      const isValid = validationKeys.length === 0;
+      const refreshFormOnChange = props.schema.refreshFormOnChange && isValid;
+      props.onChange(broadcastModel, { ...props.schema, refreshFormOnChange });
```
felixxiatransferwise, line 33: The second param here was meant to be the unchanged triggering schema, but now the refreshFormOnChange property in the triggering schema actually becomes `shouldRefreshFormOnChange`, this might be a little confusing. 

Should we broadcast another value: isFieldValid (whether the field triggering the onChange is valid)?

JulianG-TW: sure. I just thought that from the point of view of the consumer, it makes no difference.

JulianG-TW: ## Alternative 1:

```js
const isTriggerValid = validationKeys.length === 0;
props.onChange(broadcastModel, props.schema, isTriggerValid);
```
**PROs**: We don't touch the schema
**CONs**: We need to propagate this change to all other `prop.onChange` calls in other schema components.

JulianG-TW: ## Alternative 2:

```js
const isTriggerValid = validationKeys.length === 0;
props.onChange(broadcastModel, {...props.schema, isTriggerValid });
```
**PROs**: We don't need to propagate this change to all other `prop.onChange` calls in other schema components.
**CONs**: We touch the schema object.

JulianG-TW: I think I still prefer **Option 0**.

JulianG-TW:
Closed in favour of #718

PR #715, opened 2020-12-02
JsonSchemaForm onChange takes a 4th argument: isTriggeringSchemaValid

JulianG-TW:
## 🖼 Context

[ADS-1747](https://transferwise.atlassian.net/browse/ADS-1747)

In the create account form, when the user types in an email address, we're debouncing refresh by 1 second, but it's not enough:

If the users makes a pause longer than 1 second while typing the email address, we refresh the requirements, even when the email is not valid.

too-many-refreshes

The fact that focus is lost after refreshing only makes things worse.

## 🚀 Changes

`JsonSchemaForm` `onChange` callback is now called with a fourth argument, that represents whether or not the triggering field/schema is valid.

only-one-refresh

## ✅ Checklist


JulianG-TW:
Closed in favour of #718

PR #724, opened 2020-12-09
refactor!: improving accessibility of the CheckboxButton

helendurranttw:
## 🤖 Live examples

[CheckboxButton](https://transferwise.github.io/neptune-web/branch/checkbox-a11y-css/components/inputs/CheckboxButton)
[Checkbox](https://transferwise.github.io/neptune-web/branch/checkbox-a11y-css/components/inputs/Checkbox)
[CheckboxOption](https://transferwise.github.io/neptune-web/branch/checkbox-a11y-css/components/options/CheckboxOption)

## 🖼 Context

The existing CheckboxButton uses a button. To make this more accessible, we wanted to reimplement with an input.

## 🚀 Changes

* Re-implemented the CheckboxButton using an input
* CheckboxButton now requires an onChange handler (BREAKING) and optionally an aria-label
* Pulled in some Bootstrap form less files to Neptune CSS, in order to make changes in them
* Small updates to Checkbox and CheckboxOption to make better use of CheckboxButton
* Rewrote several test files using RTL

## 🤔 Considerations

To make these changes I had to pull a bunch of files in from Bootstrap. They were mostly imported as-is, with the exception of checkbox-radio.less. I've called out changes with comments.

⚠️ There is a small breaking change. CheckboxButton now requires an onChange handler, and people using it directly should provide an `aria-label`.

## ✅ Checklist


packages/components/src/checkboxButton/CheckboxButton.js
```diff
@@ -2,38 +2,31 @@ import React from 'react';
...
-    disabled={disabled || readOnly}
-  >
-    <span className="tw-checkbox-check" />
-  </button>
+import './CheckboxButton.css';
+
+const CheckboxButton = ({ checked, className, disabled, onChange, ...rest }) => (
+  <span className={classNames(className, 'np-checkbox-button')}>
+    <input type="checkbox" disabled={disabled} checked={checked} onChange={onChange} {...rest} />
+    <span className="tw-checkbox-button">
```
helendurranttw, line 10: Reusing the old classes (shared by Angular) to take advantage of existing styling.

packages/components/src/checkboxButton/CheckboxButton.js
```diff
@@ -2,38 +2,31 @@ import React from 'react';
...
+      <span className="tw-checkbox-check" />
+    </span>
+  </span>
 );
 
 CheckboxButton.propTypes = {
+  'aria-label': Types.string,
   checked: Types.bool,
-  onFocus: Types.func,
```
helendurranttw, line 21: Using `...rest` here, rather than adding these handlers one by one. I don't think they're special enough to warrant inclusion in the API.

packages/components/src/checkboxButton/CheckboxButton.js
```diff
@@ -2,38 +2,31 @@ import React from 'react';
...
-  onClick: Types.func,
-  onBlur: Types.func,
-  disabled: Types.bool,
-  readOnly: Types.bool,
   className: Types.string,
+  disabled: Types.bool,
+  onChange: Types.func.isRequired,
```
helendurranttw, line 21: On the other hand, it's important we have an onChange handler. I can see people have been using the onClick for this purpose.

packages/components/src/checkboxOption/CheckboxOption.js
```diff
@@ -10,27 +10,23 @@ const CheckboxOption = ({ media, title, content, name, checked, onChange, comple
...
-      onClick={(event) => {
-        event.preventDefault();
-        if (!disabled) {
-          onChange(!checked);
-        }
-      }}
-      button={<CheckboxButton checked={checked} disabled={disabled} />}
+      button={
+        <CheckboxButton checked={checked} disabled={disabled} onChange={() => onChange(!checked)} />
+      }
```
helendurranttw, line 15: If we pass the onChange to the CheckboxButton we don't need to fake the behaviour using the onClick

packages/css/src/less/forms/bootstrap-forms.less
```diff
@@ -0,0 +1,1053 @@
+@import (reference) '../mixins/_arrows.less';
```
helendurranttw, line 1: For the most part, this is a straight copy from Bootstrap (thus the name). I did remove vendor prefix mixins (we use autoprefixer). I'll call out the other changes I made.

packages/css/src/less/forms/bootstrap-forms.less
```diff
@@ -0,0 +1,1053 @@
...
+
+.checkbox,
+.radio {
+  > label {
+    font-size: @font-size-base;
+    display: block;
+    cursor: pointer;
+    line-height: 24px;
+    padding-top: 12px;
+    padding-bottom: 10px;
```
helendurranttw, line 906: I update these to adjust for the 1px misalignment of our font. These were causing our checkboxes / radios to look vertically off by 1px.

packages/css/src/less/forms/checkbox-radio.less
```diff
@@ -0,0 +1,243 @@
+@import (reference) '../../variables/legacy-variables.less';
```
helendurranttw, line 1: Imported from Bootstrap. Mostly unchanged, with a few additions and tweaks.

Where I edited styles I changed them to use our semantic colours. I didn't update the parts I didn't need to touch.

packages/css/src/less/forms/checkbox-radio.less
```diff
@@ -0,0 +1,243 @@
...
+  border: 1px solid @color-base-smoke-mid;
+  background: var(--color-background-screen);
+  width: 22px;
+  height: 22px;
+  border-radius: var(--radius-small);
+  display: inline-block;
+  padding: 0;
+  vertical-align: middle;
+  text-align: center;
+  transition: border 0.15s ease-in-out, background-color 0.15s ease-in-out;
```
helendurranttw, line 43: Added background-colour transition

packages/css/src/less/forms/checkbox-radio.less
```diff
@@ -0,0 +1,243 @@
...
+    width: 20px;
+    height: 20px;
+    line-height: 18px;
+    margin: 0;
+    display: none;
+    color: var(--color-background-screen);
+  }
+
+  &:hover,
+  input[type="checkbox"]:hover + &,
```
helendurranttw, line 55: Added selectors in several places to capture interaction with the new fake input.

packages/css/src/less/forms/checkbox-radio.less
```diff
@@ -0,0 +1,243 @@
...
+    background-color: var(--color-control-accent);
+
+    .tw-checkbox-check {
+      display: inline-block;
+    }
+  }
+
+  &[checked]:hover,
+  &.checked:hover,
+  input[type="checkbox"]:not(:disabled):checked:hover + & {
```
helendurranttw, line 85: Added new styles for checked + hover and checked + active. Previously the checkbox was always the same colour when checked, regardless of state.

packages/css/src/less/forms/checkbox-radio.less
```diff
@@ -0,0 +1,243 @@
...
+    &.checked {
+      background-color: var(--color-control-negative);
+    }
+
+    &[disabled] {
+      background-color: @color-base-smoke-fade;
+    }
+  }
+
+  .checkbox.has-error input[type="checkbox"]:checked + & {
```
helendurranttw, line 122: Added styles for negative + interaction states.

packages/css/src/less/grid.less
```diff
@@ -1,20 +1,141 @@
-// variables
-@import (reference) 'bootstrap/less/variables.less';
-
-// mixins
-@import (reference) 'bootstrap/less/mixins/grid.less';
-@import (reference) 'bootstrap/less/mixins/grid-framework.less';
-@import (reference) 'bootstrap/less/mixins/clearfix.less';
+@import (reference) '../variables/legacy-variables.less';
 @import (reference) './mixins/_grid.less';
-@import 'bootstrap/less/grid.less';
```
helendurranttw, line 9: Inlined grid.less from Bootstrap. We were both importing it here but also copying some of the mixins from that file directly into this file.

packages/css/src/less/forms/checkbox-radio.less
nabaroa: I understand to take the styles as they are from Bootstrap. As you mentioned, would be nice to clean them first but wasn't possible time wise. Would be nice to open tickets to work on the syntax consistency for those styles. I'd be open to work on that, feel free to assign them to me.

nabaroa:
👏 
LGTM. Taking into account that forms are large and complex styles files, makes sense to me to take them into Neptune as they are and plan to 'clean' them in next iterations (same with grid). Happy to have less and less dependencies from Bootstrap. Nice change.

nabaroa (approved):

packages/components/src/checkboxButton/CheckboxButton.less
```diff
@@ -0,0 +1,31 @@
+@import (reference) '@transferwise/neptune-css/src/variables/neptune-tokens.less';
+
+.np-checkbox-button {
+  input[type="checkbox"] {
+    // Place the actual input on top of the SVG, so people using touch to find elements will still find it
+    position: absolute;
+    width: 24px;
+    height: 24px;
```
anttio, line 8:
```suggestion
    width: var(--size-24);
    height: var(--size-24);
```

packages/components/src/checkboxButton/CheckboxButton.less
```diff
@@ -0,0 +1,31 @@
...
+.checkbox .np-checkbox-button {
+  position: absolute;
+  left: 14px;
+  top: 12px;
+
+  input[type="checkbox"], .tw-checkbox-button {
+    left: 0px;
+    top: 0px;
+  }
+}
```
anttio, line 31: We don't have any rules on nesting selectors but just for a thought should we do this:

```
.checkbox {
  .np-checkbox-button {
    ...
  }
}
```

packages/components/src/checkboxButton/CheckboxButton.less
```diff
@@ -0,0 +1,31 @@
...
+
+.checkbox .np-checkbox-button {
+  position: absolute;
+  left: 14px;
+  top: 12px;
+
+  input[type="checkbox"], .tw-checkbox-button {
+    left: 0px;
+    top: 0px;
```
anttio, line 29:
```suggestion
    left: 0;
    top: 0;
```

anttio:
image
When you change the font-size on browser the tick alignment breaks.

anttio:
image
Didn't check but is this our disabled style for secondary content?

anttio (approved):
Left few small comments. Otherwise looks good.

packages/components/src/checkbox/Checkbox.spec.js
```diff
@@ -1,116 +1,125 @@
...
+  it('calls change handler with new checked value when checkbox button is clicked', () => {
     const onChange = jest.fn();
-    component.setProps({ onChange });
+
+    const { container, rerender } = render(<Checkbox {...props} onChange={onChange} />);
+    let checkbox = container.querySelector('label');
 
     expect(onChange).not.toBeCalled();
-    clickCheckboxButton();
+    fireEvent(checkbox, new MouseEvent('click'));
```
andreapiras, line 40: Can you use `fireEvent.click(checkbox )` ?

packages/components/src/checkbox/Checkbox.spec.js
```diff
@@ -1,116 +1,125 @@
...
+    expect(container.querySelector('small')).not.toBeInTheDocument();
   });
 
-  it('calls change handler with new checked value when not disabled and checkbox button is clicked', () => {
+  it('calls change handler with new checked value when checkbox button is clicked', () => {
     const onChange = jest.fn();
-    component.setProps({ onChange });
+
+    const { container, rerender } = render(<Checkbox {...props} onChange={onChange} />);
+    let checkbox = container.querySelector('label');
```
andreapiras, line 37: can we use getByLabelText(/hello/i) here?

packages/components/src/checkbox/Checkbox.spec.js
andreapiras, line 40:
```suggestion
    fireEvent.click(checkbox )
```

packages/components/src/checkbox/Checkbox.spec.js
```diff
@@ -1,116 +1,125 @@
...
-    clickCheckboxButton();
+    fireEvent(checkbox, new MouseEvent('click'));
     expect(onChange).toBeCalledWith(true);
 
-    component.setProps({ checked: true });
-    clickCheckboxButton();
+    rerender(<Checkbox {...props} onChange={onChange} checked />);
+    checkbox = container.querySelector('label');
+
+    fireEvent(checkbox, new MouseEvent('click'));
```
andreapiras, line 46:
```suggestion
   fireEvent.click(checkbox )
```

packages/components/src/checkbox/Checkbox.spec.js
```diff
@@ -1,116 +1,125 @@
...
 
     expect(onChange).not.toBeCalled();
-    clickCheckboxButton();
+    fireEvent(checkbox, new MouseEvent('click'));
     expect(onChange).toBeCalledWith(true);
 
-    component.setProps({ checked: true });
-    clickCheckboxButton();
+    rerender(<Checkbox {...props} onChange={onChange} checked />);
+    checkbox = container.querySelector('label');
```
andreapiras, line 44: if you select the checkbox the events fires
```const checkbox = getByLabelText('hello');
    fireEvent.click(checkbox);
    expect(onChange).not.toBeCalled();```

Comment (andreapiras) [reply packages/components/src/checkbox/Checkbox.spec.js:44]: Expected number of calls: 0
    Received number of calls: 1

File: packages/components/src/checkbox/Checkbox.spec.js [hunk 20]
```diff
@@ -1,116 +1,125 @@
...
   });
 
   it('does not call change handler on checkbox button click when disabled', () => {
     const onChange = jest.fn();
-    component.setProps({ onChange, disabled: true });
 
-    expect(onChange).not.toBeCalled();
-    clickCheckboxButton();
+    const { container } = render(<Checkbox {...props} onChange={onChange} disabled />);
+    const checkbox = container.querySelector('label');
```
andreapiras, line 54: this is returning the label that contains the checkbox so we should change the name or the selector

packages/components/src/checkbox/Checkbox.spec.js
```diff
@@ -1,116 +1,125 @@
...
+
+    fireEvent(checkbox, new MouseEvent('click'));
```
andreapiras, line 56:
```suggestion
    fireEvent.click(checkbox )
```

packages/components/src/checkbox/Checkbox.spec.js
```diff
@@ -1,116 +1,125 @@
...
   it('does not call change handler on checkbox button click when readOnly', () => {
     const onChange = jest.fn();
-    component.setProps({ onChange, readOnly: true });
 
-    expect(onChange).not.toBeCalled();
-    clickCheckboxButton();
+    const { container } = render(<Checkbox {...props} onChange={onChange} readOnly />);
+    const checkbox = container.querySelector('label');
+
+    fireEvent(checkbox, new MouseEvent('click'));
```
andreapiras, line 66:
```suggestion
    fireEvent.click(checkbox )
```

packages/components/src/checkbox/Checkbox.spec.js
```diff
@@ -1,116 +1,125 @@
...
     expect(onChange).not.toBeCalled();
   });
 
-  it('has disabled class when the flag is passed', () => {
-    expect(isDisabled()).toBe(false);
+  it('has disabled class when the disabled prop is true', () => {
+    const { container } = render(<Checkbox {...props} disabled />);
+    expect(container.querySelector('.disabled')).toBeInTheDocument();
```
andreapiras, line 72: this test tells that a class is in the DOM but there's no actually any info that is in the right place or that is just one for example. I think a snapshot or this would have more infos
```
expect(container.querySelector('.checkbox')).toHaveClass('disabled');
```

packages/components/src/checkbox/Checkbox.spec.js
```diff
@@ -1,116 +1,125 @@
...
   it('renders the given label', () => {
-    expect(component.containsMatchingElement(<b>hello</b>)).toBe(true);
+    const { getByText } = render(<Checkbox {...props} />);
+    getByText('hello');
+  });
+
+  it('is enabled by default', () => {
+    const { container } = render(<Checkbox {...props} />);
+
+    expect(container.querySelector('.disabled')).not.toBeInTheDocument();
```
andreapiras, line 25: This will pass also if the element for some reason doesn't get rendered. I think a snapshot or this would have more infos
```expect(container.querySelector('.checkbox')).toHaveClass('disabled');```

File: packages/components/src/checkbox/Checkbox.spec.js [hunk 25]
```diff
@@ -1,116 +1,125 @@
...
   });
 
   it('does not render secondary text', () => {
-    expect(component.find('small')).toHaveLength(0);
+    const { container } = render(<Checkbox {...props} />);
+    expect(container.querySelector('small')).not.toBeInTheDocument();
```
andreapiras, line 30: Same as above

packages/components/src/checkbox/Checkbox.spec.js
```diff
@@ -1,116 +1,125 @@
...
-  it('has disabled class when the flag is passed', () => {
-    expect(isDisabled()).toBe(false);
+  it('has disabled class when the disabled prop is true', () => {
+    const { container } = render(<Checkbox {...props} disabled />);
+    expect(container.querySelector('.disabled')).toBeInTheDocument();
+  });
+
+  it('has error class and passes it to checkbox button when required and not checked', () => {
+    const { container, rerender } = render(<Checkbox {...props} required checked />);
+    expect(container.querySelector('.has-error')).not.toBeInTheDocument();
```
andreapiras, line 77: Should we check that a specific element does not have a class?

packages/components/src/checkbox/Checkbox.spec.js
```diff
@@ -1,116 +1,125 @@
...
+
+    rerender(<Checkbox {...props} required readOnly />);
+    expect(container.querySelector('.has-error')).not.toBeInTheDocument();
   });
 
   it('has an asterisk after the label when required', () => {
-    component.setProps({ required: true });
-    expect(component.text()).toContain('*');
+    const { getByText } = render(<Checkbox {...props} required />);
+    getByText('*');
```
andreapiras, line 93: This is asserting that an * but doesn't have any infos on where it is( `next to the label`)

packages/components/src/checkboxButton/CheckboxButton.js
andreapiras: Can we map them in the css ? 
```
.np-checkbox-button {
  .tw-checkbox-button
}
```
to not incur in any other add blocker weirdness? Or just to keep it consistent?

packages/components/src/checkboxButton/CheckboxButton.spec.js
```diff
@@ -1,62 +1,48 @@
...
-    it('has checked class', () => {
-      expect(hasCheckedClass()).toBe(true);
-    });
-
-    it('has aria-pressed as true', () => {
-      expect(hasAriaPressedAsTrue()).toBe(true);
-    });
+  it('applies aria-label if provided', () => {
+    ({ getByRole } = render(<CheckboxButton onChange={jest.fn()} aria-label="An aria label" />));
+    getByRole('checkbox', { name: 'An aria label' });
```
andreapiras, line 25: What's the benefit of this vs getByAriaLabelText?

packages/components/src/checkboxOption/CheckboxOption.spec.js
```diff
@@ -1,75 +1,69 @@
...
+    it('is not called if the option is disabled', () => {
+      const onChange = jest.fn();
+      ({ container } = render(<CheckboxOption {...defaultProps} onChange={onChange} disabled />));
 
-  const option = () => component.find(Option);
-  const buttonProp = (name) => option().prop('button').props[name];
+      const label = container.querySelector('label');
+
+      expect(onChange).not.toBeCalled();
+      fireEvent(label, new MouseEvent('click'));
```
andreapiras, line 65: fireEvent.click

packages/css/src/less/forms/checkbox-radio.less
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  .checkbox.has-error input[type="checkbox"]:checked:active + & {
+    border-color: var(--color-control-negative-active) !important;
+    background-color: var(--color-control-negative-active);
+  }
+}
+
+.tw-radio-button {
+  width: 24px;
```
andreapiras, line 139: --size-24

packages/components/src/checkboxButton/CheckboxButton.js
helendurranttw: I would argue that using the existing styling is the most effective way to ensure consistency :) Also doing that would unnecessarily double the CSS, and as I've mentioned before, it's not just in one place that you have to change it: there is also styling in several places for interactions with the checkbox (i.e. the checkbox wrapper, horizontally aligned forms with checkboxes in them etc).

packages/components/src/checkboxButton/CheckboxButton.less
helendurranttw: We could definitely make that a stylelint rule!

packages/components/src/checkbox/Checkbox.js
```diff
@@ -38,10 +36,9 @@ const Checkbox = ({
           className={classNames({ 'has-error': hasError })}
           checked={checked}
           onFocus={onFocus}
-          onClick={handleClick}
+          onChange={() => onChange(!checked)}
           onBlur={onBlur}
-          disabled={disabled}
-          readOnly={readOnly}
+          disabled={disabled || readOnly}
```
dozortsev, line 41: are you sure about this change? 🙂 there is quite a bit of difference between them https://stackoverflow.com/a/7730719/2650174, I would still allow it be able to be only `readOnly`

packages/components/src/checkbox/Checkbox.spec.js
```diff
@@ -1,116 +1,125 @@
 import React from 'react';
-import { shallow } from 'enzyme';
+import '@testing-library/jest-dom';
+import { render, fireEvent } from '@testing-library/react';
```
dozortsev, line 3: I know that this component doesn't have any i18n but for consistency please import custom `render`

example https://github.com/transferwise/neptune-web/blob/main/packages/components/src/alert/Alert.spec.js#L11

packages/components/src/checkbox/Checkbox.spec.js
```diff
@@ -1,116 +1,125 @@
...
       onChange: jest.fn(),
-      secondary: '',
     };
-    component = shallow(<Checkbox {...props} />);
   });
 
   it('renders the given label', () => {
-    expect(component.containsMatchingElement(<b>hello</b>)).toBe(true);
+    const { getByText } = render(<Checkbox {...props} />);
+    getByText('hello');
```
dozortsev, line 19: this test doesn't have any `expect(..)` so it doesn't quite do any job

packages/components/src/checkboxButton/CheckboxButton.spec.js
```diff
@@ -1,62 +1,48 @@
 import React from 'react';
-import { shallow } from 'enzyme';
+import '@testing-library/jest-dom';
+import { render, fireEvent } from '@testing-library/react';
```
dozortsev, line 3:
```suggestion
import { render, fireEvent } from '../test-utils';
```

packages/components/src/checkboxButton/CheckboxButton.story.js
```diff
@@ -0,0 +1,30 @@
...
+import { boolean } from '@storybook/addon-knobs';
+import CheckboxButton from './CheckboxButton';
+
+export default {
+  component: CheckboxButton,
+  title: 'CheckboxButton',
+};
+
+export const basic = () => {
+  const [checked, setChecked] = useState(true);
```
dozortsev, line 11: I see why you make this change but IMHO on Storybook env it's better to manipulate component via knobs rather then interact with it directly

packages/components/src/checkboxOption/CheckboxOption.js
```diff
@@ -10,27 +10,23 @@ const CheckboxOption = ({ media, title, content, name, checked, onChange, comple
...
-  id: Types.string.isRequired,
-  name: Types.string.isRequired,
-  title: Types.node.isRequired,
-  content: Types.node,
   checked: Types.bool,
-  onChange: Types.func.isRequired,
   complex: Types.bool,
+  content: Types.node,
   disabled: Types.bool,
+  id: Types.string.isRequired,
```
dozortsev, line 25: I would make this prop non required

packages/css/src/less/forms.less
```diff
@@ -1,15 +1,6 @@
...
-
-// mixins
-@import (reference) 'bootstrap/less/mixins/vendor-prefixes.less';
+@import (reference) '../variables/legacy-variables.less';
 @import (reference) './mixins/_forms.less';
-@import (reference) 'bootstrap/less/mixins/arrows.less';
-@import (reference) 'bootstrap/less/mixins/grid.less';
-@import (reference) 'bootstrap/less/mixins/tab-focus.less';
-@import 'bootstrap/less/forms.less';
+@import './forms/bootstrap-forms.less';
```
dozortsev, line 3: :feels_good:

packages/components/src/checkboxButton/CheckboxButton.js
```diff
@@ -2,38 +2,31 @@ import React from 'react';
...
+    <input type="checkbox" disabled={disabled} checked={checked} onChange={onChange} {...rest} />
+    <span className="tw-checkbox-button">
+      <span className="tw-checkbox-check" />
+    </span>
+  </span>
 );
 
 CheckboxButton.propTypes = {
+  /* Provide this if the CheckboxButton is being used without a label */
+  'aria-label': Types.string,
```
dozortsev, line 18: shouldn't we mark it as required prop for this component?

packages/components/src/checkboxOption/CheckboxOption.js
```diff
@@ -10,27 +10,23 @@ const CheckboxOption = ({ media, title, content, name, checked, onChange, comple
...
+  media: Types.node,
+  name: Types.string.isRequired,
+  onChange: Types.func.isRequired,
+  title: Types.node.isRequired,
```
dozortsev, line 29: what about `aria-label` for this component?

it could be either `aria-label` prop or pass `title` prop as a default value there `aira-label={title}`

packages/components/src/checkboxButton/CheckboxButton.js
helendurranttw: There are subtle differences between [click and change](https://stackoverflow.com/questions/58016503/click-vs-input-vs-change-for-checkboxes#:~:text=change%20happens%20once%20the%20value,and%20the%20element%20loses%20focus.&text=click%20happens%20once%20the%20element,a%20mouse%20or%20keyboard%20event.), they aren't interchangeable :)

packages/components/src/checkbox/Checkbox.spec.js
helendurranttw: getByText has a built in throw if it can't find the element it's looking for :) Though Andrea suggested I change it anyway, so that the error messages are consistent.

packages/components/src/checkbox/Checkbox.js
helendurranttw: I am, yes. checkboxes can't be set to readOnly, they only support disabled. If you look at the old code for CheckboxButton, it did this same conversion in that file :)

packages/components/src/checkbox/Checkbox.story.js
```diff
@@ -9,20 +8,22 @@ export default {
 };
 
 export const basic = () => {
-  const checked = boolean('checked', false);
+  const [checked, setChecked] = useState(true);
```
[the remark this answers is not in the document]
helendurranttw: I wanted to be able to demonstrate clicking the checkbox

packages/components/src/checkboxButton/CheckboxButton.js
```diff
@@ -2,38 +2,31 @@ import React from 'react';
...
-    onBlur={onBlur}
-    disabled={disabled || readOnly}
-  >
-    <span className="tw-checkbox-check" />
-  </button>
+import './CheckboxButton.css';
+
+const CheckboxButton = ({ checked, className, disabled, onChange, ...rest }) => (
+  <span className={classNames(className, 'np-checkbox-button')}>
+    <input type="checkbox" disabled={disabled} checked={checked} onChange={onChange} {...rest} />
```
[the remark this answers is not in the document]
helendurranttw: I figure it's fine to use it for any native attrs

packages/components/src/checkboxButton/CheckboxButton.js
helendurranttw: We can't really, because if it's in a Checkbox component it's wrapped in a label, so the aria-label isn't required.

packages/components/src/checkboxButton/CheckboxButton.spec.js
helendurranttw: I don't see getByAriaLabelText as an available method 🤔

packages/components/src/checkboxButton/CheckboxButton.story.js
helendurranttw: I don't agree, I think it's important to be able to test it the way the user interacts with it (and how our consumers will set it up)

packages/components/src/checkboxOption/CheckboxOption.js
helendurranttw: We don't need it because there is a label :)

packages/css/src/less/forms/checkbox-radio.less
helendurranttw: These are unchanged from Bootstrap - I'm not going to touch them in this PR

packages/components/src/checkbox/Checkbox.spec.js
helendurranttw: We hadn't merged main into this PR when I wrote this, but I've done that now :)

PR #726, opened 2020-12-10
docs: store scroll position on the docs

helendurranttw:
[Test the fix here](https://transferwise.github.io/neptune-web/branch/scroll-to-docs/components/inputs/Button).

## 🖼 Context

When you're using Neptune Web docs, every time you click on the name of a component, the page refreshes and the list gets scrolled to the top. It's super annoying.

## 🚀 Changes

Used a little trick I found online to push the scroll position of the nav into local storage before the page unloads, and read it in again when we're loading the nav.

## 🤔 Considerations

Ok soooooo _strictly speaking_ there's a bit of a bug in the code, but it's harmless at the moment so I'm not going to bother fixing it right now.

The bug: really, we only want the push / read of the scroll top to happen when we're in the _components_ list, but right now, the code will just do it for whatever page it's on.

This is harmless, because all the other sub-navs only have a few items - so if you apply a scroll top on there it won't do anything. Similarly, if one of their positions gets pushed and then read into the components list, it will always be 0, so it's not an issue. In fact this makes sense, because if you navigate away from the components list we want it to be reset to 0. Excellent - it's not a bug, it's a feature!

This will need revisiting in the (currently unlikely) event that one or more of the sub-navs on other pages grows longer than the bottom of the screen.

The reason this is a pain to fix is that we have to inspect the routes every time we change page and / or store a separate item in local storage for each nav on each page, and frankly I have better things to do right now.

## ✅ Checklist


packages/docs/components/Sidebar.js
```diff
@@ -37,13 +37,18 @@ const getLinks = ({ pathname, section }) => {
 };
 
 const Sidebar = ({ router: { pathname }, section }) => {
-  const [links, updateLinks] = useState([]);
-  const [filteredLinks, updateFilteredLinks] = useState('');
+  const [links, updateLinks] = useState(getLinks({ pathname, section }));
+  const [filteredLinks, updateFilteredLinks] = useState(links.map(({ content }) => content));
```
helendurranttw, line 41: This will ensure that when SSG happens there will be links there - rather than an empty box.

packages/docs/pages/_app.js
```diff
@@ -26,6 +26,13 @@ class MyApp extends App {
     if (pathname === '/' || pathname === '/_error') {
       Router.push(addBasePath('about/Home'));
     }
+
+    window.addEventListener('beforeunload', () => {
+      localStorage.setItem(
```
helendurranttw, line 31: Before navigating away from the page, get the scroll top and store it in local storage.

packages/docs/components/Sidebar.js
```diff
@@ -37,13 +37,18 @@ const getLinks = ({ pathname, section }) => {
...
 
   const [searchInput, updateSearchInput] = useState('');
   const searchEl = useRef(null);
+  const scrollableNavEl = useRef(null);
 
   useEffect(() => {
+    const top = localStorage.getItem('sidebar-scroll');
```
helendurranttw, line 48: When loading the nav list, check local storage to see if we put a scroll top in there.

andreapiras:
Maybe I m missing something but why are we not using the url here. If we set up things on user browser do we have to notify the users to be GDPR compliant?

dozortsev:
@andreapiras I think in this use case it's fine https://law.stackexchange.com/questions/30739/do-the-gdpr-and-cookie-law-regulations-apply-to-localstorage

dozortsev (approved):
As a quick temporal fix I think it's fine 👌

PR #737, opened 2020-12-15
feat!: introducing semantic styles for buttons

helendurranttw:
## 🖼 Context

[Link to demo](https://transferwise.github.io/neptune-web/branch/button-negative-changes/components/inputs/Button).
[Link to social media button docs](https://transferwise.github.io/neptune-web/branch/button-negative-changes/styles/extras/SocialMedia).

We are updating buttons to use semantic colours. We are also introducing the concept of priorities, allowing people to render filled or outlined buttons as they please. There are 3 types (accent, positive and negative), 3 priorities (primary, secondary, tertiary) and 4 states (normal, hover, active, disabled).

## 🚀 Changes

* Added the new types and a priority prop
* Added mappings for old types to new type / priority combinations, such that old buttons will continue to work
* Added deprecation notices for people using old types
* Added an upgrade tool for consumers to easily update prop names
* **BREAKING CHANGE**: Removed the `.btn-facebook` and `.btn-google` classes from `neptune.css` and added them to a new optional file: `neptune-social-media.css`
* **BREAKING CHANGE**: Removed the `.btn-download` class from `neptune.css`

## 🤔 Considerations

Existing Button instances should continue to work without changes, although you will see deprecation notices being logged. The only breaking changes in this PR are the removal of the download class and moving the social media classes.

## ✅ Checklist


packages/components/src/button/Button.js
```diff
@@ -6,38 +6,78 @@ import { logActionRequiredIf } from '../utilities';
...
+const deprecatedTypeMap = {
+  [Type.PRIMARY]: ControlType.ACCENT,
+  [Type.SECONDARY]: ControlType.ACCENT,
+  [Type.LINK]: ControlType.ACCENT,
+  [Type.PAY]: ControlType.POSITIVE,
+  [Type.DANGER]: ControlType.NEGATIVE,
+};
+
+const typeClassMap = {
+  [ControlType.ACCENT]: 'btn-accent',
```
andreapiras, line 20: are we not using np- prefix anymore?

packages/components/src/button/Button.js
```diff
@@ -6,38 +6,78 @@ import { logActionRequiredIf } from '../utilities';
...
+  [Priority.TERTIARY]: 'btn-priority-3',
+};
+
+const oldTypePriorityMap = {
+  [Type.DANGER]: Priority.SECONDARY,
+  [Type.LINK]: Priority.TERTIARY,
+  [Type.SECONDARY]: Priority.SECONDARY,
+};
+
+const establishPriority = ({ rawPriority, rawType, type }) => {
```
andreapiras, line 37: Do we need to keep this inside this file? Would it be better to move it to maybe a util folder with tests to keep this file cleaner?

packages/components/src/button/Button.story.js
```diff
@@ -30,3 +30,66 @@ export const basic = () => {
     </Button>
   );
 };
+
+export const vairants = () => {
```
andreapiras, line 34: typo

packages/components/src/button/Button.story.js
```diff
@@ -30,3 +30,66 @@ export const basic = () => {
...
+        <Button priority={Button.Priority.SECONDARY} type={Button.Type.POSITIVE} {...commonProps} />
+      </div>
+      <div className="m-b-2">
+        <div className="h4 m-b-1">Negative</div>
+        <Button priority={Button.Priority.PRIMARY} type={Button.Type.NEGATIVE} {...commonProps} />
+        <Button priority={Button.Priority.SECONDARY} type={Button.Type.NEGATIVE} {...commonProps} />
+      </div>
+
+      <div className="m-b-2 m-t-4">
+        <div className="h3 m-b-1">DEPRECATED</div>
```
andreapiras, line 73: Maybe we can separate these under a deprecated subsection

packages/components/src/button/Button.js
```diff
@@ -6,38 +6,78 @@ import { logActionRequiredIf } from '../utilities';
...
+    priority: rawPriority,
+    size,
+    type: rawType,
+    ...rest
+  } = props;
+
+  logActionRequired(props);
+
+  const type = deprecatedTypeMap[rawType] || rawType;
+  const priority = establishPriority({ type, rawPriority, rawType });
```
andreapiras, line 67: I think changing this to 
```
const {
    block,
    children,
    className,
    disabled,
    htmlType,
    loading,
    priority,
    size,
    type,
    ...rest
  } = props;
const newPriority = establishPriority({ type, priority, type });
```

would be more readable but it's just a personal preference

packages/components/src/button/Button.js
```diff
@@ -47,7 +87,29 @@ const Button = ({
   );
 };
 
-Button.Type = Type;
+const deprecatedTypeMapMessage = {
```
andreapiras, line 90: Can we have these deprecation mappings in a deprecation.js maybe?

packages/components/src/button/Button.spec.js
```diff
@@ -1,73 +1,176 @@
...
+      expect(screen.getByRole('button')).toBeDisabled();
+    });
+
+    it('sets the htmlType if set', () => {
+      render(<Button {...props} htmlType="submit" />);
+      expect(screen.getByRole('button')).toHaveAttribute('type', 'submit');
+    });
+
+    it('passes through custom classes if set', () => {
+      render(<Button {...props} className="catsarethebest" />);
```
andreapiras, line 65: They are.

packages/components/src/button/Button.spec.js
```diff
@@ -1,73 +1,176 @@
...
+    });
   });
 
-  it('calls click handler on button click if not disabled', () => {
-    expect(props.onClick).not.toBeCalled();
-    wrapper.simulate('click');
-    expect(props.onClick).toBeCalled();
+  describe('alternative states', () => {
+    it('renders as loading if loading is true', () => {
+      expect(render(<Button {...props} loading />).container).toMatchSnapshot();
```
andreapiras, line 148: shall we actually check for the span here?

packages/components/src/button/Button.spec.js
```diff
@@ -1,73 +1,176 @@
...
   });
 
-  it('adds passed class name to existing class names', () => {
-    wrapper.setProps({ className: 'm-t-2' });
-    expect(wrapper.find('button').hasClass('btn')).toBe(true);
-    expect(wrapper.find('button').hasClass('btn-md')).toBe(true);
-    expect(wrapper.find('button').hasClass('m-t-2')).toBe(true);
+  describe('deprecated types', () => {
+    it('renders primary as accent buttons and logs a warning ', () => {
+      expect(render(<Button {...props} type={Button.Type.PRIMARY} />).container).toMatchSnapshot();
```
andreapiras, line 158: I'd separate m `establishPriority` and test that individually

packages/components/src/button/Button.js
```diff
@@ -6,38 +6,78 @@ import { logActionRequiredIf } from '../utilities';
...
-  disabled,
-  htmlType,
-  loading,
-  size,
-  type,
-  ...rest
-}) => {
+import { Size, ControlType, Priority, Type } from '../common';
+
+const deprecatedTypeMap = {
```
andreapiras, line 11: Can we group all these deprecation object/array in a deprecation.js

packages/components/src/button/Button.js
helendurranttw: For new things, yes, but these are existing neptune CSS styles that we're not changing at the moment.

packages/components/src/button/Button.js
helendurranttw: Nice idea, I've moved out all the stuff relating to legacy things.

packages/components/src/button/Button.js
helendurranttw: I'm not sure the benefits would outweigh having to go and look it up when you're reading through the file and trying to understand the logic.

packages/components/src/button/Button.js
helendurranttw: Fair call, I'll do that!

packages/components/src/button/Button.js
helendurranttw: I've moved them into the legacyUtils file

andreapiras (approved):

PR #743, opened 2020-12-17
feat: ADS-1785 Add support for tabs

felixxiatransferwise:
## 🚀 Changes

 
Adds support for tabs as control.

## ✅ Checklist


packages/dynamic-flows/src/jsonSchemaForm/schemaFormControl/SchemaFormControl.js
```diff
@@ -4,10 +4,13 @@ import Types from 'prop-types';
 import { isNull, isUndefined } from '@transferwise/neptune-validation';
 import FormControl from '../../formControl';
 import { getValidModelParts } from '../../common/validation/valid-model';
+import { isOneOfObjectsSchema } from '../../common/schemaTypes/schemaTypes';
```
StevePole, line 7: Can we just use isOneOfSchema ?

I don't really like us introspecting the sub schema type and would prefer to avoid it.  Based on the way you use this, I don't think you need to check.  I can see a case for saying _never_ use tabs unless it's an oneOf of objects, but that's not what we're doing.

packages/dynamic-flows/src/jsonSchemaForm/schemaFormControl/optionMapper.js
```diff
@@ -1,10 +1,10 @@
 import { isNull, isUndefined } from '@transferwise/neptune-validation';
 import { getCurrencyFlag } from './availableCurrencyFlags';
 
-export const mapConstSchemaToOption = (schema) => {
+export const mapConstSchemaToOption = (schema, index) => {
   return {
     label: schema.title,
-    value: !isUndefined(schema.const) ? schema.const : schema.enum[0],
+    value: !isUndefined(schema.const) ? schema.const : index,
```
StevePole, line 7: Can you explain this change?  Looks like we're using a function called `mapConstSchemaToOption` for something that isn't a const schema?

felixxiatransferwise: `schema.enum[0]` will actually cause a `TypeError: Cannot read property '0' of undefined`, this is because `SchemaFormControl` will map any `oneOf` to "options"

```
   if (schema.oneOf) {
      return schema.oneOf.map(mapConstSchemaToOption);
    }
```

oneOf objects just become options with their indices as values.

Also, using `schema.enum[0]` for all of the option values when it's not a const doesn't make much sense?

packages/dynamic-flows/src/jsonSchemaForm/schemaFormControl/SchemaFormControl.js
felixxiatransferwise: I guess we can use `isOneOfSchema` but I wanted to be specific here and say that don't do this for oneOf consts, tabs certainly wouldn't make sense for that.

> I don't really like us introspecting the sub schema type
I agree, but this is currently the only way for us to differentiate oneOf consts vs oneOf schemas, perhaps this is hinting we should be more declarative in our oneOf contract?

StevePole: The broader point I was making is that you have no need to be specific here.  Your logic says if it's an object schema with >3 choices, force using the select.  But if it's not an object, we would never want to render tabs, because the content would be empty.  So you don't need to check whether it's an object schema, only that it's a oneOf.

packages/dynamic-flows/src/jsonSchemaForm/schemaFormControl/optionMapper.js
StevePole:
```
{
  enum: [1]
}
```
is a valid 'const' schema.  'const' is syntactic sugar for it.  This method is only supposed to run on const schemas, hence the name.

packages/dynamic-flows/src/jsonSchemaForm/schemaFormControl/SchemaFormControl.js
felixxiatransferwise: Isn't that assuming we won't get more types for oneOf consts later?

You are suggesting we can do:
```
if (isOneOfSchema(schema) && schema.oneOf.length > 3) {
        return FormControlType.SELECT;
      }
```
but if later, we introduce let's say `control: radio2`, this will evaluate true if it has >3 options and we would return a 'select' when we actually might not want to alter the control for this oneOf consts from `radio2`

packages/dynamic-flows/src/jsonSchemaForm/schemaFormControl/optionMapper.js
felixxiatransferwise: > is a valid 'const' schema

Sure but why was it using the first index for all options? 
If you had 
```
{
  "enum": ["red", "amber", "green", null, 42]
}
```
wouldn't that map "red" to all options?

packages/dynamic-flows/src/common/schemaTypes/schemaTypes.js
```diff
@@ -8,17 +8,22 @@ export const schemaType = {
...
+export const isOneOfSchema = (schema) => !!schema.oneOf;
 
-const isAllOfSchema = (schema) => !!schema.allOf;
+export const isAllOfSchema = (schema) => !!schema.allOf;
 
-const isBasicSchema = (schema) =>
+export const isBasicSchema = (schema) =>
   basicTypes.indexOf(schema.type) >= 0 || !!schema.enum || !!schema.const;
 
+export const isOneOfObjectsSchema = (schema) =>
```
StevePole, line 22: Can we remove this now?

packages/dynamic-flows/src/jsonSchemaForm/schemaFormControl/SchemaFormControl.js
StevePole: that's true, and something we probably should address, but I don't think what you had handles that better.  No reason you couldn't have a oneOf of objects represented by radio2.

StevePole (approved):

PR #749, opened 2020-12-22
feat!: I18n follow up

dozortsev:
## 🖼 Context

Follow up of [i18n works for neptune components](https://github.com/transferwise/neptune-web/pull/674).

## 🚀 Changes

Fixed nasty [bug](https://transferwise.slack.com/archives/C2SGDKP4Y/p1611144292068900?thread_ts=1611139542.053500&cid=C2SGDKP4Y).

---

Added clear separated support for locales and languages for `Provider` component

```jsx
<Provider i18n={
    locale: // locale codes e.g `es-ES`, `en-GB`, `en-US`, `en`, `zh-HK` etc
    messages: `../../i18n/{lang}.json` where lang is language codes: `en`, `de`, `jp`, `zh`
}>..
```

See updated docs

https://transferwise.github.io/neptune-web/branch/i18n-follow-up/about/Setup#usage

and play with component here (to see translations based on lang and different formatting of date, amount etc. based on locale value)

https://transferwise.github.io/neptune-web/branch/i18n-follow-up/components/context/Provider

---

Refactored some components to use `locale` value from global context (its newly single source of truth) rather than from props.

List of affected components:

- `Money`
- `MoneyInput`
- `PhoneNumberInput`
- `FormattedValue`
- `DynamicFieldDefinitionList`
- `DateLookup` and it's sub components (YearCalendarTable, YearCalendar, OpenButton, MonthCalendarTable, MonthCalendar, DayCalendarTable, DayCalendar)
- `DateInput` 
- `Upload`
---

In react-intl there are two ways to get reference on its context inside components, either via `useIntl` hook (generally preferred way) or `injectIntl`.

Worth mention their difference (should clarify some stuff during code review), quote from [docs](https://formatjs.io/docs/react-intl/api/#the-intl-object):

> useIntl hook: Once you've declared your IntlProvider, you can get access to the intl object via calling this hook in your functional React component
> injectIntl HOC: In class-based React components, you can wrap them with the injectIntl HOC and intl should be available as a prop.

## ✅ Checklist


packages/components/src/dateLookup/DateLookup.state.spec.js
```diff
@@ -1,64 +0,0 @@
-import React from 'react';
-import { shallow } from 'enzyme';
-
-import DateLookup from '.';
-
-// Tests for getDerivedStateFromProps results
-describe('DateLookup state', () => {
```
dozortsev, line 7: don't know how to fix this test...

```shell
    TypeError: _.default.getDerivedStateFromProps is not a function

      59 |     const onChange = jest.fn();
      60 |     const props = { value: new Date(2018, 11, 27), max: new Date(2018, 10, 1), onChange };
    > 61 |     DateLookup.getDerivedStateFromProps(props, defaultState);
```

packages/components/src/dateInput/DateInput.spec.js
```diff
@@ -1,22 +1,51 @@
...
+import { useIntl } from 'react-intl';
 import { shallow, mount } from 'enzyme';
+import { DateInput } from '..';
 
-import DateInput from '.';
-
-import { LOCALES, MONTHS_EN, MONTHS_FR } from './data/testFixtures';
 import { fakeEvent } from '../common/fakeEvents';
 
+const MONTHS_FR = [
```
dozortsev, line 8: moved some test vars from `packages/components/src/dateInput/data/testFixtures.js` here and deleted the file because they only meant to be use for this test but file (`testFixtures.js`) got bundled

packages/components/src/dateLookup/dayCalendar/DayCalendar.js
```diff
@@ -1,55 +1,53 @@
...
 
 import { MonthFormat } from '../../common';
 import Header from '../header';
 import DayCalendarTable from './table';
 
-class DayCalendar extends PureComponent {
-  selectPreviousMonth = () => {
-    const { viewMonth, viewYear } = this.props;
-    this.props.onViewDateUpdate({
+function DayCalendar({
```
dozortsev, line 10: this component doesn't have state and I prefer use `useIntl` hook to get Intl context so decided to refactor this component to be functional one

packages/components/src/dateInput/data/testFixtures.js
```diff
@@ -1,39 +0,0 @@
-// Used for testing only
```
dozortsev, line 1: but this file was always bundled

packages/components/src/phoneNumberInput/PhoneNumberInput.spec.js
```diff
@@ -250,32 +253,6 @@ describe('Given a telephone number component', () => {
...
-          component = shallow(<PhoneNumberInput {...props} locale="xx-XX" countryCode="US" />);
-          select = component.find(PREFIX_SELECT_SELECTOR);
-          input = component.find(NUMBER_SELECTOR);
-        });
-
-        it('should override locale prefix with country specific prefix', () => {
-          expect(select.props().selected.value).toBe('+1');
-        });
-      });
-    });
```
dozortsev, line 278: these tests checking `locale` prop, I deleted them as now they just redundant since this prop doesn't exist anymore

packages/components/src/upload/Upload.events.spec.js
```diff
@@ -14,48 +22,21 @@ describe('Upload (events)', () => {
 
   describe('on shallow', () => {
     beforeEach(() => {
+      component = shallow(<Upload {...props} />).dive();
       jest.restoreAllMocks();
     });
 
-    it('calls onDragEnter on dragenter event', () => {
-      const onDragEnterSpy = jest
-        .spyOn(Upload.prototype, 'onDragEnter')
```
dozortsev, line 22: Upload now is wrapped in function `injectIntl` which means that there is not access to prototype anymore (`Upload.prototype`) that is why I delete these tests

dozortsev: don't think I should provide replacement for them now due fact that we will refactor this component anyway eventually 🙂

packages/components/i18n/en.json
```diff
@@ -1,6 +1,17 @@
 {
   "neptune.CloseButton.ariaLabel": "Close",
+  "neptune.MoneyInput.Select.placeholder": "Select an option...",
   "neptune.Pagination.ariaLabel": "Pagination navigation",
   "neptune.PaginationLink.ariaLabel.active": "Go to page {pageNumber}",
-  "neptune.PaginationLink.ariaLabel.inactive": "Current page, page {pageNumber}"
-}
\ No newline at end of file
+  "neptune.PaginationLink.ariaLabel.inactive": "Current page, page {pageNumber}",
+  "neptune.Upload.csButtonText": "Select other file?",
```
helendurranttw, line 7:
```suggestion
  "neptune.Upload.csButtonText": "Upload another file?",
```

packages/components/i18n/en.json
```diff
@@ -1,6 +1,17 @@
...
+  "neptune.Upload.csFailureText": "Upload failed. Please, try again",
+  "neptune.Upload.csSuccessText": "Upload complete!",
+  "neptune.Upload.csTooLargeMessage": "Please provide a file smaller than 5MB",
+  "neptune.Upload.csWrongTypeMessage": "Please provide a supported format",
+  "neptune.Upload.psButtonText": "Cancel",
+  "neptune.Upload.psProcessingText": "Uploading...",
+  "neptune.Upload.usButtonText": "Or Select File",
```
helendurranttw, line 14:
```suggestion
  "neptune.Upload.usButtonText": "Or select a file",
```

packages/components/i18n/en.json
```diff
@@ -1,6 +1,17 @@
...
   "neptune.PaginationLink.ariaLabel.active": "Go to page {pageNumber}",
-  "neptune.PaginationLink.ariaLabel.inactive": "Current page, page {pageNumber}"
-}
\ No newline at end of file
+  "neptune.PaginationLink.ariaLabel.inactive": "Current page, page {pageNumber}",
+  "neptune.Upload.csButtonText": "Select other file?",
+  "neptune.Upload.csFailureText": "Upload failed. Please, try again",
+  "neptune.Upload.csSuccessText": "Upload complete!",
+  "neptune.Upload.csTooLargeMessage": "Please provide a file smaller than 5MB",
+  "neptune.Upload.csWrongTypeMessage": "Please provide a supported format",
```
helendurranttw, line 11:
```suggestion
  "neptune.Upload.csWrongTypeMessage": "File type not supported, please try again with a different file",
```

packages/components/src/dateInput/DateInput.js
```diff
@@ -286,7 +285,6 @@ DateInput.propTypes = {
   disabled: Types.bool,
   size: Types.oneOf([DateInput.Size.SMALL, DateInput.Size.MEDIUM, DateInput.Size.LARGE]),
   value: Types.oneOfType([Types.string, Types.instanceOf(Date)]),
-  locale: Types.string,
```
helendurranttw, line 289: Should we not log a warning for people who are passing this through? And tell them to swap to use the Provider instead?

packages/components/src/dateInput/DateInput.js
```diff
@@ -10,16 +11,13 @@ import { explodeDate, convertToLocalMidnight } from './utils';
 import { getMonthNames, isDateValid, isMonthAndYearFormat } from '../common/dateUtils';
 import './DateInput.css';
 
-const DEFAULT_LOCALE = 'en-GB';
-
-const MonthBeforeDay = ['en-US', 'ja-JP'];
+const MonthBeforeDay = ['ja'];
```
helendurranttw, line 14: Why did you remove US here?

packages/components/src/dateInput/DateInput.story.js
```diff
@@ -40,7 +39,6 @@ export const basic = () => {
         dayLabel="Day input"
         monthLabel="Month Select"
         yearLabel="Year input"
-        locale={locale}
```
helendurranttw, line 43: Same comment as above - log a warning if this is being passed?

packages/components/src/dateLookup/DateLookup.js
```diff
@@ -305,7 +305,6 @@ DateLookup.propTypes = {
   min: Types.instanceOf(Date),
   max: Types.instanceOf(Date),
   size: Types.oneOf([DateLookup.Size.SMALL, DateLookup.Size.MEDIUM, DateLookup.Size.LARGE]),
-  locale: Types.string,
```
helendurranttw, line 308: Same comment as above - log a warning if this is being passed?

packages/components/src/dateLookup/dayCalendar/DayCalendar.js
```diff
@@ -71,4 +72,4 @@ DayCalendar.defaultProps = {
   max: null,
 };
 
-export default DayCalendar;
+export default injectIntl(DayCalendar);
```
helendurranttw, line 75: Am I right in thinking that eventually we'll switch to `useIntl` everywhere?

packages/components/src/dateLookup/dayCalendar/DayCalendar.spec.js
```diff
@@ -6,6 +6,9 @@ import DayCalendar from '.';
 import Header from '../header';
 import DayCalendarTable from './table';
 
+jest.mock('react-intl', () => ({
```
helendurranttw, line 9: Is it worth pulling this out somewhere common? You end up rewriting it in several places

packages/components/src/dateLookup/dayCalendar/DayCalendar.spec.js
```diff
@@ -24,13 +27,12 @@ describe('DayCalendar', () => {
...
       viewMonth: 10,
       viewYear: 2018,
-      locale: 'xx',
       monthFormat: 'long',
       onSelect: jest.fn(),
       onLabelClick: jest.fn(),
       onViewDateUpdate: jest.fn(),
     };
-    component = shallow(<DayCalendar {...props} />);
+    component = shallow(<DayCalendar {...props} />).dive();
```
helendurranttw, line 35: 🌊🏊‍♀️

packages/components/src/dynamicFieldDefinitionList/DynamicFieldDefinitionList.js
```diff
@@ -27,7 +24,6 @@ DynamicFieldDefinitionList.Layout = Layout;
 DynamicFieldDefinitionList.propTypes = {
   model: Types.shape({}).isRequired,
   fields: Types.shape({}).isRequired,
-  locale: Types.string,
```
helendurranttw, line 30: Same comment as above - log a warning if this is being passed?

packages/components/src/dynamicFieldDefinitionList/FormattedValue/FormattedValue.spec.js
```diff
@@ -77,7 +79,7 @@ describe('FormattedValue', () => {
     });
 
     it('should display the formatted date value', () => {
-      expect(item.text()).toBe('20/12/2000');
+      expect(item.text()).toBe('12/20/2000');
```
helendurranttw, line 82: The right way around!

packages/components/src/money/Money.js
```diff
@@ -1,17 +1,16 @@
...
-const Money = ({ amount, currency, locale }) => <>{formatMoney(amount, currency, locale)}</>;
+const Money = ({ amount, currency }) => {
+  const { locale } = useIntl();
+  return <>{formatMoney(amount, currency, locale)}</>;
+};
 
 Money.propTypes = {
   amount: Types.number.isRequired,
   currency: Types.string.isRequired,
-  locale: Types.string,
```
helendurranttw, line 10: Comment as above

packages/components/src/moneyInput/MoneyInput.js
```diff
@@ -295,7 +301,6 @@ MoneyInput.propTypes = {
   amount: Types.number,
   size: Types.oneOf([MoneyInput.Size.SMALL, MoneyInput.Size.MEDIUM, MoneyInput.Size.LARGE]),
   onAmountChange: Types.func,
-  locale: Types.string,
```
helendurranttw, line 298: Comment as above

packages/components/src/moneyInput/currencyFormatting.spec.js
```diff
@@ -10,10 +10,7 @@ describe('Number formatting', () => {
   });
 
   it('parses localized numbers', () => {
-    [
-      ['1234.56', 'en-GB'],
-      ['1,23,4.56', 'en-US'],
```
helendurranttw, line 15: Worth leaving the US test, no?

packages/components/src/phoneNumberInput/PhoneNumberInput.js
```diff
@@ -167,7 +168,6 @@ PhoneNumberInput.propTypes = {
   onFocus: Types.func,
   onBlur: Types.func,
   countryCode: Types.string,
-  locale: Types.string,
```
helendurranttw, line 170: Comment as above

packages/components/src/phoneNumberInput/PhoneNumberInput.spec.js
```diff
@@ -240,33 +243,8 @@ describe('Given a telephone number component', () => {
 
       describe('and country code', () => {
         beforeEach(() => {
-          component = shallow(<PhoneNumberInput {...props} locale="es-ES" countryCode="US" />);
```
helendurranttw, line 243: Are these tests really redundant?

packages/components/src/upload/Upload.js
```diff
@@ -387,30 +393,30 @@ Upload.propTypes = {
 
 Upload.defaultProps = {
   animationDelay: 700,
-  csButtonText: 'Select other file?',
-  csFailureText: 'Upload failed.Please, try again',
-  csSuccessText: 'Upload complete!',
-  csTooLargeMessage: 'Please provide a file smaller than 5MB',
-  csWrongTypeMessage: 'Please provide a supported format',
+  csButtonText: null,
```
helendurranttw, line 396: 🎶  undefined 🎶 (for all of the nulls)

packages/components/src/upload/Upload.spec.js
```diff
@@ -242,7 +252,7 @@ describe('Upload', () => {
     });
 
     it('onSuccess is called with response when httpOptions are provided', (done) => {
-      component = mount(<Upload {...props} httpOptions={{ url: 'a-url' }} />);
+      component = mount(<Upload {...props} httpOptions={{ url: 'a-url' }} />).children();
```
helendurranttw, line 255: Why do we need children here now where we didn't before?

helendurranttw (commented):
Nice work Anton, so good to see all those `locale={locale}`s disappearing 😍

packages/components/src/dateInput/DateInput.js
dozortsev: if they start use latest components they won't be able to use them without Provider component, most likely their test will just fail with error that `intl` is not defined in React context.

this property isn't deprecated, I'm just fully removing it and how the source of truth for locale is React context.

I would not any temporal warnings...

packages/components/src/dateLookup/dayCalendar/DayCalendar.js
dozortsev: yep, I hope so,

note: `useIntl` is a hook, hooks only allowed to be used inside functional components, I hope we will get the point when all our components are function ones 🙂

packages/components/src/dateLookup/dayCalendar/DayCalendar.spec.js
dozortsev: I would love to write these down once somewhere and then reuse but jest doesn't allow use API from outside 

e.g following code 

```js
import { injectIntlMock } from '../../test-utils';

jest.mock('react-intl', () => ({
  injectIntl: injectIntlMock,
}));
```


throws an error:

```js
ReferenceError: ~/Documents/tw/code/neptune-web/packages/components/src/dateLookup/dayCalendar/DayCalendar.spec.js: The module factory of `jest.mock()` is not allowed to reference any out-of-scope variables.
Invalid variable access: injectIntlMock
```

packages/components/src/phoneNumberInput/PhoneNumberInput.spec.js
dozortsev: yep, they test prop which doesn't exist anymore (`locale`)

packages/components/src/upload/Upload.spec.js
dozortsev: `.children` is sort of an equivalent of `.dive` but for `mount(..)`

packages/docs/pages/about/Setup.mdx
```diff
@@ -50,6 +51,12 @@ export default function App() {
 }
 ```
 
+**Note:** we recommend map you locale code to ours before passing it into `Provider` and fetching translations, you can use our internal API for this:
+
+```js
+import { mapLocale, LOCALES, DEFAULT_LOCALE } from "@transferwise/components/build/es/polyfill/common/locale";
```
dozortsev, line 57: moving this close to main example to make info about these API more vivid

packages/components/src/provider/Provider.spec.js
```diff
@@ -1,12 +1,22 @@
...
 import Provider from '.';
 import closeButtonMessages from '../common/CloseButton/CloseButton.messages';
 
 describe('Provider', () => {
+  beforeAll(() => {
+    // eslint-disable-next-line no-console
+    console.error = jest.fn();
+    jest.spyOn(console, 'error').mockImplementation(() => {});
+    // eslint-disable-next-line no-console
+    jest.spyOn(console, 'warn').mockImplementation(() => {});
```
dozortsev, line 15: to prevent test failing because of unknown prop value passed in (we have custom [`console#error`](https://github.com/transferwise/neptune-web/blob/main/packages/test-config/config/setupTests.js#L11-L13) for this) and avoid spamming in test

packages/components/src/common/locale/index.js
```diff
@@ -19,10 +19,15 @@ export const LOCALES = [
...
+  if (!locale || locale.trim().length === 0) {
+    return null;
   }
-  const iso2Code = locale.slice(0, COUNTRY_ISO2_CODE_LENGTH);
+  const adjustedLocale = locale.replace('_', '-').toLowerCase();
+  const mappedLocale = LOCALES.find((value) => value.toLowerCase() === adjustedLocale);
+  if (mappedLocale) {
+    return mappedLocale;
+  }
+  const iso2Code = adjustedLocale.slice(0, COUNTRY_ISO2_CODE_LENGTH);
```
dozortsev, line 30: see test below to better understand which values it now handles

packages/components/src/provider/Provider.js
```diff
@@ -2,20 +2,20 @@ import React from 'react';
...
   let intlConfig;
-  if (localeCode === null) {
+  if (mappedLocale === null) {
     // eslint-disable-next-line no-console
     console.warn(`Unsupported locale value was provided: '${locale}', defaulting to 'en'`);
-    intlConfig = { locale: 'en', messages: en };
+    intlConfig = { locale: DEFAULT_LOCALE, messages: en };
   } else {
-    intlConfig = { locale, messages };
+    intlConfig = { locale: mappedLocale, messages };
```
dozortsev, line 18: fix of the bug https://transferwise.slack.com/archives/C2SGDKP4Y/p1611144292068900?thread_ts=1611139542.053500&cid=C2SGDKP4Y

packages/components/src/common/locale/index.js
```diff
@@ -1,28 +1,53 @@
-export const DEFAULT_LOCALE = 'en';
+const ENGLISH_LOCALE = 'en';
+export const DEFAULT_LOCALE = 'en-GB';
+export const US_LOCALE = 'en-US';
```
dozortsev, line 3: adding support for UK and US specific because we format dates, amounts etc. differently https://transferwise.slack.com/archives/C2SGDKP4Y/p1611159097088300?thread_ts=1611139542.053500&cid=C2SGDKP4Y

packages/components/package.json
```diff
@@ -24,15 +24,15 @@
...
+    "build:crowdin-source-file": "formatjs extract 'src/**/*.messages.js' --out-file i18n/en-GB.json --format i18n/formatter.js",
     "build:compile-css": "gulp compileLess --dest=src",
     "build:umd": "NODE_ENV=umd rollup -c",
     "build:umd-nopolyfill": "NODE_ENV=umd-nopolyfill rollup -c",
     "build:es": "NODE_ENV=es babel src -d build/es/polyfill --ignore '**/*.spec.js','**/*.story.js','**/__mocks__','**/test-utils'",
     "build:es-nopolyfill": "NODE_ENV=es-nopolyfill babel src -d build/es/no-polyfill --ignore '**/*.spec.js','**/*.story.js','**/__mocks__','**/test-utils'",
     "build:copy-files": "cpx 'src/**/!(db)/*.{css,json,svg}' build/es/polyfill",
     "build:copy-files2": "cpx 'src/**/!(db)/*.{css,json,svg}' build/es/no-polyfill",
-    "build:copy-lang": "cpx 'i18n/*.json' build/i18n && cpx 'i18n/*.json' build/es/i18n && cpx 'i18n/index.js' build/i18n",
+    "build:copy-lang": "cd i18n/ && cp en-GB.json en-US.json && cd .. && cpx 'i18n/*.json' build/i18n && cpx 'i18n/*.json' build/es/i18n && babel i18n/index.js -d build/i18n && cpx build/i18n/index.js build/es/i18n",
```
dozortsev, line 35: US file is just a copy of UK file

packages/components/src/common/locale/index.spec.js
```diff
@@ -0,0 +1,54 @@
...
+    it('locale codes should match with names of translations files', async () => {
+      // eslint-disable-next-line guard-for-in,no-restricted-syntax
+      for (let i = 0; i < LOCALES.length; i += 1) {
+        const locale = LOCALES[i];
+        // eslint-disable-next-line no-await-in-loop
+        const file = await import(`../../../i18n/${locale}.json`);
+        expect(file).not.toBe(null);
+      }
+    });
+  });
```
dozortsev, line 53: make sure that locale codes and files are consistent

packages/components/src/common/locale/index.js
```diff
@@ -1,28 +1,53 @@
...
+export const SPANISH_LOCALE = 'es';
+export const FRENCH_LOCALE = 'fr';
+export const HUNGARIAN_LOCALE = 'hu';
+export const INDONESIAN_LOCALE = 'id';
+export const ITALIAN_LOCALE = 'it';
+export const POLISH_LOCALE = 'pl';
+export const PORTUGUESE_LOCALE = 'pt';
+export const ROMANIAN_LOCALE = 'ro';
+export const RUSSIAN_LOCALE = 'ru';
+export const TURKISH_LOCALE = 'tr';
```
StevePole, line 15: I don't understand why we use language for locale in all of these places.  The logic that applies to en-GB and en-US should equally apply to es-ES and es-US.

packages/components/src/moneyInput/MoneyInput.js
```diff
@@ -295,7 +301,6 @@ MoneyInput.propTypes = {
   amount: Types.number,
```
StevePole, line 301: Arguably selectedCurrency could be optional and we could default based on locale.

StevePole:
General question, how does a consumer set the locale now?  I'm conscious of us coupling to some specific element of our stack.  It looks like the assumption we're making is that the consumer is also using react-intl and will have set the locale accordingly?

packages/components/src/common/locale/index.js
StevePole: I'd argue we should take the locale as given, and attempt to extract language (or country) from it for use in translation (or localisation/defaults), not limit locales to an arbitrary subset.  Are we going to edit this list every time TransferWise adds a language, sub optimal if so...

dozortsev:
@StevePole 

> General question, how does a consumer set the locale now?...

consumers set locale via Provider component, see example on [setup docs](https://transferwise.github.io/neptune-web/branch/i18n-follow-up/about/Setup#usage)

> ..I'm conscious of us coupling to some specific element of our stack. It looks like the assumption we're making is that the consumer is also using react-intl and will have set the locale accordingly?

in order to consumers to components they just need install react-intl, check this docs https://transferwise.github.io/neptune-web/branch/i18n-follow-up/about/Setup#translations-and-the-provider

packages/components/src/common/locale/index.js
dozortsev: they define list of translations which components support and also could be used for formatting, potentially in future we could use them for right-to-left support

packages/components/package.json
andreapiras, line 35: can you not cp i18n.en-GB.json instead of entering the directoring?

packages/components/src/dateInput/DateInput.js
```diff
@@ -1,25 +1,24 @@
 import React, { useState } from 'react';
+import { useIntl } from 'react-intl';
 import Types from 'prop-types';
 
 import '../common/polyfills/closest';
 import Select from '../select';
 
 import { Size, DateMode, MonthFormat } from '../common';
+import { US_LOCALE, JAPANESE_LOCALE } from '../common/locale';
```
andreapiras, line 9: JA_LOCALE

packages/components/package.json
dozortsev: I tried that, but `cp` doesn't work this way, see this comment https://askubuntu.com/questions/460206/must-copy-and-rename-file#comment1914309_460207

packages/components/src/common/locale/index.js
StevePole: The date formatter and date select support a much wider range of languages and locales than this.

StevePole: What about the es-US issue?  We support the relevant language, and we support the relevant localisation.  The only reason it wouldn't work is because of this whitelist.

packages/components/src/checkboxButton/CheckboxButton.less
```diff
@@ -24,7 +24,8 @@
   left: 14px;
   top: 12px;
 
-  input[type="checkbox"], .tw-checkbox-button {
+  input[type='checkbox'],
+  .tw-checkbox-button {
```
dozortsev, line 28: this just somehow get reformatted during `git commit`

packages/components/src/common/locale/index.js
```diff
@@ -1,30 +1,92 @@
-export const DEFAULT_LOCALE = 'en';
+export const DEFAULT_LANG = 'en';
+export const DEFAULT_LOCALE = 'en-GB';
+
 const COUNTRY_ISO2_CODE_LENGTH = 2;
 
-export const LOCALES = [
-  DEFAULT_LOCALE,
+export const SUPPORTED_LANGUAGES = [
```
StevePole, line 6: Can we take this list from here to make it easier to add languages? https://github.com/transferwise/neptune-web/pull/749/files#diff-210c6624b2820dec5012dc931347259e0ab0e7b70cfee032cc0f8be838f4d585L16

dozortsev: we easily can do that but this would mean bundle all translations files 🙂

```js
import files from '../i18n';

export const SUPPORTED_LANGUAGES = Object.keys(files);

```

dozortsev: JFYI these tests will make sure that file names and those string codes will be always consistent

https://github.com/transferwise/neptune-web/pull/749/files#diff-3c7c053d52daf9212a5d9b32aa439243d3a2940b8849eef70b976f7dee9df230R59-R85

StevePole: What if you had a file "supportLanaguages" that was imported by both?

StevePole (approved):

packages/css/src/less/forms/bootstrap-forms.less
```diff
@@ -7,7 +7,6 @@
 // Forms
 // --------------------------------------------------
 
-
```
dozortsev, line 10: I don't know why got these .less files formatting changes here 🤷‍♂️ I tried to reset them but not luck

just ignore them

packages/components/src/dateInput/DateInput.js
dozortsev: fixed

packages/components/src/moneyInput/currencyFormatting.spec.js
dozortsev: yep, fixed

packages/components/src/upload/Upload.js
dozortsev: done

packages/components/src/common/locale/index.js
```diff
@@ -1,30 +1,92 @@
...
+
+/**
+ * Verifies whether provided local is valid using JS natives {Intl#getCanonicalLocales}
+ *
+ * @param locale (`es`, `es-ES`, `en-GB`, `en`, `ja`, `ja-JP` etc)
+ * @returns {boolean}
+ */
+export const isLocaleValid = (locale) => {
+  try {
+    Intl.getCanonicalLocales(locale);
```
helendurranttw, line 87: So I might be completely misunderstanding this method, but I just tried it and I get some really weird results (this is in Chrome):

Screenshot 2021-01-28 at 14 38 30

Firefox had the same results, so I guess they're expected.. but how is 'espinica' a valid locale?

Also - this method is not supported in Internet Explorer, so you should probably use a polyfill ([this one](https://formatjs.io/docs/polyfills/intl-getcanonicallocales/) I guess?)

packages/components/src/common/locale/index.spec.js
```diff
@@ -0,0 +1,101 @@
...
+    it('given a valid locale it should return the correct value', () => {
+      expect(getCountryFromLocale('es-ES')).toBe('es');
+    });
+    it('given an invalid locale it should return false', () => {
+      expect(getCountryFromLocale('es_ES')).toBe(null);
+    });
+    it('given an invalid locale it should return false', () => {
+      expect(getCountryFromLocale('es_E')).toBe(null);
+    });
+    it('given an invalid locale it should return false', () => {
```
helendurranttw, line 97:
```suggestion
    it('given an invalid locale it should return null', () => {
      expect(getCountryFromLocale('es_ES')).toBe(null);
    });
    it('given an invalid locale it should return null', () => {
      expect(getCountryFromLocale('es_E')).toBe(null);
    });
    it('given an invalid locale it should return null', () => {
```

packages/components/src/common/locale/index.spec.js
```diff
@@ -0,0 +1,101 @@
...
+  });
+
+  describe('contract between SUPPORTED_LANGUAGES and names of translations files', () => {
+    it('number of lang codes should be equal to number of translations files', () => {
+      expect(SUPPORTED_LANGUAGES.length).toBe(Object.keys(translationFiles).length);
+    });
+
+    it('lang codes should match to names of translations object', () => {
+      // eslint-disable-next-line guard-for-in,no-restricted-syntax
+      for (const property in translationFiles) {
```
helendurranttw, line 66: You can change the syntax here and remove the need for the eslint comment:
```
     Object.keys(translationFiles).forEach((property) => {
```

packages/components/src/common/locale/index.spec.js
```diff
@@ -0,0 +1,101 @@
...
+        if (!result) {
+          // eslint-disable-next-line no-console
+          console.log(`'${property}' doesn't exist in ${[SUPPORTED_LANGUAGES]}`);
+        }
+        expect(result).toBe(true);
+      }
+    });
+
+    it('lang codes should match with names of translations files', async () => {
+      // eslint-disable-next-line guard-for-in,no-restricted-syntax
```
helendurranttw, line 77: Don't think this is necessary

packages/components/src/dateInput/DateInput.js
helendurranttw: Right - but they will still be passing in a locale prop that is now unnecessary right? Maybe we should at least tell people in the changelog?

packages/components/src/dateLookup/dayCalendar/DayCalendar.spec.js
helendurranttw: 😭

packages/components/src/moneyInput/MoneyInput.js
```diff
@@ -22,15 +24,19 @@ const formatAmountIfSet = (amount, currency, locale) => {
   return typeof amount === 'number' ? formatAmount(amount, currency, locale) : '';
 };
 
-export default class MoneyInput extends Component {
+class MoneyInput extends Component {
   constructor(props) {
     super(props);
+    // eslint-disable-next-line react/prop-types
```
helendurranttw, line 30: I can see these popping up everywhere - let's modify the eslint config at the root of neptune-web instead. I had a quick search, this should do the trick:

```
"react/prop-types": [
      "error",
      {
        "ignore": [
          "intl"
        ]
      }
    ],
```

If that works, please remove all comments like the above from the code base :)

packages/components/src/provider/Provider.js
```diff
@@ -1,30 +1,40 @@
...
 function Provider({ i18n, children }) {
-  const cache = createIntlCache();
-  const { locale, messages } = i18n;
-  const localeCode = mapLocale(locale);
-
+  const { locale, messages, defaultRichTextElements } = i18n;
+  const adjustedLocale = adjustLocale(locale);
   let intlConfig;
-  if (localeCode === null) {
+  if (adjustedLocale == null) {
```
helendurranttw, line 11: Is this supposed to be `===`?

packages/components/src/provider/Provider.spec.js
helendurranttw: For these cases you can just put `/* eslint-disable no-console */` at the top of the file - then you won't need to repeat yourself.

packages/components/src/provider/Provider.js
```diff
@@ -1,30 +1,40 @@
...
+    intlConfig = { locale: adjustedLocale, messages };
   }
-  const intl = createIntl(intlConfig, cache);
-  return <RawIntlProvider value={intl}>{children}</RawIntlProvider>;
+  return (
+    <IntlProvider
+      defaultLocale={DEFAULT_LOCALE}
+      locale={intlConfig.locale}
+      messages={intlConfig.messages}
+      defaultRichTextElements={defaultRichTextElements}
```
helendurranttw, line 25: What is this? Should we not just allow prop spreading if they want to pass extra config into IntlProvider?

packages/docs/liveEditorCode/provider.code.js
```diff
@@ -1,26 +1,30 @@
 () => {
+  const lang = defaultLocale.slice(0, 2);
```
helendurranttw, line 2: How about using getLangFromLocale here?

packages/docs/pages/about/Setup.mdx
```diff
@@ -50,6 +52,27 @@ export default function App() {
 }
 ```
 
+**Note:** we recommend map you locale code to ours before passing it into `Provider` and fetching translations, you can use our internal API for this:
```
helendurranttw, line 55: Why do we still recommend this? It should work without it now, no?

packages/docs/pages/about/Setup.mdx
```diff
@@ -33,12 +33,14 @@ import '@transferwise/components/build/main.css';
 
 import { Provider, Button } from '@transferwise/components';
 import componentsTranslations from '@transferwise/components/build/i18n';
+import { adjustLocale, getLangFromLocale } from "@transferwise/components/build/es/polyfill/common/locale";
```
helendurranttw, line 36: Why don't we expose these further up so we can do something like

```
import { adjustLocale, getLangFromLocale } from "@transferwise/components";
```

Or we could wrap them in ProviderUtils or something?

Just seems strange that we're recommended people to reach into the package like that.

yarn.lock
```diff
@@ -122,6 +122,13 @@
   dependencies:
     "@babel/highlight" "^7.8.3"
 
+"@babel/code-frame@^7.12.11":
```
helendurranttw, line 125: Are these correct? You haven't made any changes to dependencies, not sure why you've got so many updates here.

packages/components/src/moneyInput/MoneyInput.js
dozortsev: good point!

it works 👍

packages/components/src/provider/Provider.js
dozortsev: apparently some people use that https://github.com/transferwise/partner-integration-flow/blob/master/pages/_app.tsx#L209, afaik this property necessary for to be able to render HTML tags inside messages

using prop spreading here is dangerous because you can easily override IntlProvider internals, e.g following will break this component, `intl` object won't have `formatMessage` function

```
    <IntlProvider
      {...intlConfig.locale}
      defaultLocale={DEFAULT_LOCALE}
      defaultRichTextElements={defaultRichTextElements}
    >
```

packages/docs/liveEditorCode/provider.code.js
dozortsev: ok

packages/docs/pages/about/Setup.mdx
dozortsev: it's not mandatory to use them, you can pass locale straight away to Provider, but many apps have inconsistent locale codes, so we safety and less hassle we recommend to use this tools

packages/docs/pages/about/Setup.mdx
dozortsev: it's not mandatory to use them, these functions some sort of hidden gems of the library, I wouldn't expose them now this way, or at least won't be concerned about this in this PR 🙂

packages/docs/pages/_document.js
```diff
@@ -16,7 +16,6 @@ class MyDocument extends Document {
           <NextScript />
           {/* The below is necessary to stop animations running on page load in Chrome
           (https://bugs.chromium.org/p/chromium/issues/detail?id=332189) */}
```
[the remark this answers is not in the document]
dozortsev: this wasn't intentional change of this PR, I removed it because couldn't test things locally and later just forget to reset this change

code suggestion looks a bit broken, plus I prefer us addressing this thing in separate PR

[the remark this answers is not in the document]
dozortsev: so I just reset it as it on main branch

yarn.lock
dozortsev: not sure either, resetting this file

dozortsev: now that I just added polyfill (`@formatjs/intl-locale`) it got update again

packages/components/src/common/locale/index.js
dozortsev: using native `Intl` web API is probably the best that we can do to check if locale is correct

added polyfill 👍

packages/docs/pages/about/Setup.mdx
helendurranttw: I'm fine to leave them like this, but I think we should update the docs to tell people it's not necessary to use them.

packages/docs/pages/_document.js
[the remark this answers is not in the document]
helendurranttw: Fair enough, I've created a separate PR to address 👍

PR #751, opened 2020-12-23
[USEC-2856] Extend Radio option to look like a Tile

conorcussell:
## 🖼 Context

I need a Radio input which looks like a tile for [this screen ](https://www.figma.com/file/g0uFpWN5uFc6aLK3sKXSts/Registration-%2F-Login---2FA?node-id=760%3A270)

Screenshot 2020-12-23 at 12 50 40

## 🚀 Changes

- [x] Adds prop `presentation` to `RadioOption` component

## 🤔 Considerations

I was a little confused which component to edit here - `Decision` mentions that it may support Radio's and Checkboxes in the future, but that already deals with some presentational logic around lists or tiles, so it felt confusing to add this in there. 

Also, `RadioGroup` doesn't offer any way to render a `RadioOption` but only the fairly similar component `Radio`.

I was wondering if it makes sense to extend `RadioGroup` to render either `Radio` or `RadioOption`, this would be helpful as the labelling/layout requirements could be dealt with in `RadioGroup` instead of by the consumer.

## ✅ Checklist


helendurranttw:
Hey @conorcussell thanks for this - we like this approach and will integrate this work but want to revisit decision and nav option at the same time (as well as looking at the designs for radio and checkbox tiles). I'm going to close this for the moment but I've linked it in the ticket so we can integrate this when we do that piece of work.

PR #754, opened 2020-12-30
ADS-1637 Internationalisation for Dynamic Flows

felixxiatransferwise:
## 🖼 Context

Adds translations to dynamic flows.

## 🚀 Changes

 
Setup i18n.

## ✅ Checklist


crowdin.yml
```diff
@@ -6,3 +6,9 @@ files:
       two_letters_code:
         zh-HK: zh-HK
     update_option: update_as_unapproved
+  - source: /packages/components/i18n/en.json
```
dozortsev, line 9: I guess it should be `dynamic-flows` folder here

dozortsev, line 9: I guess it should be `dynamic-flows` folder here

packages/components/src/common/locale/index.js
```diff
@@ -10,6 +10,7 @@ export const LOCALES = [
   'id',
   'it',
   'ja',
+  'jp',
```
dozortsev, line 13: is this for Japanese? if so then we already have code for it `ja`

packages/components/src/common/locale/index.js
```diff
@@ -23,7 +24,7 @@ export function mapLocale(locale) {
     return locale;
   }
   const iso2Code = locale.slice(0, COUNTRY_ISO2_CODE_LENGTH);
-  if (LOCALES.includes(locale)) {
+  if (LOCALES.includes(iso2Code)) {
```
dozortsev, line 27: 👍

packages/dynamic-flows/package.json
```diff
@@ -24,13 +24,15 @@
...
     "build:compile-less": "gulp compileLess --dest=src",
     "build:es": "NODE_ENV=es babel src -d build/es/polyfill --ignore '**/*.spec.js','**/*.story.js'",
     "build:es-nopolyfill": "NODE_ENV=es-nopolyfill babel src -d build/es/no-polyfill --ignore '**/*.spec.js','**/*.story.js'",
     "build:copy-files": "cpx 'src/**/!(db)/*.{css,json,svg}' build/es/polyfill",
     "build:copy-files2": "cpx 'src/**/!(db)/*.{css,json,svg}' build/es/no-polyfill",
-    "test": "jest --env=jsdom --runInBand --config=../../node_modules/@transferwise/test-config/jest.config.js && yarn test:lint",
-    "test:watch": "jest --watch --env=jsdom --config=../../node_modules/@transferwise/test-config/jest.config.js",
+    "build:copy-lang": "cpx 'i18n/*.json' build/i18n && cpx 'i18n/*.json' build/es/i18n && cpx 'i18n/index.js' build/i18n",
+    "test": "jest --env=jsdom --runInBand && yarn test:lint",
+    "test:watch": "jest --watch --env=jsdom",
```
dozortsev, line 35: why this change? :)

packages/dynamic-flows/src/i18n/useI18n.js
```diff
@@ -0,0 +1,24 @@
+import { useEffect, useState } from 'react';
+
+import {
+  DEFAULT_LOCALE,
+  mapLocale,
+} from '@transferwise/components/build/es/polyfill/common/locale';
+
+import translations from '@transferwise/dynamic-flows/build/i18n';
```
dozortsev, line 8: here I think it should be imported from source (e.g `'../i18n'`) rather then from package

packages/dynamic-flows/src/jsonSchemaForm/JsonSchemaForm.js
```diff
@@ -1,9 +1,18 @@
 import React from 'react';
 import Types from 'prop-types';
+import { Provider } from '@transferwise/components';
```
dozortsev, line 3: I don't think DF components should use `Provider` components this way

---

`@transferwise/dynamic-flows` relies on `@transferwise/components` so when consumer installs & uses DF components it should take care of proper setup of `@transferwise/components` in their consumer codebase as well 

we mention about this in docs:

> Note: Provider must be used in repositories which bundle components (i.e. lists @transferwise/components as a dependency). For repositories which don't bundle components (typically those that list it as a peerDependency) it is unnecessary (often it's similar component libraries).

https://transferwise.github.io/neptune-web/about/Setup#usage

packages/dynamic-flows/src/jsonSchemaForm/JsonSchemaForm.js
```diff
@@ -1,9 +1,18 @@
...
 import GenericSchema from './genericSchema';
 
+import { useI18n } from '../i18n/useI18n';
+
 const JsonSchemaForm = (props) => {
-  return <GenericSchema {...props} />;
+  const i18n = useI18n(props.locale);
```
dozortsev, line 9: `useI18n` fetches translations files of DF components which you later passing into `Provider` component.. this is not right, you should pass only `@transferwise/components` translations inside the component

---

perhaps we should be more clear about this here https://transferwise.github.io/neptune-web/about/Setup#usage 🤔

packages/dynamic-flows/src/jsonSchemaForm/JsonSchemaForm.js
```diff
@@ -1,9 +1,18 @@
...
+
+  return (
+    <Provider i18n={i18n}>
```
dozortsev, line 12: here you rather should use [react-intl provider](https://formatjs.io/docs/react-intl/components#rawintlprovider) where you passing DF translations files 

or you could create custom context provider component which init react-intl provider like we have done for `@transferwise/components` (I'm talking about [Provider](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/provider/Provider.js))

it depends which kinda of `JsonSchemaForm` components is, is it main component of all DF components which will be used by all consumers of `@transferwise/dynamic-flows` package? if so then I guess follow first approach.

packages/components/src/common/locale/index.js
felixxiatransferwise: The iso 2 code for Japanese is 'jp' though? https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes

There's also multiple places in where we've used `jp-JP` so `ja` wouldn't work here.

packages/dynamic-flows/package.json
felixxiatransferwise: because i'm specifying the new `packages/dynamic-flows/jest.config.js` config which uses `../../node_modules/@transferwise/test-config/jest.config.js` as a base but also applies the new `packages/dynamic-flows/setupTests.js`

packages/dynamic-flows/src/jsonSchemaForm/JsonSchemaForm.js
felixxiatransferwise: Updated as discussed.

packages/dynamic-flows/src/jsonSchemaForm/JsonSchemaForm.js
felixxiatransferwise: Updated as discussed.

packages/components/src/common/locale/index.js
dozortsev: this list locales code based on IETF codes e.g https://gist.github.com/traysr/2001377

plus at TW we use `ja` value as well

packages/components/src/provider/Provider.js
```diff
@@ -23,7 +23,7 @@ function Provider({ i18n, children }) {
 
 Provider.propTypes = {
   i18n: Types.shape({
-    locale: Types.oneOf(LOCALES).isRequired,
+    locale: Types.string.isRequired,
```
dozortsev, line 26: what this change for? :)

packages/dynamic-flows/setupTests.js
```diff
@@ -0,0 +1,24 @@
...
+  const reactIntl = jest.requireActual('react-intl');
+  const intl = reactIntl.createIntl({
+    locale: 'en',
+  });
+
+  return {
+    ...reactIntl,
+    useIntl: () => intl,
+  };
+});
```
dozortsev, line 24: could you please explain what this config does and why need custom one?

packages/dynamic-flows/package.json
dozortsev: I see, now I'm curious why we need custom config https://github.com/transferwise/neptune-web/pull/754/files#r552562345

packages/dynamic-flows/src/jsonSchemaForm/JsonSchemaForm.story.js
```diff
@@ -35,21 +39,23 @@ export const basic = () => {
...
-      model={model}
-      errors={errors}
-      locale={locale}
-      translations={translations}
-      onChange={action('onChange')}
-      submitted={submitted}
-      disabled={disabled}
-      onPersistAsync={() => {}}
-    />
+    <Provider i18n={{ locale, messages: translations[mapLocale(locale)] }}>
```
dozortsev, line 47: here you also have to pass component's (`transferwise/components`) translations files as well

```js

import translations from '../../build/i18n';
import componentTranslations from '@transferwise/components/build/i18n';
..
const mappedLocale = mapLocale(locale);
const messages = Object.assign(translations[mappedLocale], componentTranslations[mappedLocale]);
..
return (
  <Provider i18n={{ locale: mappedLocale, messages: messages }}>

File: packages/dynamic-flows/src/jsonSchemaForm/JsonSchemaForm.story.js [hunk 12]
```diff
@@ -35,21 +39,23 @@ export const basic = () => {
   const locale = select('locale', ['en-GB', 'jp-JP'], 'en-GB');
```
dozortsev, line 39: here I would use `LOCALES` and `DEFAULT_LOCALE` from components package

```js
import { mapLocale, LOCALES, DEFAULT_LOCALE } from "@transferwise/components/build/es/polyfill/common/locale";

..

const locale = select('locale', LOCALES, DEFAULT_LOCALE);
```

packages/components/src/provider/Provider.js
felixxiatransferwise: with `Types.oneOf(LOCALES).isRequired`, your prop type would always be expecting one of LOCALES which renders `const localeCode = mapLocale(locale);` pointless as it's already going to be a 2 letter code. I believe you intended this to take in any string as the locale and then the localeCode is kept internal to the function.

packages/dynamic-flows/setupTests.js
felixxiatransferwise: refactored to use real provider

packages/dynamic-flows/package.json
```diff
@@ -24,11 +24,13 @@
     "docs": "yarn build",
     "build": "npm-run-all build:*",
     "build:clean": "rm -rf lib build",
+    "build:crowdin-source-file": "formatjs extract 'src/**/*.messages.js' --out-file i18n/en.json --format i18n/formatter.js",
     "build:compile-less": "gulp compileLess --dest=src",
     "build:es": "NODE_ENV=es babel src -d build/es/polyfill --ignore '**/*.spec.js','**/*.story.js'",
     "build:es-nopolyfill": "NODE_ENV=es-nopolyfill babel src -d build/es/no-polyfill --ignore '**/*.spec.js','**/*.story.js'",
     "build:copy-files": "cpx 'src/**/!(db)/*.{css,json,svg}' build/es/polyfill",
     "build:copy-files2": "cpx 'src/**/!(db)/*.{css,json,svg}' build/es/no-polyfill",
+    "build:copy-lang": "cpx 'i18n/*.json' build/i18n && cpx 'i18n/*.json' build/es/i18n && cpx 'i18n/index.js' build/i18n",
```
dozortsev, line 33: no need to copy JSON files twice, I think copying in `build/es/i18n` is redundant here

packages/components/src/provider/Provider.js
dozortsev: I don't see how this change make code better 🤔

having `locale: Types.oneOf(LOCALES).isRequired` still will allow pass any string at runtime on production env,

and nothing is guarantee that consumers would use our `mapLocale`, they free to use it but I'm sure the will be cases where they will have their one mapping utils to adjust their code to your or they will just pass raw value there

dozortsev (commented):
I think you should also update `JsonSchemaForm.mdx`

packages/dynamic-flows/src/test-utils/index.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import { Provider } from '@transferwise/components';
+
+function mountWithProviders(node) {
+  return mount(<Provider i18n={{ locale: 'en', messages: {} }}>{node}</Provider>);
```
dozortsev, line 6:
```suggestion
function mountWithProviders(component, locale =  'en', messages = {}) {
  return mount(<Provider i18n={{ locale, messages }}>{component}</Provider>);
```

dozortsev: suggesting to make it possible to pass custom locale and message there

felixxiatransferwise: Hmm, i usually find that we don't want to be testing messages especially since this Provider is tested externally, instead this would just test the default message is as expected. What use case is there for passing in custom locale/messages here?

packages/components/src/provider/Provider.js
felixxiatransferwise: > locale: Types.oneOf(LOCALES).isRequired still will allow pass any string at runtime

sure but here's kind of my point, `Provider` here doesn't actually require one of `LOCALES` because it does the mapping inside which means its more correct to use any string as the prop type here.

> nothing is guarantee that consumers would use our mapLocale

But we are talking about the propTypes of Provider here and Provider does use `mapLocale`?

packages/dynamic-flows/src/test-utils/index.js
dozortsev: e.g you will have component with some default text in it and will want to find that element by text (via `*ByText` or `ByLabelText`, `ByPlaceholderText` (https://testing-library.com/docs/dom-testing-library/api-queries/#queries))

packages/components/src/provider/Provider.js
dozortsev: again, how this change would make this code better?

currently by looking at this code it quickly hinting to consumer what strings they should pass inside, it also will throw warning during development env or e.g on Storybook

dozortsev: when we migrate this code on TypeScript most likely it will be [string literal types](https://stackoverflow.com/a/35312363/2650174) which will also server as auto completion in IDE and check during complication

felixxiatransferwise: > how this change would make this code better?

"because it does the mapping inside which means it's more **correct** to use any string as the prop type here."

If you were not mapping inside, then sure, the argument type should be specified as one of `LOCALES`. In reality, the contract for this function is actually any string since the it's not actually expecting a value from `LOCALES`. My point still stands for TS, if you are already specifying the the arg type in the form of string literals, then mapping inside the function becomes redundant unless it was intended as a safety net?

packages/dynamic-flows/src/test-utils/index.js
felixxiatransferwise: why do you need custom message to do that? can't you just find `*ByText(defaultText)` here? https://github.com/transferwise/neptune-web/blob/main/packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/PersistAsyncSchema.spec.js#L213

packages/components/src/provider/Provider.js
felixxiatransferwise: Updated as suggested as I don't feel strongly about this change

packages/dynamic-flows/src/test-utils/index.js
felixxiatransferwise: Updated as suggested as I don't feel strongly against the suggested change

dozortsev: sorry, but I don't mean any custom message

I mean here we passing empty messages

```js
<Provider i18n={{ locale: 'en', messages: {} }}>{node}</Provider>
```
so later [in component](https://github.com/transferwise/neptune-web/blob/ADS-1637_react-intl-for-df/packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/PersistAsyncSchema.js#L23
) during test run it will just render message id as fallback

now you should be able pass e.g English messages

```
import en from '../i18n/en.json';

mountWithProviders(< PersistAsyncSchema ..>, 'en', en)
```

to be able to test like you mentioned 

https://github.com/transferwise/neptune-web/blob/main/packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/PersistAsyncSchema.spec.js#L213

packages/components/src/provider/Provider.js
dozortsev: `mapLocale` inside `Provider` is done for actual code mapping and `locale: Types.oneOf(LOCALES)` is for developer experience

packages/docs/pages/components/dynamicForms/JsonSchemaForm.mdx
```diff
@@ -3,6 +3,7 @@ import { JsonSchemaForm } from '@transferwise/dynamic-flows';
 import code from '../../../liveEditorCode/jsonSchemaForm.code';
 
 **Note:** This component is in `@transferwise/dynamic-flows`. To use, install this package and import from `@transferwise/dynamic-flows`.
+This component requires react-intl provider with translation messages provided in `./build/i18n`. See example [here](https://transferwise.github.io/neptune-web/branch/i18y/about/Setup#usage).
```
dozortsev, line 6:
```suggestion
This component requires react-intl provider with translation messages provided in `./build/i18n`. See example [here](https://transferwise.github.io/neptune-web/about/Setup#usage).
```

dozortsev (approved):

PR #755, opened 2021-01-06
docs: update Setup docs re. using Provider component

dozortsev:
## 🖼 Context

During review of [i18n support for DF components](https://github.com/transferwise/neptune-web/pull/754) I realized some stuff which I think worth mention in Setup docs re. Provider component

## 🚀 Changes

Updated docs.

Here is rendered version for more pleasant read https://transferwise.github.io/neptune-web/branch/update-setup-docs/about/Setup but make sure to check this [comment](https://github.com/transferwise/neptune-web/pull/755/files#r552733687).


packages/docs/pages/about/Setup.mdx
```diff
@@ -56,6 +58,29 @@ By default, apps come polyfilled. If you would like to use an unpolyfilled versi
 
 You need to pass a translations object to the `Provider` component. This can be done in a few different ways, depending on the setup of your library / application.
 
+**Note:** Under the hood `Provider` component uses react-intl, it creates context using [`RawIntlProvider`](https://formatjs.io/docs/react-intl/components/#rawintlprovider)
+so if your code already uses react-intl and it's providers (e.g `RawIntlProvider` or `IntlProvider`) replace them with `Provider` passing all translations objects (`@transferwise/components` translation files, your app translation files etc) there
```
dozortsev, line 62: Just to be super clear, this is possible because one react-intl context can be use for all different libraries that rely on it's context and actually it should be done this way because defining multiple same contexts one in another will just that they will override each other.

From React docs https://reactjs.org/docs/context.html#contextprovider

> ..Providers can be nested to override values deeper within the tree.

```js
import { RawIntlProvider } from 'react-intl';
import { Provider } from '@transferwise/components';

const ConsumerApp = () => (
  <RawIntlProvider ...>
    <Provider i18n={...}> // here react-intl context defined inside Provider will override above context
      ..
    </Provider>
  </TranslationProvider>
);
```

dozortsev: note: they opt out and use `RawIntlProvider` instead `Provider` and it will work but if in future we extend functionally of `Provider` component (e.g with theming context) they will have to switch to `Provider` so it's better to educate consumer to use `Provider` since the beginning

packages/docs/pages/about/Setup.mdx
```diff
@@ -32,15 +32,17 @@ import '@transferwise/icons/lib/styles/main.min.css';
...
-import translations from '@transferwise/components/build/i18n';
+import componentTranslations from '@transferwise/components/build/i18n';
 
-export default function App() {
-  const currentLocale = ...
+import appTranslations from '...';
 
+export default function App() {
+  const locale = ...
+  const messages = Object.assing(componentTranslations[locale], appTranslations[locale]);
```
andreapiras, line 41: This one can probably be {...componentTranslations[locale], ...appTranslations[locale]}

packages/docs/pages/about/Setup.mdx
```diff
@@ -32,15 +32,17 @@ import '@transferwise/icons/lib/styles/main.min.css';
 import '@transferwise/components/build/main.css';
 
 import { Provider, Button } from '@transferwise/components';
-import translations from '@transferwise/components/build/i18n';
+import componentTranslations from '@transferwise/components/build/i18n';
```
andreapiras, line 35: componentsTranslations

packages/docs/pages/about/Setup.mdx
```diff
@@ -32,15 +32,17 @@ import '@transferwise/icons/lib/styles/main.min.css';
...
 
-export default function App() {
-  const currentLocale = ...
+import appTranslations from '...';
```
andreapiras, line 37: `import appTranslations from 'path-to-app-translations'`

packages/docs/pages/about/Setup.mdx
```diff
@@ -56,6 +58,29 @@ By default, apps come polyfilled. If you would like to use an unpolyfilled versi
 
 You need to pass a translations object to the `Provider` component. This can be done in a few different ways, depending on the setup of your library / application.
 
+**Note:** Under the hood `Provider` component uses react-intl, it creates context using [`RawIntlProvider`](https://formatjs.io/docs/react-intl/components/#rawintlprovider)
```
andreapiras, line 61: Is Provider a component ? or a `provider` ?

andreapiras, line 61:
```suggestion
**Note:** `Provider` uses react-intl which creates context using [`RawIntlProvider`](https://formatjs.io/docs/react-intl/components/#rawintlprovider).
If your code uses `RawIntlProvider` or `IntlProvider` they can be replaced with `Provider` with all necessary translations (`@transferwise/components` + App translations).
Translations can be combined using rest operator (as shown on above example) or using [`formatjs compile`](https://formatjs.io/docs/guides/distribute-libraries#declaring-with-a-convention)).
If your App uses doesn't use react-intl to handle translations the setup should be handled separately.
If you are using retranslate library a setup example can be:
```

packages/docs/pages/about/Setup.mdx
andreapiras, line 62:
```suggestion

```

packages/docs/pages/about/Setup.mdx
```diff
@@ -56,6 +58,29 @@ By default, apps come polyfilled. If you would like to use an unpolyfilled versi
...
+so if your code already uses react-intl and it's providers (e.g `RawIntlProvider` or `IntlProvider`) replace them with `Provider` passing all translations objects (`@transferwise/components` translation files, your app translation files etc) there
+(you can combine translations all together using e.g `Object#assign` (as shown on above example) or use [`formatjs compile`](https://formatjs.io/docs/guides/distribute-libraries#declaring-with-a-convention) this is really up to you).
```
andreapiras, line 63:
```suggestion

```

packages/docs/pages/about/Setup.mdx
dozortsev: this is already clear enough what do we mean

packages/docs/pages/about/Setup.mdx
dozortsev: 👌

packages/docs/pages/about/Setup.mdx
dozortsev: why you suggesting to remove this?

andreapiras: This would be replaced by my commit suggestion

andreapiras: This is related to my previous commit suggestion.

packages/docs/pages/about/Setup.mdx
dozortsev: > ..If your code uses `RawIntlProvider` or `IntlProvider` they can be replaced with `Provider` with all necessary translations (`@transferwise/components` + App translations).

they **should** be replaced, they better use our Provider

dozortsev: > ..If your App uses doesn't use react-intl to handle translations the setup should be handled separately.

a bit weird one :)

packages/docs/pages/about/Setup.mdx
dozortsev: I see

packages/docs/pages/about/Setup.mdx
```diff
@@ -56,6 +58,29 @@ By default, apps come polyfilled. If you would like to use an unpolyfilled versi
 
 You need to pass a translations object to the `Provider` component. This can be done in a few different ways, depending on the setup of your library / application.
 
+**Note:** `Provider` component uses react-intl which creates context using [`RawIntlProvider`](https://formatjs.io/docs/react-intl/components/#rawintlprovider).
+If your code already uses react-intl and its providers (e.g `RawIntlProvider` or `IntlProvider`) replace them with `Provider` passing all translations objects (`@transferwise/components` translation files, your app translation files, etc) there
+(Translations can be combined using rest operator (as shown on above example) or using [`formatjs compile`](https://formatjs.io/docs/guides/distribute-libraries#declaring-with-a-convention)).
```
andreapiras, line 63:
```suggestion
If your code uses react-intl, its providers `RawIntlProvider` or `IntlProvider` should be replaced with `Provider` which accepts a prop `messages` that has to contains all your app + `@transferwise/components` translations 
Translations can be combined using rest operator (as shown on above example) or using [`formatjs compile`](https://formatjs.io/docs/guides/distribute-libraries#declaring-with-a-convention)).
```

dozortsev: > ..If your code uses react-intl, its providers `RawIntlProvider` or `IntlProvider`

if someone use react-intl this doesn't guarantee that they also use its providers components, that why I have "and" between them

> ..`Provider` which accepts a prop `messages`..

why describe Provider component API here? this part redundant here

> ...that has to contains all your app + `@transferwise/components` translations

there could be cases where not only translations of those those two should be passed inside, translations of others libraries that expose their translations files (instead of bundling them) should be passed as well, that is why I have I say about that following way "(`@transferwise/components` translation files, your app translation files, etc)"

packages/docs/pages/about/Setup.mdx
```diff
@@ -56,6 +58,29 @@ By default, apps come polyfilled. If you would like to use an unpolyfilled versi
 
 You need to pass a translations object to the `Provider` component. This can be done in a few different ways, depending on the setup of your library / application.
 
+**Note:** `Provider` component uses react-intl which creates context using [`RawIntlProvider`](https://formatjs.io/docs/react-intl/components/#rawintlprovider).
```
dozortsev, line 61:
```suggestion
**Note:** `Provider` component uses react-intl which creates a context using [`RawIntlProvider`](https://formatjs.io/docs/react-intl/components/#rawintlprovider).
```

packages/docs/pages/about/Setup.mdx
dozortsev: added "a" in between "creates context" 👍

PR #758, opened 2021-01-13
ADS-1820 Trigger validation on submit

felixxiatransferwise:
## 🖼 Context

When we have an oneOf schemas with shared property names like this:
```
{
  "title": "Choose schema",
  "oneOf": [
    {
      "type": "object",
      "title": "Option A",
      "properties": {
        "a": {   // <-- // <-- "a" is optional here
          "title": "A Number",
          "type": "number"
        }
      }
    },
    {
      "type": "object",
      "title": "Option B",
      "properties": {
        "a": {    // <-- "a" is required here
          "title": "B Number",
          "type": "number"
        }
      },
      "required": ["a"].
    }
  ]
}
```

When we submit, switch options and submit again, we don't get a validation message even though we should.

https://user-images.githubusercontent.com/55701870/104480936-0e2de200-55bd-11eb-8dec-22831a456daa.mov

## 🚀 Changes

 
With the changes, we validate on submit. Resulting in the correct behaviour:

https://user-images.githubusercontent.com/55701870/104480828-ec345f80-55bc-11eb-8fb7-11e070125e23.mov

## ✅ Checklist


packages/dynamic-flows/src/jsonSchemaForm/basicTypeSchema/BasicTypeSchema.js
```diff
@@ -78,7 +78,7 @@ const BasicTypeSchema = (props) => {
   const isHidden = props.schema.hidden || isConst;
 
   useEffect(onSchemaChange, [props.schema]);
-  useEffect(onModelChange, [props.model]);
+  useEffect(onModelChange, [props.model, props.submitted]);
```
JulianG-TW, line 81: Just a wondering...
Should we rename the `onModelChange` function since now it's triggered by more than just changes in the model?

JulianG-TW (approved):

packages/dynamic-flows/src/jsonSchemaForm/JsonSchemaForm.e2e.spec.js
```diff
@@ -182,16 +189,69 @@ describe('E2E: Given a component for rendering a JSON schema form', () => {
...
   describe('when disabled', () => {
+    it('should disable the controls', () => {
+      render(<JsonSchemaForm {...props} disabled />);
+
+      expect(screen.getByLabelText('Number')).toBeDisabled();
+      expect(screen.getByRole('textbox')).toBeDisabled();
+      screen.getAllByRole('button').forEach((btn) => {
+        expect(btn).toBeDisabled();
+      });
+    });
```
StevePole, line 200: This section is nowhere near as clean as it used to be. 
image

Can we find a way to do this that's a bit more uniform?

packages/dynamic-flows/src/jsonSchemaForm/JsonSchemaForm.e2e.spec.js
```diff
@@ -182,16 +189,69 @@ describe('E2E: Given a component for rendering a JSON schema form', () => {
...
 
-    it('should disable the controls', () => {
-      expect(component.find('input[type="text"]').props().disabled).toBe(true);
-      expect(component.find('input[type="number"]').props().disabled).toBe(true);
-      expect(component.find(Checkbox).props().disabled).toBe(true);
-      expect(component.find(Radio).first().props().disabled).toBe(true);
-      expect(component.find(Select).props().disabled).toBe(true);
+    describe('when the schema where the property is optional is selected and submitted without value', () => {
+      it('should not display validation errors', () => {
+        const { rerender } = render(<JsonSchemaForm {...props} />);
```
StevePole, line 237: This line is repeated, you could do it in beforeEach

packages/dynamic-flows/src/jsonSchemaForm/JsonSchemaForm.e2e.spec.js
```diff
@@ -182,16 +189,69 @@ describe('E2E: Given a component for rendering a JSON schema form', () => {
...
+
+      expect(screen.getByLabelText('Number')).toBeDisabled();
+      expect(screen.getByRole('textbox')).toBeDisabled();
+      screen.getAllByRole('button').forEach((btn) => {
+        expect(btn).toBeDisabled();
+      });
+    });
+  });
+
+  describe('when oneOf schemas share the same property name', () => {
```
StevePole, line 203: This seems a slightly convoluted test, is there not a simpler way to recreate the scenario?

StevePole: I think you could just pass in an object with a single prop, then pass in the same object with the prop now required?

Otherwise it's a bit confusing what is actually being tested here, is it something special about oneOf?

packages/dynamic-flows/src/jsonSchemaForm/JsonSchemaForm.e2e.spec.js
felixxiatransferwise: 👍  i've refactored so we assert all elements with relevant control roles are disabled.

packages/dynamic-flows/src/jsonSchemaForm/JsonSchemaForm.e2e.spec.js
felixxiatransferwise: This is the exact scenario that is failing in production: when an oneOf contains multiple object schemas but some of the object schemas share property names and these properties have different validation requirements. I'd like to test this exact case to prevent regression.

I've renamed the test and consts to make it clearer.

StevePole (approved):

PR #759, opened 2021-01-14
feat: whole alert is tappable

andreapiras:
## 🖼 Context

On touch devices we decided to make the whole alert tappable while on desktop only the link provided is clickable

## 🚀 Changes

 
Add onTouchStart, onTouchEnd, onTouchMove

## ✅ Checklist


packages/components/src/alert/Alert.js
```diff
@@ -37,8 +38,33 @@ const Alert = (props) => {
...
+  const handleTouchMove = () => {
+    setShouldFire(false);
+  };
+
+  const handleTouchEnd = () => {
+    if (shoudlFire) {
+      if (action.target === '_blank') {
+        window.top.open(action.href);
+      } else {
+        window.top.location.assign(action.href);
```
helendurranttw, line 54: I think this could benefit from a comment

packages/components/src/alert/Alert.js
```diff
@@ -37,8 +38,33 @@ const Alert = (props) => {
   const Icon = iconTypeMap[mappedType];
   const iconEl = icon ? cloneElement(icon, { size: 24 }) : <Icon size={24} />;
 
+  const handleTouchStart = () => {
+    setShouldFire(true);
+  };
```
helendurranttw, line 43:
```suggestion
  const handleTouchStart = () => setShouldFire(true);
```

packages/components/src/alert/Alert.js
```diff
@@ -37,8 +38,33 @@ const Alert = (props) => {
...
+
+  const handleTouchMove = () => {
+    setShouldFire(false);
+  };
```
helendurranttw, line 47:
```suggestion
  const handleTouchMove = () => setShouldFire(false);
```

packages/components/src/alert/Alert.js
```diff
@@ -23,6 +23,7 @@ const iconTypeMap = {
 };
 
 const Alert = (props) => {
+  const [shoudlFire, setShouldFire] = useState(false);
```
helendurranttw, line 26: Typo

packages/components/src/alert/Alert.spec.js
```diff
@@ -254,4 +254,84 @@ describe('Alert', () => {
...
+        assign: jest.fn(),
+      };
+    });
+
+    afterAll(() => {
+      window.location = location;
+    });
+
+    describe('when target is not blank', () => {
+      it('redirect window on tap', () => {
```
helendurranttw, line 273:
```suggestion
      it('loads action on tap', () => {
```

packages/components/src/alert/Alert.spec.js
```diff
@@ -254,4 +254,84 @@ describe('Alert', () => {
...
+        };
+        const { getByRole } = render(<Alert action={action} message={message} />);
+
+        fireEvent.touchStart(getByRole('alert'));
+        expect(window.location.assign).not.toHaveBeenCalled();
+        fireEvent.touchEnd(getByRole('alert'));
+        expect(window.location.assign).toHaveBeenCalledWith(action.href);
+      });
+
+      it('doesn`t redirect on scroll', () => {
```
helendurranttw, line 287:
```suggestion
      it('doesn`t redirect on touch move', () => {
```

packages/components/src/alert/Alert.spec.js
```diff
@@ -254,4 +254,84 @@ describe('Alert', () => {
...
+        };
+        const { getByRole } = render(<Alert action={action} message={message} />);
+
+        fireEvent.touchStart(getByRole('alert'));
+        expect(window.open).not.toHaveBeenCalled();
+        fireEvent.touchEnd(getByRole('alert'));
+        expect(window.open).toHaveBeenCalledWith(action.href);
+      });
+
+      it('doesn`t opens a new window on scroll', () => {
```
helendurranttw, line 320: I don't think the target should impact the behaviour - think this test is a duplicate

helendurranttw (commented):
Looking good - but I just tested it, and if you click the dismiss cross, it will dismiss and also load the action. We should prevent that behaviour if the user clicks the cross.

andreapiras:
> Looking good - but I just tested it, and if you click the dismiss cross, it will dismiss and also load the action. We should prevent that behaviour if the user clicks the cross.

@helendurranttw I've fixed the bug thank you

packages/components/src/alert/Alert.js
```diff
@@ -37,8 +39,33 @@ const Alert = (props) => {
...
+  const handleTouchStart = () => setShouldFire(true);
+
+  const handleTouchMove = () => setShouldFire(false);
+
+  const handleTouchEnd = (event) => {
+    if (!closeRef.current.contains(event.target)) {
+      if (shouldFire && action) {
+        if (action.target && action.target === '_blank') {
+          // window.top returns the topmost window in the hierarchy of window objects.
+          // In this case is used to redirection the window if the event is triggered from within iframes.
```
dozortsev, line 51: no need to leave this comment in code, we should assume that people knows about this,

this should be more of a PR comment to simplify code review rather than comment in code 🙂

packages/components/src/alert/Alert.js
```diff
@@ -23,7 +23,9 @@ const iconTypeMap = {
 };
 
 const Alert = (props) => {
+  const [shouldFire, setShouldFire] = useState(false);
   const { arrow, action, children, className, icon, onDismiss, message, type } = props;
+  const closeRef = useRef(null);
```
dozortsev, line 28: would call this `closeButtonRef`

packages/components/src/alert/Alert.js
```diff
@@ -37,8 +39,33 @@ const Alert = (props) => {
   const Icon = iconTypeMap[mappedType];
   const iconEl = icon ? cloneElement(icon, { size: 24 }) : <Icon size={24} />;
 
+  const handleTouchStart = () => setShouldFire(true);
+
+  const handleTouchMove = () => setShouldFire(false);
+
+  const handleTouchEnd = (event) => {
+    if (!closeRef.current.contains(event.target)) {
```
dozortsev, line 47: are we checking here that event is coming not from close button? if so, I would either add comment here or extract this expression in variable e.g `isAlertTap`

packages/components/src/alert/Alert.js
```diff
@@ -37,8 +39,33 @@ const Alert = (props) => {
...
+      if (shouldFire && action) {
+        if (action.target && action.target === '_blank') {
```
dozortsev, line 49:
```suggestion
      if (shouldFire && action?.href) {
        if (action?.target=== '_blank') {
```

packages/components/src/alert/Alert.spec.js
```diff
@@ -8,12 +8,24 @@ import {
...
-import { render, cleanup, screen, userEvent } from '../test-utils';
+import { render, cleanup, screen, userEvent, fireEvent } from '../test-utils';
 import Alert from './Alert';
 
+jest.mock('react', () => {
+  const originReact = jest.requireActual('react');
+  const mUseRef = jest.fn();
+  return {
+    ...originReact,
+    useRef: mUseRef,
```
dozortsev, line 19: feels a bit strange to mock `useRef`, seems like you trying to test implementation of React component rather outcome (DOM), sort of Enzyme approach.. can you try avoid mock it?

packages/components/src/alert/closeButton/CloseButton.js
```diff
@@ -5,16 +5,20 @@ import { Cross as CrossIcon } from '@transferwise/icons';
...
-  <button
-    type="button"
-    className={classNames('np-close-button close', className)}
-    aria-label="Close"
-    onClick={onClick}
-  >
-    <CrossIcon size={size} />
-  </button>
-);
+export const CloseButton = React.forwardRef((props, ref) => {
```
dozortsev, line 8: do you really need use `.forwardRef` here? I thought that `useRef` in parent component should be enough

dozortsev, line 8: can we use already existing close button component (`common/CloseButton.js`)?

andreapiras: Yes because you need to know if the element clicked is the button itself.

packages/components/src/alert/Alert.js
andreapiras: @helendurranttw asked for a comment

packages/components/src/alert/closeButton/CloseButton.js
andreapiras: They don't have currently the same implementation, I'd rather do it in a separate ticket

andreapiras: addressed here https://github.com/transferwise/neptune-web/pull/766/files#diff-13fed0b7d412562b84d801dacc715ba1a2b988e7d19b240c52392dcef83f2fe5R7

packages/components/src/alert/Alert.js
dozortsev: can we still make this change? otherwise why no?

packages/components/src/alert/Alert.js
```diff
@@ -37,8 +39,33 @@ const Alert = (props) => {
   const Icon = iconTypeMap[mappedType];
   const iconEl = icon ? cloneElement(icon, { size: 24 }) : <Icon size={24} />;
 
+  const handleTouchStart = () => setShouldFire(true);
+
+  const handleTouchMove = () => setShouldFire(false);
+
+  const handleTouchEnd = (event) => {
+    if (closeButtonRef?.current && !closeButtonRef.current.contains(event.target)) {
```
dozortsev, line 47: are we checking here that event is coming not from close button? if so, I would either add comment here or extract second expression in variable e.g `isAlertTap`

packages/components/src/alert/Alert.js
```diff
@@ -54,7 +81,12 @@ const Alert = (props) => {
         )}
       </div>
       {onDismiss && (
-        <CloseButton onClick={onDismiss} size={CloseButton.Size.SMALL} className="m-l-2" />
+        <CloseButton
```
dozortsev, line 84: I guess we should merge this PR only after this one https://github.com/transferwise/neptune-web/pull/766 due to `CloseButton` & `Alert` changes overlap

packages/components/src/alert/Alert.js
dozortsev: can we please delete this comment

dozortsev (approved):

packages/components/src/alert/Alert.js
dozortsev: cc @helendurranttw

dozortsev (approved):
worth keeping in mind https://github.com/transferwise/neptune-web/pull/759#discussion_r565447432

packages/components/src/alert/Alert.js
helendurranttw: Sure, fine with me

helendurranttw:
The close button on the alert is misaligned - probably due to merging the close button. Could you fix it?

Before:
Before

Now:
After

Check Figma if you're in any doubt.

PR #760, opened 2021-01-15
refactor!: popover refactor

andreapiras:
## 🖼 Context
As part of component's platform alignment Popover has been refactored. A new position framework has been introduced to support the screen position of the popover.

## 🚀 Changes

- Refactored popover which now use [Popper](https://popper.js.org/)
- Created ResponsivePanel which switches between Panel and 
- Created Panel which is the container of Popover content
- Created BottomSheet which is the mobile version of Panel
- Create useAttachEvent hook to detect click outside of a specific elements
- Deprecated Positions:  LEFT_TOP, RIGHT_TOP,BOTTOM_RIGHT,BOTTOM_LEFT
- Improved a11y of Popover
  
## 🤔 Considerations
BREAKING CHANGES:
- children: changed type
- classNames: changed type and name
- containsFocusableElement: removed
- content: changed type
- removed some functionalities : trigger wrapping, iOS legacy code please check confluence document for more info

## ✅ Checklist


packages/components/src/popover/Detached.js
```diff
@@ -0,0 +1,112 @@
...
+import classnames from 'classnames';
+import throttle from 'lodash.throttle';
+import BottomSheet from './BottomSheet';
+
+import { Position, Breakpoint } from '../common';
+import './Detached.css';
+
+const THROTTLE_MS = 100;
+
+const Detached = ({
```
helendurranttw, line 13: How about "FloatingPanel"?

packages/components/src/popover/Detached.js
```diff
@@ -0,0 +1,112 @@
...
+Detached.Position = {
+  TOP: Position.TOP,
+  RIGHT: Position.RIGHT,
+  BOTTOM: Position.BOTTOM,
+  LEFT: Position.LEFT,
+};
+
+Detached.defaultProps = {
+  arrow: false,
+  extraStyles: undefined,
```
helendurranttw, line 80: As this is just a container, I'd hope we're able to set any desired styles on the content, avoiding the need for this

packages/components/src/popover/Detached.js
helendurranttw: Actually - I'm less and less convinced these two components need to be separate. Can Popover be our base component used by Dropdown and DateLookup?

packages/components/src/popover/Popover.js
```diff
@@ -1,147 +1,96 @@
...
-  componentWillUnmount() {
-    this.close();
-  }
+import { Position } from '../common';
+import Detached from './Detached';
 
-  closePopoverOnOutsideClick = (event) => {
-    const popoverClicked = this.popoverElement.contains(event.target);
-    const triggerClicked = this.triggerElement.contains(event.target);
+const Popover = ({ children, content, fallbackPlacements, flip, intialOpen, placement, title }) => {
```
helendurranttw, line 9: I'd like to simplify this API. To start, I'm not a fan of using children to pass the anchor. Could we pass a reference to the anchor element, and instead use children for content?

What does "flip" do that placement can't do?

Should we make this stateless, and just accept `open` instead of `initialOpen`?

What are fallbackPlacements? Material have a really nice way of [handling placement](https://material-ui.com/components/popover/). Maybe we can do something similar?

packages/components/src/popover/Detached.js
andreapiras: Popover wraps the button, I think it would be weird for Popover to wrap SelectField? no

packages/components/src/popover/Detached.js
```diff
@@ -0,0 +1,112 @@
...
+import './Detached.css';
+
+const THROTTLE_MS = 100;
+
+const Detached = ({
+  arrow,
+  children,
+  extraStyles,
+  fallbackPlacements,
+  flip,
```
dozortsev, line 18: when we would need flip effect? do you have any use cases?

packages/components/src/popover/Detached.js
```diff
@@ -0,0 +1,112 @@
...
+
+Detached.Position = {
+  TOP: Position.TOP,
+  RIGHT: Position.RIGHT,
+  BOTTOM: Position.BOTTOM,
+  LEFT: Position.LEFT,
+};
+
+Detached.defaultProps = {
+  arrow: false,
```
dozortsev, line 79: how consumer will be able to specify position of arrow?

packages/components/src/popover/Detached.js
dozortsev: yeah, I guess having mandatory [`className` prop](https://github.com/transferwise/neptune-web/blob/main/rfc/0001-always-pass-classname.md) should be enough, no?

packages/components/src/popover/Detached.js
```diff
@@ -0,0 +1,112 @@
...
+  extraStyles: undefined,
+  fallbackPlacements: Detached.Position.TOP,
+  flip: true,
+  intialOpen: false,
```
dozortsev, line 83: seems like redundant prop (don't see it being used inside component)

packages/components/src/popover/Detached.js
dozortsev: just found out that Popover it defines specific structure for it's body via `title` and `content` prop about which Detached should bother at all that is one more reason to keep those to components separately 🙂

packages/components/src/popover/Detached.js
```diff
@@ -0,0 +1,112 @@
...
+  offset: [0, 12],
```
dozortsev, line 84: is this 12px?

for this I think we should create abstraction which will be more convenient for consumer. e.g:

```js
const Offset = {
   NONE: [0, 0],
   SMALL: [0, 12],
   ..
}
```
we shouldn't expose Popper API via our component (because tomorrow we might decide switch on another library, and having abstraction will allow us do that without introducing breaking changes), we should abstract it via our "Neptune" things, above example might be bad but I guess you got the idea 🙂

packages/components/src/popover/Detached.less
```diff
@@ -0,0 +1,38 @@
+@import (reference) '@transferwise/neptune-css/dist/less/zindex.less';
+
+.np-dropdown {
```
dozortsev, line 3: I would give same name as name of component it belongs to, `.np-detached` 🙂

packages/components/src/popover/Detached.less
```diff
@@ -0,0 +1,38 @@
...
+  z-index: @zindex-popover;
+  padding: var(--size-16) var(--size-24);
+  background-color: var(--color-background-elevated);
+  background-clip: padding-box;
+  border: 1px solid var(--color-background-elevated);
+  border-radius: var(--radius-small);
+  box-shadow: 0px 0px 40px rgba(34, 48, 73, 0.2);
```
dozortsev, line 10: shadow should also be coming from our Neptune CSS variables (no need to bother about this in this PR, I just wanted to say about this, you can ignore this comment 😄)

dozortsev, line 10: shouldn't shadow be configurable? I mean `<Detached shadow={ true | false } />`

(don't have strong opinion on this, just double checking)

packages/components/src/popover/Popover.js
dozortsev: I think `title` prop is necessary, it enforces consistency

discussion: https://transferwise.slack.com/archives/C01EH9WVASZ/p1605801330045800?thread_ts=1605794545.043400&cid=C01EH9WVASZ

dozortsev (commented):
great progress 👍

packages/components/src/popover/Detached.js
helendurranttw: Haha I like BlueFadeScreen 😜

Detached isn’t descriptive of its purpose. It’s a state - something can be attached or detached to something else. I would argue that it's less descriptive than FloatingPanel (which at least gives a sense of what the component might look like / be used for). Open to alternatives if you have other ideas!

I'm not sure the title / content is important enough to prevent us from merging these components.

Let's have a chat and discuss.

packages/components/src/popover/Detached.js
```diff
@@ -0,0 +1,112 @@
...
+  extraStyles,
+  fallbackPlacements,
+  flip,
+  offset,
+  placement,
+  referenceElement,
+}) => {
+  const [arrowElement, setArrowElement] = useState(null);
+  const [popperElement, setPopperElement] = useState(null);
+  const [isMobile, setIsMobile] = useState(() => window?.innerWidth < Breakpoint.SMALL);
```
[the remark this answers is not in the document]
andreapiras: useClientWidth refers to the parent element but we want popover to switch based on the fact that the window is on mobile size or not. Maybe I can pass a window ref to this hook and I can achieve the same

packages/components/src/popover/Detached.js
andreapiras: Removed - Popper is going to handle this automatically

packages/components/src/popover/Detached.js
andreapiras: Yes but then every time someone uses a popover it has to be careful about sizes ?
image

andreapiras: According to figma the popover has to be 252px, but that may not apply to menus. Should we exposed a `type=(default|menu|select|etc)` to manage the styles?

packages/components/src/popover/Popover.js
andreapiras: If we pass anchor , if you have 10 popovers you have to do the same for each element, I don't think it would be very smooth

packages/components/src/common/hooks/useIsClickOutside/useIsClickOutside.js
```diff
@@ -0,0 +1,20 @@
...
+    const updateIsClickOutside = (event) => {
+      if (ref && !ref.current.contains(event.target)) {
+        setIsClickOutside(true);
+      } else {
+        setIsClickOutside(false);
+      }
+    };
+    document.addEventListener('click', updateIsClickOutside, true);
+
+    return () => document.removeEventListener('click', updateIsClickOutside, true);
```
dozortsev, line 16: what this `true` means here? according to [docs](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) you third param is object

packages/components/src/popover/BottomSheet/BottomSheet.spec.js
```diff
@@ -0,0 +1,31 @@
...
+    return open ? <div className="dimmer">{children}</div> : null;
+  };
+});
+
+jest.mock('../../slidingPanel', () => {
+  // eslint-disable-next-line react/prop-types
+  return ({ open, children }) => {
+    return open ? <div className="sliding-panel">{children}</div> : null;
+  };
+});
```
dozortsev, line 18: why we need these mocks?

andreapiras: It is a good practice to not retest component. Here all we need to know is the high level behaviour of slidingPanel which has already been unit tested so no need to retest it.

packages/components/src/popover/Detached/Detached.js
```diff
@@ -0,0 +1,76 @@
...
+  const [popperElement, setPopperElement] = useState(null);
+
+  const modifiers = [
+    {
+      name: 'computeStyles',
+      options: {
+        adaptive: false,
+      },
+    },
+    { name: 'offset', options: { offset: [0, 12] } },
```
andreapiras, line 22: We need to define either a set of Offset

packages/components/src/popover/Detached/Detached.js
```diff
@@ -0,0 +1,76 @@
...
+    { name: 'flip' },
+  ];
+
+  if (arrow) {
+    modifiers.push({ name: 'arrow', options: { padding: 5, element: arrowElement } });
```
andreapiras, line 27: Popper currently offers multiple option to display arrow that we need to discuss @helendurranttw @dozortsev

packages/components/src/popover/Popover.js
```diff
@@ -1,147 +1,117 @@
...
-      this.close();
-    }
-  };
+import { Position, Breakpoint } from '../common';
+import BottomSheet from './BottomSheet';
+import Detached from './Detached';
+import { useClientWidth, useIsClickOutside } from '../common/hooks';
 
-  handleKeyUp = (event) => {
-    if (event.target.nodeName !== 'BUTTON' && event.keyCode === KeyCodes.ENTER) {
```
andreapiras, line 39: I don't think we should keep this, but rather enforce pass a button in the docs

packages/components/src/popover/Popover.js
```diff
@@ -1,147 +1,117 @@
...
-      'tw-popover',
-      this.style('popover'),
-      this.style('animate'),
-      this.style('in'),
-      { [this.style('scale-down')]: !isOpen },
-      this.style(placement),
-    );
+  const popoverContent = (
+    <div className="np-popover__content" id={memoizedId} aria-hidden={!open} role="tooltip">
+      <span className="sr-only">Tooltip:</span>
```
andreapiras, line 38: This can be removed, I found it in an example and quite liked it. If we want to keep it we should translate. Try with voiceOver

packages/components/src/popover/Detached/Detached.js
```diff
@@ -0,0 +1,76 @@
...
+  arrow: true,
+  open: false,
+  preferredPlacement: Detached.Placement.TOP,
+};
+
+Detached.propTypes = {
+  arrow: Types.bool,
+  children: Types.node.isRequired,
+  open: Types.bool,
+  preferredPlacement: Types.oneOf([
```
dozortsev, line 67: why it's called "preferred"? is it because if not enough space it will default to some other position

dozortsev: we tend to use word 'position' rather than 'placement' could you please rename this prop to comply and consistent with API of with other components which has positioning feature

packages/components/src/popover/Detached/Detached.js
```diff
@@ -0,0 +1,76 @@
...
+
+  const modifiers = [
+    {
+      name: 'computeStyles',
+      options: {
+        adaptive: false,
+      },
+    },
+
+    { name: 'flip' },
```
dozortsev, line 22: what these for? 🙂

packages/components/src/popover/Popover.js
```diff
@@ -1,147 +1,117 @@
...
-    removeClickClassFromDocumentOnIos();
-    document.removeEventListener('click', this.closePopoverOnOutsideClick, true);
-  };
+const Popover = ({ children, content, preferredPlacement, title }) => {
+  logActionRequiredIf(
+    `Popover has deprecated the ${preferredPlacement} value for the 'preferredPlacement' prop. Please use ${deprecatePlacements[preferredPlacement]} instead.`,
+    deprecatePlacements[preferredPlacement],
+  );
+  const memoizedId = useMemo(() => uniqid('np-popover-'), []);
+  const [clientWidth] = typeof window !== 'undefined' ? useClientWidth({ ref: window }) : [0];
```
dozortsev, line 19: why we don't return `[0]` [FlowNav](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/flowNavigation/FlowNavigation.js) and [SizeSwapper](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/sizeSwapper/SizeSwapper.js#L16)? how their cases are different?

dozortsev: maybe we could we put `typeof window !== 'undefined'` check inside `useClientWidth` hook and if its `false` return `[0]` from there?

packages/components/src/popover/Popover.js
```diff
@@ -1,147 +1,117 @@
...
 
-  toggle = () => (this.state.isOpen ? this.close() : this.open());
+  const isMobile = clientWidth && clientWidth < Breakpoint.SMALL;
 
-  createTrigger = () => {
-    const { children: child, containsFocusableElement } = this.props;
-    const { isOpen } = this.state;
-    const wrappedChild = wrapInDOMElementIfNecessary(child);
+  useIsClickOutside({
+    ref: [outsideClickRef, triggerRef],
```
dozortsev, line 26: what `ref` in this case is? could you please rename it?

packages/components/src/common/hooks/useIsClickOutside/useIsClickOutside.js
```diff
@@ -0,0 +1,21 @@
+import { useEffect } from 'react';
+
+export const useIsClickOutside = ({ ref, callBack }) => {
```
dozortsev, line 3: I don't really understand how this works.., could you please at least add comments above the function

packages/components/src/popover/Detached/Detached.js
dozortsev: would not use word 'preferred'

packages/components/src/popover/Detached/Detached.js
```diff
@@ -0,0 +1,76 @@
+import React, { useState, forwardRef } from 'react';
```
dozortsev, line 1: it should sit it it's dedicated folder `scr/detached`

dozortsev (commented):
would perhaps also add `onClose` as optional prop for `Detached`, will be useful when we would use it e.g in [DateLookup](https://github.com/transferwise/neptune-web/blob/popover-refactor/packages/components/src/dateLookup/DateLookup.js#L286)

packages/components/src/popover/BottomSheet/BottomSheet.js
```diff
@@ -0,0 +1,30 @@
+import React, { forwardRef } from 'react';
+import Types from 'prop-types';
+import Dimmer from '../../dimmer';
+import SlidingPanel from '../../slidingPanel';
+import './BottomSheet.css';
+
+const BottomSheet = forwardRef((props, ref) => {
```
dozortsev, line 7: why it needs `ref`?

packages/components/src/common/panel/Panel.js
```diff
@@ -0,0 +1,76 @@
...
+  arrow: Types.bool,
+  children: Types.node.isRequired,
+  open: Types.bool,
+  preferredPlacement: Types.oneOf([
+    Panel.Placement.TOP,
+    Panel.Placement.RIGHT,
+    Panel.Placement.LEFT,
+    Panel.Placement.BOTTOM,
+  ]),
+  triggerRef: Types.shape({ current: Types.shape({}) }).isRequired,
```
dozortsev, line 73: please expose `className` prop

packages/components/src/common/panel/Panel.js
```diff
@@ -0,0 +1,76 @@
...
+  arrow: true,
+  open: false,
+  preferredPlacement: Panel.Placement.TOP,
+};
+
+Panel.propTypes = {
+  arrow: Types.bool,
+  children: Types.node.isRequired,
+  open: Types.bool,
+  preferredPlacement: Types.oneOf([
```
dozortsev, line 67: can we please call this position?

packages/components/src/common/panel/Panel.less
```diff
@@ -0,0 +1,44 @@
+@import (reference) '@transferwise/neptune-css/src/less/variables/_typography.less';
+@import (reference) '@transferwise/neptune-css/dist/less/zindex.less';
+
+.np-detached {
```
dozortsev, line 4: panel

packages/components/src/common/bottomSheet/BottomSheet.js
```diff
@@ -0,0 +1,30 @@
+import React, { forwardRef } from 'react';
+import Types from 'prop-types';
+import Dimmer from '../../dimmer';
+import SlidingPanel from '../../slidingPanel';
+import './BottomSheet.css';
+
+const BottomSheet = forwardRef((props, ref) => {
+  const { children, open } = props;
```
dozortsev, line 8:
```suggestion
const BottomSheet = forwardRef(({ children, open } , ref) => {
```

packages/components/src/common/responsivePanel/ResponsivePanel.js
```diff
@@ -0,0 +1,66 @@
...
+      ) : (
+        <Panel
+          arrow={arrow}
+          open={open}
+          preferredPlacement={preferredPlacement}
+          ref={ref}
+          triggerRef={triggerRef}
+        >
+          {children}
+        </Panel>
```
dozortsev, line 32: why this can't be just inside Panel?

andreapiras: We may reuse Panel on its own

packages/components/src/popover/Detached/Detached.js
andreapiras: This is the current used API, i keep it this way to not introduce deprecation but we can. I would call it placement just because in popper API it's named placement too

packages/components/src/popover/BottomSheet/BottomSheet.js
```diff
@@ -0,0 +1,30 @@
...
+  const { children, open } = props;
+
+  return (
+    <Dimmer open={open}>
+      <SlidingPanel open={open} position="bottom" transparent>
```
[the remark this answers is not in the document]
andreapiras: > would perhaps also add `onClose` as optional prop for `Detached`, will be useful when we would use it e.g in [DateLookup](https://github.com/transferwise/neptune-web/blob/popover-refactor/packages/components/src/dateLookup/DateLookup.js#L286)

Would it be better to managed that from the trigger?

packages/components/src/popover/Popover.js
andreapiras: what do you mean?

packages/components/src/popover/BottomSheet/BottomSheet.js
andreapiras: To detect whether the user has clicked inside or outside the component. This will prevent the bottomsheet to close if people click on it

packages/components/src/popover/Popover.js
```diff
@@ -1,147 +1,117 @@
...
+          });
+        })}
+      </span>
+
+      {isMobile ? (
+        <BottomSheet open={open} ref={outsideClickRef}>
+          {popoverContent}
+        </BottomSheet>
+      ) : (
+        <Detached
```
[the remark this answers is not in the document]
andreapiras: I've implemented this inside ResponsivePanel

packages/components/src/common/panel/Panel.js
andreapiras: @dozortsev Do we need it now? Shall we expose it once we have a use case?

packages/components/src/common/panel/Panel.js
andreapiras: Popper uses the word placement. Shall we keep that to be aligned?

packages/components/src/common/responsivePanel/ResponsivePanel.js
```diff
@@ -0,0 +1,61 @@
...
+};
+
+ResponsivePanel.defaultProps = {
+  preferredPlacement: ResponsivePanel.Placement.TOP,
+};
+
+ResponsivePanel.propTypes = {
+  children: Types.node.isRequired,
+  preferredPlacement: Types.oneOf([
+    ResponsivePanel.Placement.TOP,
```
andreapiras, line 51: I'm leaving out BOTTOM_START BOTTOM_END until we'll need it for Select

packages/components/src/popover/BottomSheet/BottomSheet.js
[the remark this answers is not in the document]
dozortsev: how trigger component will know that Panel just closed?

packages/components/src/common/bottomSheet/BottomSheet.less
```diff
@@ -0,0 +1,4 @@
+.np-bottom-sheet {
```
helendurranttw, line 1: Where did you get this from? Are these styles shared with other floating things (like popover?) Should this be a utility class?

packages/components/src/common/bottomSheet/BottomSheet.js
```diff
@@ -0,0 +1,30 @@
+import React, { forwardRef } from 'react';
+import Types from 'prop-types';
+import Dimmer from '../../dimmer';
+import SlidingPanel from '../../slidingPanel';
+import './BottomSheet.css';
+import { Position } from '../propsValues/position';
+
+const BottomSheet = forwardRef(({ children, open }, ref) => (
```
helendurranttw, line 8: Cleeean ✨

packages/components/src/common/hooks/useIsClickOutside/useIsClickOutside.js
```diff
@@ -0,0 +1,30 @@
...
+ * @usage `useIsClickOutside({refs: [insideClick, triggerRef], callBack: (val) => setOpen(val)});`
+ *
+ * */
+export const useIsClickOutside = ({ refs, callBack }) => {
+  useEffect(() => {
+    const updateIsClickOutside = (event) => {
+      let isClickOutsideElements = true;
+
+      // If the click is triggered from at least one of the elements provided set isClickOutsideElements to false
+      refs.forEach((el) => {
```
helendurranttw, line 18: If `isClickOutsideElements = false;` you may as well end the loop (use `some` instead of `forEach`).

packages/components/src/common/hooks/useIsClickOutside/useIsClickOutside.js
```diff
@@ -0,0 +1,30 @@
...
+        if (el?.current?.contains(event.target)) {
+          isClickOutsideElements = false;
+        }
+      });
+
+      callBack(isClickOutsideElements);
+    };
+    document.addEventListener('click', updateIsClickOutside);
+
+    return () => document.removeEventListener('click', updateIsClickOutside);
```
helendurranttw, line 28: Love the idea of this hook, but at the moment I'm not sure it is sufficiently sophisticated. It seems like once you've set up the hook, that event listener is going to run on every click.

How could you avoid that? You need a way to start listening when you open the element and stop once you've clicked outside.

packages/components/src/common/panel/Panel.js
helendurranttw: I think it's sensible to do it now.

packages/components/src/common/panel/Panel.js
```diff
@@ -0,0 +1,80 @@
...
+import { usePopper } from 'react-popper';
+import { Position } from '..';
+import './Panel.css';
+
+const Panel = forwardRef(({ arrow, children, open, placement, triggerRef }, ref) => {
+  const [arrowElement, setArrowElement] = useState(null);
+  const [popperElement, setPopperElement] = useState(null);
+
+  const modifiers = [
+    // This option, enabled by default, tells Popper to use the most suitable CSS properties
```
helendurranttw, line 14: I think you can trim this a bit

packages/components/src/common/panel/Panel.js
```diff
@@ -0,0 +1,80 @@
...
+    // to position the popper (either top and left, or bottom and right).
+    // This allows the popper content to change, and reduce the likelihood of needing to recompute
+    // the popper position.
```
helendurranttw, line 17: This is a bit confusing - for one thing you're describing the `adaptive` property, so it should probably go above line 21 - but also you've turned adaptive off, so the benefits you list aren't enabled.

packages/components/src/common/panel/Panel.js
```diff
@@ -0,0 +1,80 @@
...
+      },
+    },
+    // This flips the placement if that placement does not fit, the opposite placement will be used.
+    { name: 'flip' },
+  ];
+
+  if (arrow) {
+    modifiers.push({ name: 'arrow', options: { padding: 5, element: arrowElement } });
+    // This lets you displace a popper element from its reference element.
+    modifiers.push({ name: 'offset', options: { offset: [0, 12] } });
```
helendurranttw, line 31: Should this not be configurable? Seems like the user might want to control this.

Also would you still want an offset of 12 if there is no arrow?

packages/components/src/common/panel/Panel.less
```diff
@@ -0,0 +1,50 @@
+@import (reference) '@transferwise/neptune-css/src/less/variables/_typography.less';
```
helendurranttw, line 1: Don't think this is necessary?

packages/components/src/common/panel/Panel.less
```diff
@@ -0,0 +1,50 @@
...
+@import (reference) '@transferwise/neptune-css/dist/less/zindex.less';
+
+.np-panel {
+  z-index: @zindex-popover;
+  padding: var(--size-16) var(--size-24);
+  background-color: var(--color-background-elevated);
+  border: 1px solid var(--color-background-elevated);
```
helendurranttw, line 8: Do you need this border?

packages/components/src/common/panel/Panel.less
```diff
@@ -0,0 +1,50 @@
...
+  border-radius: var(--radius-small);
+  box-shadow: 0px 0px 40px rgba(34, 48, 73, 0.2);
```
helendurranttw, line 10: Once again - where is this coming from - should it be standardised or shared?

packages/components/src/common/panel/Panel.less
```diff
@@ -0,0 +1,50 @@
...
+  opacity: 0;
+  visibility: hidden;
+
+  &--open {
+    transition: opacity 0.2s cubic-bezier(0.6, 0.2, 0.1, 1),
```
helendurranttw, line 15: Same for these - should these be variables?

helendurranttw: Or can we use the standard fade classes provided by neptune css?

packages/components/src/common/panel/Panel.spec.js
```diff
@@ -0,0 +1,42 @@
...
+  const props = {
+    arrow: true,
+    children: <div>children</div>,
+    open: false,
+    preferredPlacement: Panel.Placement.TOP,
+    triggerRef: {},
+  };
+
+  it('renders', () => {
+    const { container } = render(<Panel {...props} />);
```
helendurranttw, line 15: Maybe one that tests the content is rendered?

packages/components/src/common/responsivePanel/ResponsivePanel.js
helendurranttw: In Panel you called this `placement` - best be consistent between the two. Though I think I prefer position for both tbh

packages/components/src/common/responsivePanel/ResponsivePanel.js
```diff
@@ -0,0 +1,61 @@
...
+import { Position, Breakpoint } from '..';
+import BottomSheet from '../bottomSheet';
+import Panel from '../panel';
+import { useClientWidth } from '../hooks';
+
+const ResponsivePanel = forwardRef(
+  ({ arrow, children, open, preferredPlacement, triggerRef }, ref) => {
+    const [clientWidth] = typeof window !== 'undefined' ? useClientWidth({ ref: window }) : [null];
+
+    const isMobile = clientWidth && clientWidth < Breakpoint.SMALL;
```
helendurranttw, line 13: Why aren't you using SizeSwapper here?

packages/components/src/common/responsivePanel/ResponsivePanel.js
```diff
@@ -0,0 +1,61 @@
...
+
+ResponsivePanel.propTypes = {
+  children: Types.node.isRequired,
+  preferredPlacement: Types.oneOf([
+    ResponsivePanel.Placement.TOP,
+    ResponsivePanel.Placement.RIGHT,
+    ResponsivePanel.Placement.LEFT,
+    ResponsivePanel.Placement.BOTTOM,
+  ]),
+  arrow: Types.bool.isRequired,
```
helendurranttw, line 56: Don't think this should be required - default to false?

packages/components/src/common/panel/Panel.js
```diff
@@ -0,0 +1,80 @@
...
+};
+
+Panel.defaultProps = {
+  arrow: true,
+  open: false,
+  placement: Panel.Placement.TOP,
+};
+
+Panel.propTypes = {
+  arrow: Types.bool,
```
helendurranttw, line 68: In fact I feel weird about panels having arrows. If you want an arrow, wouldn't you use a popver?

packages/components/src/common/responsivePanel/ResponsivePanel.js
```diff
@@ -0,0 +1,61 @@
...
+const ResponsivePanel = forwardRef(
+  ({ arrow, children, open, preferredPlacement, triggerRef }, ref) => {
+    const [clientWidth] = typeof window !== 'undefined' ? useClientWidth({ ref: window }) : [null];
+
+    const isMobile = clientWidth && clientWidth < Breakpoint.SMALL;
+
+    return (
+      <>
+        {isMobile ? (
+          <BottomSheet open={open} ref={ref}>
```
helendurranttw, line 18: This is great, will be very useful 👍

packages/components/src/drawer/Drawer.spec.js
```diff
@@ -61,15 +61,15 @@ describe('Drawer', () => {
   });
 
   it('renders slidingPanel with right props', () => {
-    const slidingPanel = component.find('SlidingPanel');
```
helendurranttw, line 64: Does this test really prove that you've rendered a slidingPanel? 🤔

packages/components/src/popover/Popover.js
```diff
@@ -1,147 +1,105 @@
...
-    }
-  };
-
-  open = () => {
-    this.setState({ isOpen: true });
-
-    addClickClassToDocumentOnIos();
-    document.addEventListener('click', this.closePopoverOnOutsideClick, true);
+const Popover = ({ children, content, preferredPlacement, title }) => {
+  logActionRequiredIf(
```
helendurranttw, line 13: You should use the deprecatedProp utility :)

helendurranttw, line 13: You might consider moving this to the bottom (see Alert for an example). It just cleans up the component a bit.

packages/components/src/popover/Popover.js
```diff
@@ -1,147 +1,105 @@
...
-  classNames: Types.objectOf(Types.string),
+  title: Types.string,
 };
 
-Popover.defaultProps = {
-  title: null,
-  containsFocusableElement: false,
-  preferredPlacement: Popover.Placement.RIGHT,
-  classNames: {},
+export const deprecatePlacements = {
```
helendurranttw, line 98: *deprecatedPlacements

packages/components/src/popover/Popover.js
```diff
@@ -1,147 +1,105 @@
...
-
-    addClickClassToDocumentOnIos();
-    document.addEventListener('click', this.closePopoverOnOutsideClick, true);
+const Popover = ({ children, content, preferredPlacement, title }) => {
+  logActionRequiredIf(
+    `Popover has deprecated the ${preferredPlacement} value for the 'preferredPlacement' prop. Please use ${deprecatePlacements[preferredPlacement]} instead.`,
+    deprecatePlacements[preferredPlacement],
+  );
+  const [open, setOpen] = useState(false);
+  const memoizedId = useMemo(() => uniqid('np-popover-'), []);
```
helendurranttw, line 18: I think you just want useState here. From the React docs:

> You may rely on useMemo as a performance optimization, not as a semantic guarantee. In the future, React may choose to “forget” some previously memoized values and recalculate them on next render, e.g. to free memory for offscreen components.

packages/components/src/popover/Popover.js
helendurranttw: That would be a breaking change

packages/components/src/popover/Popover.js
```diff
@@ -1,147 +1,105 @@
...
-  handleKeyUp = (event) => {
-    if (event.target.nodeName !== 'BUTTON' && event.keyCode === KeyCodes.ENTER) {
-      this.toggle();
-    }
-  };
-
-  open = () => {
-    this.setState({ isOpen: true });
-
-    addClickClassToDocumentOnIos();
```
helendurranttw, line 47: I presume we still need this? Or have you checked that we no longer do?

packages/components/src/popover/Popover.js
```diff
@@ -1,147 +1,105 @@
...
-  };
-
-  toggle = () => (this.state.isOpen ? this.close() : this.open());
-
-  createTrigger = () => {
-    const { children: child, containsFocusableElement } = this.props;
-    const { isOpen } = this.state;
-    const wrappedChild = wrapInDOMElementIfNecessary(child);
-
-    const focusableProps = !containsFocusableElement && {
```
helendurranttw, line 65: You've deleted this but not replicated the behaviour, which is necessary. If the trigger is not focusable, we need to wrap it in one.

packages/components/src/popover/Popover.js
```diff
@@ -1,147 +1,105 @@
...
-      { [this.style('scale-down')]: !isOpen },
-      this.style(placement),
-    );
+  const placement = deprecatePlacements[preferredPlacement] || preferredPlacement;
+
+  return (
+    <>
+      <span className="np-popover d-inline-block" ref={triggerRef}>
+        {/* setting ref on child won't work for React components without forwardRef  */}
+        {Children.map(children, (child) => {
```
helendurranttw, line 37: I'm not sure this is the right behaviour. The old behaviour was to wrap multiple or non-focusable elements in a wrapper. I think that's probably a more sensible default than applying a click handler to every element in the children.

packages/components/src/popover/Popover.js
```diff
@@ -1,147 +1,105 @@
...
+  title: undefined,
+};
 
 Popover.propTypes = {
-  children: Types.oneOfType([Types.element, Types.string]).isRequired,
-  title: Types.oneOfType([Types.element, Types.string]),
-  content: Types.oneOfType([Types.element, Types.string, Types.func]).isRequired,
-  containsFocusableElement: Types.bool,
+  children: Types.node.isRequired,
+  content: Types.node.isRequired,
```
helendurranttw, line 80: This is a breaking change! Is there a good reason to drop support for content as a function?

helendurranttw: I see you've also removed `containsFocusableElement` - that would be another breaking change I think we should avoid. I believe this was added to support using the popover for selecting categories

packages/components/src/slidingPanel/SlidingPanel.js
```diff
@@ -1,53 +1,65 @@
...
-        'sliding-panel--fixed': slidingPanelPositionFixed,
-      },
-      'sliding-panel',
-    )}
-    appear
-    unmountOnExit
-  >
-    <div className="sliding-panel">{children}</div>
-  </CSSTransition>
+const SlidingPanel = forwardRef(
```
helendurranttw, line 9: Would this count as a breaking change? This component is exposed

helendurranttw:
A few comments on usability:

* If you try and use the bottom sheet on mobile, you'll find you can scroll the content under the dimmer. Sometimes the bottom sheet moves up and down as well
* Can we implement drag to close for the bottom sheet, the way it works on mobile?
* The bottom sheet should close when you press escape
* Popover should close when you press escape

packages/components/src/common/responsivePanel/ResponsivePanel.js
andreapiras: Happy to do that, the reason why I left placement is because popper uses placements, I left preferredPalcement in Responsive to avoid a deprecation for now, but again happy to rename it

packages/components/src/common/bottomSheet/BottomSheet.less
andreapiras: from figma 
image. Let me check if they are shared in case I ll make it a utils

packages/components/src/common/panel/Panel.js
andreapiras: I though we said we were going to revisit this once we'll reuse this for Select? Am I wrong?

packages/components/src/common/panel/Panel.less
andreapiras: It's coming from Figma. @gusso can we shared this with Bottomsheet and standardize it?

packages/components/src/common/panel/Panel.less
andreapiras: This comes from the select styles `.open>.dropdown-menu` which, I can maybe reuse

packages/components/src/common/panel/Panel.spec.js
andreapiras: Is that not covered by snapshot?

packages/components/src/drawer/Drawer.spec.js
andreapiras: I think this test is pretty useless tbh as it's testing react functionalities. Also expect(slidingPanel).toHaveLength(1); and the other props checks should be sufficient to target the sliding panel or an exact same component with the same interface

packages/components/src/popover/Popover.js
andreapiras: I know, do you think we should go ahead with it or reintroduce this code. Should the popover be opened by other elements rather than a button? I guess opening it from a link would be wrong as links are meant for navigation, opening from non interactive elements like icons would be wrong too. WDYT?

packages/components/src/popover/Popover.js
andreapiras: I did via [browserStack](https://live.browserstack.com/dashboard?utm_source=chrome&utm_medium=extension&utm_campaign=quick-launch#os=iOS&os_version=12.0&device_browser=safari&zoom_to_fit=true&full_screen=true&url=https%3A%2F%2Ftransferwise.github.io%2Fneptune-web%2Fbranch%2Fpopover-refactor%2Fcomponents%2Falerts%2FPopover&speed=1&host_ports=google.com%2C80%2C0) test on iOS and it is working as expected

packages/components/src/popover/Popover.js
andreapiras: As part of the consideration above, should we only allow buttons for this or interactive elements. Should it be a concern of Popover to wrap the element? It feel weird to me that we are delegating this to popover

packages/components/src/popover/Popover.js
andreapiras: Why would you pass multiple elements as trigger? Is it a pattern that we want to recommend. Also if you pass multiple element which element is the popover going to be attached?

packages/components/src/common/responsivePanel/ResponsivePanel.js
andreapiras: SizeSwapper swaps based on parent size, we can extend it to accept window. It also add flex classes that I don't think they are needed here. It can be extended, I don't have strong opinions about

packages/components/src/popover/Popover.js
andreapiras: As part of our alignment meeting, we said that popover should contain only html (bold, link, etc), what would it be a case to pass in a function?

andreapiras: I feel these elements should not be allowed. If I was building this component from scratch I'd not think about these types. Why should people pass in a function?

packages/components/src/slidingPanel/SlidingPanel.js
andreapiras: No it should behave as before

packages/components/src/common/panel/Panel.js
andreapiras: Panel contains all the popper logic which also manages the arrow positioning etc, I'm not sure how I would separate them

packages/components/src/popover/Popover.js
andreapiras: @dozortsev also tested it on his iphone and confirmed bottomSheet works as expected

packages/components/src/popover/Popover.js
andreapiras: I refactored the API and now Popover is a controlled stateless component that accepts open true|false, in this way consumer can implement their logic to open/close

packages/components/src/common/hooks/useIsConditionMet/useIsConditionMet.js
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @param {bool} attachListener - when to attach the event.
+ * @param {function} callBack - callBack called when a condition is met.
+ * @param {function} condition - condtion to check,
+ * @param {string} eventType - type of event to attach to document.
+ *
+ * @usage `useIsConditionMet({refs: [insideClick, triggerRef], callBack: (val) => setOpen(val)});`
+ *
+ * */
+export const useIsConditionMet = ({ attachListener, callBack, condition, eventType }) => {
```
andreapiras, line 14: I can pass an array of conditions here

packages/components/src/common/panel/Panel.js
helendurranttw: Ah ok fair enough 👍

andreapiras:
@helendurranttw 
> * If you try and use the bottom sheet on mobile, you'll find you can scroll the content under the dimmer. Sometimes the bottom sheet moves up and down as well
This is more a bug with Dimmer, I think we should look at it separately

> * Can we implement drag to close for the bottom sheet, the way it works on mobile?
In progress
> * The bottom sheet should close when you press escape
Done
> * Popover should close when you press escape
Don

packages/components/src/popover/Detached/Detached.js
andreapiras: We decided to set manage these once we refactor select

packages/components/src/popover/Detached/Detached.js
andreapiras: We decided to set manage these once we refactor select

packages/components/src/popover/Detached/Detached.js
andreapiras: renamed to position

packages/components/src/popover/Popover.js
andreapiras: this has been completely refactored

packages/components/src/common/panel/Panel.js
andreapiras: renamed

packages/components/src/common/panel/Panel.less
andreapiras: dropdown-menu has a bunch of other css that interferes with popper css

packages/components/src/common/responsivePanel/ResponsivePanel.js
andreapiras: Extended sizeswapper and used it

packages/components/src/popover/Popover.js
andreapiras: This is not used anymore

packages/components/src/popover/Popover.js
andreapiras: I check the usage on search console and found out that containsFocusableElement is used in only one instance which I reached out to owner and provided for alternative

packages/components/src/popover/Popover.js
```diff
@@ -1,147 +1,151 @@
...
-      'data-toggle': 'popover',
-      onClick: this.toggle,
-      onKeyUp: this.handleKeyUp,
-      ref: (element) => {
-        this.triggerElement = element;
-      },
-    });
-  };
-
-  style = (className) => this.props.classNames[className] || className;
```
andreapiras, line 82: Spoke with @phildobsontw about this. This code was needed because Popover was using bootstrap classes that on external sites were clashing with consumer css. New popover is not relying anymore on this classes so this code can be removed

packages/components/src/common/hooks/useClientWidth/useClientWidth.js
```diff
@@ -8,13 +8,17 @@ const useIsomorphicLayoutEffect = typeof window !== 'undefined' ? useLayoutEffec
...
-  };
-
   useIsomorphicLayoutEffect(() => {
+    const updateClientWidth = () => {
+      if (ref) {
+        // if ref is referencing to window we check the window.innerWidth
+        if (ref.innerWidth) {
+          setClientWidth(ref.innerWidth);
+        } else if (ref.current) {
+          setClientWidth(ref.current.clientWidth);
```
dozortsev, line 18:
```suggestion
        // when `ref` is a window
        if (ref.innerWidth) {
          setClientWidth(ref.innerWidth);
        }
        // when `ref` is an element
        else if (ref.current) {
          setClientWidth(ref.current.clientWidth);
```

packages/components/src/common/panel/Panel.js
andreapiras: Maybe I can change the default to false?

packages/components/src/common/panel/Panel.less
andreapiras: I'm discussing how to centralize them in neptune-tokens channel

packages/components/src/common/responsivePanel/ResponsivePanel.js
```diff
@@ -0,0 +1,64 @@
+import React, { forwardRef } from 'react';
+import Types from 'prop-types';
+
+import { Position } from '..';
+import BottomSheet from '../bottomSheet';
+import Panel from '../panel';
+import SizeSwapper from '../../sizeSwapper';
+
+const ResponsivePanel = forwardRef(({ arrow, children, open, position, triggerRef }, ref) => {
+  const windowRef = typeof window !== 'undefined' && window;
```
dozortsev, line 10: in TS world this code will be weird as `windowRef` can be `false` or `window` I would suggest here ternary operator for such things

dozortsev, line 10:
```suggestion
  const windowRef = typeof window === 'undefined' ? undefined : window;
```

dozortsev: https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx ...and not only in JSX 😉

packages/components/src/common/responsivePanel/ResponsivePanel.js
```diff
@@ -0,0 +1,64 @@
...
+    },
+    {
+      items: [
+        <Panel
+          arrow={arrow}
+          open={open}
+          position={position}
+          ref={ref}
+          triggerRef={triggerRef}
+          key="panel"
```
dozortsev, line 28: sorry, what this `key` and above one `key="bottomSheet"` for?

packages/components/src/common/panel/Panel.js
```diff
@@ -0,0 +1,73 @@
+import React, { useState, forwardRef } from 'react';
+import Types from 'prop-types';
+import classnames from 'classnames';
+import { usePopper } from 'react-popper';
+
+import { Position } from '..';
+import './Panel.css';
+
+const Panel = forwardRef(({ arrow, children, className, open, position, triggerRef }, ref) => {
```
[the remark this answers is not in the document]
dozortsev: wondering whether we could avoid having `triggerRef` at all?

have you tried pass `children` in popper?

```js
usePopper(children, popperElement, {
```

packages/components/src/popover/Popover.js
```diff
@@ -1,147 +1,151 @@
...
+    callBack,
+    attachListener: open,
+  });
+
+  useAttachEvent({
+    eventType: 'keydown',
+    condition: (event) => event.keyCode === keyCodes.ESCAPE,
+    callBack,
+    attachListener: open,
+  });
```
dozortsev, line 44: I think this logic should sit inside ResponsivePanel to enforce consistency for all dropdown/popovers

I guess we want all of them to be closed when user clicks outside and close it when user press `Esc` etc.

e.g [Select has exactly same logic](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/select/Select.js#L199-L200) and [DateLookup](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/dateLookup/DateLookup.js#L179-L180)

we don't want to define exact same `useAttachEvent` for `keydown` and `useAttachEvent` for `click` multiple times right?

andreapiras: Good point, let me move it inside ResponsivePanel

packages/components/src/common/panel/Panel.js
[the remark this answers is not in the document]
andreapiras: that doesn't work children is a completely different object

packages/components/src/common/panel/Panel.js
```diff
@@ -0,0 +1,73 @@
...
+  const [arrowElement, setArrowElement] = useState(null);
+  const [popperElement, setPopperElement] = useState(null);
+
+  const modifiers = [];
```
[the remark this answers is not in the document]
andreapiras: Sorry can you explain a bit more here?

packages/components/src/common/panel/Panel.js
```diff
@@ -0,0 +1,73 @@
...
+
+  if (arrow) {
+    modifiers.push({ name: 'arrow', options: { padding: 5, element: arrowElement } });
+    // This lets you displace a popper element from its reference element.
+    modifiers.push({ name: 'offset', options: { offset: [0, 16] } });
```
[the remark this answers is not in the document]
andreapiras: I created a const POPOVEROFFSET = [0, 16]; We said we'll sort this once we get to Select refactor

packages/components/src/common/panel/Panel.js
```diff
@@ -0,0 +1,73 @@
...
+
+  return (
+    // Popper recommends to use the popper element as a wrapper around an inner element that can have any CSS property transitioned for animations.
+    <div
+      ref={setPopperElement}
+      style={{ ...styles.popper }}
+      {...attributes.popper}
+      className={classnames('np-panel')}
+    >
+      <div
```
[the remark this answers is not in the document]
andreapiras: No it won't behave properly, you'll get css weird behaviours. Comment on line 27 also explains why

packages/components/src/common/responsivePanel/ResponsivePanel.js
andreapiras: But this exactly the outcome. I don't understand. Is it just a code style change you are suggesting?
if window is not defined 
typeof window !== 'undefined' && window; => false
if window is defined 
typeof window !== 'undefined' && window; => window

packages/components/src/common/responsivePanel/ResponsivePanel.js
andreapiras: This fixes 
image

packages/components/src/common/hooks/useAttachEvent/useAttachEvent.js
```diff
@@ -0,0 +1,29 @@
+import { useEffect } from 'react';
+
+/**
+ * useAttachEvent attaches an event to the document and return the value of a condition provided.
+ *
+ * @param {bool} attachListener - condition to be met for the listener to be attached.
+ * @param {function} callBack - callBack called when the provided condition is met.
```
helendurranttw, line 7: Callback is usually all lower case

packages/components/src/common/hooks/useAttachEvent/useAttachEvent.js
```diff
@@ -0,0 +1,29 @@
+import { useEffect } from 'react';
+
+/**
+ * useAttachEvent attaches an event to the document and return the value of a condition provided.
```
helendurranttw, line 4: Could you expand on this a little? "returns the value of a condition provided" doesn't quite make sense

packages/components/src/popover/Popover.js
```diff
@@ -1,147 +1,133 @@
...
-              : content}
-          </p>
-        </div>
-      </>
-    );
-  }
-}
+Popover.defaultProps = {
+  className: undefined,
+  position: Popover.Position.TOP,
```
helendurranttw, line 82: Is this a purposeful change from top to right? I don't see much reason to change this.

packages/docs/liveEditorCode/popover.code.js
```diff
@@ -1,14 +1,9 @@
...
+  position={Popover.Position.TOP}
+  title="Guaranteed rate"
 >
-  <Button
-    onClick={() => console.log('clicked')}
-    disabled={false}
-    block={false}
-    size={Button.Size.Medium}
-  >
+  <Button type={Button.Type.PRIMARY} onClick={() => console.log('here')}>
```
helendurranttw, line 6: Maybe change the text of the log?

packages/components/src/summary/Summary.spec.js
```diff
@@ -8,25 +8,29 @@ describe('Summary', () => {
...
-        }}
-        icon={<strong>icon</strong>}
-        status={Summary.Status.DONE}
-        title="title"
-      />,
-    );
-    expect(asFragment()).toMatchSnapshot();
+  it('renders full component', async () => {
+    let container;
+    await waitFor(() => {
```
helendurranttw, line 13: Why is this now needed?

packages/components/src/common/panel/Panel.less
andreapiras: @helendurranttw I created a ticket for that to be added in neptune-tokens

packages/components/src/summary/Summary.spec.js
andreapiras: because popover now is based on popper which performs state updates

packages/components/src/common/panel/Panel.js
```diff
@@ -0,0 +1,79 @@
+import React, { useState, forwardRef } from 'react';
+import Types from 'prop-types';
+import classnames from 'classnames';
+import { usePopper } from 'react-popper';
+
+import { Position } from '..';
+import './Panel.css';
+
+const POPOVEROFFSET = [0, 16];
```
dozortsev, line 9:
```suggestion
const POPOVER_OFFSET = [0, 16];
```

packages/components/src/common/panel/Panel.js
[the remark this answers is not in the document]
dozortsev: word "modifiers" is a bit confused at quick glance not clear what it belong or used for, for me it looks like some properties of popper so I'm suggesting call it this way 🙂

packages/components/src/common/panel/Panel.js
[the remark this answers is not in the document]
dozortsev: how about we call it `childrenRef`?

packages/components/src/common/panel/Panel.js
```diff
@@ -0,0 +1,81 @@
...
+  ({ arrow, children, className, open, position: placement, triggerRef }, ref) => {
+    const [arrowElement, setArrowElement] = useState(null);
+    const [popperElement, setPopperElement] = useState(null);
+
+    const modifiers = [];
+
+    if (arrow) {
+      modifiers.push({ name: 'arrow', options: { element: arrowElement } });
+      // This lets you displace a popper element from its reference element.
+      modifiers.push({ name: 'offset', options: { offset: POPOVEROFFSET } });
```
andreapiras, line 21:
```suggestion
      modifiers.push({ name: 'offset', options: { offset: POPOVEROFFSET } });
```
```suggestion
      modifiers.push({ name: 'offset', options: { offset: POPOVER_OFFSET } });
```

packages/components/src/common/panel/Panel.js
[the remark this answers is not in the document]
dozortsev: we mostly have used word "trigger" among our conversations with designers to make things clear but in code I would prefer call them more consistent to each other (`children` and `childrenRef`)

packages/components/src/common/panel/Panel.less
```diff
@@ -0,0 +1,59 @@
...
+    opacity: 0;
+    visibility: hidden;
+    padding: var(--size-16) var(--size-24);
+    background-color: var(--color-background-elevated);
+    border-radius: var(--radius-small);
+    box-shadow: 0px 0px 40px rgba(34, 48, 73, 0.2);
+    transition: opacity 0.2s cubic-bezier(0.6, 0.2, 0.1, 1),
+      transform 0.2s cubic-bezier(0.6, 0.2, 0.1, 1), visibility 0s linear 0s;
+    transform: translateY(-10px) scale(0.95);
+    z-index: -@zindex-popover;
```
andreapiras, line 19:
```suggestion
```

packages/components/src/common/panel/Panel.less
```diff
@@ -0,0 +1,58 @@
...
+      transform 0.2s cubic-bezier(0.6, 0.2, 0.1, 1), visibility 0s linear 0s;
+    transform: translateY(-10px) scale(0.95);
+  }
+  &--open {
+    z-index: @zindex-popover;
+    .np-panel__content {
+      transform: none;
+      opacity: 1;
+      visibility: visible;
+      height: auto;
```
andreapiras, line 26:
```suggestion
```

packages/components/src/common/responsivePanel/ResponsivePanel.js
```diff
@@ -0,0 +1,84 @@
...
+};
+
+ResponsivePanel.defaultProps = {
+  arrow: false,
+  onClose: undefined,
+  position: ResponsivePanel.Position.TOP,
+};
+
+ResponsivePanel.propTypes = {
+  onClose: Types.func,
```
dozortsev, line 71: I guess `onClose` should be required

packages/components/src/popover/Popover.js
```diff
@@ -1,131 +1,73 @@
...
-      this.style('animate'),
-      this.style('in'),
-      { [this.style('scale-down')]: !isOpen },
-      this.style(placement),
-    );
+const Popover = ({ children, className, content, preferredPlacement, title }) => {
+  const triggerRef = useRef(null);
+  const [open, setOpen] = useState(false);
+
+  const onClose = (val) => val && setOpen(false);
```
dozortsev, line 14: why we need `val &&` here?

dozortsev, line 14:
```suggestion
  const onClose = () => setOpen(false);
```

can it be like this?

packages/components/src/common/panel/Panel.js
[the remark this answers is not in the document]
andreapiras: I rather leave it that way as it is inline with Popper API

packages/components/src/common/panel/Panel.js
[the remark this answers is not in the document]
andreapiras: to be honest I'd leave as it is because now we are passing children but if we refactor and pass the trigger as prop rather than children this won't make sense anymore

packages/components/src/common/bottomSheet/BottomSheet.less
andreapiras: I created a ticket to centralize this in tokens

packages/components/src/popover/Popover.js
```diff
@@ -1,131 +1,73 @@
...
+    <span className={classnames('np-popover', className)}>
+      <span className="d-inline-block" ref={triggerRef}>
+        {cloneElement(children, {
+          onClick: () => {
+            if (children?.props?.onClick) {
+              children.props.onClick();
+            }
+            setOpen(!open);
+          },
+        })}
```
dozortsev, line 26: why this logic should be defined here rather than done by `onClick` of trigger element/component?

same as it's done e.g in [Modal](https://transferwise.github.io/neptune-web/branch/popover-refactor/components/overlays/Modal)

andreapiras: The main reason is because this component is not stateless. When I asked in the guild for feedback seemed like that majority of people wanted a "statefull" component. to avoid add boilerplate to their Apps

andreapiras: Another feedback was that stateless component may have led to inconsistent implementation regarding onClose. In this way we can force components to always behave in a consistent way

packages/components/src/common/panel/Panel.less
```diff
@@ -0,0 +1,56 @@
...
+
+.np-panel {
+  // This is needed because since the __content is visible:hidden the element will still occupy some space even if is not
+  // visible. That means that if there is an element in proximity may be not clickable because it is underneath the invisible
+  // layer.
+  z-index: -@zindex-popover;
+  &__content {
+    opacity: 0;
+    visibility: hidden;
+    padding: var(--size-16) var(--size-24);
```
dozortsev, line 11: padding should be defined by consumer of Panel component as it will be different from consumer to consumer

in your case this padding should be in `Popover.less`

packages/components/src/common/panel/Panel.spec.js
helendurranttw: Fair enough

packages/components/src/common/hooks/useClientWidth/useClientWidth.js
```diff
@@ -8,18 +8,29 @@ const useIsomorphicLayoutEffect = typeof window !== 'undefined' ? useLayoutEffec
...
+        // when `ref` is an element
+        else if (ref.current) {
+          setClientWidth(ref.current.clientWidth);
+        }
+      }
+    };
+    // This assignment saves a reference to the function so it will be the same passed to both addEventListener removeEventListener.
+    // If throttle gets passed directly to both add and removeEventListenet the results will be that the event
+    // won't get removed even if the component is unmounted.
+    const attachedFunction = throttle(updateClientWidth, throttleMs);
```
helendurranttw, line 27: 👍

packages/components/src/common/panel/Panel.js
```diff
@@ -0,0 +1,81 @@
+import React, { useState, forwardRef } from 'react';
+import Types from 'prop-types';
+import classnames from 'classnames';
+import { usePopper } from 'react-popper';
+
+import { Position } from '..';
+import './Panel.css';
+
+const POPOVER_OFFSET = [0, 16];
```
helendurranttw, line 9: Popper?

helendurranttw: Is the 16 to account for the arrow? What if there is no arrow?

packages/components/src/common/responsivePanel/ResponsivePanel.js
```diff
@@ -0,0 +1,84 @@
...
+import { Position } from '..';
+import { useConditionalListener } from '../hooks';
+import BottomSheet from '../bottomSheet';
+import Panel from '../panel';
+import SizeSwapper from '../../sizeSwapper';
+
+import keyCodes from '../keyCodes';
+
+const ResponsivePanel = ({ arrow, onClose, children, open, position, triggerRef }) => {
+  const windowRef = typeof window === 'undefined' ? undefined : window;
```
helendurranttw, line 13: How is `windowRef` different to window? 🤔

packages/components/src/common/responsivePanel/ResponsivePanel.js
```diff
@@ -0,0 +1,84 @@
...
+  children: Types.node.isRequired,
+  position: Types.oneOf([
+    ResponsivePanel.Position.TOP,
+    ResponsivePanel.Position.RIGHT,
+    ResponsivePanel.Position.LEFT,
+    ResponsivePanel.Position.BOTTOM,
+  ]),
+  arrow: Types.bool,
+  open: Types.bool,
+  triggerRef: Types.shape({ current: Types.shape({}) }).isRequired,
```
helendurranttw, line 81: It feels a bit weird passing the trigger ref all the way through. If you have to pass in the trigger, why not handle the open / close state?

Could we get away without passing this in?

packages/components/src/common/panel/Panel.js
andreapiras: If there's no arrow it will look like that 
image
Which will cover the select case. 
Maybe I should call it ARROW_OFFSET ?

packages/components/src/common/responsivePanel/ResponsivePanel.js
andreapiras: I just used a const here to group the check logic, if I pass directly window I'll get errors on SSR.
image

packages/components/src/common/responsivePanel/ResponsivePanel.js
andreapiras: Having the state here will cause that half state (open/close) will need to be managed in Popover and half state inside ResponsivePanel

gusso (approved):

packages/components/src/popover/Popover.js
dozortsev: we should consider implement same API approach in Modal component then as currently [whether it's opened or not controlled by anchor/trigger element](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/modal/Modal.js#L194)

packages/components/src/common/panel/Panel.less
```diff
@@ -0,0 +1,69 @@
...
+    .np-panel__arrow {
+      top: -6px;
+    }
+  }
+
+  &[data-popper-placement^='left'] > &__content {
+    transform: translateX(10px) scale(0.95);
+    .np-panel__arrow {
+      right: -6px;
+    }
```
dozortsev, line 67: can't these be combined? to avoid repeating of 

```css
    transform: translateX(10px) scale(0.95);
    .np-panel__arrow {
      right: -6px;
```

dozortsev (commented):
I would add badge with word "updated" to emphasise and highlight this refactoring/breaking change

similar example of same badge you can find in [`Provider.mdx`](https://github.com/transferwise/neptune-web/blob/main/packages/docs/pages/components/context/Provider.mdx) file

```js
  badge: { type: 'updated', expiryDate: '..' }
```

Screen Shot 2021-02-26 at 2 53 08 PM

packages/components/src/common/panel/Panel.js
helendurranttw: I think maybe given we're only using it once we should just inline it.

I do think the offset should be configurable, but we can add that later when it becomes necessary

packages/components/src/common/responsivePanel/ResponsivePanel.js
```diff
@@ -0,0 +1,87 @@
...
+
+import { Position } from '..';
+import { useConditionalListener } from '../hooks';
+import BottomSheet from '../bottomSheet';
+import Panel from '../panel';
+import SizeSwapper from '../../sizeSwapper';
+
+import keyCodes from '../keyCodes';
+
+const ResponsivePanel = ({ arrow, onClose, className, children, open, position, anchorRef }) => {
```
helendurranttw, line 12:
```suggestion
const ResponsivePanel = ({ anchorRef, arrow, children, className, onClose, open, position }) => {
```

packages/components/src/common/responsivePanel/ResponsivePanel.js
```diff
@@ -0,0 +1,87 @@
...
+  const windowRef = typeof window === 'undefined' ? undefined : window;
+
+  const ref = useRef(null);
+
+  useConditionalListener({
+    eventType: 'click',
+    callback: (event) =>
+      onClose(![ref, anchorRef].some((el) => el?.current?.contains(event.target))),
```
helendurranttw, line 20: You don't document anywhere what onClose receives. Would be good to document this so future consumers can use it.

packages/components/src/popover/Popover.js
```diff
@@ -136,12 +88,14 @@ Popover.propTypes = {
...
-  classNames: Types.objectOf(Types.string),
+  title: Types.node,
 };
 
-Popover.defaultProps = {
-  title: null,
-  containsFocusableElement: false,
-  preferredPlacement: Popover.Placement.RIGHT,
-  classNames: {},
+export const deprecatedPlacements = {
```
helendurranttw, line 94: Why are you exporting this?

packages/components/src/common/panel/Panel.less
andreapiras: They all have different values
transform: translateY(10px) scale(0.95);
transform: translateY(-10px) scale(0.95);
transform: translateX(10px) scale(0.95)
transform: translateX(-10px) scale(0.95)

packages/components/src/popover/Popover.less
```diff
@@ -1,11 +1,27 @@
...
+  .np-popover__content {
+    font-size: @font-size-small;
+  }
+  &.np-panel {
+    .np-popover__title {
+      .h5;
+    }
+    .np-popover__content {
+      padding: var(--size-16) var(--size-24);
+      max-width: 252px;
```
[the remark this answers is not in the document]
andreapiras: image

But this should be just width

packages/components/src/common/responsivePanel/ResponsivePanel.js
andreapiras: Changed the logic so the function only gets called when the condition is met

packages/docs/pages/components/alerts/Popover.mdx
```diff
@@ -7,4 +7,10 @@ import code from '../../../liveEditorCode/popover.code';
 
 export const meta = {
   name: 'Popover',
+  badge: {
+    type: 'Updated',
+    badge: {
+      expiryDate: '2021-03-31',
+    },
+  },
```
helendurranttw, line 15:
```suggestion
```

helendurranttw (approved):

PR #761, opened 2021-01-18
feat: add stacked option to RadioGroup component

matebek-tw:
## 🖼 Context

https://transferwise.atlassian.net/browse/HANDL-550

We are having a custom radio-group component implemented in `help-center` [contact form](https://transferwise.com/help/contact) (login required to see the component).

image

This PR is aimed to implement this behaviour so everyone can use this slightly modified version of the `RadioGroup` component if they want to.

image

## 🚀 Changes

- Add `radio-group.less` and `radio-group--stacked css selector`
- Add optional `stacked` property to `RadioGroup` component

## ✅ Checklist


oliverviljamaa:
The purpose seems similar to what we already have with [https://transferwise.github.io/neptune-web/components/inputs/RadioGroup](https://transferwise.github.io/neptune-web/components/inputs/RadioGroup).

Do we need the variation?

oliverviljamaa:
The purpose seems similar to what we already have with [RadioGroup](https://transferwise.github.io/neptune-web/components/inputs/RadioGroup).

Do we need the variation?

If we believe this is better, perhaps it should be the _only_ option?

helendurranttw:
Hey @matebek-tw, thanks so much for the contribution! As @oliverviljamaa points out, we already have a component that handles this situation.

Could you explain your rationale for implementing this variation?

Thanks!

dozortsev (commented):
Hey

thanks for considering to make this change

from above Jira ticket:

> What we see in the screenshot started as an experiment, and, in the sake of shipping faster, we decided to keep our designs by implementing them in our own codebase, instead of updating Neptune. Then, once proved that the new UI/UX was indeed better than what we had, it was fully rolled out...

could you please bring more context and/or proofs re. new design? this should help us evaluate this change and whether we should have only one design ([agree with Oliver](https://github.com/transferwise/neptune-web/pull/761#issuecomment-762600024)) for radio options 🙂

matebek-tw:
I added this option to the existing RadioGroup component, you can check it out in action [here](https://transferwise.github.io/neptune-web/branch/radio-group-stacked/components/inputs/RadioGroup) by adding the `stacked` option as a property to the component.

This variation lets you "stack" the options logically and visually together, rather than having a margin and a border separating them. We liked how it made the questions flow a bit more easier to the eyes on the contact form. Each question has its answers contained into one block. This is how our contact form would look like with or without this option available.

image
image

I agree as well that maybe this should be the default and perhaps the only option for radio groups, but that would require to update the CSS primitives around `.radio` which might have a bigger impact on other things I am not aware of, and I am not confident doing so (and not sure tbh where those styles are actually declared), but I am open to bring this to that direction with a little assistance.

gusso:
Hey @matebek-tw, we discussed on Slack a while ago that we should consider if this could be the only variant by visualising the change in other screens/features, do you have any updates if this happened?

matebek-tw:
Hey, not really, the process stuck at the [proposal](https://transferwise.atlassian.net/wiki/spaces/DS/pages/1390151773/How+to+make+a+proposal) stage, our team planning to re-think the flow where we are using these components, so I guess this won't be prioritised soon. We can close this PR though as even if we move forward with the proposal, the implementation will be completely different.

PR #762, opened 2021-01-19
DS-810 Deprecate css embedded logo

nabaroa:
## 🖼 Context

📎 https://transferwise.atlassian.net/browse/DS-810

Navigation logo is consumed through CSS, with base64 background images. The idea is to deprecate this usage and recommend the consumers to use the logos as assets.

**The reasons:**

- Keeping the logo in CSS could be convenient but it’s not the better option performance wise. To render the logo on the page, all the neptune.css file must be downloaded first, which is a render blocker resource (https://web.dev/lighthouse-total-blocking-time/ ). The background base64 logos create 15KB of extra CSS. If we use the logos as assets, max 3KB per logo and can be cached.
- Accessibility and semantic.

## 🚀 Changes

- Deprecation of logo as CSS class

## ✅ Checklist


packages/css/src/less/logos.less
```diff
@@ -0,0 +1,4 @@
+.logo__image {
+  max-width: 138px;
+  height: var(--size-24);
+}
```
anttio, line 4: Do we need this? Can't we use `width` and `height` attributes on the `img`? If we need the CSS then maybe consider moving it closer to the consumer LESS, i.e. the flowNavigation styles.

nabaroa: That would be the cleanest option for sure, but unfortunately with only the `width` attribute we can get weird sizes of the logo. 

image

This logo style is going to be used outside flowNavigation component, as in public-nav, footer, or other navigation options, so I'd leave it not scoped to a certain component.

anttio: Makes sense. Could we then rethink the naming of this class? Something more specific than `.logo__image` as I can see this ending up used on other logos than the site/brand logo.

nabaroa:
I removed the change from FlowNavigation from this PR. It has been addressed in another task.

packages/css/src/less/navbar.less
```diff
@@ -17,7 +17,6 @@
 @import (reference) 'bootstrap/less/mixins/tab-focus.less';
 @import 'bootstrap/less/utilities.less';
 @import 'bootstrap/less/dropdowns.less';
-@import 'bootstrap/less/logos.less';
```
andreapiras, line 20: Is anyone using this css? This could be a Breaking change

nabaroa: Yes, it is a breaking change. I added the definition in the commit description. Do I need to do something else?

andreapiras: We may need to merge this into next? @helendurranttw

andreapiras (approved):

PR #766, opened 2021-01-22
fix: fixes for flownavigation

andreapiras:
## 🖼 Context

This PR removed Logo Full component and its usage
## 🚀 Changes

- Removed Logo full from FlowNavigation and Overlay Header
- Removed x padding for Stepper in FlowNavigation mobile
- Add p-t-3 for FlowNavigation and OverlayHeader
- Reintroduced AvatarWrapper

## 🤔 Considerations

This should be a Breaking change but due to the fact that this change has been just introduced and very likely never adopt it we ll release as minor

## ✅ Checklist


packages/components/src/flowNavigation/FlowNavigation.js
```diff
@@ -35,7 +35,6 @@ const FlowNavigation = ({ activeStep, avatar, logo, onClose, onGoBack, steps })
         />
       )}
       <Logo
```
dozortsev, line 37: why we still need keep this component here

packages/components/src/index.js
```diff
@@ -57,4 +57,3 @@ export { default as Switch } from './switch';
 export { default as Decision } from './decision';
 export { default as Nudge } from './nudge';
 export { default as InlineAlert } from './inlineAlert';
-export { default as Logo } from './common/logo';
```
dozortsev, line 60: should we now also delete this component?

dozortsev (commented):
if we don't need to delete Logo component from FlowNav then implementation of FlowNav is a bit weird:

```jsx
leftContent = {
  <>
    <div className="hidden-xs">{logo}</div>
    <div className="visible-xs">
      {onGoBack && activeStep > 0 && (
        <BackButton ... />
      )}
      <Logo
        className={classNames('np-flow-navigation--logo', {
          'np-flow-navigation--logo__hidden': activeStep,
          'np-flow-navigation--logo__display': !activeStep,
        })}
      />
    </div>
  </>
}
```

---

from PR description:

> ..This should be a Breaking change but due to the fact that this change has been just introduced and very likely never adopt it we ll release as major

do you actually meant minor? ([major.minor.patch])

packages/components/src/flowNavigation/FlowNavigation.js
andreapiras: This return the flag logo, we decided to keep it for now because the new logo contains a flag

packages/components/src/index.js
andreapiras: We still need it to output the flag

andreapiras:
> if we don't need to delete Logo component from FlowNav [#766 (comment)](https://github.com/transferwise/neptune-web/pull/766#discussion_r563718817) then implementation of FlowNav is a bit weird:
> 
> ```js-jsx
> leftContent = {
>   <>
>     <div className="hidden-xs">{logo}</div>
>     <div className="visible-xs">
>       {onGoBack && activeStep > 0 && (
>         <BackButton ... />
>       )}
>       <Logo
>         className={classNames('np-flow-navigation--logo', {
>           'np-flow-navigation--logo__hidden': activeStep,
>           'np-flow-navigation--logo__display': !activeStep,
>         })}
>       />
>     </div>
>   </>
> }
> ```
> 
> from PR description:
> 
> > ..This should be a Breaking change but due to the fact that this change has been just introduced and very likely never adopt it we ll release as major
> 
> do you actually meant minor? ([major.minor.patch])

yes sorry minor.

Regarding the implementation there is animation applied to the logo that makes it roll when the step is zero. 
What do you find weird exactly?

dozortsev:
> ..What do you find weird exactly?

```jsx
leftContent = {
  <>
    <div className="hidden-xs">{logo}</div>
    ...
      <Logo
        className={classNames('np-flow-navigation--logo', {
          'np-flow-navigation--logo__hidden': activeStep,
          'np-flow-navigation--logo__display': !activeStep,
        })}
      />
    </div>
  </>
}
```

logo now is coming from two different sources, aren't you finding this weird?

packages/components/src/flowNavigation/FlowNavigation.js
dozortsev: how this flag is different from other logos assets? why it should sit inside component while others should be passed via prop `logo`?

packages/components/src/common/closeButton/CloseButton.less
```diff
@@ -1,6 +1,14 @@
 .np-close-button {
```
andreapiras, line 1: alert css was overriding FN css

packages/components/src/flowNavigation/FlowNavigation.less
```diff
@@ -10,10 +10,13 @@
 
   &__stepper {
     .p-b-0;
-    .p-x-3;
+    .p-x-0;
     @media (--screen-sm) {
       .p-x-5;
     }
+    @media (--screen-lg) {
+      max-width: 562px;
```
andreapiras, line 18: previous FN had limited width

packages/components/src/alert/Alert.js
```diff
@@ -4,7 +4,7 @@ import classNames from 'classnames';
 import requiredIf from 'react-required-if';
 import { InfoCircle, CheckCircle, CrossCircle, Alert as AlertIcon } from '@transferwise/icons';
 import { Sentiment, Size } from '../common';
-import CloseButton from './closeButton';
+import CloseButton from '../common/closeButton';
```
andreapiras, line 7: I merged the two closeButton because they had the same className and they override each other

packages/components/src/common/closeButton/CloseButton.js
```diff
@@ -8,27 +8,40 @@ import messages from './CloseButton.messages';
...
   const intl = useIntl();
+  const { onClick, className, size } = props;
   return (
     <button
+      ref={ref}
       type="button"
-      className={classNames('np-close-button', 'btn-link', 'text-no-decoration', className)}
+      className={classNames(
+        'np-close-button close btn-link text-no-decoration',
+        { 'np-close-button--large': size === CloseButton.Size.LARGE },
```
andreapiras, line 20: This logic was actually missing in both closebutton

helendurranttw: When would we have a large close icon?

packages/components/src/flowNavigation/FlowNavigation.js
helendurranttw: Maybe one day we can take it out and make it completely configurable, but for now since we only have one use case for this component and it will always be a flag, let's leave it there to ease the burden on our consumers.

packages/components/src/common/closeButton/CloseButton.js
andreapiras: in alert

packages/components/src/flowNavigation/FlowNavigation.js
dozortsev: 👍

packages/components/src/alert/Alert.js
dozortsev: 👏

dozortsev (approved):
looks good

sbrandwoo:
Should this PR also add the bottom border back? 

I think we are also missing some padding on either side.

sbrandwoo:
To be a bit more specific, I'd like us to add back all styling that all consumers will need to re-add to the component. This includes the bottom border. If we consumers need to do it manually we'll probably end up with a bottom border on send and balances, and nowhere else.

For padding it should include the same horizontal padding as it had before, it currently touches the edges of the screen:
Selection_215

Also worth nothing that the stepper is now vertically aligned and the profile icon is not vertically aligned, this makes it look unbalanced when there are no steps:

Selection_216

packages/components/src/common/logo/Logo.js
```diff
@@ -1,38 +0,0 @@
-import React from 'react';
```
fergusjordantw, line 1: I still think it would be a good idea to export a `LogoWrapper` component which will let consumers easily plug any logo into which gracefully takes care of sizing

andreapiras: Hey @fergusjordantw what the component would do? Just setting the size?

andreapiras:
> To be a bit more specific, I'd like us to add back all styling that all consumers will need to re-add to the component. This includes the bottom border. If consumers need to do it manually we'll probably end up with a bottom border on send and balances, and nowhere else.
> 
> For padding it should include the same horizontal padding as it had before, it currently touches the edges of the screen:
> Selection_215
> 
> Also worth noting that the stepper is now vertically aligned and the profile icon is not vertically aligned, this makes it look unbalanced when there are no steps:
> 
> Selection_216

Hey @sbrandwoo , I added the padding and the border. If you don't need the stepper and the border please use OverlayHeader

andreapiras:
> horizontal padding

Hey @sbrandwoo, So I'm adding the horizontal padding and I added back the done property to show hide the border bottom

packages/components/src/common/logo/Logo.js
fergusjordantw: Resolved offline

conorcussell (approved):

packages/docs/pages/components/navigation/FlowNavigation.mdx
```diff
@@ -1,12 +1,12 @@
 import { LiveEditorBlock, GeneratePropsTable } from '../../../utils';
-import { FlowNavigation, Avatar, Logo } from '@transferwise/components';
+import { FlowNavigation, Avatar } from '@transferwise/components';
 import { Profile as ProfileIcon, Briefcase as BriefcaseIcon } from '@transferwise/icons';
 
 import code from '../../../liveEditorCode/flowNavigation.code';
```
helendurranttw, line 5: Might be a good idea to add the ideal image dimensions here?

packages/docs/pages/components/overlays/OverlayHeader.mdx
```diff
@@ -1,13 +1,13 @@
 import { LiveEditorBlock, GeneratePropsTable } from '../../../utils';
-import { OverlayHeader, Avatar, Logo } from '@transferwise/components';
+import { OverlayHeader, Avatar } from '@transferwise/components';
 import { Briefcase as BriefcaseIcon, Profile as ProfileIcon } from '@transferwise/icons';
 import code from '../../../liveEditorCode/overlayHeader.code';
 
 Simple header component that displays a responsive logo an avatar and a close button.
```
helendurranttw, line 6: And here

packages/components/src/common/avatarWrapper/AvatarWrapper.js
```diff
@@ -0,0 +1,48 @@
+import React, { useState, useEffect } from 'react';
```
helendurranttw, line 1: You should add this back to the docs, and maybe use it in the example in docs?

sbrandwoo:
> If you don't need the stepper and the border please use OverlayHeader

We do need the border. For example we use FlowNavigation with no steps as a loading screen until the flow has loaded, then the steps are added. We also have a lot of flows that don't use steps because they are so short - I'd prefer if all the flows looked the same still.

andreapiras:
> > If you don't need the stepper and the border please use OverlayHeader
> 
> We do need the border. For example we use FlowNavigation with no steps as a loading screen until the flow has loaded, then the steps are added. We also have a lot of flows that don't use steps because they are so short - I'd prefer if all the flows looked the same still.

Ok just added and renamed the prop as showBorder which by default is true and you can control the border with it.

packages/components/src/flowNavigation/FlowNavigation.less
```diff
@@ -1,19 +1,32 @@
 @import (reference) '@transferwise/neptune-css/src/less/addons/_spacing-utilities.less';
 
+@flag: url(data:image/svg+xml;base64,PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0idXRmLTgiPz4KPCFET0NUWVBFIHN2ZyBQVUJMSUMgIi0vL1czQy8vRFREIFNWRyAxLjEvL0VOIiAiaHR0cDovL3d3dy53My5vcmcvR3JhcGhpY3MvU1ZHLzEuMS9EVEQvc3ZnMTEuZHRkIj4KPHN2ZyB2ZXJzaW9uPSIxLjEiIGlkPSJ0cmFuc2Zlcndpc2UtZmxhZyIgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIiB4bWxuczp4bGluaz0iaHR0cDovL3d3dy53My5vcmcvMTk5OS94bGluayIgeD0iMHB4IgoJIHk9IjBweCIgdmlld0JveD0iMCAwIDQ4IDUyIiBlbmFibGUtYmFja2dyb3VuZD0ibmV3IDAgMCA0OCA1MiIgeG1sOnNwYWNlPSJwcmVzZXJ2ZSI+CjxwYXRoIGZpbGw9IiMwMEM2RkYiIGQ9Ik01LjgsMGw4LjIsMTMuNkwwLjEsMjcuMmgyNC4xbDIuNi01LjRIMTNsOC4yLTguM2wtNS04aDIyLjZMMTkuMiw1Mmg2LjdsMjItNTJINS44eiIvPgo8L3N2Zz4=);
+
 .np-flow-navigation {
-  min-width: 320px;
-  .p-b-1;
+  width: 100%;
+  min-width: 100%;
```
andreapiras, line 7:
```suggestion
  
```

packages/components/src/flowNavigation/FlowNavigation.less
```diff
@@ -1,19 +1,32 @@
 @import (reference) '@transferwise/neptune-css/src/less/addons/_spacing-utilities.less';
 
+@flag: url(data:image/svg+xml;base64,PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0idXRmLTgiPz4KPCFET0NUWVBFIHN2ZyBQVUJMSUMgIi0vL1czQy8vRFREIFNWRyAxLjEvL0VOIiAiaHR0cDovL3d3dy53My5vcmcvR3JhcGhpY3MvU1ZHLzEuMS9EVEQvc3ZnMTEuZHRkIj4KPHN2ZyB2ZXJzaW9uPSIxLjEiIGlkPSJ0cmFuc2Zlcndpc2UtZmxhZyIgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIiB4bWxuczp4bGluaz0iaHR0cDovL3d3dy53My5vcmcvMTk5OS94bGluayIgeD0iMHB4IgoJIHk9IjBweCIgdmlld0JveD0iMCAwIDQ4IDUyIiBlbmFibGUtYmFja2dyb3VuZD0ibmV3IDAgMCA0OCA1MiIgeG1sOnNwYWNlPSJwcmVzZXJ2ZSI+CjxwYXRoIGZpbGw9IiMwMEM2RkYiIGQ9Ik01LjgsMGw4LjIsMTMuNkwwLjEsMjcuMmgyNC4xbDIuNi01LjRIMTNsOC4yLTguM2wtNS04aDIyLjZMMTkuMiw1Mmg2LjdsMjItNTJINS44eiIvPgo8L3N2Zz4=);
+
 .np-flow-navigation {
-  min-width: 320px;
-  .p-b-1;
+  width: 100%;
+
```
andreapiras, line 7:
```suggestion
```

PR #767, opened 2021-01-22
[USEC-2872] Simplify RadioButton HTML by using a span.

conorcussell:
## 🖼 Context

While trying to fix a bug in the registration page, I ran into some strange behaviour with the RadioButton component. 

The `RadioGroup` is rendered on the server-side with one selectedValue "personal". When client rendering takes over and updates the props to select "business" - the `className` + `aria-pressed` attributes on the html for the element `button.tw-radio-button` are not updated, which leads to an incorrect UI.

Screenshot 2021-01-22 at 17 01 21

Screenshot 2021-01-22 at 17 06 20

I'm suspecting it has something to do with keys and React thinking the DOM node doesn't need to be re-rendered but can't quite understand what the issue was after much fiddling.

Anyway, while looking into it I noticed that the accessibility panel was complaining (I think because the button has a negative tabindex property this is a false negative and it's actually fine in the real world) but switching the button to a span actually fixed my issue as well as making storybook plugin a11y happy. 

Screenshot 2021-01-22 at 16 51 43
Screenshot 2021-01-22 at 16 52 12

## 🚀 Changes

- [x] Simplifies the `RadioButton` component by using a `span` instead of a `button` and removing redundant `aria-pressed` and `tabindex` attributes.

## ✅ Checklist


packages/components/src/common/RadioButton/RadioButton.less
```diff
@@ -7,3 +7,11 @@ input[type='radio'] {
     }
   }
 }
+
+.tw-radio-button.disabled {
+  background-color: var(--gray-base);
```
conorcussell, line 12: This variable doesn't exist in neptune-tokens but it's needed to keep the background colour of the radio button the same as it is now. I can add it to that project if you would prefer?

packages/components/src/common/RadioButton/RadioButton.spec.js
```diff
@@ -10,25 +10,17 @@ describe('Radio button', () => {
   });
 
   describe('by default', () => {
-    it('has aria-pressed as false', () => {
```
dozortsev, line 13: would you mind re-write this test for this component using testing-library rather than updating Enzyme one? 🙂

the test for this component shouldn't be big, shouldn't take much of your time 😇

(we try to get rid of Enzyme https://github.com/transferwise/neptune-web/blob/main/packages/components/CONTRIBUTING.md#testing)

packages/components/src/common/RadioButton/RadioButton.less
dozortsev: cc @andreapiras

dozortsev (approved):
looks good for me

packages/components/src/common/RadioButton/RadioButton.less
dozortsev: cc @andreapiras

packages/components/src/common/RadioButton/RadioButton.js
```diff
@@ -16,16 +16,9 @@ const RadioButton = ({ id, value, name, checked, onChange, disabled, readOnly })
       onChange={() => (!checked ? onChange(value) : null)}
```
dozortsev, line 16: can we put this `input` element inside `span.tw-radio-button`? to make markup of this component aligned with very similar component - [CheckboxButton](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/checkboxButton/CheckboxButton.js#L8-L13)

packages/components/src/common/RadioButton/RadioButton.js
```diff
@@ -16,16 +16,9 @@ const RadioButton = ({ id, value, name, checked, onChange, disabled, readOnly })
...
     />
-    <button
-      type="button"
-      className={classNames('tw-radio-button', { checked })}
-      disabled={disabled || readOnly}
-      aria-pressed={checked}
-      tabIndex="-1"
-      onClick={() => (!checked ? onChange(value) : null)}
-    >
+    <span className={classNames('tw-radio-button', { checked, disabled: disabled || readOnly })}>
```
andreapiras, line 19: How is the onClick triggered now?

packages/components/src/common/RadioButton/RadioButton.less
andreapiras: @conorcussell According to this table https://github.com/transferwise/neptune-web/blob/main/packages/css/docs/upgrading.md 
The replacement is `--color-background-neutral`

conorcussell: Amazing, thanks @andreapiras will update that

packages/components/src/common/RadioButton/RadioButton.js
dozortsev: it's on `input` element (see line 16) 

don't know why it was duplicated previously

conorcussell: Yeah, this component doesn't actually work if it's used without a label, but since it's an internal component I think that's fine. All it's consumers wrap it in a label

packages/components/src/common/RadioButton/RadioButton.less
conorcussell: Fixed in https://github.com/transferwise/neptune-web/pull/767/commits/1cca1f786e34fe4cf5592ba1406d36c00ba25d57

packages/components/src/common/RadioButton/RadioButton.spec.js
conorcussell: Fixed in https://github.com/transferwise/neptune-web/pull/767/commits/6d18550bacd27f45ef2cd7d35e416530ac8766de

packages/components/src/common/RadioButton/RadioButton.js
conorcussell: Fixed in https://github.com/transferwise/neptune-web/pull/767/commits/b62bc5df0ad32275401c0033e93eb0f0593bc79b

packages/components/src/common/RadioButton/RadioButton.spec.js
```diff
@@ -1,108 +1,100 @@
...
+            <span
+              class="tw-radio-check"
+            />
+          </span>
+        </span>
+      </div>
+    `);
   });
 
-  describe('by default', () => {
```
conorcussell, line 12: This whole rendering expectation is now covered with a snapshot test

packages/components/src/common/RadioButton/RadioButton.spec.js
```diff
@@ -1,108 +1,100 @@
...
+            <span
+              class="tw-radio-check"
+            />
+          </span>
+        </span>
+      </div>
+    `);
   });
 
-  describe('when checked', () => {
```
conorcussell, line 34: This expectation is covered by a snapshot on line 34

packages/components/src/common/RadioButton/RadioButton.spec.js
```diff
@@ -1,108 +1,100 @@
...
+            <span
+              class="tw-radio-check"
+            />
+          </span>
+        </span>
+      </div>
+    `);
   });
 
-  describe('when disabled', () => {
```
conorcussell, line 52: This expectation is covered by a snapshot on line 61

packages/components/src/common/RadioButton/RadioButton.spec.js
```diff
@@ -1,108 +1,100 @@
...
-    it('has disabled input', () => {
-      expect(inputIsDisabled()).toBe(true);
-    });
-
-    it('has disabled button', () => {
-      expect(buttonIsDisabled()).toBe(true);
-    });
-  });
-
-  it('has id when passed', () => {
```
conorcussell, line 66: The 3 following expectations are covered by the default snapshot

packages/components/src/common/RadioButton/RadioButton.js
dozortsev: no, sorry, I meant placing `input` inside already existing `span`, no need create extra elements

something something like this:

```jsx
<span className={classNames('tw-radio-button', { checked, disabled: disabled || readOnly })}>
    <input
      type="radio"
      className="sr-only"
     ...
     <span className="tw-radio-check" />
</span>     
```

packages/components/src/common/RadioButton/RadioButton.spec.js
```diff
@@ -1,108 +1,100 @@
...
 
 describe('Radio button', () => {
-  let component;
-  beforeEach(() => {
-    component = shallow(<RadioButton name="a-name" />);
+  it('renders correctly', () => {
+    const { container } = render(<RadioButton id="an-id" name="a-name" value="a-value" />);
+
+    expect(container).toMatchInlineSnapshot(`
+      <div>
```
dozortsev, line 11: we tend to store snapshot files as separate files, could you please follow this convention?

packages/components/src/common/RadioButton/RadioButton.js
conorcussell: This won't work, in order to style it correctly the span needs to be adjacent to the input
```
input[type='radio'] {
  &:focus,
  &:active {
    & + .tw-radio-button {
      border-color: var(--color-border-focus);
      outline: none;
    }
  }
}

PR #768, opened 2021-01-25
feat: markdown tooltips for dynamic flows

felixxiatransferwise:
## 🖼 Context

We've updated the tooltips proposal https://transferwise.atlassian.net/wiki/spaces/DF/pages/1535679076/DFv3+Tooltips, this PR implements the decision.

## 🚀 Changes

 
`schema.help` will show up as popover (expands on click):
Screenshot 2021-01-25 at 13 04 56

`schema.description` will show up as `InlineAlert` on focus:
Screenshot 2021-01-25 at 13 05 12

## ✅ Checklist


packages/dynamic-flows/src/jsonSchemaForm/basicTypeSchema/BasicTypeSchema.js
```diff
@@ -99,6 +103,16 @@ const BasicTypeSchema = (props) => {
               {props.schema.title}
             </label>
           )}
+          {hasHelp && (
+            <Popover
+              content={<Markdown>{props.schema.help.markdown}</Markdown>}
+              preferredPlacement={Popover.Placement.TOP}
+            >
+              <span className="m-l-1">
```
felixxiatransferwise, line 111: There is a an issue with the "dotted" lines underscoring the icon (see screenshots in description). If we add this style to the icon styles, that dotted line extends to the title. Using this approach under that issue is resolved.

packages/dynamic-flows/src/jsonSchemaForm/basicTypeSchema/BasicTypeSchema.js
```diff
@@ -85,11 +87,13 @@ const BasicTypeSchema = (props) => {
     'has-error':
       ((props.submitted || !changed) && !!props.errors) ||
       ((props.submitted || (changed && blurred)) && validations.length),
-    'has-info': focused && props.schema.help,
+    'has-info': focused && props.schema.description,
```
JulianG-TW, line 90: Probably makes no difference, but we probably want a boolean here. Should we use `!!` ?
```
`'has-info': focused && !!props.schema.description,`
```

JulianG-TW: I think the `has-error` can also result in the expression evaluating to `validations.length` which is a Number instead of boolean.

packages/dynamic-flows/src/jsonSchemaForm/basicTypeSchema/BasicTypeSchema.js
```diff
@@ -2,6 +2,8 @@ import React, { useState, useEffect } from 'react';
 import Types from 'prop-types';
 import classNames from 'classnames';
 
+import { Markdown, Popover } from '@transferwise/components';
+import { HelpCircle as HelpIcon } from '@transferwise/icons';
```
JulianG-TW, line 6: I think I would make a custom component for the Popover.
Perhaps even better if it's in a separate file, so that `BasicTypeSchema` doesn't have to import these "implementation details".

JulianG-TW (approved):

packages/dynamic-flows/src/jsonSchemaForm/basicTypeSchema/BasicTypeSchema.js
felixxiatransferwise: 👍  very good spot! thanks, fixed

PR #779, opened 2021-01-29
docs: update setup docs

helendurranttw:
Updating docs for translations and the provider - [READ HERE](https://transferwise.github.io/neptune-web/branch/setup-docs/about/Setup#translations-and-the-provider).


dozortsev (approved):
okey

dozortsev (commented):
tests seems failed due to out of memory issue

it's safe to merge it into `i18n-follow-up` branch anyway

PR #780, opened 2021-01-29
Bump peer dependencies

JulianG-TW:
## 🖼 Context

Some packages have outdated peerDependencies. I think this is because these are not automatically set by `lerna publish`.

## 🚀 Changes

This PR bumps three peerDependencies to match the version in devDependencies.

## 🤔 Considerations

Can we change the Lerna config to do this automatically?
I found something [in the docs](https://github.com/lerna/lerna/tree/main/commands/publish#--graph-type-alldependencies), but it wasn't very clear, and I'm not sure if it's the right thing to do.

## ✅ Checklist

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -82,7 +82,7 @@
   },
   "peerDependencies": {
     "@transferwise/icons": "^2.6.0",
-    "@transferwise/neptune-css": "^5.0.1",
+    "@transferwise/neptune-css": "^7.0.1",
     "currency-flags": "^2.1.1",
     "prop-types": "^15.7.2",
     "react": "^16",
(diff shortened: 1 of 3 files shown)


helendurranttw:
Hey @JulianG-TW , we're doing a release today so will update this with the correct versions and merge when we're done.

dozortsev (commented):
@JulianG-TW 

>  Some packages have outdated peerDependencies. I think this is because these are not automatically set by lerna publish.

thanks for looking at this, but we have automation for that, example of one latest post release commits https://github.com/transferwise/neptune-web/commit/9250804c3521e158510466f2d64b0888e835722f

dozortsev:
https://github.com/transferwise/neptune-web/pull/780#pullrequestreview-579362958

PR #784, opened 2021-02-01
chore: bump circleci resource class to try and stop our oom errors

helendurranttw:
What it says on the tin.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -76,7 +76,7 @@ aliases:
 jobs:
   setup:
     <<: *container
-    resource_class: xlarge
+    resource_class: large
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -76,6 +76,7 @@ aliases:
 jobs:
   setup:
     <<: *container
+    resource_class: large
```
dozortsev, line 79: seems only small/medium plans available for us https://circleci.com/docs/2.0/configuration-reference/#resource_class

medium is default one so we hit the edge, need find another solution

helendurranttw:
Superseded by https://github.com/transferwise/neptune-web/pull/792

PR #782, opened 2021-01-29
feat: Use default value in OneOf const schemas

JulianG-TW:
## 🖼 Context

The default value for const schemas within a `OneOf` was not being used.

Consider the following schema where the default value is "EUR":

```json
{
  "type": "object",
  "displayOrder": ["currency"],
  "properties": {
    "currency": {
      "title": "Currency",
      "type": "string",
      "oneOf": [
        { "title": "EUR", "icon": { "name": "flag-eur" }, "description": "Euro", "const": "EUR" },
        {
          "title": "GBP",
          "icon": { "name": "flag-gbp" },
          "description": "British pound",
          "const": "GBP"
        },
        {
          "title": "USD",
          "icon": { "name": "flag-usd" },
          "description": "United States dollar",
          "const": "USD"
        },
        {
          "title": "ARS",
          "icon": { "name": "flag-ars" },
          "description": "Argentine peso",
          "const": "ARS"
        },
        {
          "title": "AUD",
          "icon": { "name": "flag-aud" },
          "description": "Australian dollar",
          "const": "AUD"
        }
      ],
      "validationMessages": { "required": "Please enter currency." },
      "refreshFormOnChange": true,
      "default": "EUR"
    }
  },
  "required": ["currency"]
}
```
When this schema is used in a `JsonSchemaForm` without a model, the drop down does not pre-select the default value:

Screenshot 2021-01-29 at 17 18 30

## 🚀 Changes

We now use the default value when there's no model and the child-schemas are const.

Screenshot 2021-01-29 at 17 18 35

## ✅ Checklist

- [x] Make PR title meaningful and follow [the commit lint format](https://github.com/transferwise/neptune-web/blob/master/CONTRIBUTING.md#versioning-and-commit-lint) (it will appear in the changelog)
- [x] Changes are tested and all tests pass
- [x] Changes meet [accessibility standards](https://github.com/transferwise/neptune-web/blob/main/ACCESSIBILITY.md) and there are no violations in the console
- [x] Changes work in all supported browsers (don't forget IE11)
- [x] You've updated the documentation if necessary

PR #788, opened 2021-02-01
Broadcasting default value for  OneOfSchema during initialisation

JulianG-TW:
## 🖼 Context

In #782 we started using the default value in OneOf schemas, but we forgot to dispatch `onChange` during initialisation. 

As a result the model is not updated, and if the form is submitted without changes, validation errors will show up:

validation error displaying for currency field even when there's a currency selected

## 🚀 Changes

We're dispatching `onChange` when the schema "changes", if there's no model and there's a valid default value.

## ✅ Checklist


packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.js
```diff
@@ -106,6 +108,12 @@ const OneOfSchema = (props) => {
 
   // When the schema we receive from parent changes
   useEffect(() => {
+    const modelIndex = getValidIndexFromModel(props.schema, props.model);
+    const defaultIndex = getValidIndexFromDefault(props.schema);
+    if (modelIndex < 0 && defaultIndex >= 0) {
```
felixxiatransferwise, line 113: shall we make this condition `modelIndex == -1` instead of `< 0`, doesn't make a difference but feels clearer to me that it's just not found.

felixxiatransferwise (approved):

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.js
JulianG-TW: changed it! thanks!

PR #790, opened 2021-02-02
fix: flow navigation alignments

andreapiras:
## 🖼 Context

This PR brings back some alignments from old version until some more in depth design work will be made on FN

## ✅ Checklist


packages/components/src/flowNavigation/FlowNavigation.js
```diff
@@ -53,11 +59,16 @@ const FlowNavigation = ({ activeStep, avatar, logo, onClose, onGoBack, done, ste
...
-          'np-flow-navigation--large': !isSmall,
-          'np-flow-navigation--small': isSmall,
-        })}
+        className={classNames(
+          'np-flow-navigation__content p-x-3',
+          {
+            'np-flow-navigation--hidden': !clientWidth,
+            'xs-max': clientWidth < Breakpoint.SMALL,
+          },
+          Object.keys(containerBreakpoints).filter(
```
fergusjordantw, line 68: Just a thought but should these be prefixed by `np-flow-navigation--` for namespacing safety?

packages/components/src/flowNavigation/FlowNavigation.js
```diff
@@ -12,12 +12,18 @@ import { useClientWidth } from '../common/hooks';
 
 import './FlowNavigation.css';
 
+// Size switches on parent container which may or may not have the same size as the window.
+const containerBreakpoints = {
+  sm: Breakpoint.SMALL,
+  lg: Breakpoint.LARGE,
+  xl: Breakpoint.EXTRA_LARGE,
```
fergusjordantw, line 19: With this being 1200 and a max-width of 1164, won't this never be the case?

fergusjordantw (commented):
Small things

packages/components/src/flowNavigation/FlowNavigation.js
andreapiras: This is based on the parent container that currently is 100% of window

packages/components/src/flowNavigation/FlowNavigation.js
andreapiras: Yup makes sense, they are still defined inside np-flow-navigation. I wanted to eventually make them reusable but it probably a bit premature

fergusjordantw (approved):
Looks good

PR #792, opened 2021-02-02
chore: CI build speed up as a fix for OOM issues

dozortsev:
## 🖼 Context

Our CI build recently have started constantly struggling of OOM issues on top of that it's been already felt a bit slow.

## 🚀 Changes

- Separated code linting task from tests, now it is two separate jobs, this should make more resources (memory) for linting job (which was mostly main cause of OOM).  Also, both jobs now run simultaneously which is faster.

- Added `--maxWorkers=4` and `--runInBand` to all Jest runners. I played with `--maxWorkers=4` vs `--runInBand` and first option makes tests run much faster while `--runInBand` is good for linting (check [Jest docs](https://jestjs.io/docs/en/troubleshooting#tests-are-extremely-slow-on-docker-andor-continuous-integration-ci-server) for more details).

- Removed redundant `*persist-workspace` from some jobs which also saves about 5-20s on each job.

- Separated bundlesize check from tests.

---

Note: I considered to give more resources for job(s) (via [`resource_class`](https://circleci.com/docs/2.0/configuration-reference/#resource_class)) and enable [`parallelism`](https://circleci.com/docs/2.0/configuration-reference/#parallelism) but these approaches are not accessible for us due to limitations of free plan which we currently use.

## Results

Before             |  After
:-------------------------:|:-------------------------:
  |  

Note: usually results vary from build to build.

Here you can see significant speed up for linting and testing job. Plus potentially fix for OOM (although during testing CI build I haven't encountered OOM I'm still not sure for 100%, think that we need release changes and see them in practice with "real" PRs in day to day life).

See all CI test builds on this branch here: https://app.circleci.com/pipelines/github/transferwise/neptune-web?branch=ci-improvements

## ✅ Checklist


.circleci/config.yml
```diff
@@ -45,8 +45,21 @@ aliases:
   - &test
     name: Test packages
     command: yarn test
-    environment:
-      NODE_OPTIONS: --max_old_space_size=4096
+
+  - &test-exp
```
andreapiras, line 49: what is exp for?

package.json
```diff
@@ -62,7 +62,7 @@
   "scripts": {
     "setup": "yarn install --frozen-lockfile && yarn build",
     "start": "echo 'Run yarn setup to install dependencies and build all packages.'",
-    "build": "lerna run build --stream",
```
andreapiras, line 65: This will give you messages on the actual tasks why has it been removed?

packages/components/package.json
```diff
@@ -34,12 +34,12 @@
     "build:copy-files2": "cpx 'src/**/!(db)/*.{css,json,svg}' build/es/no-polyfill",
     "build:copy-lang": "cpx 'i18n/*.json' build/i18n && cpx 'i18n/*.json' build/es/i18n && cpx 'i18n/index.js' build/i18n",
     "docs": "yarn build",
-    "test": "jest --env=jsdom --runInBand --config=../../node_modules/@transferwise/test-config/jest.config.js && yarn test:lint",
+    "test": "jest --env=jsdom --runInBand --config=../../node_modules/@transferwise/test-config/jest.config.js",
     "test:watch": "jest --watch --env=jsdom --config=../../node_modules/@transferwise/test-config/jest.config.js",
-    "test:lint": "jest --config ../../node_modules/@transferwise/test-config/test/jest.lint.js",
+    "lint": "jest --config ../../node_modules/@transferwise/test-config/test/jest.lint.js",
```
andreapiras, line 39: I think this one should be test:lint and line 41 lint

packages/docs/package.json
```diff
@@ -16,9 +16,9 @@
     "docs:deploy": "npm-run-all deploy:*",
     "deploy:export": "next export -o docs-build && touch docs-build/.nojekyll",
     "deploy:gh-pages": "deploy-to-github-pages -d docs-build -m main",
-    "lint": "eslint components/ pages/ utils/",
-    "lint:fix": "yarn lint --fix",
-    "test": "jest --config ../../node_modules/@transferwise/test-config/test/jest.lint.js"
+    "lint-temp": "eslint components/ pages/ utils/",
```
andreapiras, line 19: why temp?

packages/test-config/test/jest.lint.js
```diff
@@ -1,6 +1,5 @@
 module.exports = {
   rootDir: process.env.PWD,
-  displayName: 'lint',
```
dozortsev, line 3: this use to be useful when linting and test were run as one job to be able easier to differentiate but not anymore, plus the smaller output the faster process

Screen Shot 2021-02-04 at 4 40 59 PM

packages/components/package.json
```diff
@@ -34,12 +34,11 @@
...
     "docs": "yarn build",
-    "test": "jest --env=jsdom --runInBand --config=../../node_modules/@transferwise/test-config/jest.config.js && yarn test:lint",
+    "test": "jest --env=jsdom --runInBand --config=../../node_modules/@transferwise/test-config/jest.config.js",
     "test:watch": "jest --watch --env=jsdom --config=../../node_modules/@transferwise/test-config/jest.config.js",
-    "test:lint": "jest --config ../../node_modules/@transferwise/test-config/test/jest.lint.js",
+    "lint": "jest --runInBand --config ../../node_modules/@transferwise/test-config/test/jest.lint.js",
     "test:visual": "yarn chromatic CHROMATIC_PROJECT_TOKEN || true",
-    "lint": "eslint 'src/**/*.js'",
-    "lint:fix": "yarn lint --fix",
+    "lint:fix": "eslint 'src/**/*.js' --fix",
```
dozortsev, line 41: as mentioned in PR description, I separated test and linting tasks to be separate workflow jobs and in order to achive that I had to split them in NPM scripts as well

package.json
```diff
@@ -65,11 +65,12 @@
...
     "docs": "lerna run docs --stream",
     "docs:deploy": "lerna run docs:deploy",
-    "test": "lerna run test --stream && bundlesize",
+    "test": "lerna run test --parallel --no-prefix",
+    "bundlesize": "bundlesize",
     "clean": "npm-run-all clean:*",
     "clean:modules": " lerna exec -- rm -rf ./node_modules && rm -rf node_modules  && rm -rf .next",
     "clean:lock": "find . -name '*.lock' -or -name '*-lock.json' -type f -delete",
-    "lint": "lerna run lint",
+    "lint": "lerna run lint --parallel --since main --no-prefix",
```
dozortsev, line 73: - [`--parallel`](https://github.com/lerna/lerna/tree/main/commands/run#--parallel) 
- [`--no-prefix`] - remove sub-package prefix, it's really not necessary here, the less output the faster the process
- [`--since`](https://www.npmjs.com/package/@lerna/filter-options#--since-ref) perform `lint` commad for all packages that have changed since `main`

.circleci/config.yml
```diff
@@ -38,15 +38,21 @@ aliases:
...
+  - &conventional-commit
     name: Conventional Commits check
     command: npx commitlint -f $(git log --author=CircleCI -1 --format=format:"%H") -V
 
   - &test
-    name: Test packages
+    name: Run tests across all packages
     command: yarn test
-    environment:
-      NODE_OPTIONS: --max_old_space_size=4096
```
dozortsev, line 49: this is redundant as it's 4096 by default

.circleci/config.yml
```diff
@@ -189,18 +225,6 @@ workflows:
...
-      # hence it requires `conventional-commit-lint` job
-      - test-main:
-          context: frontend-build
-          requires:
-            - setup
-            - conventional-commit-lint
-          filters:
-            branches:
-              only:
-                - main
```
dozortsev, line 202: now CI workflow on main branch would be similar to PR builds

helendurranttw: I guess the point of the difference was to excuse errors in the commit history in branches, because you can squash and merge and the commits themselves become irrelevant. Are we happy to add this check to branches?

package.json
helendurranttw: Amazing! `since` is so handy!

packages/validation/package.json
```diff
@@ -18,11 +18,10 @@
...
   },
   "scripts": {
-    "test": "jest --env=jsdom --runInBand --config=../../node_modules/@transferwise/test-config/jest.config.js && yarn test:lint",
-    "test:watch": "jest --watch --env=jsdom --config=../../node_modules/@transferwise/test-config/jest.config.js",
-    "test:lint": "jest --config ../../node_modules/@transferwise/test-config/test/jest.lint.js",
-    "lint": "eslint 'src/**/*.js'",
-    "lint:fix": "yarn lint --fix",
+    "test": "jest --env=node --maxWorkers=4 --config=../../node_modules/@transferwise/test-config/jest.config.js",
+    "test:watch": "jest --watch --env=node --config=../../node_modules/@transferwise/test-config/jest.config.js",
+    "lint": "jest --maxWorkers=4 --config ../../node_modules/@transferwise/test-config/test/jest.lint.js",
```
helendurranttw, line 23: What is the difference between linting using eslint directly and running a lint via jest?

.circleci/config.yml
andreapiras: to be honest, if we remove rebase and merge this check becomes redundant and can be removed from branches ONLY. It needs to stay on main because changelog needs CC

packages/validation/package.json
dozortsev: I don't really know, don't remember who added this

but I guess, it allows treat eslint linting as tests, e.g fail build if some "test(s)" failed and use Jest features like [watch mode](https://github.com/jest-community/jest-runner-eslint#toggle---fix-in-watch-mode) during development, it also make them more similar (consistent) with actual tests which is good

dozortsev: @andreapiras any you might want to add?

.circleci/config.yml
dozortsev: hmm, I'm a bit confused..

> ..Are we happy to add this check to branches?

we do have this check in branches

> .. if we remove rebase and merge this check becomes redundant and can be removed from branches ONLY

yeah, make sense, although this check is super tiny and fast so.. I would leave it as for `main` as for other branches 🙂

dozortsev: now CC check became necessary for beta releases (which also done via Lerna) so we need check every commit as it might become useful for someone to publish release out of it

example:
https://app.circleci.com/pipelines/github/transferwise/neptune-web/7949/workflows/65bf5682-5388-4dce-bc80-a0aba3b5db8e

CC check is very tiny and fast so I won't bother about it much, if we remove it we won't gain something significant

packages/validation/package.json
andreapiras: I think the main gain was to centralize the runner into the test-config package but apart from that I don't know about any other benefit

.circleci/config.yml
```diff
@@ -38,15 +38,21 @@ aliases:
     command: |
       npx lerna publish --canary --preid beta --dist-tag beta --yes
```
andreapiras, line 39: Should this have a `--no-changelog` https://github.com/lerna/lerna/blob/main/commands/version/README.md#--no-changelog

andreapiras: and also this --no-git-tag-version

.circleci/config.yml
```diff
@@ -89,14 +95,19 @@ jobs:
...
+
+  bundlesize:
+    <<: *container
+    resource_class: small
+    steps:
+      - *attach-workspace
+      - run: *bundlesize
 
   visual-test:
     <<: *container
```
andreapiras, line 113: *persist-workspace can be removed from here too

andreapiras (approved):

andreapiras:
Looks great good job

.circleci/config.yml
dozortsev: I did consider them but Lerna doesn't do those things when publish beta version so it seems pointless to add them, perhaps this is because of `--canary` flag

PR #753, opened 2020-12-26
fix: fix bug in MoneyInput component

farah:
Fixes #752

Stopping the propagation of this event to prevent bubbling seems to do the trick.

Issue #752, opened 2020-12-26
MoneyInput component is broken


farah:
I tried upgrading this package recently and among other things, this component stopped working as expected.

Basically the dropdown closes immediately after opening. For some odd reason it calls 

https://github.com/transferwise/neptune-web/blob/415cc4e17b123a20960f09645bf1cacb74290705/packages/components/src/select/Select.js#L279-L283

immediately after 

https://github.com/transferwise/neptune-web/blob/415cc4e17b123a20960f09645bf1cacb74290705/packages/components/src/select/Select.js#L273-L277

I was unable to provide a Codesandbox link as it throws an error when importing from `@transferwise/components`.

You can reproduce this easily by just bootstrapping a cra app and adding the following dependencies:

```
    "@transferwise/components": "^31.5.1",
    "@transferwise/formatting": "^2.2.0",
    "@transferwise/icons": "^2.6.0",
    "@transferwise/neptune-css": "^5.1.0",
    "@transferwise/neptune-tokens": "^1.1.0",
```
`App.js`

```
import { MoneyInput } from "@transferwise/components";
function App() {
  return (
    <div className="App">
      <label htmlFor="money-input">Money input label</label>
      <MoneyInput
        id="money-input"
        amount={1000}
        locale="en-GB"
        size="lg"
        onAmountChange={(value) => console.log("amount changed", value)}
        onCurrencyChange={() => alert("currency changed")}
        addon={null}
        searchPlaceholder="Type a currency or country"
        onCustomAction={() => alert("Custom action")}
        customActionLabel="Custom action label"
        currencies={[
          {
            header: "Popular currencies",
          },
          {
            value: "EUR",
            label: "EUR",
            note: "Euro",
            currency: "eur",
            searchable: "Spain, Germany, France, Austria",
          },
          {
            value: "GBP",
            label: "GBP",
            note: "British pound",
            currency: "gbp",
            searchable: "England, Scotland, Wales",
          },
        ]}
        selectedCurrency={{
          value: "EUR",
          label: "EUR",
          note: "Euro",
          currency: "eur",
          searchable: "Spain, Germany, France, Austria",
        }}
      />
    </div>
  );
}

export default App;
```

I've attached a giphy of the behaviour below.

![moneyinput-bug](https://user-images.githubusercontent.com/50220822/103153249-5d190080-47e3-11eb-8249-0fac4c334683.gif)

May be related to #572

@dozortsev

farah:
[...]

PR #798, opened 2021-02-08
fix: close Select's dropdowns

dozortsev:
## 🖼 Context

<details>
  <summary>Atm it's possible to open multiple Select's dropdowns which is considered as buggy behaviour, examples: </summary>
  
image

Screen Shot 2021-02-05 at 11 55 48 AM

</details>

The above issue was introduced in fix https://github.com/transferwise/neptune-web/pull/572 for another thing in Select, but contributor was having that issue because he was using component with React 17 ([I haven't confirmed this from him](https://github.com/transferwise/neptune-web/pull/753#issuecomment-776862708) but I was able to reproduce this issue my self by setup latest CRA), where React introduced breaking changes around [events delegation](https://reactjs.org/blog/2020/10/20/react-v17.html#changes-to-event-delegation) ([that is also why we couldn't reproduce this bug in our docs website nor on our Storybook as both rely on React 16](https://github.com/transferwise/neptune-web/issues/752#issuecomment-751363963)).

Atm I couldn't find a way how to make this component work with both versions of React (I did try follow suggestion from [React docs](https://reactjs.org/blog/2020/08/10/react-v17-rc.html#fixing-potential-issues) and some other [examples](https://stackoverflow.com/questions/7398290/unable-to-understand-usecapture-parameter-in-addeventlistener/7398447#7398447)), but it's either works correct on React 17 but doesn't work on 16 and visa-versa, not sure it's even possible, perhaps major refactoring/whole rewrite would help.
 
## 🚀 Changes

Simple revert the change made in https://github.com/transferwise/neptune-web/pull/572

## ✅ Checklist


packages/components/src/select/Select.js
```diff
@@ -270,8 +270,8 @@ export default class Select extends Component {
     document.removeEventListener('click', this.handleDocumentClick, false);
   }
 
-  handleButtonClick = (e) => {
-    stopPropagation(e);
+  handleButtonClick = () => {
+    // stopPropagation(e);
```
dozortsev, line 274: revering fix https://github.com/transferwise/neptune-web/pull/753 which introduced another bug

helendurranttw (commented):
This looks fine to me, though you should check to see you don't re-introduce the bug that the change you reverted was trying to fix 😅

dozortsev:
yep, in this PR I will try fix the issue https://github.com/transferwise/neptune-web/pull/753 just in another way

packages/docs/liveEditorCode/select.code.js
```diff
@@ -49,9 +37,6 @@
         {
           value: 8,
           label: 'A thing with searchable alternatives',
-          /**
-           * searchStrings is an array of strings that maps additional searchable terms to a given option.
-           */
```
dozortsev, line 54: deleting redundant comments from docs

helendurranttw (approved):

PR #807, opened 2021-02-15
fix: reference error when file type is null

arikanmstf-tw:
## 🖼 Context

Resolves: https://transferwise.atlassian.net/browse/BS-721
Rollbar: https://rollbar.com/TransferWise/batch-payments/items/4072/?item_page=0&

When OS is windows, getFileType util returns nulll; and (null).indexOf causes errors.
Instead getFileType should return empty string.

## 🤔 Considerations

27c9e466-0a81-424d-a7f0-284bcbac2f4b

## ✅ Checklist


packages/components/src/upload/utils/getFileType/getFileType.js
```diff
@@ -1,19 +1,19 @@
 export const getFileType = (file, file64) => {
   if (!file && !file64) {
-    return null;
+    return '';
   }
 
   if (file && file.type && file.type !== '') {
-    return file.type;
+    return file.type || '';
```
dozortsev, line 7: I think it's better to use [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) in such cases

arikanmstf-tw: sure, done

dozortsev (approved):

PR #811, opened 2021-02-17
feat: add aria-labels to Summary statuses

helendurranttw:
## 🖼 Context

We're adding aria labels to help non-sighted users understand the status badges.

## 🚀 Changes

Added aria labels. Also updated semantic colour tokens.

## ✅ Checklist


packages/components/src/summary/Summary.js
```diff
@@ -41,10 +51,13 @@ const Summary = ({
   const Badge = status && BadgeIcons[status];
 
   return (
-    <Element className={classNames('np-summary d-flex align-items-start', className)}>
+    <Element
+      className={classNames('np-summary d-flex align-items-start', className)}
+      aria-label={status && intl.formatMessage(messages[statusLabels[status]])}
```
andreapiras, line 56: Much better 👌🏻

packages/components/src/summary/Summary.messages.js
```diff
@@ -0,0 +1,16 @@
+import { defineMessages } from 'react-intl';
+
+export default defineMessages({
+  statusNotDone: {
+    id: 'neptune.Summary.statusNotDone',
+    defaultMessage: 'Item to do',
```
andreapiras, line 6: Did we agree on the labels? Do we need the word Item?

helendurranttw: The reason I added this was I think it sounds more natural:
* “list, 3 items”
*     “item done, 1 of 3"
*          “You have verified your address
*     “item pending, 2 of 3”
*          “We’re verifying your identity”
      etc…

As opposed to:
* “list, 3 items”
*     “done, 1 of 3"
       etc…

I've asked @StephanieTW , will wait to see what she says.

andreapiras (approved):

packages/components/src/summary/Summary.messages.js
StephanieTW: This makes sense to me

PR #815, opened 2021-02-22
chore: enable TS for docs package

dozortsev:
## ✅ Checklist


packages/docs/tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+  "compilerOptions": {
+    "target": "es5",
+    "lib": ["dom", "dom.iterable", "esnext"],
+    "allowJs": true,
+    "skipLibCheck": true,
+    "strict": false,
```
sbrandwoo, line 7: I highly recommend at least setting `"strictNullChecks": true,`

packages/docs/.eslintrc
```diff
@@ -9,6 +9,19 @@
...
+    "import/extensions": ["error", "ignorePackages", {
+      "js": "never",
+      "ts": "never",
+      "tsx": "never"
+    }]
+  },
+  "settings": {
+    "import/resolver": {
+      "node": {
+        "extensions": [".js", ".ts", ".tsx", ".json"]
```
dozortsev, line 23: make ESLint understand TypeScript files (`.ts` and `.tsx`)

packages/docs/pages/components/context/Provider.tsx
```diff
@@ -0,0 +1,61 @@
...
+  'fr',
+  'ru',
+  'de',
+  'tr',
+].map((locale) => ({
+  value: locale,
+  label: locale,
+}));
+
+export default function PageContent() {
```
dozortsev, line 30: so basically [Provider](https://transferwise.github.io/neptune-web/branch/enable-ts-for-docs/components/context/Provider) component page is now on TS

packages/docs/.eslintrc
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "@transferwise",
   "rules": {
+    "react/require-default-props": [0],
```
dozortsev, line 4: disabling it as it appear in TS files, and there seems not other way how to workaround this

Screen Shot 2021-03-02 at 11 07 02 AM

packages/docs/utils/pageUtils.js
```diff
@@ -2,6 +2,13 @@ import React from 'react';
...
 import { format, formatDistance } from 'date-fns';
 
+/**
+ * Check if date is past date
+ *
+ * @param {Date | null | undefined} endDate test date
+ * @return {boolean} `true` if date is less than today or `null` / `undefined`,
+ * otherwise return `false`
+ */
 export function isExpired(endDate) {
```
dozortsev, line 12: added JSDoc in a few places to see how TS picks types from it

packages/docs/tsconfig.json
dozortsev: I enabled it

packages/docs/components/Badge.tsx
```diff
@@ -0,0 +1,22 @@
...
+  children: ReactNode;
+  expiryDate?: Date;
+  className?: string;
+};
+
+export default function Badge({
+  children,
+  expiryDate = undefined,
+  className = undefined,
+}: BadgeProps) {
```
sbrandwoo, line 15: You can alternatively declare it like this, using the `FC` type:
```typescript
const Badge: FC<BadgeProps> = (props) => { }
```

There's two main things you get from this:

1.  `FC` knows that all components can also take a `children` prop without one being explicitly defined.

For example below both definitions will run fine but the second one doesn't compile because it doesn't know `children` is always valid:

```typescript
export type BadgeProps = {
  expiryDate?: Date;
  className?: string;
};

export const Badge1: FC<BadgeProps> = props => {
  return <div>{props.children}</div>;
};

export function Badge2(props: BadgeProps) {
  // ERROR: Property 'children' does not exist on type 'BadgeProps'.ts(2339)
  return <div>{props.children}</div>;
}
```

2. `FC` also defines a return type of the function as `ReactElement` which gives more type safety.

For example below the first example correctly errors, since `Date` is an invalid return type for a component:

```typescript
// ERROR
// Type '(props: PropsWithChildren<BadgeProps>) => Date' is not assignable to type 'FC<BadgeProps>'.
// Type 'Date' is missing the following properties from type 'ReactElement<any, any>': type, props, keyts(2322)
export const Badge1: FC<BadgeProps> = props => {
  return new Date();
};

export function Badge2(props: BadgeProps) {
  return new Date();
}
```

packages/docs/utils/pageUtils.js
sbrandwoo: Cool, I had no idea you can do this

packages/docs/utils/pageUtils.js
```diff
@@ -2,6 +2,13 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { format, formatDistance } from 'date-fns';
 
+/**
+ * Check if date is past date
+ *
+ * @param {Date | null | undefined} endDate test date
+ * @return {boolean} `true` if date is less than today or `null` / `undefined`,
```
sbrandwoo, line 9: If the comment is correct this should be `@return {boolean | null}`.

sbrandwoo:
It's a good habit to always define a return type to your functions. For example in the below the compiler can pick out that the function can return undefined in some situations:

```
export function isExpired(endDate: Date | null): boolean | null {
  // ERROR: Object is possibly 'undefined'
  return endDate?.getTime() <= Date.now();
}
```

and this tells you the code should probably be

```
export function isExpired(endDate: Date | null): boolean | null {
  return endDate ? endDate.getTime() <= Date.now() : null;
}
```

There's a check in ESLint to ensure you don't forget and it's part of our usual set up. You can see https://github.com/transferwise/navigation-ui/commit/71424aac26c7322a8dcc0a20fac8306ab88e270a for an idea of what rules we're using elsewhere.

packages/docs/utils/pageUtils.js
dozortsev: yep,

in any TS file:

Screen Shot 2021-03-02 at 1 48 09 PM

packages/docs/components/Badge.tsx
dozortsev: second point is really great one but re. first one I'm not sure we want every component have `children` prop.

Apparently using `FC` is actually not so encouraging to use https://github.com/facebook/create-react-app/pull/8177

packages/docs/utils/pageUtils.js
dozortsev: sorry, it should never return `null`, if consumer pass there `Date | null | undefined` it will always return `boolean`

I updated comment:

```js
 * @param {Date | null | undefined} endDate test date
 * @return {boolean} `true` if date is less than today or equal `null` / `undefined`, otherwise returns `false`
 ```

File: packages/docs/components/Badge.tsx (hunk 6 above)
Comment (dozortsev) [reply packages/docs/components/Badge.tsx]: We definitely will enable [rules for TS code](https://github.com/typescript-eslint/typescript-eslint) and especially [`explicit-function-return-type`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-function-return-type.md) 👍

Comment (dozortsev) [reply packages/docs/components/Badge.tsx]: Thanks for suggestion it's good to know, maybe we actually would use `FC` in all component but atm the pattern is haven't been decided yet.

I added `ReactElement` as return type for all TS React components https://github.com/transferwise/neptune-web/pull/815/commits/9ba8ed8c981c6fd44fa5f18ceead258d4822dd8c

File: packages/docs/CONTRIBUTING.md [hunk 8]
```diff
@@ -57,3 +57,14 @@ import { GeneratePropsTable } from '../../utils';
...
+# TypeScript Support
+
+This package supports TypeScript, eventually we want all codebase be written on TS, so it's encouraged to at least create new files on TS (`.ts`, `.tsx`). Feel free to refactor files that you touch on TS as well.
+
+If you don't feel quite confident with TypeScript then you have to:
+
+- put [`// @ts-check`](https://www.typescriptlang.org/docs/handbook/intro-to-js-ts.html#ts-check) at the top of JavaScript files (this is the easiest way to enable TS checking per file)
+- and write [JSDoc](https://jsdoc.app/) (in particular [`@param` and `@return`](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html)) for some API.
+
+This will help us build more reliable and consistent code to migrate on TS faster.
```
dozortsev, line 70: rendered version https://github.com/transferwise/neptune-web/blob/enable-ts-for-docs/packages/docs/CONTRIBUTING.md#typescript-support

packages/docs/CONTRIBUTING.md
```diff
@@ -57,3 +57,14 @@ import { GeneratePropsTable } from '../../utils';
 ```
 
 This will read the component PropTypes and generate a table showing what is supported and the defaults.
+
```
dozortsev, line 60: rendered version https://github.com/transferwise/neptune-web/blob/enable-ts-for-docs/packages/docs/CONTRIBUTING.md#typescript-support

packages/docs/components/Badge.tsx
oliverviljamaa: I prefer explicit `children`, so no `FC` as well, albeit having been a proponent in the past.

I don't think there's a good solution — `ReactElement` still doesn't support actually valid null, string, and number types, for example, resulting in unnecessary fragments. Perhaps we should roll our own `FC` alternative?

packages/docs/.eslintrc
oliverviljamaa: You could use `overrides` to disable that rule for `.ts` and `.tsx` files only.

oliverviljamaa (approved):
Exciting 👏

sbrandwoo (approved):

packages/docs/.eslintrc
dozortsev: awesome, didn't know about this, done https://github.com/transferwise/neptune-web/pull/815/commits/ea0b4ab545c2a405efe22dbb813cc56f345d6644

packages/docs/.eslintrc
```diff
@@ -24,5 +23,13 @@
         "extensions": [".js", ".ts", ".tsx", ".json"]
       }
     }
-  }
+  },
+  "overrides": [
+    {
+      "files": ["**/*.tsx"],
+      "rules": {
+        "react/require-default-props": "off"
```
dozortsev, line 31: this rule is redundant for TS files

PR #822, opened 2021-02-24
feat!: Add prop to tabs to stop animation on click and make animation false by default

ksunandp:
## 🖼 Context

Animation between panels in Tabs causes some issues both functionality and visually when used inside cards. Account details on /user/account currently uses a stripped down version - https://github.com/transferwise/receive-components/blob/master/lib/BankDetails/Tabs/Tabs.tsx#L8. And for the new activity cards the Activity Team are building, we want to use tabs but without the animation.

## 🚀 Changes

Add a prop `animatePanels` which is always true unless specified. When true, it will effectively make animation between panels instant. The line will still animate so it is not as jarring a change, but the content itself will switch instantly.

## 🤔 Considerations

Is this the right thing to do? Or should tabs always animate?

## ✅ Checklist


fergusjordantw:
Looks good, although I'm wondering if this should also effect the animation for swiping between tabs? Currently it does and it feels a bit weird for it to move to the new tab instantly during swipe (mind you it's not terrible).

ksunandp:
> Looks good, although I'm wondering if this should also effect the animation for swiping between tabs? Currently it does and it feels a bit weird for it to move to the new tab instantly during swipe (mind you it's not terrible).

How could we do that, simple as checking `state.isSwiping` is false?

fergusjordantw:
> How could we do that, simple as checking state.isSwiping is false?

I think that would do it, yes. I'm not saying we _need_ to do that though, it does make sense that if you're setting `animatePanels` to false that it would prevent all panel animation. I'm just wondering if it makes sense to keep it when swiping for your use-case (in which case I'd rename the prop to something else). Another alternative being that setting `animatePanels` would prevent swiping altogether as it's now a more static component to interact with (would also probably require a prop rename)

ksunandp:
> > How could we do that, simple as checking state.isSwiping is false?
> 
> I think that would do it, yes. I'm not saying we _need_ to do that though, it does make sense that if you're setting `animatePanels` to false that it would prevent all panel animation. I'm just wondering if it makes sense to keep it when swiping for your use-case (in which case I'd rename the prop to something else). Another alternative being that setting `animatePanels` would prevent swiping altogether as it's now a more static component to interact with (would also probably require a prop rename)

I would be happy to disable the animation only on click and leaving the swipe, so that mobile web feels fluid. At least until further testing is done. Maybe `animatePanelsOnClick` would be more appropriate prop name in this case. And add the `!isSwiping` check.

hendrikmolder:
> > > How could we do that, simple as checking state.isSwiping is false?
> > 
> > 
> > I think that would do it, yes. I'm not saying we _need_ to do that though, it does make sense that if you're setting `animatePanels` to false that it would prevent all panel animation. I'm just wondering if it makes sense to keep it when swiping for your use-case (in which case I'd rename the prop to something else). Another alternative being that setting `animatePanels` would prevent swiping altogether as it's now a more static component to interact with (would also probably require a prop rename)
> 
> I would be happy to disable the animation only on click and leaving the swipe, so that mobile web feels fluid. At least until further testing is done. Maybe `animatePanelsOnClick` would be more appropriate prop name in this case. And add the `!isSwiping` check.

If we'd keep the animation on swiping, wouldn't it still block us from using it inside Cards as the animation issue/bug remains?

ksunandp:
> If we'd keep the animation on swiping, wouldn't it still block us from using it inside Cards as the animation issue/bug remains?
@hendrikmolder @fergusjordantw  Yeah, just tested this on an ipad, and the swipe still causes issues, think we can disable the swipe and animation then. Although I'm not sure they should be the same prop. Maybe another prop of `canSwipePanels`? Which I can add here. Or do you think we merge the two?

ksunandp:
@fergusjordantw @hendrikmolder Made the change to `animatePanelsOnClick` this means on swipe the panels will still animate. We can then use the `changeTabOnSwipe` prop that already exists to stop that.

@andreapiras Should be ready now. Also any ideas why my commits don't trigger builds 🤔

fergusjordantw (approved):

helendurranttw (approved):

PR #827, opened 2021-03-01
refactor: return noscroll class handler to Dimmer

joe-sayer:
## 🖼 Context

Scroll/NoScroll functions were moved out of Dimmer because of some problems in send-flow where unopened modals that were rendered in the React tree with open={false} were unmounted after another modal is opened, resulting in the open modal being un-scrollable.

Original PR: https://github.com/transferwise/neptune-web/pull/572

We would like to bring these features back to Dimmer so we don't have to duplicate this logic in children (Drawer,Modal, BottomSheet etc).

## 🚀 Changes

 

The `addNoScrollBodyClass` and `removeNoScrollBodyClass` functions have been returned to inside Dimmer along with the tests for the function calls.

Pre-fix

https://user-images.githubusercontent.com/16908937/109623238-3a85c980-7b3d-11eb-9246-4b3c36d507bb.mp4

Post-fix

https://user-images.githubusercontent.com/16908937/109622474-6e142400-7b3c-11eb-9ee5-916dee61e684.mp4

## ✅ Checklist


packages/components/src/dimmer/Dimmer.js
```diff
@@ -11,64 +12,73 @@ import KEY_CODES from '../common/keyCodes';
...
+    document.addEventListener('keydown', handleOnKeyDown);
+  };
+
+  const onExit = () => {
+    removeNoScrollBodyClass();
+    document.removeEventListener('keydown', handleOnKeyDown);
+  };
+
+  useEffect(() => {
+    onEnter();
```
andreapiras, line 72: @fergusjordantw maybe we should rename these to onEnter  -> onMount and onExit  -> onUnmount ?

fergusjordantw: Yep, can do

packages/components/src/dimmer/Dimmer.spec.js
```diff
@@ -31,6 +30,26 @@ describe('Dimmer', () => {
     jest.clearAllMocks();
   });
 
+  it('calls addNoScrollBodyClass on mount', () => {
+    expect(addNoScrollBodyClass).not.toHaveBeenCalled();
+    component = mount(<Dimmer {...props} />);
+    expect(addNoScrollBodyClass).toHaveBeenCalled();
+  });
+
+  it('calls removeNoScrollBodyClass on unmount', () => {
```
fergusjordantw, line 39: Duplicate test

fergusjordantw: While not necessary, it might be good to add a test which simulates the other kinda off behaviour that we used to have which is now fixed. When onClick triggered we would run cleanup as well as running it on unmount. Would be good to add a test that simulates the behaviour of an onClick triggering `open={false}` getting passed in and asserting that we only call `removeNoScrollBodyClass` once.

packages/components/src/dimmer/Dimmer.rtl.spec.js
```diff
@@ -0,0 +1,45 @@
...
+import { addNoScrollBodyClass, removeNoScrollBodyClass } from '../common';
+
+import { render, cleanup } from '../test-utils';
+import { Dimmer } from './Dimmer';
+
+jest.mock('../common');
+
+jest.mock('react-transition-group/CSSTransition', () => (props) =>
+  props.in ? props.children : null,
+);
```
fergusjordantw, line 11: If these are now in RTL, personally I wouldn't mock this anymore. It made sense when it was in the enzyme tests but here we should be able to achieve the same thing using something like this to wait for it to be mounted and the inverse to wait for unmount.

```
await waitFor(() => {
  expect(screen.queryByRole("presentation")).toBeInTheDocument();
});
```

packages/components/src/dimmer/Dimmer.rtl.spec.js
```diff
@@ -0,0 +1,45 @@
...
+    expect(removeNoScrollBodyClass).not.toHaveBeenCalled();
+    const { unmount } = render(<Dimmer {...props} />);
+    unmount();
+    expect(removeNoScrollBodyClass).toHaveBeenCalled();
+  });
+
+  it('calls removeNoScrollBodyClass only once when the children is unmounted', () => {
+    const { rerender, unmount } = render(<Dimmer {...props} />);
+    expect(removeNoScrollBodyClass).not.toHaveBeenCalled();
+    expect(removeNoScrollBodyClass).not.toBeCalled();
```
fergusjordantw, line 39: Can remove one of those assertions

fergusjordantw (approved):
Few small comments now that the tests are in RTL, but will approve so as not to block you

PR #830, opened 2021-03-03
feat!: Add base url to dynamic flows

felixxiatransferwise:
## 🚀 Changes

 
We need a way to specify different baseUrls for asynchronous calls like validation async and persist async for different dynamic flows consumers.

## ✅ Checklist


packages/dynamic-flows/src/common/contexts/baseUrlContext/BaseUrlContext.js
```diff
@@ -0,0 +1,5 @@
+import { createContext } from 'react';
+
+export const BaseUrlContext = createContext({
+  baseUrl: null,
+});
```
JulianG-TW, line 5: AFAIK we can make the context hold a `string`. Not sure why we need an object here, unless we expect more things to be added to this context in the near future.

I would create a custom `useBaseUrl` hook here and export only that.

This could have two benefits:
1. Avoid having to import 2 things in the components that consume this.
2. We can also throw with a more friendly error message, should the context return undefined/null.

JulianG-TW (approved):

PR #834, opened 2021-03-05
feat: add target and callback for summary

luujam2:
## 🖼 Context

for Summary components we would like to have the option to open the link in a new tab as well as fire analytics on click

## 🚀 Changes

 
- `action.target` accepts a string for the target on the anchor tag
- `action.callback` accepts a function which will get called on click of the anchor tag if provided

## ✅ Checklist


packages/components/src/summary/Summary.js
```diff
@@ -106,6 +112,8 @@ Summary.propTypes = {
     text: Types.node.isRequired,
     href: Types.string.isRequired,
     'aria-label': Types.string,
+    target: Types.string,
+    callback: Types.func,
```
dozortsev, line 116: it should be called `onClick`

dozortsev (approved):

PR #836, opened 2021-03-05
feat: add forwardRef, fix MoneyInput accepting invalid input, fix es-ES locale bug

fergusjordantw:
## 🖼 Context

This PR aims to fix up a few things, but at the moment I haven't done any tests since I'd like to open it more for a discussion around the approach I've taken (I was planning on doing some exploration and opening a thread but then just decided to continue). 

https://transferwise.atlassian.net/browse/DS-555
https://transferwise.atlassian.net/browse/DS-864

#### To fix: https://transferwise.atlassian.net/browse/DS-864

Changes our check for the `groupSeparator` to use 10k instead of 1k when trying to get the thousand delimiter. It seems `es-ES` doesn't comma separate `1000` as a value, but does separate `10000`. Updating this check to use `10000` & get the delimiter off the third character fixes the issue and doesn't seem to negatively affect a few of the other locales I tested.

Playing with values in the console to visualise it:

Screenshot 2021-03-05 at 17 18 54

#### To fix: https://transferwise.atlassian.net/browse/DS-555

This is the part of the PR I'm questioning the most. We want to stop users from inputting text values to prevent values like `$1000` since they won't trigger amount `onAmountChange` and customers can get confused as to why the quote isn't updating. 

I can basically think of two main approaches to this: 

1. Prevent user input of unwanted stuff `onKeyDown` before the change fires
2. Set the parsed value to state in `onChange`

I decided to go with the first one since the second approach would require us to store/reset the caret position since otherwise it would jump to the end of the input on each change. The annoying (but also positive I guess?) thing about the first approach is that it requires us to very explicitly say what we'll accept. As you can see, the list is already pretty big just to support a few ways of interacting with the input like selecting all, using arrow keys, etc. so I'm worried this might be too prohibitive an approach. I'm also not sure what impact this will have on accessibility. But, that's what this PR is mostly for, to discuss the approach!

#### To do an entirely unrelated thing

I've added a forwardRef to let us pass a `ref` into the actual input DOM node. It works fine but reads a bit messy because it's a class component so I cant add it to the actual declaration and need to re-export `Size` on the `forwardRef`. I'm guessing you guys might want this either formatted a different way or added with a different approach, just let me know (pretty please don't make me refactor the entire thing to a function component)!

## ✅ Checklist


sbrandwoo:
> Prevent user input of unwanted stuff onKeyDown before the change fires
Does this affect pasting of values?

fergusjordantw:
> Does this affect pasting of values?

No, that still works (at least on Mac) because of the check for `metaKey` on the event. That said, I hadn't actually considered that specific use case (added the meta check for select all) which does highlight the danger in this approach that we can miss certain user behaviour that we should cater to.

packages/components/src/moneyInput/MoneyInput.js
```diff
@@ -52,6 +54,30 @@ class MoneyInput extends Component {
     }
   }
 
+  handleKeyDown = (event) => {
+    const { keyCode, metaKey, key } = event;
+    const isNumberKey = !Number.isNaN(parseInt(key, 10)); // keyCode >= 48 && keyCode <= 57;
+
+    if (
+      !(
+        metaKey ||
```
mikknarusk, line 63: can we extract this check into a function to give it a descriptive name? I had a bit of trouble understanding which way this check works

fergusjordantw:
> Does this affect pasting of values?

I think I previously misunderstood this as asking if it would prevent paste, which it doesn't. No, this isn't going to parse the pasted value in any way, so if you past an incorrect value of `$1000` then that is what will get input and we still won't have correctly parsed it/triggered an onAmountChange handler. To fix this we could either add a dedicated paste handler to parse the values there or update the parsing in `setAmount` to be more exhaustive in stripping out text at the start of the string

packages/components/src/moneyInput/MoneyInput.js
sbrandwoo: Googling `metaKey` it's the special key on macs. Other keys need to be listened to for Windows etc?

sbrandwoo:
I think you had the right understanding the first time. I didn't want to prevent copy/paste from working.

Potential third option: `onKeyPress` we strip out unwanted characters and pass valid numbers to `onAmountChange`. And `onBlur` we update the input to with a formatted number.

So while typing the user will see "$1000", there's a call to `onAmountChange(1000)` and when the user moves away from the input we reformat it to `1,000`.

packages/components/src/moneyInput/MoneyInput.js
fergusjordantw: Will add `ctrlKey` and `altKey` to be safe.

fergusjordantw:
> Potential third option: onKeyPress we strip out unwanted characters and pass valid numbers to onAmountChange. And onBlur we update the input to with a formatted number.
>
> So while typing the user will see "$1000", there's a call to onAmountChange(1000) and when the user moves away from the input we reformat it to 1,000.

Also a good option. If we aren't going to parse pasted input, then this would be a better option since it would be a more consistent user experience for what the user sees when they type "$1000" and blur vs when they paste "$1000" and blue. The only downside being that we're still allowing users to input invalid values and trying to react to that instead of proactively preventing it. 

Either way we go we'll need to improve our parsing for what we do on blur since at the moment it only strips out text that come after numbers, not before.

packages/components/src/moneyInput/MoneyInput.js
```diff
@@ -52,6 +54,32 @@ class MoneyInput extends Component {
...
+      keyCode === keyCodes.RIGHT ||
+      keyCode === keyCodes.ENTER ||
+      keyCode === keyCodes.ESCAPE ||
+      keyCode === keyCodes.TAB
+    );
+  };
+
+  handleKeyDown = (event) => {
+    if (!this.isInputAllowedForKeyEvent(event)) {
+      event.preventDefault();
```
arikanmstf-tw, line 79: this allows entering multiple commas, dots right ?  something like;

`12.123.4.4,123123,233` .

my solution was allowing only one `DOT` and converting other commas to dots as well; please check it out [here](https://github.com/transferwise/batch-payments-flow/blob/bs-419/packages/template-instructions-page/src/components/BankTemplateFieldSample/BankTemplateFieldSample.js#L7).

fergusjordantw: Yes, it does allow entering multiple commas/dots. We shouldn't limit what the user is inputting to only a single separator. `1,000,000` has multiple commas that are entirely valid. Also we can't just assume dots are the decimal separator, it changes per locale

helendurranttw:
See #914

PR #839, opened 2021-03-08
feat: Conditionally prepend baseUrl

felixxiatransferwise:
## 🚀 Changes

 
We need conditionally prepend the URL so dynamic flows will support both baseUrl for both absolute and relative URLs.

## ✅ Checklist


JulianG-TW (approved):
👍  
Regarding tests: I think it would be best if we could assert on the api call made by the PersistAsyncSchema in PersistAsyncSchema.spec.js

PR #845, opened 2021-03-10
feat: rtl support

joe-sayer:
## 🖼 Context

This PR is necessary because currently there is no support for locales that read right to left.

## 🚀 Changes

 
Added support for:
- Checkbox
- DateInput
- InputWithDisplayFormat
- PhoneNumberInput
- Switch

## ✅ Checklist


packages/components/src/inputWithDisplayFormat/InputWithDisplayFormat.spec.js
```diff
@@ -1,14 +1,23 @@
 import React from 'react';
 
+import { shallow } from 'enzyme';
```
dozortsev, line 3: we should avoiding using Enzyme https://github.com/transferwise/neptune-web/blob/main/packages/components/CONTRIBUTING.md#testing

PR #846, opened 2021-03-10
feat!: enable TypeScript, provide types for all components at `@transferwise/components`

dozortsev:
### tl;dr list of changes

- Enabled TypeScript support for `components/` sub-package (for now TS files will be transpiled by Babel via [`@babel/preset-typescript`](https://babeljs.io/docs/en/babel-preset-typescript).
- Migrated a couple components to TypeScript (`Loader`, `Chip`, and `Badge`).
- Integrated [`react-to-typescript-definitions`](https://www.npmjs.com/package/react-to-typescript-definitions) to generate TypeScript types for JS components based on [`prop-types`](https://www.npmjs.com/package/prop-types).
- Expose common and component specific constants (as TS enums) from `@transferwise/components`.
- Release notes & migration guide as a [blog post](https://transferwise.github.io/neptune-web/branch/enable-ts-for-components/blog/2021-04-30-components-v36-release-guide---typescript-support).
- Codemod to address breaking changes.

### In More Details

We want to provide **accurate** type declarations for all component already now, without waiting until while library would be written on TypeScript, so after a bit of research decided give a try of react-to-typescript-definitions, a library that generates TypeScript declarations (a content for `.d.ts`) based on `prop-types`.

The library generates pretty accurate types, ~95% of them are correct, leaving JSDoc in place.

<details>
  <summary>A couple examples:</summary>

**`Button.d.ts`**

```typescript
import * as React from 'react';

export type ButtonHtmlType = "submit" | "reset" | "button";

export type ButtonPriority = "primary" | "secondary" | "tertiary";

export type ButtonType = "accent" | "positive" | "negative" | "primary" | "pay" | "secondary" | "danger" | "link";

export type ButtonSize = "xs" | "sm" | "md" | "lg";

export interface ButtonProps {
    block?: boolean;
    children: React.ReactNode;
    className?: string;
    disabled?: boolean;
    htmlType?: ButtonHtmlType;
    loading?: boolean;
    onClick?: any;
    priority?: ButtonPriority;
    /**
     * @deprecated `Types.PRIMARY`, `Types.PAY`, `Types.SECONDARY`, `Types.DANGER`, `Types.LINK`
     */
    type?: ButtonType;
    /**
     * @deprecated `Sizes.EXTRA_SMALL`
     */
    size?: ButtonSize;
}

declare const Button: React.FC<ButtonProps>;

export default Button;
```  

**`Popover.d.ts`**

```typescript
import * as React from 'react';

export type PopoverPreferredPlacement = "top" | "right" | "bottom" | "left" | "left-top" | "right-top" | "bottom-right" | "bottom-left";

export interface PopoverProps {
    children: React.ReactNode;
    className?: string;
    content: React.ReactNode;
    /**
     * `'left-top'` / `'right-top'` are deprecated use `Position.TOP` / `'top'` instead,
     * `'bottom-right'` / `'bottom-left'` are deprecated use `Position.BOTTOM` / `'bottom'` instead
     */
    preferredPlacement?: PopoverPreferredPlacement;
    title?: React.ReactNode;
}

declare const Popover: React.FC<PopoverProps>;

export default Popover;
```

**`Summary.d.ts`**

```typescript
import * as React from 'react';

export interface SummaryAction {
    text: React.ReactNode;
    href: string;
    "aria-label"?: string;
    target?: string;
    onClick?: (...args: any[])=>any;
}

export interface SummaryInfo {
    "aria-label": string;
    content: React.ReactNode;
    presentation?: "POPOVER" | "MODAL";
    title?: React.ReactNode;
}

export type SummaryStatus = "notDone" | "done" | "pending";

export interface SummaryProps {
    /**
     * Action displayed at the bottom of the Summary
     */
    action?: SummaryAction;
    /**
     * Decides which html element should wrap the Summary
     */
    as?: string;
    /**
     * Extra classes applied to Summary
     */
    className?: string;
    /**
     * @deprecated please use description instead
     */
    content?: any;
    /**
     * Summary description
     */
    description?: React.ReactNode;
    /**
     * @deprecated please use info instead
     */
    help?: any;
    /**
     * Infos displayed on help Icon click inside Popover or Modal
     */
    info?: SummaryInfo;
    /**
     * @deprecated please use icon instead
     */
    illustration?: any;
    /**
     * Main Summary Icon
     */
    icon?: any;
    /**
     * Decides the badge applied to Icon
     */
    status?: SummaryStatus;
    /**
     * Summary title
     */
    title: React.ReactNode;
}

declare const Summary: React.FC<SummaryProps>;

export default Summary;
```
</details>

Note: if we avoid using the library `tsc` generates not so good type declarations, in many cases it applies `any` and it removes JSDoc

<details>
  <summary>A couple examples (of same components as above):</summary>

**`Button.d.ts`**

```typescript
export default Button;
declare function Button(props: any): JSX.Element;
declare namespace Button {
    export { Priorities as Priority };
    export const Type: {
        ACCENT: ControlTypes.ACCENT;
        POSITIVE: ControlTypes.POSITIVE;
        NEGATIVE: ControlTypes.NEGATIVE;
        PRIMARY: Types.PRIMARY;
        PAY: Types.PAY;
        SECONDARY: Types.SECONDARY;
        DANGER: Types.DANGER;
        LINK: Types.LINK;
    };
    export namespace Size {
        const EXTRA_SMALL: Sizes;
        const SMALL: Sizes;
        const MEDIUM: Sizes;
        const LARGE: Sizes;
    }
    export namespace propTypes {
        const block: PropTypes.Requireable<boolean>;
        const children: PropTypes.Validator<PropTypes.ReactNodeLike>;
        const className: PropTypes.Requireable<string>;
        const disabled: PropTypes.Requireable<boolean>;
        const htmlType: PropTypes.Requireable<string>;
        const loading: PropTypes.Requireable<boolean>;
        const onClick: any;
        const priority: PropTypes.Requireable<string>;
        const type: PropTypes.Requireable<string>;
        const size: PropTypes.Requireable<string>;
    }
    export namespace defaultProps {
        const block_1: boolean;
        export { block_1 as block };
        const className_1: any;
        export { className_1 as className };
        const disabled_1: boolean;
        export { disabled_1 as disabled };
        const htmlType_1: string;
        export { htmlType_1 as htmlType };
        const loading_1: boolean;
        export { loading_1 as loading };
        const priority_1: Priorities;
        export { priority_1 as priority };
        const size_1: Sizes;
        export { size_1 as size };
        const type_1: ControlTypes;
        export { type_1 as type };
    }
}
```

**`Popover.d.ts`**

```typescript
export default Popover;
declare function Popover({ children, className, content, preferredPlacement, title }: {
    children: any;
    className: any;
    content: any;
    preferredPlacement: any;
    title: any;
}): JSX.Element;
declare namespace Popover {
    namespace defaultProps {
        const className: any;
        const preferredPlacement: Positions;
        const title: any;
    }
    namespace propTypes {
        export const children: PropTypes.Validator<PropTypes.ReactNodeLike>;
        const className_1: PropTypes.Requireable<string>;
        export { className_1 as className };
        export const content: PropTypes.Validator<PropTypes.ReactNodeLike>;
        const preferredPlacement_1: PropTypes.Requireable<string>;
        export { preferredPlacement_1 as preferredPlacement };
        const title_1: PropTypes.Requireable<PropTypes.ReactNodeLike>;
        export { title_1 as title };
    }
}
```

**`Summary.d.ts`**
```typescript
export default Summary;
declare function Summary({ action, as: Element, className, content, description, help, icon, illustration, info, status, title, }: {
    action: any;
    as: any;
    className: any;
    content: any;
    description?: any;
    help: any;
    icon: any;
    illustration: any;
    info?: any;
    status: any;
    title: any;
}): JSX.Element;
declare namespace Summary {
    namespace propTypes {
        const action: PropTypes.Requireable<PropTypes.InferProps<{
            text: PropTypes.Validator<PropTypes.ReactNodeLike>;
            href: PropTypes.Validator<string>;
            'aria-label': PropTypes.Requireable<string>;
            target: PropTypes.Requireable<string>;
            onClick: PropTypes.Requireable<(...args: any[]) => any>;
        }>>;
        const as: PropTypes.Requireable<string>;
        const className: PropTypes.Requireable<string>;
        const content: (props: any, propName: any, ...rest: any[]) => any;
        const description: PropTypes.Requireable<PropTypes.ReactNodeLike>;
        const help: (props: any, propName: any, ...rest: any[]) => any;
        const info: PropTypes.Requireable<PropTypes.InferProps<{
            'aria-label': PropTypes.Validator<string>;
            content: PropTypes.Validator<PropTypes.ReactNodeLike>;
            presentation: PropTypes.Requireable<string>;
            title: PropTypes.Requireable<PropTypes.ReactNodeLike>;
        }>>;
        const illustration: (props: any, propName: any, ...rest: any[]) => any;
        const icon: any;
        const status: PropTypes.Requireable<string>;
        const title: PropTypes.Validator<PropTypes.ReactNodeLike>;
    }
    namespace defaultProps {
        const action_1: any;
        export { action_1 as action };
        const as_1: string;
        export { as_1 as as };
        const className_1: any;
        export { className_1 as className };
        const content_1: any;
        export { content_1 as content };
        const help_1: any;
        export { help_1 as help };
        const illustration_1: any;
        export { illustration_1 as illustration };
        const status_1: any;
        export { status_1 as status };
    }
}
import PropTypes from "prop-types";
```

</details>

👌 So as a consumer you get all benefits of TypeScript already now for all components 💯 

Screen Shot 2021-04-15 at 5 35 50 PM

---

Screen Shot 2021-04-15 at 5 34 22 PM

---

Screen Shot 2021-04-15 at 5 19 39 PM

---

Screen Shot 2021-04-15 at 5 27 37 PM

### Breaking Change

**Note: we do provide a codemod for this release, so it should be easy and fast to migrate.**

`react-to-typescript-definitions` doesn't "pay attention" to following assignments when parse files (you have seen this pattern in many of components):

```javascript
Alert.ArrowPosition = ArrowPosition;
Alert.Size = { SMALL: Size.SMALL, LARGE: Size.LARGE };
Alert.Type = Sentiment;
```

which means that it won't be possible to use those things (`Alert.Size.SMALL`, `Button.Type.INFO` etc.) at TypeScript applications for components (`@transferwise/components`) that are written in JavaScript.

Taking into account above issue and amount of components that are written on JavaScript probably the best should would be get rid of such pattern and expose common constants (as TS enums) separately, right from the package:

**Before**

```javascript
import { Alert, Button, Avatar, Popover } from '@transferwise/components';

<Alert type={Alert.Sentiment.INFO} />
<Alert type="info" />

<Button size={Button.Size.EXTRA_LARGE} />
<Button size="xl" />

<Avatar theme={Avatar.Theme.DARK} />
<Logo type={Logo.ProfileType.PERSONAL} />

<Popover preferredPlacement={Popover.Position.LEFT} />
```

**After**

```javascript
import { Alert, Button, Avatar, Popover, Size, Sentiment, Theme, Position, ProfileType } from '@transferwise/components';

<Alert type={Sentiment.INFO} />
<Alert type="info" />

<Button size={Size.EXTRA_LARGE} />
<Button size="xl" />

<Avatar theme={Theme.DARK} />
<Logo type={ProfileType.PERSONAL} />

<Popover preferredPlacement={Position.LEFT} />
```

This is something that we actually considered do at some point anyway. At the end of the day this gives us following benefits:

- saves us a bit of bundle size

- allow to write more concise code as at consumer side 
    ```javascript
    <Loader size={Size.SMALL} />
    // vs
    <Loader size={Loader.Size.SMALL} />
    ```
    as here in the package, we won't need write that [custom](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/alert/Alert.js#L115-L117) [mapping](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/button/Button.js#L52-L59) [anymore](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/popover/Popover.js#L60-L69).

- consumers will be able to reuse those constants in their apps without need of importing entire component(s) and/or writing exactly same constants

- this would sort of push us to keep those enums consistent (we have multiple duplicate, similar but inconsistent instances atm, here is just one example: ([`Position`](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/common/propsValues/position.js) vs [`Placement`](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/common/propsValues/placement.js) vs [`ArrowPosition`](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/alert/withArrow/withArrow.js#L5-L12))


packages/components/tsconfig.json
```diff
@@ -9,14 +10,28 @@
...
-  "exclude": ["node_modules"]
+  "include": ["**/*.ts", "**/*.tsx"],
+  "exclude": [
+    "node_modules",
+    "build",
+    "scripts",
+    "test",
+    "public",
+    "**/*.story.*",
+    "**/*.spec.*",
```
sbrandwoo: You might want TypeScript to work when writing tests, but for the tests and stories not to appear in the compiled output. For this reason Crab has a `tsconfig.json` for type checking and a `tsconfig-build.json` for building, which extends the base file with the excludes.

packages/components/CONTRIBUTING.md
```diff
@@ -39,6 +39,12 @@ As a reminder: to load the docs, run `docs` **from the root of the monorepo**, n
 
 Please familiarise yourself with our [accessibility guidelines](https://github.com/transferwise/neptune-web/blob/main/ACCESSIBILITY.md) and ensure your changes meet accessibility standards.
 
+# TypeScript Support
+
+We encourage you write new components on TypeScript, please consider refactor component(s) to TypeScript if you do major changes/refactoring.
+
+If you do migrate some component(s) to TypeScript please make sure to double check and if necessary remove it from [list of component for custom type generation process](https://github.com/transferwise/neptune-web/blob/main/packages/components/scripts/generate-type-declarations.js#L10).
+
```
dozortsev, line 47: link: https://github.com/transferwise/neptune-web/blob/enable-ts-for-components/packages/components/scripts/generate-type-declarations.js#L10

packages/components/package.json
```diff
@@ -5,6 +5,7 @@
   "license": "Apache-2.0",
   "main": "./build/umd/polyfill/main.js",
   "module": "./build/es/polyfill/index.js",
+  "types": "./build/es/polyfill/index.d.ts",
```
dozortsev, line 8: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#including-declarations-in-your-npm-package

packages/components/package.json
```diff
@@ -30,15 +31,16 @@
     "build:compile-css": "gulp compileLess --dest=src",
     "build:umd": "NODE_ENV=umd rollup -c",
     "build:umd-nopolyfill": "NODE_ENV=umd-nopolyfill rollup -c",
-    "build:es": "NODE_ENV=es babel src -d build/es/polyfill --ignore '**/*.spec.js','**/*.story.js','**/__mocks__','**/test-utils'",
-    "build:es-nopolyfill": "NODE_ENV=es-nopolyfill babel src -d build/es/no-polyfill --ignore '**/*.spec.js','**/*.story.js','**/__mocks__','**/test-utils'",
+    "build:es": "NODE_ENV=es babel src --extensions '.tsx,.js,.ts' -d build/es/polyfill --ignore '**/*.spec.js','**/*.story.js','**/__mocks__','**/test-utils'",
+    "build:es-nopolyfill": "NODE_ENV=es-nopolyfill babel src --extensions '.tsx,.js,.ts' -d build/es/no-polyfill --ignore '**/*.spec.js','**/*.story.js','**/__mocks__','**/test-utils'",
```
dozortsev, line 35: Babel transpiles JS and TS files

packages/components/package.json
```diff
@@ -30,15 +31,16 @@
...
+    "build:generate-types": "tsc --declarationDir 'build/es/polyfill' && tsc --declarationDir 'build/es/no-polyfill' && node ./scripts/generate-type-declarations.js",
```
dozortsev, line 36: TypeScript generates type definitions for all files

and `react-to-typescript-definitions` rewrites type definitions for JS files

packages/components/src/alert/Alert.js
```diff
@@ -132,9 +132,9 @@ Alert.propTypes = {
   /** The presence of the onDismiss handler will trigger the visibility of the close button */
   onDismiss: PropTypes.func,
   /** The type dictates which icon and colour will be used */
-  type: PropTypes.oneOf(Object.values(Alert.Type)),
+  type: PropTypes.oneOf(['negative', 'neutral', 'positive', 'warning', 'info', 'error', 'success']),
 
-  /** @DEPRECATED */
+  /** @deprecated no arrow for `Alert` component anymore, consider to use `InlineAlert` component */
```
dozortsev, line 137: according to JSDoc spec https://jsdoc.app/tags-deprecated.html

packages/components/src/card/Card.js
```diff
@@ -69,7 +69,7 @@ const Card = ({
 const hasChildren = ({ children }) => children;
 
 Card.propTypes = {
-  as: PropTypes.elementType,
+  as: PropTypes.string,
```
dozortsev, line 72: react-to-typescript-definitions doesn't support `elementType` defaulting to `any` type,

since we pass there string value `<Card as="div" .. />` I just decided to change it to `PropTypes.string` which also would for the types generating tool

packages/components/src/common/Option/Option.js
```diff
@@ -64,7 +64,7 @@ Option.propTypes = {
   content: PropTypes.node,
   onClick: PropTypes.func,
   href: requiredIf(PropTypes.string, (props) => props.as === 'a'),
-  as: PropTypes.elementType,
+  as: PropTypes.string,
```
dozortsev, line 67: same as https://github.com/transferwise/neptune-web/pull/846/files#r608755434

packages/components/src/common/propsValues/sentiment.ts
```diff
@@ -0,0 +1,19 @@
...
+   */
+  INFO = 'info',
+  /**
+   * @deprecated
+   */
+  ERROR = 'error',
+  /**
+   * @deprecated
+   */
+  SUCCESS = 'success',
```
dozortsev, line 18: it's important to follow [JSDoc spec](https://jsdoc.app/tags-deprecated.html) and use lower case some tools rely use that

e.g VSCode and IntelliJ strikethrough such values
Screen Shot 2021-04-07 at 4 08 11 PM

packages/components/src/alert/Alert.js
```diff
@@ -132,21 +123,14 @@ Alert.propTypes = {
   /** The presence of the onDismiss handler will trigger the visibility of the close button */
   onDismiss: PropTypes.func,
   /** The type dictates which icon and colour will be used */
-  type: PropTypes.oneOf(Object.values(Alert.Type)),
+  type: PropTypes.oneOf(['negative', 'neutral', 'positive', 'warning', 'info', 'error', 'success']),
 
-  /** @DEPRECATED */
+  /** @deprecated no arrow for `Alert` component anymore, consider to use [`InlineAlert`](https://transferwise.github.io/neptune-web/components/alerts/InlineAlert) component */
   arrow: deprecated(
```
dozortsev, line 129: JSDoc allow Markdown syntax, here who this component would look like at VSCode

Screen Shot 2021-04-08 at 6 28 27 PM

packages/components/src/common/locale/index.js
```diff
@@ -1,4 +1,15 @@
+/**
+ * Default language
+ *
+ * @type {string}
+ */
```
dozortsev, line 5: TypeScript can read type from JSDoc when generate `.d.ts` file, in this file I needed to slightly adjust types, this won't be a breaking change

packages/components/tsconfig.json
dozortsev: we will address as part a bit later, in separate PR https://transferwise.atlassian.net/browse/DS-1003

packages/components/src/alert/withArrow/alertArrowPositions.ts
```diff
@@ -0,0 +1,9 @@
+// TODO: consider to move this enum into component file once we migrate it on TypeScript or replace with some common enum
```
conorcussell, line 1: I think in practise `enum` is quite unwieldy/irritating

Generally I think using union types instead works better - see: https://fettblog.eu/tidy-typescript-avoid-enums/#string-enums-are-named-types

[This post from Bloomberg also agrees](https://www.techatbloomberg.com/blog/10-insights-adopting-typescript-at-scale/)

dozortsev: okey, I have a look at these articles,

we not going to use enums for types declarations, we plan to use `type` (see how I implemented [`Loader`](https://github.com/transferwise/neptune-web/blob/enable-ts-for-components/packages/components/src/loader/Loader.tsx), [`Badge`](https://github.com/transferwise/neptune-web/blob/enable-ts-for-components/packages/components/src/badge/Badge.tsx)), these enums are for passing data into props `<Alert arrow={ArrowPosition.TOP} />

conorcussell: Yeah I just had a look at the examples and noticed that as a consumer I would be able to do `<Button type="submit">` and that's the main thing I was worried about so its probably all good!

packages/components/src/common/propsValues/size.ts
```diff
@@ -0,0 +1,13 @@
+export enum Sizes {
```
sbrandwoo, line 1: Enums shouldn't be plural. If you consider a function that takes an enum value, it takes a single Size, not multiple Sizes:

```typescript
function render(size: Sizes) {}
```

dozortsev: okey, I see they use singular form at [enums documentation](https://www.typescriptlang.org/docs/handbook/enums.html)

packages/components/src/alert/withArrow/alertArrowPositions.ts
oliverviljamaa: @dozortsev Do we need the enums at all when we can type safely do `size="md"`?

packages/components/.eslintrc
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": ["@transferwise", "../../.eslintrc"],
+  "ignorePatterns": ["**/*.ts", "**/*.tsx"]
```
oliverviljamaa, line 3: Do I understand correctly that we ignore TypeScript files from being linted? 🤔 If yes, why?

packages/components/CONTRIBUTING.md
```diff
@@ -39,6 +39,12 @@ As a reminder: to load the docs, run `docs` **from the root of the monorepo**, n
 
 Please familiarise yourself with our [accessibility guidelines](https://github.com/transferwise/neptune-web/blob/main/ACCESSIBILITY.md) and ensure your changes meet accessibility standards.
 
+# TypeScript Support
+
+We encourage you write new components on TypeScript, please consider rewrite components to TypeScript if you do major changes/refactoring.
```
oliverviljamaa, line 44: I recommend being even stricter with this 😊

packages/components/README.md
```diff
@@ -24,20 +24,29 @@ npm install @transferwise/components react react-dom prop-types currency-flags
...
+    </Provider>
   );
 }
 ```
 
+### TypeScript | Type Declarations
+
+This package exposes [type declarations](https://www.typescriptlang.org/docs/handbook/2/type-declarations.html) for all components.
+
+Note: types for some of components are not 100% accurate (some of them will be just [`any`](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#any)), this is due to fact that many of components written on JavaScript and types generated by [react-to-typescript-definitions](https://www.npmjs.com/package/react-to-typescript-definitions) based of components `PropTypes` and unfortunatly the library doesn't resolve all `PropTypes` features and use cases (e.g custom validators, `PropTypes.objectOf`, [requireIf](https://www.npmjs.com/package/react-required-if), etc). But bear with us as eventually it would be fully on TypeScript.
```
oliverviljamaa, line 45: Nice documentation 👏

packages/components/package.json
oliverviljamaa: Did you find Babel with TypeScript easier to integrate than TypeScript directly?

packages/components/src/alert/Alert.js
oliverviljamaa: Awesome!

packages/components/.eslintrc
dozortsev: yes, this is because eslint shows inappropriate errors at TS files, we definitely plan to enable it, I was hope do that via `@transferwise/eslint-config` package

oliverviljamaa (commented):
Awesome work on this 💪

packages/components/.eslintrc
fergusjordantw: Is it not possible to just override some of those rules you don't want for a specific pattern for the moment? We do this in send-flow to remove rules that would be overkill for stuff like tests https://github.com/transferwise/send-flow/blob/master/.eslintrc.js#L105-L112

packages/components/src/alert/withArrow/alertArrowPositions.ts
dozortsev: good question, I see following benefits/reason to have them exposed:

- I've seen people create custom and almost exactly the same enums in their repos, exposing from here would avoid creating duplicates. Note: those enums are not only for `@transferwise/components` you can use them for your own custom components
- this way we will be able enforce consistency, as some people call them `small` | `medium` instead `sm` | `md` in their apps, or `top_left` | `top_right` vs `left-up` | `right-up`, some of them more valuable for consistency e.g `Breakpoint` - thresholds for switching between mobile view and desktop, `Sentiment` (`NEGATIVE` - red, `NEUTRAL` - blue, `POSITIVE` - green) used a lot (consistency of naming)
- TS enable JSDoc, so we can add documentation to them for a better/guided usage

   Screen Shot 2021-04-21 at 2 30 37 PM

packages/components/.eslintrc
dozortsev: I see, thanks, we plan to address ESLint and many other TS related things as separate PR https://transferwise.atlassian.net/browse/DS-1003

oliverviljamaa (approved):
👏

packages/components/package.json
```diff
@@ -26,19 +27,19 @@
...
     "build:umd": "NODE_ENV=umd rollup -c",
     "build:umd-nopolyfill": "NODE_ENV=umd-nopolyfill rollup -c",
-    "build:es": "NODE_ENV=es babel src -d build/es/polyfill --ignore '**/*.spec.js','**/*.story.js','**/__mocks__','**/test-utils'",
-    "build:es-nopolyfill": "NODE_ENV=es-nopolyfill babel src -d build/es/no-polyfill --ignore '**/*.spec.js','**/*.story.js','**/__mocks__','**/test-utils'",
-    "build:copy-files": "cpx 'src/**/!(db)/*.{css,json,svg}' build/es/polyfill",
-    "build:copy-files2": "cpx 'src/**/!(db)/*.{css,json,svg}' build/es/no-polyfill",
-    "build:copy-lang": "cpx 'i18n/*.json' build/i18n && cpx 'i18n/*.json' build/es/i18n && cpx 'i18n/index.js' build/i18n",
+    "build:es": "NODE_ENV=es babel src --extensions '.tsx,.js,.ts' -d build/es/polyfill --ignore '**/*.spec.js','**/*.story.js','**/__mocks__','**/test-utils'",
+    "build:es-nopolyfill": "NODE_ENV=es-nopolyfill babel src --extensions '.tsx,.js,.ts' -d build/es/no-polyfill --ignore '**/*.spec.js','**/*.story.js','**/__mocks__','**/test-utils'",
+    "build:generate-types": "tsc && node ./scripts/generate-type-declarations.js",
```
dozortsev, line 36: TypeScript generates `.d.ts` files for TS files and react-to-typescript-definitions does same for JS files

packages/components/package.json
dozortsev: oh only now noticed this comment,

yep, it's easier to get started with TS this way, you don't need to change your build pipeline and configs much

I mostly followed this guide https://www.typescriptlang.org/docs/handbook/babel-with-typescript.html#babel-for-transpiling-tsc-for-types

sbrandwoo: I think if you want to be able to import `.less` files and  such from TypeScript files, then Babel + some bundler is the only option.

Raw `tsc` only works if you are doing something more simple.

sbrandwoo (approved):
:partying_face:

czebe (approved):
F I N A L L Y !

packages/components/package.json
oliverviljamaa: Oh, yep, I meant TypeScript through Rollup.

sbrandwoo (approved):

bundlesize.config.json
```diff
@@ -2,19 +2,19 @@
...
     },
     {
       "path": "./packages/components/build/umd/no-polyfill/main.js",
-      "maxSize": "120kB"
+      "maxSize": "124kB"
     },
     {
       "path": "./packages/components/build/main.css",
-      "maxSize": "10kB"
+      "maxSize": "9kB"
```
helendurranttw, line 13: Why did the CSS output sizes change?

packages/components/src/avatar/Avatar.spec.js
```diff
@@ -20,7 +21,7 @@ describe('Avatar', () => {
 
   it('renders a small sized emoji Avatar with outline', () => {
     const { getByText } = render(
-      <Avatar type={Avatar.Type.EMOJI} size={Avatar.Size.SMALL} outlined>
+      <Avatar type={AvatarType.EMOJI} size={Size.SMALL} outlined>
```
helendurranttw, line 24: I'm not sure there's anything we can do about this, but it seems a little strange to have one prop avatar specific, and one not (AvatarType but Size). Maybe AvatarType could be renamed? 🤔

packages/components/src/button/Button.js
```diff
@@ -67,29 +58,20 @@ Button.propTypes = {
...
   // eslint-disable-next-line
   onClick: requiredIf(PropTypes.func, (props) => props.htmlType !== 'submit'),
-  priority: PropTypes.oneOf([
-    Button.Priority.PRIMARY,
-    Button.Priority.SECONDARY,
-    Button.Priority.TERTIARY,
-  ]),
-  /** @DEPRECATED Type.PRIMARY, Type.PAY, Type.SECONDARY, Type.DANGER, Type.LINK */
+  priority: PropTypes.oneOf(['primary', 'secondary', 'tertiary']),
+  /** @deprecated `Type.PRIMARY`, `Type.PAY`, `Type.SECONDARY`, `Type.DANGER`, `Type.LINK` */
```
helendurranttw, line 62: Maybe we should update this deprecation notice to be raw values, since that's what will be autocompleted?

packages/components/src/button/Button.spec.js
```diff
@@ -4,10 +4,12 @@ import '@testing-library/jest-dom';
 import { render, cleanup, screen, userEvent } from '../test-utils';
 
 import Button from '.';
+import { ControlType, Type, Priority, Size } from '../common';
 
-const { ACCENT, POSITIVE, NEGATIVE } = Button.Type;
-const { PRIMARY, SECONDARY, TERTIARY } = Button.Priority;
-const { SMALL, MEDIUM, LARGE } = Button.Size;
+const { ACCENT, POSITIVE, NEGATIVE } = ControlType;
+const { PAY, LINK, DANGER } = Type;
```
helendurranttw, line 10: What is the difference between ControlType and Type?

packages/components/src/common/commonProps.ts
```diff
@@ -0,0 +1,9 @@
+export type CommonProps = {
+  /**
+   * Space-separated list of the case-sensitive CSS classes,
```
helendurranttw, line 3: Nice!

packages/components/src/common/propsValues/size.ts
```diff
@@ -0,0 +1,13 @@
+export enum Size {
+  EXTRA_SMALL = 'xs',
+  SMALL = 'sm',
+  MEDIUM = 'md',
+  LARGE = 'lg',
+  EXTRA_LARGE = 'xl',
+}
+
+export type SizeExtraSmall = 'xs';
```
helendurranttw, line 9: I'm wondering why you're doing this here but not in other prop defs?

bundlesize.config.json
dozortsev: it didn't change, the bundle is `8.33KB` while boundary was `10` so I just decided adjusted it to a bit lower boundary, when I was adjusting other thresholds, probably shouldn't have done that to avoid introducing any noise PR changes 😄

packages/components/src/avatar/Avatar.spec.js
dozortsev: yeah, eventually and ideally we should get rid of component specific enums and have only generic ones, I left [comments](https://github.com/transferwise/neptune-web/blob/enable-ts-for-components/packages/components/src/avatar/avatarTypes.ts#L1) in each of those custom enums saying that we should get rid of them in

I think this enum alignment can be done as part of separate PR

packages/components/src/button/Button.spec.js
dozortsev: I guess you introduced them here as part of https://github.com/transferwise/neptune-web/pull/737

Screen Shot 2021-04-28 at 11 09 03 AM

packages/components/src/common/propsValues/size.ts
dozortsev: I define these types for Size because Size types were necessary for me when I refactored some components in TS as part of this PR, to be specific for [`Loader`](https://github.com/transferwise/neptune-web/blob/enable-ts-for-components/packages/components/src/loader/Loader.tsx#L7) and [`Badge`](https://github.com/transferwise/neptune-web/blob/enable-ts-for-components/packages/components/src/badge/Badge.tsx#L10) components they have `size` prop

e.g I also needed same types for Theme and did same there https://github.com/transferwise/neptune-web/blob/enable-ts-for-components/packages/components/src/common/theme.ts

packages/components/src/button/Button.spec.js
dozortsev: note: [`Type`](https://github.com/transferwise/neptune-web/blob/enable-ts-for-components/packages/components/src/common/propsValues/type.ts#L2) is deprecated and I think it should be deleted as part of https://transferwise.atlassian.net/browse/DS-1027

packages/components/src/button/Button.js
dozortsev: okey, done https://github.com/transferwise/neptune-web/pull/846/commits/f7b24d4a92b070102e8a5f0f9df061bc94c002e1

conorcussell (approved):

packages/components/babel.config.js
```diff
@@ -77,7 +81,6 @@ module.exports = {
   sourceType: 'unambiguous',
   plugins: [
     '@babel/plugin-proposal-class-properties',
-    '@babel/plugin-proposal-optional-chaining',
```
helendurranttw, line 80: Why did this get removed?

packages/components/bin/neptune-upgrader.js
```diff
@@ -94,11 +95,15 @@ const run = async () => {
   const coreCommand = `${codeshiftPath} ${files} -t=${transform} ${options}`;
 
   if (confirm) {
+    console.log('🚀 Running codemod, please wait a couple sec...');
```
helendurranttw, line 98:
```suggestion
    console.log('🚀 Running codemod, please wait a few seconds...');
```

packages/components/scripts/generate-type-declarations.js
```diff
@@ -0,0 +1,87 @@
...
+ * (same folder as we specifing in `tsconfig.json` via `declarationDir`)
+ */
+const typesDistFolder = 'build/types';
+
+/**
+ * list of JS files that uses prop-types (basically React components)
+ * and that are exposed from `@transferwise/components` library
+ */
+const jsFiles = [
+  { filename: 'Provider', folder: 'provider' },
```
helendurranttw, line 15: @dozortsev did you try and do this using globs, like Andrea suggested?

packages/components/src/loader/Loader.tsx
```diff
@@ -0,0 +1,51 @@
...
+import classnames from 'classnames';
+import { Size, SizeExtraSmall, SizeSmall, SizeMedium, SizeLarge, SizeExtraLarge } from '../common';
+
+import './Loader.css';
+
+type SizeType = SizeExtraSmall | SizeSmall | SizeMedium | SizeLarge | SizeExtraLarge;
+
+type Props = {
+  /**
+   * @deprecated this is legacy property, use `size` property instead
```
helendurranttw, line 11:
```suggestion
   * @deprecated use `size` instead
```

packages/components/src/loader/Loader.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  const legacySize: SizeType = small ? Size.EXTRA_SMALL : size;
+
+  return (
+    <div
+      className={classnames(style('tw-loader'), style(`tw-loader--${legacySize}`))}
+      data-testid={restProps['data-testid']}
+    >
+      {Array(5)
+        // @ts-expect-error
```
helendurranttw, line 41: Why is this needed?

packages/components/src/logo/logoTypes.ts
```diff
@@ -0,0 +1,5 @@
+// TODO: consider to move this enum into component file once we migrate it on TypeScript or replace with some common enum
+export enum Type {
```
helendurranttw, line 2: You may as well call this LogoType, no?

packages/components/src/markdown/Markdown.js
```diff
@@ -65,10 +65,48 @@ function stripNodes({ blockList, parsed }) {
...
 Markdown.propTypes = {
   children: PropTypes.string.isRequired,
-  as: PropTypes.elementType,
+  as: PropTypes.string,
   className: PropTypes.string,
-  allowList: PropTypes.arrayOf(PropTypes.oneOf(NODE_TYPE_LIST)),
-  blockList: PropTypes.arrayOf(PropTypes.oneOf(NODE_TYPE_LIST)),
+  allowList: PropTypes.arrayOf(
+    PropTypes.oneOf([
+      'block_quote',
```
helendurranttw, line 72: Can we not even spread the array here? E.g. ...NODE_TYPE_LIST

helendurranttw: (same for the blockList)

packages/components/upgrades/2021-02-v34-35.js
```diff
@@ -18,21 +20,15 @@ const transform = (file, api, options) => {
...
-  const log = (message, level = 0) => {
-    if (options.verbose >= level) {
-      api.report(message);
-    }
-  };
-
   const j = api.jscodeshift;
   const root = j(file.source);
 
-  log('✨ All aboard the transformation train 🚂', 2);
```
helendurranttw, line 30: You did NOT remove the sparkles from my log messages! Put them back! 😂

packages/components/upgrades/2021-02-v34-35.js
```diff
@@ -7,9 +9,9 @@ const TYPE_REPLACEMENTS = {
 };
 
 const PRIORITY_ADDITIONS = {
-  secondary: 'Button.Priority.SECONDARY',
```
helendurranttw, line 10: I'm not sure this is correct. I guess the way I was imagining that these would run, is one version at a time - so you upgrade from 34 - 35, run the script, commit the changes, then do the next step.

In retrospect I guess this might be ok.. I think this could be a slippery slope though. Having to go back and edit old upgrade scripts doesn't seem like a good idea to me.

packages/components/upgrades/2021-04-v35-36-typescript-support.js
```diff
@@ -0,0 +1,289 @@
...
+    component: 'Typeahead',
+    enums: [sizeCase, { old: 'Type', new: 'Sentiment' }],
+  },
+  {
+    component: 'Upload',
+    enums: [{ old: 'UploadStep', new: 'UploadStep' }],
+  },
+];
+
+const ButtonTypeBreakingChange = {
```
helendurranttw, line 129: I think we should enforce old scripts to be run before new ones, so you shouldn't have to do this.

packages/components/upgrades/2021-04-v35-36-typescript-support.js
```diff
@@ -0,0 +1,289 @@
...
+                  value: componentsPackageName,
+                },
+              })
+              .forEach((paths) => {
+                paths.value.specifiers.push(j.importSpecifier(j.identifier(newEnum)));
+              });
+          }
+
+          // do we still need import component?
+          const isComponentNotUsedAtJsx =
```
helendurranttw, line 212: Hmm I don't think we should be doing this - this seems outside the scope of the upgrade, no?

packages/components/upgrades/2021-04-v35-36-typescript-support.js
```diff
@@ -0,0 +1,289 @@
...
+                openingElement: {
+                  name: {
+                    type: 'JSXIdentifier',
+                    name: component,
+                  },
+                },
+              })
+              .size() === 0;
+
+          if (isComponentNotUsedAtJsx) {
```
helendurranttw, line 224: And this

helendurranttw: Or can this be caused by our changes?

packages/components/upgrades/2021-04-v35-36-typescript-support.js
helendurranttw: You could remove a lot of code if we just say - please run the other script first.

packages/components/upgrades/2021-04-v35-36-typescript-support.js
```diff
@@ -0,0 +1,289 @@
...
+              .forEach((paths) => {
+                paths.value.specifiers.push(j.importSpecifier(j.identifier(newProp)));
+              });
+          }
+          return j.jsxIdentifier(`${newProp}.${enumItemValue}`);
+        });
+    },
+  );
+
+  return fileTree.toSource(defaultPrintOptions);
```
helendurranttw, line 288: Nice job with your first codemod! 🎉

packages/components/upgrades/2021-02-v34-35.js
helendurranttw: I take it back! I found it! 😂 ❤️ ✨ ```api.report(`✨ ${message}`);```

packages/components/upgrades/utils/index.js
```diff
@@ -0,0 +1,17 @@
...
+};
+
+export function getLiteralType(parser) {
+  return parser === 'tsx' ? 'StringLiteral' : 'Literal';
+}
+
+export const log = (verbose, message, level = 0) => {
+  if (verbose >= level) {
+    // eslint-disable-next-line no-undef
+    api.report(`✨ ${message}`);
```
helendurranttw, line 15: 😊

packages/docs/liveEditorCode/radioGroup.code.js
```diff
@@ -7,21 +7,18 @@
       value: 'radio-1',
       label: 'Radio1',
       secondary: 'Secondary line 1',
-      name: 'name',
```
helendurranttw, line 10: These are required, according to the code?

packages/docs/pages/about/Setup.mdx
```diff
@@ -31,9 +31,9 @@ import '@transferwise/icons/lib/styles/main.min.css';
 // This is only necessary if you are using the UMD build
 import '@transferwise/components/build/main.css';
 
-import { Provider, Button } from '@transferwise/components';
+import { Provider, Button, Size } from '@transferwise/components';
 import componentTranslations from '@transferwise/components/build/i18n';
-import { getLangFromLocale } from '@transferwise/components/build/es/polyfill/common/locale';
+import { getLangFromLocale } from '@transferwise/components';
```
helendurranttw, line 36: Much nicer 😌

helendurranttw: Though now you can merge with line 34!

packages/docs/pages/components/context/Provider.tsx
```diff
@@ -1,14 +1,20 @@
...
+} from '@transferwise/components';
+
+// @ts-expect-error
 import messages from '@transferwise/components/build/es/no-polyfill/common/closeButton/CloseButton.messages';
 import { DocLink } from '../../../utils/pageUtils';
 import { LiveEditorBlock, GeneratePropsTable } from '../../../utils';
 import { Meta } from '../../../utils/sections';
 
-// @ts-ignore
+// @ts-expect-error
```
helendurranttw, line 17: What's the issue here?

helendurranttw:
The one last thing I'd like resolved before merging is the naming of and use of the common types. I feel like in some places we're going a bit too far - is the orientation of a chevron really the same as the positioning of a drawer?

As for naming - I worry a little that such generic names like position, size, type and width may conflict with other types in the future. Can we be a little more specific?

It's looking great, almost there!

packages/components/src/common/propsValues/width.ts
```diff
@@ -0,0 +1,4 @@
+export enum Width {
+  BLOCK = 'block',
```
helendurranttw, line 2: These don't seem like necessarily intuitive values for width, to me. It could equally be small / large, or 100, 200, 300. Maybe this type can be more specific?

packages/components/babel.config.js
dozortsev: because this is simply redundant, our babel config needs clean up, but for now I'm putting this back as this nothing related with TS changes, sorry for noise changes in PR

packages/components/scripts/generate-type-declarations.js
dozortsev: yes, I did try `glob` but it doesn't simply code, it maybe makes list of files look a bit smaller but I get more hustle with filtering and string transforming at [this part](https://github.com/transferwise/neptune-web/pull/846/files#diff-214eea1d68f4dbd75793dc5213a168ea9ccd9c1459cc3e10569e0e02b0e14e3dR70-R87)

using simple allow list (aka whitelist) of components is much simpler, clear what components we still generate types using that tool and avoid generating for unnecessary files

packages/components/src/loader/Loader.tsx
dozortsev: first argument of [`.fill`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fill) function is required and TS is complaining about it, I simple migrated this component to TS and don't want to fix this thing now as afraid to break something so I just tell TS to ignore this issue using [`// @ts-expect-error`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments)

packages/components/src/logo/logoTypes.ts
dozortsev: yeah, make sense

packages/components/src/markdown/Markdown.js
dozortsev: no, `react-to-typescript-definitions` works in similar way as jscodeshift using AST to scan file

packages/components/upgrades/2021-02-v34-35.js
dozortsev: I like ✨

packages/components/upgrades/2021-02-v34-35.js
dozortsev: ah, sorry, I didn't mean to make this change, thanks for find this!

packages/components/upgrades/2021-04-v35-36-typescript-support.js
dozortsev: yeah, I also think this way https://github.com/transferwise/neptune-web/pull/846#discussion_r622490654

Here I need this because in some cases we need `Button.Type` while in some `Button.ControlType` this is because when people migrate they still didn't opt out of `@deprecated` things in `Button`

packages/components/upgrades/2021-04-v35-36-typescript-support.js
dozortsev: this is when people import component just to use enum

packages/components/upgrades/2021-04-v35-36-typescript-support.js
dozortsev: https://github.com/transferwise/neptune-web/pull/846#discussion_r623307549

packages/components/upgrades/2021-04-v35-36-typescript-support.js
dozortsev: thanks 🙇

packages/components/upgrades/utils/index.js
dozortsev: ✨

packages/docs/liveEditorCode/radioGroup.code.js
dozortsev: this is little bug in docs, this prop here is totally redundant, `RadioGroup` doesn't have this prop [here](https://github.com/transferwise/neptune-web/blob/enable-ts-for-components/packages/components/src/radioGroup/RadioGroup.js#L46-L51)

packages/docs/pages/about/Setup.mdx
dozortsev: good point, thanks for noticing

packages/docs/pages/components/context/Provider.tsx
dozortsev:
```
neptune-web/packages/docs/liveEditorCode/provider.code.js' is not a module.ts(2306)
```

packages/components/src/common/propsValues/width.ts
dozortsev: I took this enum from Tabs component https://github.com/transferwise/neptune-web/blob/main/packages/components/src/tabs/Tabs.js#L553

this is about width being block similar to [`display: block`](https://developer.mozilla.org/en-US/docs/Web/CSS/display) and auto means width based of inner content.

dozortsev:
> I feel like in some places we're going a bit too far - is the orientation of a chevron really the same as the positioning of a drawer?

yes, they are same ([`Chevron`](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/chevron/Chevron.js#L54) and [`Drawer`](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/drawer/Drawer.js#L48) both based on `Position`). Those that have inconsistent values with common enums have their custom enum (note: temporarily, needs to be fixed separately) e.g `AlertArrowPosition`.


> As for naming - I worry a little that such generic names like position, size, type and width may conflict with other types in the future. Can we be a little more specific?

I mean they haven't been conflicting until now, why would they do that in future? at consumer side people can use aliases to avoid naming clashes

packages/docs/pages/blog/2021-04-26-typescript-support-for-components.tsx
```diff
@@ -0,0 +1,154 @@
...
+      </li>
+      </ul>
+      <h2>Misc</h2>
+      <p>Components that are missing support of types are:
+      <code>Upload</code>, <code>MoneyInput</code>, <code>Dimmer</code> and <code>Snackbar</code>. This is because they wrapped into HoC right at <code>export</code>
+      statement which react-to-typescript-definitions does not resolve and kinda skips generating types for them. But I guess we would improve/fix that any time soon in next releases.
+      </p>
+    </div>
+  );
+}
```
andreapiras, line 154:
```suggestion
/* eslint-disable react/jsx-curly-brace-presence */
/* eslint-disable prettier/prettier */
import React, { ReactElement } from 'react';
import {Alert,Sentiment} from '@transferwise/components'
import { Meta } from '../../utils/sections';

export const meta: Meta = {
  name: 'Neptune Web v36.0.0 release',
  date: '2021-04-26',
  authors: [
    {
      name: 'Anton Dozortsev',
      githubUsername: 'dozortsev',
    },
  ],
};

export default function PageContent(): ReactElement {
  return (
    <div>
      <br />
      TypeScript setup for <code>@transferwise/components</code> library.
      <br />
      <br />
      <h2>Features</h2>
      <p />
      <b>Types</b>
      <p>
        The library now exposes types for the majority of components (see <a href='#exception-list'>exceptions</a>).
        Types are generated by <a href="https://www.npmjs.com/package/react-to-typescript-definitions">react-to-typescript-definitions</a> which generates, with an accuracy of ~90%,&nbsp;
        <a href="https://www.typescriptlang.org/docs/handbook/2/type-declarations.html">type declarations</a> based on <a href="https://reactjs.org/docs/typechecking-with-proptypes.html">PropTypes</a>
      </p>
      <b>JSDoc</b>
      <p>
        Types support will enable <a href="https://jsdoc.app/">JSDoc</a> annotation.<br/>
        Extra helper documentation will appear in your editor by hovering over a component or its props.
      </p>
      <b>TypeScript Support</b>
      <p>
      After the release new components from <code>@transferwise/components</code> can be written in TypeScript.<br/>
      TS code will be transpiled by Babel (via <a href="https://babeljs.io/docs/en/babel-preset-typescript">@babel/preset-typescript</a>).<br/>
      When contributing please consider rewriting your component(s) with TypeScript otherwise please follow these{' '}
      <a href="https://github.com/transferwise/neptune-web/blob/main/packages/components/CONTRIBUTING.md#js-component-rules">rules for JS components and their props-types</a>{' '}
      to allow <code>react-to-typescript-definitions</code> generates correct types.
      </p>
      <h2 id="breaking-changes">Breaking Changes</h2>
      <p />
      <b>Detached Enums</b><br/><br/>
      <Alert type={Sentiment.WARNING}><b>Note:</b> This breaking change will affect a lot of apps, in this release we&apos;ll provide a codemod script to update consumer&apos;s codesbases and ease the migration process (see migration checklist section).
      </Alert ><br/>
      <p>
      Removed components&apos; static properties like the example below <code>Alert.Size.SMALL</code>:
      <pre>
        {"import { Alert, Popover, Button } from '@transferwise/components';"}<br/>
        <br/>
        {"<Alert size={Alert.Size.SMALL} />"}<br/>
        {"<Popover preferredPlacement={Popover.Placement.TOP} />"}<br/>
        {"<Button type={Button.Type.POSITIVE} />"}<br/>
      </pre>
      instead enums are exposed separately and you will need to import them as follow:
      <pre>
        {"import { Alert, Popover, Button, Size, Position, Sentiment } from '@transferwise/components';"}<br/>
        <br/>
        {"<Alert size={Size.SMALL} />"}<br/>
        {"<Popover preferredPlacement={Position.TOP} />"}<br/>
        {"<Button type={Sentiment.POSITIVE} />"}<br/>
      </pre>
      
      
      Raw values can also be used although we don&apos;t recommend it:
      <pre>
        {'<Alert size="sm" />'}<br/>
        {'<Popover preferredPlacement="top" />'}<br/>
        {'<Logo type="positive" />'}<br/>
      </pre>
      <b id="reasons-for-enums">The use of the exposed enums is encouraged for a couple of reasons:</b>
      <ul>
        <li>
          Enums can be reused across your app also outside <code>@transferwise/components</code> scope preventing duplicated definitions.
        </li>
        <li>
          Enums will enforce consistency across TW codebase by helping developer to align on variables&apos; values.
          <br/><br/>Check the following examples for unconsistent definitions:
           <ul>
            <li><code>small</code> | <code>medium</code> vs <code>sm</code> | <code>md</code></li>
            <li><code>top_left</code> | <code>top_right</code> vs <code>left-up</code> | <code>right-up</code></li>
           </ul>  
           <br/>
          Some examples of reusable enums will be: 
          <ul>
            <li><code>Breakpoint</code> - thresholds for switching between mobile view and desktop,</li>
            <li><code>Sentiment</code> - <code>NEGATIVE</code> - red, <code>NEUTRAL</code> - blue, <code>POSITIVE</code> - green)</li>
          </ul>
        </li>
        <br/>
        <li>Enums usage will improve editor&apos;s JSDoc support</li>
      </ul>
      </p>
      <h2 id="migration-checklist">Migration Checklist</h2>
      <ul>
        <li>
        Please check that all your dependencies versions are equal or higher than this version release.
        </li>
        <li>Delete any custom types declarations for <code>@transferwise/components</code> from <code>global.d.ts</code> / <code>declarations.d.ts</code> and similar files as the library will now provide types.</li>
        <li>Adjust the usage of enums either manually or by running codemod:
          <pre>
          $ yarn run neptune-upgrade-util<br/>
          ? ✨ Which upgrade would you like to run?<br/>
          ...<br/>
            ▶ 2021-04-v35-36-typescript-support.js<br/>
          </pre>
        </li>
      <li>
        If ESLint is enabled for TypeScript files your <code>eslint-disable-*</code> comments can be deleted as they won&apos;t be necessary anymore.
        <ul>
          <li><a href="https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unsafe-assignment.md"><code>@typescript-eslint/no-unsafe-assignment</code></a></li>
          <li><a href="https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unsafe-member-access.md"><code>@typescript-eslint/no-unsafe-member-access</code></a></li>
        </ul>
      </li>
      </ul>

     <Alert type={Sentiment.WARNING}> Mostly applicable for TypeScript projects:</Alert>
      <ul>
      <li>
        Translations and utils will be exported from the library.
        <br />
        Before
        <pre>
          {"import componentTranslations from '@transferwise/components/build/i18n';"}<br/>
          {"import { getLangFromLocale } from '@transferwise/components/build/es/polyfill/common/locale';"}
        </pre>
        After
        <pre>
          {"import { translations, getLangFromLocale } from '@transferwise/components';"}
        </pre>
        You can still use old imports but TypeScript compiler will raise a warning.
      </li>
      <li>
       It will be possible to have wrong types TypeScript warnings, in that case please fix them manually or ignore them via <code>{"// @ts-expect-error"}</code>.
      </li>
      <li>
        You probably would find some redundant props that have been deleted in previous releases of some of components that you still use
        (very often case would be a <code>locale</code> prop), delete them as they dead code by now.
      </li>
      </ul>
      <h2 id="exception-list">Exceptions</h2>
      <p>Some HOC components haven&apos;t been fully migrated yet:
      <ul>
        <li><code>Upload</code></li>
        <li><code>MoneyInput</code></li>
        <li><code>Dimmer</code></li>
        <li><code>Snackbar</code></li>
      </ul>
      This is because react-to-typescript-definitions is unable to generate types for these specific HOC exports. This will be taken care of in next releases.
      </p>
    </div>
  );
}
```

packages/components/CONTRIBUTING.md
```diff
@@ -39,6 +39,76 @@ As a reminder: to load the docs, run `docs` **from the root of the monorepo**, n
 
 Please familiarise yourself with our [accessibility guidelines](https://github.com/transferwise/neptune-web/blob/main/ACCESSIBILITY.md) and ensure your changes meet accessibility standards.
 
+# TypeScript Support
+
+We strongly encourage you write new components on TypeScript, please consider rewrite components to TypeScript if you do major changes/refactoring.
```
andreapiras, line 44:
```suggestion
We encourage new components to be written in TypeScript, please also consider to rewrite components in TypeScript during major refactoring.
```

dozortsev: 👌

packages/components/CONTRIBUTING.md
```diff
@@ -39,6 +39,76 @@ As a reminder: to load the docs, run `docs` **from the root of the monorepo**, n
...
+
+If you do migrate some component(s) to TypeScript please make sure to double check and if necessary remove it from [list of component for custom type generation process](https://github.com/transferwise/neptune-web/blob/main/packages/components/scripts/generate-type-declarations.js#L10).
+
```
andreapiras, line 47:
```suggestion
While migrating a component to TypeScript please remove it, if necessary, from [list of component for custom type generation process](https://github.com/transferwise/neptune-web/blob/main/packages/components/scripts/generate-type-declarations.js#L10).

```

packages/components/CONTRIBUTING.md
```diff
@@ -39,6 +39,76 @@ As a reminder: to load the docs, run `docs` **from the root of the monorepo**, n
...
+### JS Component Rules
+
+If you don't rewrite component on TypeScript and just simply add new feature or fix bug make sure to ask yourself "how this would work in TypeScript world" this would help us later easier migrate this component to TS.
+
```
andreapiras, line 51:
```suggestion
For JS new features or bug fixes please always keep in mind how they would behave in a Typescript environment as this would help us to migrate them to TS.
```

packages/components/CONTRIBUTING.md
```diff
@@ -39,6 +39,76 @@ As a reminder: to load the docs, run `docs` **from the root of the monorepo**, n
...
+Also, please follow these rules, this will ensure that type declarations for JS components generated correctly by [`react-to-typescript-definitions`](https://www.npmjs.com/package/react-to-typescript-definitions).
+
+Note: these rules necessary only for components that are exposed from the package, but it will great if you follow them for internal components as well for consistency and in case we decide export internal component it would be less of an effort todo that.
```
andreapiras, line 54:
```suggestion
Please follow these rules to ensure that type declarations for JS components gets generated correctly by [`react-to-typescript-definitions`](https://www.npmjs.com/package/react-to-typescript-definitions).

Note: Please try to follow these rules for both exposed and internal components to improve codebase consistency and to reduce the effort in case we decide expose an internal component.
```

packages/components/CONTRIBUTING.md
```diff
@@ -39,6 +39,76 @@ As a reminder: to load the docs, run `docs` **from the root of the monorepo**, n
...
+
+- Make sure `export default` is clear, you shouldn't wrap component in any functions and other stuff, as this will lead to fully skipping generating `.d.ts` file or types with TS error `JSX element type 'SomeComponent' does not have any construct or call signatures.`
+
```
andreapiras, line 57:
```suggestion
- Please avoid wrapping components and have a clear `export default` as otherwise this will lead to fully skipping generating `.d.ts` file or types with TS error `JSX element type 'SomeComponent' does not have any construct or call signatures.`
```

packages/components/CONTRIBUTING.md
```diff
@@ -39,6 +39,76 @@ As a reminder: to load the docs, run `docs` **from the root of the monorepo**, n
...
+// do
+SomeComponent.propTypes = {
+  size: PropTypes.oneOf(['xs', 'sm', 'md']),
+  position: PropTypes.oneOf(['top', 'left', 'right', 'bottom']),
+}
+SomeComponent.defaultProps = {
+  size: Size.SMALL,
+}
+```
+- no need to create following props inside components, they won't work at TypeScript consumer projects
```
andreapiras, line 86:
```suggestion
- There is no need to create static properties as they won't work in TypeScript consumer apps
```

packages/components/CONTRIBUTING.md
```diff
@@ -39,6 +39,76 @@ As a reminder: to load the docs, run `docs` **from the root of the monorepo**, n
...
+// do
+import PropTypes from 'prop-types';
+```
+- avoid using custom validator in prop-types, this will lead to type `any` for those props
+```javascript
+// don't
+onClick: requiredIf(PropTypes.func, ...),
+content: deprecated(PropTypes.node, ...),
+```
+- Ultimately, alwasy check generated types, run `yarn build` inside `packages/components` folder then head to `/build/types/index.d.ts` file, from there, navigate to your component `.d.ts` file and check accuracy of types, if types aren't what meant to be then tweak prop-types in the component and repeat these steps again.
```
andreapiras, line 108:
```suggestion
- Ultimately, always check the generated types by: 
  - run `yarn build` inside `packages/components`
  - head to `/build/types/index.d.ts`,
  - search for your component type declaration in `.d.ts` 
  - if types aren't the expected ones please tweak the prop-types in the component and repeat these steps again.
```

packages/components/src/common/propsValues/control.ts
```diff
@@ -0,0 +1,11 @@
+export enum ControlType {
+  ACCENT = 'accent',
```
andreapiras, line 2: how come we can't just use Sentiment?
also in case we can't can we at least do
export enum ControlType {
  ACCENT = Sentiment.ACCENT,

dozortsev: I don't know, this PR is not about aligning enums, doing `ACCENT = Sentiment.POSITIVE` can bring even more confusion

packages/docs/pages/blog/2021-04-26-typescript-support-for-components.tsx
dozortsev: thanks, I accepted many of these suggestions but not all this is because suggested sentence is changing original meaning and some times we lose the point

packages/components/CONTRIBUTING.md
dozortsev: 👌

packages/components/CONTRIBUTING.md
dozortsev: 👌

packages/components/CONTRIBUTING.md
dozortsev: 👍

packages/components/CONTRIBUTING.md
dozortsev: 👍

packages/components/CONTRIBUTING.md
dozortsev: 👍

packages/components/CONTRIBUTING.md
dozortsev: 🆗

PR #853, opened 2021-03-16
chore: Stylelinting Level Up 🎨

dozortsev:
## 🖼 Context

We have [stylelint](https://stylelint.io/) configured in neptune-web repo but it wasn't really used, its possible push code with violations, loads of existing files were required fixes of violations, we didn't run stylelinting neither as part of CI build or test or something like that.. So this PR address all those issues!

## 🚀 Changes

- enable running linting for changed style files (`.less`, and `.css`) on pre-commit event (via `lint-staged`)
- fixed (and disabled) existing style issues
- added running stylelinting as part of `yarn lint` command

## ✅ Checklist


packages/components/src/snackbar/Snackbar.less
```diff
@@ -45,7 +45,7 @@
       margin: -14px -16px -14px 0;
       background: none;
       outline: 0;
-      font: inherit;
+      font-style: inherit;
```
dozortsev, line 48: `font-style` is first value of `font` https://www.w3schools.com/cssref/pr_font_font.asp

packages/components/src/moneyInput/MoneyInput.less
```diff
@@ -35,8 +35,9 @@
 
   span:not(.tw-icon) {
     min-width: 2.6em; // fits a 3 letter string, ZMW being the longest one
+
     @media (--screen-sm) {
-      min-width: calc(~'2.6em' + var(--size-32));
+      min-width: calc(2.6em + var(--size-32));
```
dozortsev, line 40: stylelint was complaining and failing at this line 

```
$ stylelint --allow-empty-input './src/**/*.less' -f verbose
<input css 1>
 1:1  ✖  Unknown word   CssSyntaxError
```

This code was added here https://github.com/transferwise/neptune-web/pull/666 to fix following bug

Screenshot 2020-10-13 at 10 29 21

I tested this thing with my changes, it looks as necessary

Screen Shot 2021-03-17 at 4 15 23 PM

packages/css/package.json
```diff
@@ -36,7 +36,8 @@
     "copy:images": "cpx 'src/img/**' dist/img",
     "copy:fonts": "cpx 'src/fonts/*' dist/fonts",
     "prettier": "prettier --write src/less/*.less",
-    "lintless": "stylelint ./src/less",
+    "lint": "yarn lintless",
+    "lintless": "stylelint './src/(less|props|variables)/**' '!**/bootstrap-forms.less'",
```
dozortsev, line 40: here I excluded `bootstrap-forms.less`, I don't know why but stylelint throws some weird errors for this file although it looks legit, so I think it's find to exclude it

```shell
$ stylelint './src/(less|props|variables)/**'

src/less/forms/bootstrap-forms.less
 83:1  ✖  Cannot parse selector   parseError
 83:1  ✖  Cannot parse selector   parseError
 83:1  ✖  Cannot parse selector   parseError
 83:1  ✖  Cannot parse selector   parseError
 83:1  ✖  Cannot parse selector   parseError
 83:1  ✖  Cannot parse selector   parseError
 83:1  ✖  Cannot parse selector   parseError
```

packages/components/src/lab/mobileNav/MobileNav.less
```diff
@@ -167,7 +169,7 @@
   flex: 0 0 auto;
   display: inline-block;
   width: 100%;
-  font-family: Averta;
+  font-family: Arial, sans-serif;
```
anttio, line 172: Should this still be Averta instead of Arial?

packages/components/src/snackbar/Snackbar.less
anttio: Worth checking out that this doesn't break anything. What this does is `font-style: inherit` instead of the original `font: inherited inherited inherited ...`. So the shorthand sets `inherited` to all.

packages/css/src/less/core/_scaffolding.less
```diff
@@ -195,8 +195,8 @@ textarea {
 
 input,
 textarea {
-  &:input-placeholder,
-  &::input-placeholder,
+  // stylelint-disable selector-no-vendor-prefix
+  &:-ms-input-placeholder,
```
anttio, line 199: Would autoprefixer do this automatically?

packages/css/src/less/forms/bootstrap-forms.less
```diff
@@ -76,7 +77,7 @@ input[type='number']::-webkit-inner-spin-button {
 input[type='radio'],
 input[type='checkbox'] {
   margin: 4px 0 0;
-  margin-top: 1px \9; // IE8-9
+  margin-top: 1px;
```
anttio, line 80: This probably breaks things. The `\9` is a hack to target old IEs. Most likely this line can be removed.

packages/components/src/lab/mobileNav/MobileNav.less
dozortsev: yeah indeed, I reverted it back and disabled the [rule](https://stylelint.io/user-guide/rules/font-family-no-missing-generic-family-keyword) for this like

packages/css/src/less/core/_scaffolding.less
dozortsev: I don't really know whether we run these less files through autoprefixer, maybe @andreapiras or @helendurranttw?

andreapiras: you need to check the css generated file and see if the prefixed values have been generated.
Having the ms- prefix here seems that we are targeting a smaller set

packages/css/src/less/forms/bootstrap-forms.less
```diff
@@ -192,6 +195,23 @@ output {
   textarea& {
     height: 144px;
   }
+
+  + .text-muted,
```
andreapiras, line 199: Where is this coming from?

packages/css/src/less/core/_scaffolding.less
dozortsev: ignore that, I added that prefix by mistake

PR #854, opened 2021-03-16
fix!: default behaviour of DateLookup component

joe-sayer:
## 🖼 Context

Ensures that DateLookup behaviour is functional when only required props are passed.

## 🚀 Changes

 
- BREAKING CHANGE: Make value prop required

## 🤔 Considerations

- Although this is a breaking change, there are probably no use cases where DateLookup is used without providing a value prop

## ✅ Checklist


helendurranttw:
Please check codesearch to make sure of your hypothesis. I'll tag this for release with the next major just to be safe.

packages/components/src/dateLookup/DateLookup.js
```diff
@@ -303,7 +303,7 @@ class DateLookup extends PureComponent {
 }
 
 DateLookup.propTypes = {
-  value: PropTypes.instanceOf(Date),
+  value: PropTypes.instanceOf(Date).isRequired,
```
dozortsev, line 306: did you figure out why it doesn't show date in this case?

https://user-images.githubusercontent.com/4015628/121344612-ddf43080-c91b-11eb-8e89-2b43a442546b.mov

perhaps we could just fix that inside component without changing it's API

dozortsev:
Will be done at https://github.com/transferwise/neptune-web/pull/1065

PR #863, opened 2021-03-18
feat: add ariaContent prop to Popover

rerdos:
## 🖼 Context
[DS-978](https://transferwise.atlassian.net/browse/DS-978)

In our use-case, when passing a focusable content to the component, it causes assistive technologies to focus it twice. Also the contents are very visual, so plainly reading them would not work well.

## 🚀 Changes

- Provide an ariaContent prop as a descriptor in case Popover contains an illustration, or any other content that would otherwise require further description for screen readers.

## 🤔 Considerations

- Tried with `aria-content`, but eslint complains `content` is not a valid aria attribute.

Kitten

## ✅ Checklist

### DIFF
--- a/packages/components/src/popover/Popover.js
+++ b/packages/components/src/popover/Popover.js
@@ -8,7 +8,7 @@ import { logActionRequiredIf } from '../utilities';
 
 import './Popover.css';
 
-const Popover = ({ children, className, content, preferredPlacement, title }) => {
+const Popover = ({ children, className, ariaContent, content, preferredPlacement, title }) => {
   logActionRequired({ preferredPlacement });
   const anchorRef = useRef(null);
   const [open, setOpen] = useState(false);
@@ -29,7 +29,7 @@ const Popover = ({ children, className, content, preferredPlacement, title }) =>
         {open && (
           <span role="status" className="sr-only">
             {title}
-            {content}
+            {ariaContent || content}
           </span>
         )}
       </span>
@@ -70,6 +70,7 @@ Popover.Placement = {
 
 Popover.defaultProps = {
   className: undefined,
+  ariaContent: undefined,
   preferredPlacement: Popover.Placement.RIGHT,
   title: undefined,
 };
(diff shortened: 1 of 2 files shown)


packages/components/src/popover/Popover.js
```diff
@@ -8,7 +8,7 @@ import { logActionRequiredIf } from '../utilities';
 
 import './Popover.css';
 
-const Popover = ({ children, className, content, preferredPlacement, title }) => {
+const Popover = ({ children, className, ariaContent, content, preferredPlacement, title }) => {
```
dozortsev, line 11: does such attribute actually exist? (I can't find it)

if not then I wouldn't use `aria` prefix here, as it's reserved for a11y attributes

rerdos: I tried with `aria-content`, but eslint complains content is not a valid aria attribute, that's why camel cased it.

I'm happy to call it something else - do you have any suggestion?

dozortsev: well, perhaps I would prefer to disable rule for the line

packages/components/src/popover/Popover.js
```diff
@@ -29,7 +29,7 @@ const Popover = ({ children, className, content, preferredPlacement, title }) =>
         {open && (
           <span role="status" className="sr-only">
             {title}
-            {content}
+            {ariaContent || content}
```
dozortsev, line 32: I don't quite understand how this fix would solve the issue? could you please maybe show some screenshot?

dozortsev: @andreapiras is this issue because of popper? if so then I think we should address this at lower level component (e.g Panel) rather than only fix it for Popover

andreapiras: @dozortsev this is allowing people to take control of the aria content read by the screen reader. If nothing is passed by default the content gets read otherwise the provided ariaContent gets read

andreapiras:
https://github.com/transferwise/neptune-web/pull/904

PR #873, opened 2021-03-29
Export Tile component 

gittymcgit:
## 🖼 Context
We'd like to use Tiles, so I'm getting them exported from components
Here it is https://transferwise.github.io/neptune-web/branch/export-tile-component/components/options/Tile 
image

## 🚀 Changes
Exported Tile component and added a docs page for it
 

## 🤔 Considerations
I had to add `display: block` to `.np-tile` because otherwise it only looks alright when parent component is flex (due to being a link)

## ✅ Checklist


packages/docs/pages/components/options/Tile.mdx
```diff
@@ -0,0 +1,12 @@
...
+import { Tile } from '@transferwise/components';
+import code from '../../../liveEditorCode/tile.code';
+
+<LiveEditorBlock code={code} scope={{ Tile }} />
+<GeneratePropsTable componentName="Tile" />
+
+
+export const meta = {
+  name: 'Tile',
+};
```
dozortsev, line 11: if we decide to export Tile then could you please write this file as TypeScript file

example: https://github.com/transferwise/neptune-web/blob/main/packages/docs/pages/components/context/Provider.tsx

PR #875, opened 2021-03-31
feat: support grid use-case in Decision

gittymcgit:
## 🖼 Context
[NSP-1](https://transferwise.atlassian.net/browse/NSP-1)

image

Adding a grid support as we agreed
## 🚀 Changes
For grid case, I added flex-wrap class on the container. Also, had to change margins from being on the left side to be right and bottom, specifically for this case. Didn't change `LIST_BLOCK` presentation. 
Fixed the width of the tile on 300px for `md`, and 260px on `sm`

## 🤔 Considerations
I had to do `!important` on `margin-bottom` since `.decision` has `margin-bottom: 0 !important` that I'm not confident enough to remove.  I could do margin top instead of bottom, but then I'd have to do negative margin on the container to even it out.

## ✅ Checklist


packages/components/src/decision/Decision.less
```diff
@@ -3,7 +3,17 @@
     margin-left: var(--size-16);
   }
 
+  &.np-decision--grid &__tile--small {
+    margin: 0 var(--size-16) var(--size-16) 0;
+    margin-bottom: var(--size-16) !important;
```
gittymcgit, line 8: https://github.com/transferwise/neptune-web/blob/main/packages/css/src/less/decision.less#L7 this is the culprit

packages/docs/pages/components/navigation/Decision.tsx
```diff
@@ -0,0 +1,33 @@
+import React, { ReactElement } from 'react';
+import { Decision, Avatar } from '@transferwise/components';
+import { LiveEditorBlock, GeneratePropsTable } from '../../../utils';
+
+// @ts-ignore
+import code from '../../../liveEditorCode/decision.code';
+
+export default function PageContent(): ReactElement {
```
dozortsev, line 8: 😍 ❤️ 🙇

helendurranttw:
Hey @gittymcgit, thanks for these changes! There's a small issue with the margin when the items wrap on to the next line - see it in the attached screenshot.

Not sure off a surefire fix off the top of my head, but you could try either a negative margin on the wrapping element or try using grids + grip-gap.

Screenshot 2021-04-01 at 13 10 45

helendurranttw (approved):
Thanks @gittymcgit !

PR #876, opened 2021-03-31
feat: Promoted oneOf checkbox

felixxiatransferwise:
## 🚀 Changes

 
- Adds promoted oneOf controls
- Adds checkbox control for promoted oneOfs

## ✅ Checklist


packages/dynamic-flows/src/jsonSchemaForm/promotedOneOfSchema/control/PromotedOneOfCheckboxControl.js
```diff
@@ -0,0 +1,56 @@
...
+      setSelection(selectionWhenUnchecked);
+    }
+  }, [checked]);
+
+  return (
+    <div className="form-group">
+      <Checkbox label={title} onChange={setChecked} checked={checked} />
+    </div>
+  );
+};
```
JulianG-TW, line 28: I think we can avoid keeping local state for `checked` and instead derive its value from `selection === selectionWhenChecked`.
This would remove the need for `useEffect`.

```js
  const checked = selection === selectionWhenChecked;
  
  const toggleSelection () => 
    setSelection(checked ? selectionWhenUnchecked : selectionWhenChecked);
  
  return (
    <div className="form-group">
      <Checkbox label={title} onChange={toggleSelection} checked={checked} />
    </div>
  );
};
```
I haven't tried this. Probably not a good idea to merge from GitHub UI.

JulianG-TW (approved):

PR #886, opened 2021-04-08
chore: exclude 'bootstrap-forms.less' from lint-staged too

dozortsev:
## 🖼 Context

https://github.com/transferwise/neptune-web/pull/853/files#r596206435

## 🚀 Changes

Exclude `bootstrap-forms.less` from verifying at `git commit` hook (via `lint-staged`).


helendurranttw:
I understand why we need this but I don't like it. I'd prefer to disable stylelint on the file. I think in any case we should probably just refactor the file to fix the stylelint errors. I'm going to approve this as a temporary fix and let's create an issue to fix the file :)

helendurranttw (approved):

PR #887, opened 2021-04-08
feat: Add dynamic flow component

StevePole:
## 🖼 Context

We need to implement a web client for V3 dynamic flows

This PR includes:
- Loading of and transitioning between dynamic forms
- Dynamic refresh after user action
- Loading state
- Error boundary

Considerations:
- `action`s need a property in the JSON that determines whether validation is required to execute action (solved for now: validation only required on action of type POST, not GET etc.)

## 🚀 Changes

This introduces the dynamic flow component that utilises existing dynamic layout and dynamic form components.  We also introduce a 'stepService' for handling the transition between steps of a flow.

## ✅ Checklist


packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -0,0 +1,161 @@
...
+import { convertStepToLayout, inlineFormSchemas } from './layoutService';
+
+import { request } from './stepService';
+
+/**
+ * ## DynamicFlow
+ *
+ * The dynamic flow component is responsible for the asynchronous actions occuring
+ * within a dynamic flow, managing transitions between steps as well as refreshing
+ * requirements.  It does control any view logic, but takes the step definition
```
brettyukich, line 14: does -> doesn't

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -0,0 +1,161 @@
...
+      console.log('catch', error); // eslint-disable-line
+      setValidations(error.validation); // TODO this not thrown
+    }
+  };
+
+  const onModelChange = async (newModel, isValid, triggerSchema) => {
+    setModel(newModel);
+    setModelIsValid(isValid);
+
+    if (triggerSchema.refreshRequirements) {
```
diego-transferwise, line 67: This field is called [refreshRequirementsOnChange](https://github.com/transferwise/neptune-web/blob/387cad79018f4294fd9722eb613bce9e8cf41ac1/packages/docs/liveEditorCode/dynamicFieldDefinitionList.code.js#L21)

diego-transferwise, line 67: This field is called [refreshRequirementsOnChange](https://github.com/transferwise/neptune-web/blob/387cad79018f4294fd9722eb613bce9e8cf41ac1/packages/docs/liveEditorCode/dynamicFieldDefinitionList.code.js#L21)

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -0,0 +1,202 @@
...
+  const fetchRefresh = (action, data) => {
+    return request(action, data, baseUrl)
+      .then(validateResponse)
+      .then((response) => {
+        setStepSpecification(response);
+      })
+      .catch(handleFetchError);
+  };
+
+  const validateResponse = (response) => new Promise((resolve) => resolve(response));
```
StevePole, line 69: What's the purpose of this?

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -0,0 +1,202 @@
...
+      const action = { url: stepSpecification.refreshFormUrl, method: 'POST' };
+
+      fetchRefresh(action, combineModels(updatedModels));
+    }
+  };
+
+  const onAction = async (action) => {
+    const { exit, data, method } = action;
+
+    if (exit) {
```
StevePole, line 96: Eddie is just iterating on this specification, but I believe the intent is now that we exit _after_ any actions are performed.  So if there's a submission, which should wait for a successful response before exiting.

StevePole: https://transferwise.atlassian.net/wiki/spaces/DF/pages/1946032207/Flow+Completion+Revisited?atlOrigin=eyJpIjoiN2ZmNjM1OTA0NmZkNDdjYWE0ODEwMTE0N2M1MGEzNTgiLCJwIjoiY29uZmx1ZW5jZS1jaGF0cy1pbnQifQ

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -0,0 +1,202 @@
...
+      ...combineModels(models),
+      ...data,
+    };
+
+    if (isSubmissionMethod(method)) {
+      setSubmitted(true);
+
+      if (modelIsValid) {
+        fetchStep(action, submissionData).finally(() => {
+          setSubmitted(false);
```
StevePole, line 111: I haven't been writing code recently so you'll have to remind me 😂 .  Does this run in the case of failure?  Because if the response fails, we should leave submitted as true.  It's only when we get a successful response that we move to the next step of the flow, where submitted should be false.

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -0,0 +1,202 @@
...
+        });
+      }
+      return;
+    }
+
+    fetchStep(action);
+  };
+
+  const onPersistAsync = () => {};
```
StevePole, line 120: I can help with this, it's not all that complex.

packages/dynamic-flows/src/flow/errorBoundary/index.js
```diff
@@ -0,0 +1,74 @@
...
+
+      return (
+        <>
+          {hasError && (
+            <Alert
+              action={{
+                text: 'Retry',
+                href: window.location.href,
+              }}
+              message="Oops. Something went wrong..."
```
StevePole, line 44: Would really like to change this.  I've asked (again) for some copy guidance https://transferwise.slack.com/archives/C2XG7R3EZ/p1621963665053000

Whatever it is, can we translate it?

StevePole (commented):
Few points/questions.  In general looks good 👍

packages/dynamic-flows/src/flow/DynamicFlow.js
richardbata: Right now it has no purpose, its for the future if validations needed. Let me clean up for now then

packages/dynamic-flows/src/flow/DynamicFlow.js
richardbata: Correct, we should only change submitted to false if submission is successful, let me change it

packages/dynamic-flows/src/flow/DynamicFlow.js
richardbata: Cool, I can remove the handler

packages/dynamic-flows/src/flow/errorBoundary/index.js
richardbata: Should we change this once we get the copy?

packages/dynamic-flows/src/flow/DynamicFlow.js
richardbata: @eddiewoodleytw did we take a decision on which option we're going with? How should the Web implementation behave for now?

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -0,0 +1,202 @@
...
+  useEffect(() => {
+    if ((stepSpecification || {}).model) {
+      setModels({ [INITIALIZATION_SCHEMA_PROPERTY]: stepSpecification.model });
+    }
+  }, [stepSpecification]);
+
+  const fetchStep = async (action, data) => {
+    setLoading(true);
+
+    return request(action, data, baseUrl)
```
arikanmstf-tw, line 47: how can we add oauthToken and other headers here? I think we'll need them

packages/dynamic-flows/src/flow/stepService/RestStepService.js
```diff
@@ -0,0 +1,22 @@
+async function request(action, data, baseUrl) {
+  const { url, method } = action;
+
+  return fetch(baseUrl + url, {
+    method,
+    headers: {
+      'Content-Type': 'application/json',
```
arikanmstf-tw, line 7: maybe a way to have additional headers here? related to my other comment above

packages/dynamic-flows/src/flow/stepService/RestStepService.js
```diff
@@ -0,0 +1,22 @@
+async function request(action, data, baseUrl) {
```
arikanmstf-tw, line 1: and here, I actually always prefer having one parameter only, ( `request({ action, data, baseUrl })` ) , so If you want to add/remove some in the future, you wont break the existing.

packages/dynamic-flows/src/flow/DynamicFlow.js
richardbata: For authentication we could take the tokens from the window within the request function. For extra headers, we can add a parameter

richardbata: I'll try to set this up on custom env to see what authentication we need exactly. Do you think we can add the authentications in a separate PR?

packages/dynamic-flows/src/flow/errorBoundary/index.js
richardbata: I added support for translations once we get the copy

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -0,0 +1,242 @@
...
+      .finally(() => {
+        setLoading(false);
+      });
+  };
+
+  const fetchRefresh = (action, data) => {
+    return request({ action, data, baseUrl })
+      .then(getJson)
+      .then((json) => {
+        setStepSpecification(json);
```
StevePole, line 70: You could simplify, feel free to ignore though!

packages/dynamic-flows/src/flow/DynamicFlow.spec.js
```diff
@@ -0,0 +1,624 @@
...
+          data: { a: 2, c: true },
+          baseUrl,
+        });
+      });
+    });
+
+    // No such thing at the moment
+    // describe('when we recieve a flow termination', () => {
+    //   beforeEach(() => {
+    //     stepResolve({ exit: true }); // TODO what is the format for exit?
```
StevePole, line 511: Could we now enable this test?

packages/dynamic-flows/src/layout/DynamicLayout.js
```diff
@@ -103,12 +103,13 @@ DynamicLayout.propTypes = {
   baseUrl: Types.string.isRequired,
   onAction: Types.func.isRequired,
   onModelChange: Types.func.isRequired,
-  onPersistAsync: Types.func.isRequired,
+  onPersistAsync: Types.func,
```
StevePole, line 106: I think we can remove this?  It's handled internally now, I don't think we pass it down or do anything with it?

StevePole (commented):
Couple little comments, but very minor.  Happy to approve if you convert out of draft.

StevePole:
Correction, it is out of draft, but I can't approve because I opened it.

packages/dynamic-flows/src/flow/DynamicFlow.js
richardbata: Could combine `then`s, but duties are separated this way I guess

packages/dynamic-flows/src/layout/DynamicLayout.js
richardbata: I would remove this field from all the components in a separate PR

packages/dynamic-flows/i18n/index.js
```diff
@@ -1,4 +1,31 @@
...
+import pl from './pl.json';
+import pt from './pt.json';
+import ro from './ro.json';
+import ru from './ru.json';
+import tr from './tr.json';
+import zh from './zh.json';
 
-export default { en, ja };
+export default {
+  en,
```
diego-transferwise, line 17: What about the other languages?

packages/dynamic-flows/src/common/api/utils.js
```diff
@@ -6,6 +6,10 @@ export function isStatus422(status) {
   return status === 422;
 }
 
+export function getJson(response) {
+  return response.json();
```
diego-transferwise, line 10: What's the point of having this function?

richardbata: Its a common logic used for all fetch requests. Of course its possible to call `.json()` from the following `then`s also

packages/dynamic-flows/src/layout/DynamicLayout.js
felixxiatransferwise: doesn't the client still need to know about the internal persist async status so it can better make decisions about blocking/delaying submissions?

packages/dynamic-flows/src/flow/errorBoundary/ErrorBoundary.js
```diff
@@ -0,0 +1,78 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { Alert, Sentiment } from '@transferwise/components';
+import { injectIntl } from 'react-intl';
+
+import messages from './ErrorBoundary.messages';
+
+const withErrorBoundary = (WrappedComponent) => {
```
felixxiatransferwise, line 8: Why still use HOC for this? isn't it better to use a context hook now?

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -0,0 +1,235 @@
...
+ */
+const DynamicFlow = (props) => {
+  const { baseUrl, flowUrl, onClose, onStepChange, locale, onError } = props;
+
+  const [stepSpecification, setStepSpecification] = useState({});
+  const [models, setModels] = useState({});
+  const [modelIsValid, setModelIsValid] = useState(true);
+  const [loading, setLoading] = useState(true);
+  const [submitted, setSubmitted] = useState(false);
+  const [validations, setValidations] = useState();
```
JulianG-TW, line 30: In recipient-form we were keeping separate "states", and had race conditions which made us refactor into a single state that would be updated in one go. https://github.com/transferwise/recipient-form/pull/18

I'd like to try and reproduce that race condition here, but storybook is currently broken for me. I get the `Oops. Something went wrong...` alert.

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -0,0 +1,243 @@
...
+      resolve(json);
+    });
+
+  const onModelChange = (newModel, formSchema, triggerModel, triggerSchema) => {
+    const { $id } = formSchema;
+
+    const updatedModels = updateModels($id, newModel);
+
+    if ((triggerSchema || {}).refreshRequirementsOnChange) {
+      const action = { url: stepSpecification.refreshFormUrl, method: 'POST' };
```
JulianG-TW, line 117: I think we should check if `triggerSchema.refreshFormUrl` is defined first and only fall back to `stepSpecification.refreshFormUrl`.

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -0,0 +1,243 @@
...
+
+      fetchRefresh(action, combineModels(updatedModels));
```
JulianG-TW, line 119: This may benefit from debouncing for text input "schemas". we don't want to refresh the step on every keystroke as users type in their email address, for instance.

We should also use abort signal, to abort previous requests. [Here's a custom useAbortController hook](https://github.com/transferwise/recipient-form/blob/4f5f8fd9eed62024bf2984213e2ce79d26a549aa/src/dfv3-form/useJsonSchemaForm/useAbortController.ts) and [here's how it can be used](https://github.com/transferwise/recipient-form/blob/4f5f8fd9eed62024bf2984213e2ce79d26a549aa/src/dfv3-form/useJsonSchemaForm/useJsonSchemaFormState.tsx#L75-L78).

JulianG-TW (approved):
Looks good. I left a few comments for future improvements based on our experience with recipient creation in dfv3.

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -0,0 +1,245 @@
...
+
+  const INITIALIZATION_SCHEMA_PROPERTY = 'initial';
+
+  useEffect(() => {
+    const action = { url: flowUrl, method: 'GET' };
+    fetchStep(action);
+  }, [baseUrl, flowUrl]);
+
+  useEffect(() => {
+    if ((stepSpecification || {}).model) {
```
helendurranttw, line 39: If you find yourself doing this a lot you could add the `@babel/plugin-proposal-optional-chaining` plugin to your babel config. We did this in components. It allows you to safely rewrite that to `if (stepSpecification?.model)`.

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -0,0 +1,245 @@
...
+
+  const checkForExitCondition = (response) =>
+    new Promise((resolve) => {
+      const { data, headers } = response;
+
+      const exitHeader = 'X-DF-Exit';
+
+      if (headers && headers.get(exitHeader)) {
+        onClose(data);
+        return;
```
helendurranttw, line 97: Should you be resolving here?

packages/dynamic-flows/src/flow/errorBoundary/ErrorBoundary.js
```diff
@@ -0,0 +1,78 @@
...
+    render() {
+      const { intl } = this.props;
+      const { hasError, isFatalError } = this.state;
+
+      return (
+        <>
+          {hasError && (
+            <Alert
+              action={{
+                text: 'Retry',
```
helendurranttw, line 45: This needs to be localised.

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -0,0 +1,245 @@
...
+// eslint-disable-next-line
+DynamicFlow.propTypes = {
+  baseUrl: Types.string,
+  flowUrl: Types.string,
+  httpClient: Types.shape({
+    request: Types.func,
+  }),
+  onClose: Types.func.isRequired,
+  onStepChange: Types.func,
+  locale: Types.string,
```
helendurranttw, line 232: So, about locale. To use Neptune components correctly (e.g. with translations and direction support), the consuming app needs to wrap all instances of components in a [Provider](https://transferwise.github.io/neptune-web/components/context/Provider).

Since DynamicFlow uses components, it itself will need to be wrapped in the Provider. The assumption then is that the consumer will pass `locale` to the Provider, and in this component you can consume it using `react-intl` as normal, via:

```
import { useIntl } from 'react-intl';
const { locale } = useIntl();
```

And delete this locale prop.

We haven't yet set up Neptune to fail gracefully if Provider isn't set, though it's something we're looking into. This isn't likely to be a scenario we encounter for our apps, though we should think about how to handle this better for our 3rd party consumers. I think for the purposes of this PR I think it's fine to presume Provider will be set and you can safely retrieve the locale using the above.

packages/dynamic-flows/src/flow/DynamicFlow.story.js
```diff
@@ -0,0 +1,44 @@
...
+import { mockHttpClient } from './httpClient';
+
+export default {
+  component: DynamicFlow,
+  title: 'DynamicFlow',
+};
+
+export const basic = () => {
+  const steps = {
+    decision: '/decision',
```
helendurranttw, line 15: Great to have a rich example in Storybook. I'm getting an error when I try and click on Choose an existing recipient. It looks like DynamicReview is expecting the component prop to have a fields prop, but it's getting a definitions prop instead. Possibly either an issue with LayoutService.convertStepReviewToDynamicReview or the mapping in DynamicReview?

(I realise this may be an existing issue unrelated to your changes!)

packages/dynamic-flows/src/flow/errorBoundary/ErrorBoundary.messages.js
```diff
@@ -0,0 +1,9 @@
+import { defineMessages } from 'react-intl';
+
+export default defineMessages({
+  errorAlert: {
+    id: 'dynamicFlows.ErrorBoundary.errorAlert',
+    defaultMessage: 'Oops. Something went wrong...',
+    description: 'Description of Dynamic flows error alert',
```
helendurranttw, line 7: I think these descriptions are used to help the translators know what to write - so maybe something like "Generic error message for when something has gone wrong" would be more helpful 😊

packages/dynamic-flows/src/flow/DynamicFlow.js
StevePole: No worries, it changed!  Before it was 
`.then((json) => { setStepSpecification(json)) }`
which could have been
`.then(setStepSpecification)`

packages/dynamic-flows/src/layout/DynamicLayout.js
StevePole: Maybe 🤔  I'm not sure this is actually plumbed in at the moment though.  I don't know that we have the callback available at layout level to trigger it.  I wouldn't worry for this PR, we can add later if needed.

packages/dynamic-flows/src/flow/DynamicFlow.story.js
StevePole: This was a bug, I created the original spec, but the field name got changed after my implementation.  Someone just fixed the bug, maybe we still have an out of date example JSON?

packages/dynamic-flows/src/flow/DynamicFlow.js
StevePole: I probably wrote this before I knew about that, I expect this prop is unused and we can remove it?

For 3rd parties I don't expect them to use these components, because it's unlikely they're going to buy into our design system wholesale.  I would see us either creating a pluggable architecture where we default to native controls, but let you pass in react replacements.  Or we'll create another example package which is a fork of this with the custom controls removed.

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.js
```diff
@@ -147,6 +147,13 @@ const OneOfSchema = (props) => {
     setId(generateId());
   }, [props.schema]);
 
+  useEffect(() => {
+    const modelIndex = getActiveSchemaIndex(props.schema, props.model);
+    if (modelIndex !== schemaIndex) {
+      onChooseNewSchema(getActiveSchemaIndex(props.schema, props.model));
```
helendurranttw, line 153:
```suggestion
      onChooseNewSchema(modelIndex);
```

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -0,0 +1,245 @@
...
+
+    return newModels;
+  };
+
+  const combineModels = (formModels) =>
+    Object.values(formModels).reduce((prev, model) => ({ ...prev, ...model }), {});
+
+  const areModelsValid = (formModels, schemas) =>
+    schemas.reduce(
+      (validSoFar, schema) => isValidSchema(formModels[schema.$id] || {}, schema) && validSoFar,
```
helendurranttw, line 180: Do you need to default `formModels[schema.$id]` to `{}`? I'd imagine undefined would just return false? Worth checking though.

One other small thing, you could use `some` instead of `reduce`, letting you bail out of the loop if you encounter an invalid model :)

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -0,0 +1,245 @@
...
+    const layout = step.layout ? step.layout : convertStepToLayout(step);
+
+    return inlineReferences(layout, step.schemas, step.actions);
+  };
+
+  const components = getComponents(stepSpecification);
+
+  return (
+    <div className="m-a-3 m-t-5">
+      {!components && <p>No layout</p>}
```
helendurranttw, line 203: Is this necessary? Maybe we can just render nothing if there is nothing to render. You could move the components && to the outside div.

packages/dynamic-flows/src/flow/errorBoundary/ErrorBoundary.messages.js
richardbata: Oh nice, I didn't realise its seen by the translators

packages/dynamic-flows/src/flow/DynamicFlow.story.js
richardbata: I adjusted the Storybook not to show errors.

packages/dynamic-flows/src/flow/DynamicFlow.js
richardbata: Resolving throws exceptions later in the promise chain, and actually we'd expect component to be cleaned up after onClose, hence its less noisy to leave the promise pending and wait until the DynamicFlow is cleaned up

packages/dynamic-flows/src/flow/DynamicFlow.js
richardbata: The defaulting is required as is not with the helper functions

packages/dynamic-flows/src/flow/DynamicFlow.spec.js
```diff
@@ -0,0 +1,683 @@
...
+
+  const navigateAction = {
+    label: 'Skip',
+    method: 'GET',
+    url: '/next',
+  };
+
+  const actions = [successAction, navigateAction, failureAction];
+
+  // const reviewFields = {
```
helendurranttw, line 43: Guess this can go now :)

packages/dynamic-flows/src/flow/DynamicFlow.spec.js
```diff
@@ -0,0 +1,683 @@
...
+      title: 'We create the thing!',
+      description: 'You now do stuff with the thing',
+      actionTitle: 'Continue',
+    },
+    result: {
+      exitValue: 'value',
+    },
+  };
+
+  const formUrl = '/form'; // eslint-disable-line
```
helendurranttw, line 138: Can't see why you'd need these disables 🤔

packages/dynamic-flows/src/flow/DynamicFlow.spec.js
```diff
@@ -0,0 +1,683 @@
...
+    it('should load the step specification using the supplied URL', () => {
+      expect(mockClient.request).toHaveBeenCalledWith({
+        action: { url: decisionUrl, method: 'GET' },
+      });
+    });
+  });
+
+  describe('when there is JS step', () => {
+    it('should use the DynamicJS component', () => {
+      // TODO
```
helendurranttw, line 254: 🙃

helendurranttw (approved):
Looks good to me - there are a few minor issues in the spec to fix up, but apart from that looks good!

PR #888, opened 2021-04-08
feat: include action references in layout service

diego-transferwise:
## 🚀 Changes
Now layout service can inline the layout of action references

## ✅ Checklist


packages/dynamic-flows/src/flow/layoutService/LayoutService.js
```diff
@@ -217,23 +223,33 @@ function inlineFormSchema(formComponent, schemas, model) {
...
-    left: inlineFormSchemas(columnsComponent.left, schemas, model),
-    right: inlineFormSchemas(columnsComponent.right, schemas, model),
+    left: inlineReferences(columnsComponent.left, schemas, actions, model),
+    right: inlineReferences(columnsComponent.right, schemas, actions, model),
   };
 }
 
-function getSchemaById(schemas, id) {
-  return schemas.find((schema) => schema.$id === id);
+function getById(references, id) {
```
StevePole, line 251: Not sure these are 'references' they are the thing being referenced...  I actually wonder if there should be a getSchemaById and a getActionById.  I appreciate the implementation is the same, but they are different things, and could diverge in future.

packages/dynamic-flows/src/flow/layoutService/LayoutService.spec.js
```diff
@@ -247,6 +247,27 @@ describe('Given a utility service for handling dynamic layouts', () => {
...
     ];
 
+    const actions = [
+        {
+          $id: "#submitMyDetails",
+          title: "Continue",
+          url: "/v3",
+          type: "primary",
+          method: "POST",
+          skipValidation: false,
```
StevePole, line 257: Is this a thing in V3?  I don't think it is...

StevePole (commented):
Couple small questions

packages/dynamic-flows/src/flow/layoutService/LayoutService.spec.js
eddiewoodleytw: I don't think it is either, definitely not implemented on iOS client.

diego-transferwise: I got it from the backend library, how outdate is it? https://github.com/transferwise/dynamic-interfaces/blob/master/v3/src/main/java/com/transferwise/df/v3/steps/internal/Action.java#L30

packages/dynamic-flows/src/flow/layoutService/LayoutService.js
diego-transferwise: It makes sense.

packages/dynamic-flows/src/flow/layoutService/LayoutService.spec.js
StevePole: In principle I think we should remove it.  The focus with JSON schema is on specifying what make a valid model, so it's a bit backwards to skip validating it.  We have 'GET' method actions, if we want to navigate without submitting a model and therefore without validation.

StevePole (approved):
LGTM 👍

PR #889, opened 2021-04-09
chore: add Provider component as global decorator to DF Storybook

dozortsev:
## 🚀 Changes

Move `Provider` component setup from `JsonSchemaForm` Strorybook to global config and define it as a decorator.

Screen Shot 2021-04-09 at 4 45 14 PM


packages/dynamic-flows/src/jsonSchemaForm/JsonSchemaForm.story.js
```diff
@@ -31,7 +21,6 @@ export const basic = () => {
   const schemas = {
     simple: simpleSchema,
     promotedOneOf: promotedOneOfSchema,
-    promotedOneOfCheckbox: promotedOneOfCheckboxSchema,
```
helendurranttw, line 34: Why did you remove this?

dozortsev: fixed https://github.com/transferwise/neptune-web/pull/889/commits/1cd8c397c20399aa691ea659ee46503bdaf72e51

packages/dynamic-flows/.storybook/preview.js
```diff
@@ -45,4 +54,24 @@ const CenterDecorator = (storyFn) => (
   </div>
 );
 
-export const decorators = [StrictModeDecorator, CenterDecorator];
+// list is not exhaustive but should enough for testing diff edge cases
+// feel free to add more during development
+const severalExamplesOfSupportedLocales = [DEFAULT_LOCALE, 'ja-JP'];
```
StevePole, line 59: Could we include the other ones we had before?  Some were chosen as they had some outlier behaviour

    'en-US',
    'ja-JP',
    'zh-HK',
    'es',
    'fr',
    'ru',
    'de',
    'tr',

packages/dynamic-flows/.storybook/preview.js
```diff
@@ -45,4 +54,24 @@ const CenterDecorator = (storyFn) => (
...
+
+const ProviderDecorator = (storyFn) => {
+  const locale = select('locale (global)', severalExamplesOfSupportedLocales, DEFAULT_LOCALE);
+  const lang = getLangFromLocale(locale);
+  // eslint-disable-next-line fp/no-mutating-assign
+  const supportedLangs = Object.assign(
```
StevePole, line 65: Spread?  I'm still paranoid about browsers that don't support Object.assign...

packages/dynamic-flows/.storybook/preview.js
dozortsev: I included only those two because atm DF supports only them https://github.com/transferwise/neptune-web/blob/main/packages/dynamic-flows/i18n/index.js

PR #891, opened 2021-04-13
feat: Make the DateLookup clearable

sghabi:
## 🖼 Context

We want to make the DateLookup clearable. We need this for scheduled transfers: we want to add the capability for customer to enter an and date. So they also need to remove it if needed.

## ✅ Checklist

## How it looks like (in `send-flow`):
sch-7-datecomponent


packages/components/src/dateLookup/openButton/OpenButton.js
```diff
@@ -6,9 +6,36 @@ import { formatDate } from '@transferwise/formatting';
...
+  placeholder,
+  label,
+  monthFormat,
+  disabled,
+  onClick,
+  isClearable,
+  onReset,
+}) => {
   const { locale } = useIntl();
+  const onResetCall = (event) => {
```
sghabi, line 24: remove this

packages/components/src/dateLookup/DateLookup.js
```diff
@@ -281,6 +281,10 @@ class DateLookup extends PureComponent {
         <OpenButton
           {...{ selectedDate, size, placeholder, label, monthFormat, disabled }}
           onClick={this.open}
+          isClearable={this.props.isClearable}
+          onReset={() => {
```
andreapiras, line 285: Hey thanks for contributing!
I would pass a different onClick if it's clearable

packages/components/src/dateLookup/DateLookup.js
```diff
@@ -281,6 +281,10 @@ class DateLookup extends PureComponent {
         <OpenButton
           {...{ selectedDate, size, placeholder, label, monthFormat, disabled }}
           onClick={this.open}
+          isClearable={this.props.isClearable}
```
andreapiras, line 284: Hey thanks for contributing!
What about something like this?
```
<OpenButton
          {...{ selectedDate, size, placeholder, label, monthFormat, disabled }}
          onClick={this.open}
          isClearable={this.props.isClearable}
        >
          <button
            className="clear-btn"
            onClick={(e) => {
              this.stopPropagation(e);
              this.props.onChange(null);
            }}
          >
            <CrossCircle filled size={24} />
          </button>
        </OpenButton>
```

and then inside OpenButton

{children || <Chevron orientation={Chevron.Orientation.BOTTOM} disabled={disabled} />}

packages/components/src/dateLookup/DateLookup.view.spec.js
```diff
@@ -58,6 +59,12 @@ describe('DateLookup view', () => {
     expect(component.find('.dropdown-menu')).toHaveLength(1);
   });
 
+  describe('when is clearable', () => {
+    it('passes onReset prop to open button', () => {
+      expect(openButton().prop('onReset')).not.toBe(null);
```
andreapiras, line 64: I'd test that when button is clicked onChange gets called with null

packages/components/src/dateLookup/openButton/OpenButton.js
```diff
@@ -6,9 +6,36 @@ import { formatDate } from '@transferwise/formatting';
...
+    stopPropagation();
+    onReset();
+  };
+  const stopPropagation = (event) => {
```
andreapiras, line 28: If you follow my suggestion this can be moved into parent

packages/components/src/dateLookup/openButton/OpenButton.js
```diff
@@ -6,9 +6,36 @@ import { formatDate } from '@transferwise/formatting';
...
+    if (!event) {
```
andreapiras, line 29: Is this necessary, I assume that since this func is called onClick event is always defined

packages/components/src/dateLookup/openButton/OpenButton.less
```diff
@@ -0,0 +1,13 @@
+@import (reference) '@transferwise/neptune-css/src/variables/neptune-tokens.less';
+
+.open-button {
```
andreapiras, line 3: this class is not in OpenButton

packages/components/src/dateLookup/openButton/OpenButton.js
```diff
@@ -6,9 +6,36 @@ import { formatDate } from '@transferwise/formatting';
...
+      return;
+    }
+    event.stopPropagation();
```
andreapiras, line 32: event.stopPropagation(); seems already to be doing the trick what do u need event.nativeEvent.stopImmediatePropagation(); for?

packages/components/src/dateLookup/openButton/OpenButton.js
sghabi: I had issue in tests complaining event was null. I think it's maybe because initially i used span instead of button. But removed it now

packages/components/src/dateLookup/openButton/OpenButton.js
sghabi: this is how it's done in `Typeahead` : https://github.com/transferwise/neptune-web/blob/2e2e30df618e30c96cee3ee995c1ec93136c11c1/packages/components/src/typeahead/Typeahead.js#L198
I thought maybe this handle more cases  (better for old browser maybe ?) but I actually don't know ^^

packages/components/src/dateLookup/DateLookup.js
```diff
@@ -312,6 +337,7 @@ DateLookup.propTypes = {
   onChange: PropTypes.func.isRequired,
   onFocus: PropTypes.func,
   onBlur: PropTypes.func,
+  isClearable: PropTypes.bool,
```
dozortsev, line 340: please rename this prop to `clearable`, we follow HTML name for boolean props (`disabled`, `hidden` etc)

packages/components/src/dateLookup/DateLookup.js
```diff
@@ -281,7 +291,22 @@ class DateLookup extends PureComponent {
...
+          {this.props.isClearable ? (
+            <button
+              onClick={(e) => {
+                this.stopPropagation(e);
+                this.props.onChange(null);
+              }}
+              className="clear-btn btn-sm btn-unstyled d-inline-block"
+              type="button"
+            >
+              <CrossCircle filled size={24} />
```
dozortsev, line 304: I think here you should use `size={16}` as Chevron's icon is also 16

packages/components/src/dateLookup/DateLookup.js
```diff
@@ -281,7 +291,22 @@ class DateLookup extends PureComponent {
...
+            </button>
```
dozortsev, line 305: wondering whether we should extract this into separate component - `ClearButton` 🤔  similar to `CloseButton`

it would encapsulate, [aria-label](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/common/closeButton/CloseButton.js#L23) with message "Clear", icon, maybe something else

dozortsev (commented):
please enable this feature at [docs example](https://github.com/transferwise/neptune-web/blob/main/packages/docs/liveEditorCode/dateLookup.code.js) to make it more discoverable

packages/components/src/dateLookup/DateLookup.js
andreapiras: Hey @dozortsev , I was in the process of doing this, but then I discuss it with @helendurranttw and we realize that it would have been probably more of an hassle than  a benefit. If we create this as separate component it should be a `button` but then we would not be able to use it inside `OpenButton` because you cannot put a button as descendant of another button. So we may have to change the root element in `OpenButton` but there are so many bootstrap classes involved that for now it is probably better to leave it alone

dozortsev: I see that now it's based on [`span`](https://github.com/transferwise/neptune-web/pull/891/files#diff-4dbbede956101d11e769358e05aace88f703f2a320434d84f070b1bf5fab26f9R53-R62) so it could be extracted

packages/components/src/dateLookup/openButton/OpenButton.spec.js
```diff
@@ -91,6 +98,18 @@ describe('OpenButton', () => {
     expect(button().hasClass('btn-lg')).toBe(true);
   });
 
-  const button = () => component.find('button');
+  it('shows chevorn button when onClear is not provided', () => {
```
sghabi, line 101: type: `chevron`

packages/components/src/dateLookup/DateLookup.story.js
```diff
@@ -26,24 +26,26 @@ export const basic = () => {
       max={new Date(maxvalue)}
       min={new Date(minvalue)}
       monthFormat={monthFormat}
-      onChange={(v) => action(v)}
```
sghabi, line 29: why did we have this in the first place ? I also change it locally to change to have a `setValue` (to test) but wasn't sure it was ok to change

packages/components/src/dateLookup/openButton/OpenButton.less
```diff
@@ -0,0 +1,18 @@
+.np-open-button {
+  .clear-btn {
+    transition: color 0.15s ease-in-out;
+    color: var(--color-control-accent);
+    height: 100%;
+    position: absolute;
+    top: 0;
```
sghabi, line 7: should this be defined based on the size of the encapsulating button ? Like I did here: https://github.com/transferwise/neptune-web/pull/891/commits/027f278285c90ac77e900f3c80488e27bd82c74d 
(not sure what are the best practices ^^)

andreapiras: Hey, I did use an 100% container and centered the element inside using utility classes.

packages/components/src/dateLookup/DateLookup.js
andreapiras: I think the ClearButton should be a button. We should change OpenButton instead but I think it's out of scope of this PR

packages/components/src/dateLookup/DateLookup.js
andreapiras: I'm using 16 for Small DateLookup and 24 for other sized cc:@sghabi

packages/components/src/dateLookup/DateLookup.story.js
andreapiras: It was an old setup, yeah it's ok to push this :)

packages/components/src/dateLookup/openButton/OpenButton.js
```diff
@@ -33,7 +47,20 @@ const OpenButton = ({ selectedDate, size, placeholder, label, monthFormat, disab
           {placeholder}
         </span>
       )}
-      <Chevron orientation={Chevron.Orientation.BOTTOM} disabled={disabled} />
+      {!onClear ? (
```
dozortsev, line 50: I would suggest have here same prop here in this component as in DateLookup - `clearable`, as having two on click event handlers `onClick` and `onClear` is a bit weird

packages/components/src/dateLookup/openButton/OpenButton.js
```diff
@@ -33,7 +47,20 @@ const OpenButton = ({ selectedDate, size, placeholder, label, monthFormat, disab
...
+        <Chevron orientation={Chevron.Orientation.BOTTOM} disabled={disabled} />
+      ) : (
+        <span
+          onClick={onClear}
+          onKeyDown={onClear}
+          role="button"
+          tabIndex={0}
+          className="clear-btn d-flex align-items-center"
```
dozortsev, line 58: I guess it should be grey color and disabled (not clickable) when DateLookup is disabled


https://user-images.githubusercontent.com/4015628/115037588-88287c80-9ec6-11eb-8c46-6e721b029859.mov

andreapiras: I don't think we should have the clear button at all when is disabled

packages/components/src/dateLookup/openButton/OpenButton.js
andreapiras: I know it is but they are in fact two distinct actions that exists at the same time. When clear button is there user can still select date by clicking on the label

gusso (approved):

packages/components/src/dateLookup/openButton/OpenButton.js
helendurranttw: I agree. It doesn't look like there are any tests for this case?

packages/components/src/dateLookup/DateLookup.testingLibrary.spec.js
```diff
@@ -171,4 +177,16 @@ describe('DateLookup (events)', () => {
...
     });
   });
+
+  describe('when is clearable', () => {
+    beforeEach(() => {
+      props = { value: date, onChange: jest.fn(), clearable: true };
+      ({ container } = render(<DateLookup {...props} />));
+    });
+
+    it('onChange gets called with null when reset button is clicked', () => {
```
helendurranttw, line 187: Could use some more tests here

packages/components/src/dateLookup/openButton/OpenButton.js
```diff
@@ -2,17 +2,47 @@ import React from 'react';
...
+import { isEventType } from '@transferwise/neptune-validation';
+
+import messages from './OpenButton.messages';
 
 import Chevron from '../../chevron';
 
 import { Size, MonthFormat } from '../../common';
+import './OpenButton.css';
+
+const OpenButton = ({
```
helendurranttw, line 15: Seems weird that this is called "OpenButton" now that it opens and clears 🤔

packages/docs/liveEditorCode/dateLookup.code.js
```diff
@@ -10,7 +10,8 @@
       label="label"
       monthFormat="long"
       disabled={false}
-      onChange={v => setDate(v)}
+      onChange={(v) => setDate(v)}
+      clearable={false}
```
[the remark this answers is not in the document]
andreapiras: I put clearable={false} to make people aware it does exists, but it should not be the default behavior

helendurranttw (approved):

PR #899, opened 2021-04-16
Fix animation in drawer

gittymcgit:
## 🖼 Context

I am experiencing an issue with Drawer's animation when rendering any component that uses SizeWrapper inside of the Drawer (e.g. Decision)
I think this is caused by reflow, caused by reading clinetWidth of the element by SIzeWrapper. 
https://transferwise.slack.com/archives/CJNBX9CP6/p1618578261041600 – video is in the thread
## 🚀 Changes
I wrapped this operation in requestAnimationFrame. It seems like the issue is resolved by that. 

## 🤔 Considerations
requestAnimationFrame is widely supported, but if we think that we shouldn't use it, timeout would do it  (https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame)

## ✅ Checklist


packages/components/src/common/hooks/useClientWidth/useClientWidth.js
```diff
@@ -28,7 +28,7 @@ export const useClientWidth = ({ ref, throttleMs = THROTTLE_MS }) => {
 
     window.addEventListener('resize', attachedFunction, true);
 
-    updateClientWidth();
+    window.requestAnimationFrame(updateClientWidth);
```
dozortsev, line 31: could you please add comment here explaining why we use `requestAnimationFrame` here? this would be quite useful for future cases when someone would debug/read this code

gittymcgit: Sure!

dozortsev (approved):
Reproduced the bug and tested the fix, look good!

Thanks for contributing 🙇

PR #902, opened 2021-04-20
feat: make Card expandable with keyboard

hendrikmolder:
## 🖼 Context

We use `Card` for the _main content_ in quite a few pages/apps – e.g. Direct Debits, account details etc. But some customers may not be able to see that content as toggling `Card` via keyboard has been impossible.

## 🚀 Changes
 

- Make `Card` expandable with keyboard
- Display focus styles only when focussing `Card` with keyboard

## 🤔 Considerations

I'm not sure why we only expanded card when we had children – I'd prefer to see empty card rather than nothing happening on click/keyDown event. Wondering if changing this behaviour would be good? Does anybody use `Card` without children?

**Edit:** Keeping the current behaviour – only improving accessibility in this PR.

## ✅ Checklist


packages/components/src/card/Card.js
```diff
@@ -20,20 +21,27 @@ const Card = ({
...
       className={classNames('tw-card list-group-item p-a-0', className, { active: isOpen })}
       id={id}
       data-testid={rest['data-testid']}
     >
-      {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events,jsx-a11y/interactive-supports-focus, jsx-a11y/no-static-element-interactions */}
       <div
         className={classNames('p-a-panel tw-card__panel', {
           'tw-card__panel--inactive': !children,
         })}
-        role={children ? 'button' : null}
```
hendrikmolder, line 35: See my question under _Considerations_ 😬

hendrikmolder: Maybe not worth changing tbh

packages/components/src/card/Card.less
```diff
@@ -32,6 +32,10 @@
         cursor: default;
       }
     }
+
+    &:focus:not(:focus-visible) {
+      outline: none;
+    }
```
hendrikmolder, line 38: Hides outline when element gets focussed using a mouse click...

Some browsers might still show outline _on click_ as well – https://caniuse.com/?search=focus-visible

packages/components/src/card/Card.js
oliverviljamaa: I don't know of an use case without `children` either, but if we want to remove the check, we should do it everywhere, e.g. also in `onClick`.

packages/components/src/card/Card.spec.js
```diff
@@ -107,6 +107,11 @@ describe('Card', () => {
       expect(defaultProps.onClick).toBeCalledWith(!defaultProps.isExpanded);
     });
 
+    it('calls onClick with inverse of current isExpanded value when toggle panel is toggled with keyboard', () => {
+      keyboardTogglePanel(true);
+      expect(defaultProps.onClick).toBeCalledWith(!defaultProps.isExpanded);
```
oliverviljamaa, line 112: Should probably be `onExpand` in the future, perhaps worth leaving a TODO?

packages/components/src/card/Card.js
hendrikmolder: Yeah, curious to heard DS input here, too.

But maybe not worth coupling with these tiny accessibility improvements 🤷  just didn't want to keep the ignore rules for eslint a11y..

packages/components/src/card/Card.less
NewmanJWise: An interesting alternative is putting a div inside the panel div, the div fills the entire of the inside of the parent div so the user only clicks on the child div with no outline whilst the keyboard only navigates to the parent div with the outline

packages/components/src/card/Card.js
```diff
@@ -20,20 +21,27 @@ const Card = ({
...
-        onClick={() => children && onClick(!isExpanded)}
+        role="button"
+        onClick={() => children && onClick(!isExpanded)} // TODO: Consider renaming to onExpand as Card can be expanded with keyboard
+        tabIndex={0}
+        onKeyDown={(event) => {
```
conorcussell, line 39: Do you think it should communicate its state to screen readers using `aria-expanded`? [Reference](https://inclusive-components.design/collapsible-sections/)

Also, how does the component work with screenreaders? It looks like the children are just hidden with CSS so I guess the whole contents of each card will be read out when tabbing through, rather than just the label which will be visible to non-screenreader users.

hendrikmolder: Good point! Haven't tested that yet, but will look into 🙇

packages/components/src/card/Card.js
```diff
@@ -20,20 +21,27 @@ const Card = ({
...
+          if (TOGGLE_KEYS.includes(event.key) && children) {
+            event.preventDefault();
+            onClick(!isExpanded);
```
dozortsev, line 42: wait, I think here we shouldn't invoke `onClick` handler, I would expose `onKeyDown` prop for this

packages/components/src/card/Card.js
dozortsev: I check original PR https://github.com/transferwise/neptune-web/pull/301 and couldn't find any discussion/decision re. this, perhaps would leave it for now

packages/components/src/card/Card.js
dozortsev, line 42: could you please extract this in separate function (e.g `onKeyDownHanlder`)

packages/components/src/card/Card.js
dozortsev: `children` rendered only when component is expanded https://github.com/transferwise/neptune-web/blob/main/packages/components/src/card/Card.js#L52-L60

packages/components/src/card/Card.js
oliverviljamaa: Why? It's nice and close to the similar `onClick` here 🤔

packages/components/src/card/Card.js
hendrikmolder: Alright, keeping the current behaviour then.

packages/components/src/card/Card.js
hendrikmolder: Added `aria-expanded`

packages/components/src/card/Card.js
andreapiras: I think this may be the reason 'tw-card__panel--inactive': !children,

packages/components/src/card/Card.js
dozortsev: well, it's a little bit big, make reading JSX a bit harder

dozortsev: but not strong opinion, I guess we can move forward with that

dozortsev (approved):

PR #904, opened 2021-04-21
chore: enhance Popover a11y

andreapiras:
## 🖼 Context

This Pr enhances a11y for Popover by focusing on the content when popover is open and by trapping the focus so screen readers won't be able to leave the area until the Popover is open

## ✅ Checklist


packages/components/src/common/responsivePanel/ResponsivePanel.js
```diff
@@ -50,14 +34,15 @@ const ResponsivePanel = ({ anchorRef, arrow, children, className, onClose, open,
...
+          onClose={onClose}
         >
           {children}
         </Panel>,
       ],
       breakpoint: Breakpoint.SMALL,
     },
   ];
-  return <SizeSwapper items={items} ref={windowRef} />;
+  return <SizeSwapper items={items} ref={windowRef} inline />;
```
andreapiras, line 45: This still works because the element are swiped based on the window size and not the parent container. If we were using the parent container as ref we won't be able to use inline as

packages/components/src/popover/Popover.js
```diff
@@ -42,7 +36,11 @@ const Popover = ({ children, className, content, preferredPlacement, title }) =>
         className="np-popover__container"
       >
         <div className="np-popover__content" aria-hidden={!open} role="tooltip">
-          {title && <div className="np-popover__title m-b-1">{title}</div>}
+          {title && (
+            <div role="heading" aria-level={1} className="np-popover__title m-b-1">
```
helendurranttw, line 40: I think there's only supposed to be one H1 per page, maybe this could be an H2?

helendurranttw: Or I guess since we're trapping focus it's kind of like a new surface.. maybe an H1 is ok.

helendurranttw (approved):
Nice work!

PR #905, opened 2021-04-22
chore (Chip): RTL support

joe-sayer:
## 🖼 Context

This PR is necessary in order for Chip to support RTL locales.

## 🚀 Changes

 

Used utility classes when `isRTL` is true and applied some custom styling.

## ✅ Checklist


packages/components/src/chip/Chip.js
```diff
@@ -4,12 +4,30 @@ import classNames from 'classnames';
...
+
   return (
-    <div className={classNames('chip', { 'has-error': hasError }, className)}>
+    <div
+      className={classNames(
+        'chip',
+        'd-inline-flex',
+        'align-items-center',
+        'm-t-1',
+
```
dozortsev, line 20: hasn't ESLint complain about this empty line?

packages/components/src/chip/Chip.js
```diff
@@ -4,12 +4,30 @@ import classNames from 'classnames';
...
+        { 'has-error': hasError, 'p-r-1 p-l-2': !isRTL, 'p-l-1 p-r-2': isRTL },
```
dozortsev, line 21: I guess we would write a lot of these `!isRTL` so I think it would make sense to tell `useDirection` hook to provide `isLTR`

generally I suggest you avoid using `!` it make code a bit harder to read/understand.

dozortsev (approved):

packages/components/src/chip/Chip.js
joe-sayer: No problems with ESLint

PR #910, opened 2021-04-23
feat(docs): global search

dozortsev:
## Demo

https://transferwise.github.io/neptune-web/branch/search/about/Home

search

## Changes

- Integrated [Algolia Search](https://docsearch.algolia.com) in docs as per their [React app example](https://codesandbox.io/s/docsearch-react-v3-playground-619yg).
- Use [Logo](https://transferwise.github.io/neptune-web/components/content/Logo) component instead of custom implementation at the docs.
- Delete custom search at components page.
- Update favicon to Neptune logo.

## Considerations

https://github.com/transferwise/neptune-web/pull/910#discussion_r773826542

## Checklist


packages/docs/components/Layout.js
```diff
@@ -87,7 +91,9 @@ const Layout = ({ children, router: { pathname } }) => {
     <div className="Content" role="main">
       {page ? (
         <>
-          <h1 className="colored-dot">{page.component.meta.name}</h1>
+          <h1 id={page.component.meta.name} className="colored-dot">
+            {page.component.meta.name}
```
dozortsev, line 95: atm we have a little issue when search for page titles the search doesn't show their names in list items, this is because they missing `id` attr. This is basically a fix to the issue but it will get applied only when we merge this PR in `main`.

Screen Shot 2021-12-22 at 11 42 45 AM
Screen Shot 2021-12-22 at 11 40 54 AM

packages/docs/components/Layout.js
```diff
@@ -13,6 +19,8 @@ import Meta from './Meta';
 import Sidebar from './Sidebar';
 import ThreeColumnLayout from './layout/threeColumnLayout';
 
+import '@docsearch/css';
```
matebek-tw, line 22: Shouldn't we include all the css in `_app.js` or in `main.css` to keep the imports in the same place?

matebek-tw:
I believe we don't need the event listener to store the scroll position in [_app.js](https://github.com/transferwise/neptune-web/pull/910/files#diff-a0d9bc311950c5ae7272124753ac11bff70a24b7f431a8902da25cc867a16612L38) anymore.

matebek-tw (approved):

packages/docs/components/Layout.js
dozortsev: yeah, good point, moving it to `_app.js`

PR #914, opened 2021-04-26
feat: fix MoneyInput accepting invalid input

arikanmstf-tw:
Quotes from the old [PR](https://github.com/transferwise/neptune-web/pull/836)

## 🖼 Context

> This PR aims to fix up a few things, but at the moment I haven't done any tests since I'd like to open it more for a discussion around the approach I've taken (I was planning on doing some exploration and opening a thread but then just decided to continue). 

https://transferwise.atlassian.net/browse/DS-555
https://transferwise.atlassian.net/browse/DS-864

## 🚀 Changes

> #### To fix: https://transferwise.atlassian.net/browse/DS-864

>  Changes our check for the `groupSeparator` to use 10k instead of 1k when trying to get the thousand delimiter. It seems `es-ES` doesn't comma separate `1000` as a value, but does separate `10000`. Updating this check to use `10000` & get the delimiter off the third character fixes the issue and doesn't seem to negatively affect a few of the other locales I tested.

>  Playing with values in the console to visualise it:

Screenshot 2021-03-05 at 17 18 54

> #### To fix: https://transferwise.atlassian.net/browse/DS-555

> This is the part of the PR I'm questioning the most. We want to stop users from inputting text values to prevent values like `$1000` since they won't trigger amount `onAmountChange` and customers can get confused as to why the quote isn't updating. 

>  I can basically think of two main approaches to this: 

>  1. Prevent user input of unwanted stuff `onKeyDown` before the change fires

>  2. Set the parsed value to state in `onChange`

> I decided to go with the first one since the second approach would require us to store/reset the caret position since otherwise it would jump to the end of the input on each change. The annoying (but also positive I guess?) thing about the first approach is that it requires us to very explicitly say what we'll accept. As you can see, the list is already pretty big just to support a few ways of interacting with the input like selecting all, using arrow keys, etc. so I'm worried this might be too prohibitive an approach. I'm also not sure what impact this will have on accessibility. But, that's what this PR is mostly for, to discuss the approach!

## Tests

Tested with browserstack at following browsers:

* IE 11 (Win10)
* Edge 90  (Win10)
* Firefox 88  (Win10)
* Safari 14 (Mac 10)
* Chrome  90 (Mac 10) 

Real device:
* Safari Mobile (IOS)
* Chrome Mobile (IOS)

## ✅ Checklist


packages/components/src/moneyInput/currencyFormatting.spec.js
```diff
@@ -25,4 +25,20 @@ describe('Number formatting', () => {
...
+
+  test.each([
+    ['0---9(s.t.r!@#$%^&*()_+^[]{};’,<>?', 9],
+    ['$12.341', 12.34],
+    ['$12.347', 12.35],
+    ['.asd121', 0.12],
+    ['.asd129', 0.13],
+    [',asd129', 129],
+    ['123.123.123', 123.12],
+    ['asda', NaN],
```
arikanmstf-tw, line 37: shall we return 0 if parsed value is NaN ?  means; shall we

keep the existing value when user pastes string without any number (`asda`);
or replace existing value with 0

dozortsev (commented):
I guess this PR closes https://github.com/transferwise/neptune-web/pull/836 right?

cc @fergusjordantw

packages/components/src/moneyInput/MoneyInput.js
```diff
@@ -52,6 +53,48 @@ class MoneyInput extends Component {
...
   }
 
+  isInputAllowedForKeyEvent = (event) => {
+    const { keyCode, metaKey, key } = event;
+    const isNumberKey = !Number.isNaN(parseInt(key, 10));
+
+    return (
+      isNumberKey ||
+      metaKey ||
+      keyCode === keyCodes.BACKSPACE ||
```
andreapiras, line 63: Hey can we try to reuse this util to verify the key? https://github.com/transferwise/neptune-web/blob/main/packages/validation/src/event-validators/isKey.js

andreapiras: I should put a comment in the keycodes file to deprecate

packages/components/src/moneyInput/MoneyInput.js
```diff
@@ -52,6 +53,48 @@ class MoneyInput extends Component {
...
+      keyCode === keyCodes.LEFT ||
+      keyCode === keyCodes.RIGHT ||
+      keyCode === keyCodes.ENTER ||
+      keyCode === keyCodes.ESCAPE ||
+      keyCode === keyCodes.TAB
+    );
+  };
+
+  handleKeyDown = (event) => {
+    if (!this.isInputAllowedForKeyEvent(event)) {
```
andreapiras, line 77: I am a bit confused, are we trying to allow only digits here?. If that's the case can we just not use a regular expression to filter event.target.value and remove isInputAllowedForKeyEvent completely?

packages/components/src/moneyInput/MoneyInput.js
```diff
@@ -52,6 +53,48 @@ class MoneyInput extends Component {
...
+    }
+  };
+
+  handlePaste = (event) => {
+    const paste = (event.clipboardData || window.clipboardData).getData('text');
+    const parsed = isEmpty(paste)
+      ? null
+      : parseAmount(paste, this.props.selectedCurrency.currency, this.locale);
+
+    if (!Number.isNaN(parsed)) {
```
andreapiras, line 88: You can use isNumber from @transferwise/neptune-validation

packages/components/src/moneyInput/MoneyInput.js
```diff
@@ -52,6 +53,48 @@ class MoneyInput extends Component {
     }
   }
 
+  isInputAllowedForKeyEvent = (event) => {
+    const { keyCode, metaKey, key } = event;
+    const isNumberKey = !Number.isNaN(parseInt(key, 10));
```
andreapiras, line 58: You can use isNumber from @transferwise/neptune-validation

packages/components/src/moneyInput/MoneyInput.js
```diff
@@ -52,6 +53,48 @@ class MoneyInput extends Component {
...
+    if (!this.isInputAllowedForKeyEvent(event)) {
+      event.preventDefault();
+    }
+  };
+
+  handlePaste = (event) => {
+    const paste = (event.clipboardData || window.clipboardData).getData('text');
+    const parsed = isEmpty(paste)
+      ? null
+      : parseAmount(paste, this.props.selectedCurrency.currency, this.locale);
```
andreapiras, line 86: Can we try and reuse this https://github.com/transferwise/formatting/blob/master/src/currency/currency.ts#L57

arikanmstf-tw: formatAmount returns string with group separator. ( ie: 123456 => 123,456) , however parseAmount returns valid float so consumer safely use it.

fergusjordantw: I think this should more closely align with how `onAmountChange` works. We should be setting the state to be the "money input value" and then calling `props.onAmountChange` with the parsed float. For paste we should just be stripping out unwanted characters. So if I paste in 1,000.12, I don't think it should reformat it to 1000.12. 

onAmountChange stores the user inputted value in state as `formattedAmount` and calls `props.onAmountChange` with the parsed float. I don't think this should deviate from that, the only thing we want to do here is make sure that the user inputted value is only valid characters (decimals, commas, digits).

packages/components/src/moneyInput/MoneyInput.js
```diff
@@ -52,6 +53,48 @@ class MoneyInput extends Component {
     }
   }
 
+  isInputAllowedForKeyEvent = (event) => {
```
fergusjordantw, line 56: Missing tests for all of this. Can we write some proper integration tests with RTL to test all of this?

packages/components/src/moneyInput/MoneyInput.js
arikanmstf-tw: When we talked in our last meeting, we aggreed that on paste, everything (including group separator) will be removed. and only digits and decimal separator will stay. did you remember ? @fergusjordantw @helendurranttw

fergusjordantw: I don't remember that specifically, next time lets write down decisions from the meeting. @helendurranttw can you confirm that's what we want?

packages/components/src/moneyInput/MoneyInput.js
arikanmstf-tw: done

packages/components/src/moneyInput/MoneyInput.js
arikanmstf-tw: done

packages/components/src/moneyInput/MoneyInput.js
andreapiras: Don't worry ignore my comment is was an optional change anyway

packages/components/src/moneyInput/MoneyInput.js
```diff
@@ -52,6 +54,48 @@ class MoneyInput extends Component {
...
+    const { keyCode, metaKey, key } = event;
+    const isNumberKey = isNumber(parseInt(key, 10));
+
+    return (
+      isNumberKey ||
+      metaKey ||
+      keyCode === keyCodes.BACKSPACE ||
+      keyCode === keyCodes.COMMA ||
```
helendurranttw, line 65: keyCodes are actually [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode), so we've been started moving towards [key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key).

We have a similar file of key mappings to the one we have of keyCodes.

I think there are probably a couple you're missing here. `delete` springs to mind, but I think there is also a `clear` button on some keyboards that might be helpful. Possibly also `home` and `end`?

Maybe you could keep a whitelist and then check if the value is in the list, rather than extending this or statement?

packages/components/src/moneyInput/MoneyInput.js
helendurranttw: As per our discussion on Slack, this is the expected behaviour:

**On keypress**
* prevent them from typing anything except numbers, the decimal separator or group separator

**On blur**
* remove invalid characters (e.g. duplicate decimal separators, group separators) and round to 2dp
* format the number using group separators

**On paste**
* remove all characters except numbers and the first decimal separator

I don't think you need to round on paste, since you're going to round on blur.

packages/components/src/moneyInput/currencyFormatting.spec.js
helendurranttw: I think we should leave whatever was there before

packages/components/src/moneyInput/MoneyInput.js
helendurranttw: I think there is some value in matching what we allow the user to type and what we accept on paste. So for example, if the user can type this

123.456 (which will be rounded to 123.46 on blur)
123.45.67 (which will be truncated to 123.45 on blur)

We could allow the same on paste:

$123.456 => 123.456 (on paste) => 123.46 (on blur)
$123.45.67 => 123.45.67 (on paste) => 123.45 (on blur)

This might feel more predictable, since it mimics the behaviour of the input when typing. That said, if we're going to soon limit the user from typing things like duplicate decimal separators.. maybe there's no point 😅

packages/components/src/moneyInput/MoneyInput.js
helendurranttw: So we're moving key validation to neptune validation (the package called validation in the neptune web monorepo). You can see the map of existing keys in `common/keyMap.js`. Once you've added the new keys you want to support in there, you can use [this util](https://github.com/transferwise/neptune-web/blob/main/packages/validation/src/event-validators/isKey.js) to check whether the key pressed is one of your list.

packages/components/src/moneyInput/currencyFormatting.spec.js
arikanmstf-tw: before, there was no logic on paste, this is a new thing

packages/components/src/moneyInput/MoneyInput.js
arikanmstf-tw: there is an update needed in order to use isKey, right? because not all keys are there

helendurranttw: Yeah, you'll need to add them. Once you've done that you should be able to remove your use of keyCodes :)

packages/components/src/moneyInput/MoneyInput.spec.js
```diff
@@ -531,4 +534,27 @@ describe('Money Input', () => {
...
+
+    beforeEach(() => {
+      rtlComponent = render(<MoneyInput {...props} amount={null} />);
+      jest.clearAllMocks();
+    });
+
+    test.each([
+      ['asd', ''],
+      ['1a2s3d', '123'],
+      ['±!@#$^*_+?><,', ''],
```
fergusjordantw, line 552: I think this should really include some examples of not just numbers being allowed but also decimals and commas

packages/components/src/moneyInput/MoneyInput.spec.js
```diff
@@ -531,4 +534,27 @@ describe('Money Input', () => {
       });
     });
   });
+
+  describe('when typing', () => {
+    let rtlComponent;
+    beforeAll(() => {
+      numberFormatting.formatAmount = jest.fn((num) => num);
```
fergusjordantw, line 541: Why do we need to mock formatAmount? Would it be possible to do a `jest.requireActual()` here to bypass the above mock so that we can test the entire integration?

packages/components/src/moneyInput/MoneyInput.js
```diff
@@ -52,6 +54,46 @@ class MoneyInput extends Component {
...
   }
 
+  isInputAllowedForKeyEvent = (event) => {
+    const { keyCode, metaKey, key } = event;
+    const isNumberKey = isNumber(parseInt(key, 10));
+
+    return (
+      isNumberKey ||
+      metaKey ||
+      isKey({ keyType: key, event }) ||
```
helendurranttw, line 64: This isn't quite right. The idea of this utility is to check to see if the event is a key you already know about. For example:

```
const { TAB, ESCAPE } = Key;

isKey({ keyType: TAB, event })
```

At the moment the key map only supports checking for space, tab, enter and escape. You probably don't want to support spaces, though I guess the others are fine.

arikanmstf-tw: btw isKey doesn’t support space right now, because at keyMap file, it’s key is SPACE  but it should be ' '
like this;
export const keyMap = {
  [' ']: { key: [' ', 'Spacebar'], keyCode: 32 },

I'll remove SPACE from here anyways 👍

packages/components/src/moneyInput/currencyFormatting.spec.js
helendurranttw: Right, I meant keep the existing value.

packages/components/src/moneyInput/MoneyInput.spec.js
arikanmstf-tw: there is an issue with testing library, COMMA and DOT keyCodes are different, and this causes the problem, I found the bug report at [testing-library](https://github.com/testing-library/user-event/issues/456) , it’s resolved at v13, however at neptune components testing library version is v11,

TLDR;
cant add the test entered  134,ada3  => resolved 134,3

from test:
```
console.log
    { keyCode: 44, key: ',', allowed: false }
  console.log:
    { keyCode: 46, key: '.', allowed: false }
```

packages/components/src/moneyInput/MoneyInput.js
```diff
@@ -52,6 +70,40 @@ class MoneyInput extends Component {
     }
   }
 
+  isInputAllowedForKeyEvent = (event) => {
+    const { keyCode, metaKey, key } = event;
+    const isNumberKey = isNumber(parseInt(key, 10));
+
+    return (
+      isNumberKey ||
+      metaKey ||
```
fergusjordantw, line 79: Can we test this on windows? I know this works for the command key on mac but we need to make sure we aren't removing the ability to paste. Might need to include ctrlKey here

arikanmstf-tw: very true. added it

helendurranttw (approved):

fergusjordantw (changes requested):
When pasting `100.000,12` into it when the locale is es (`.` being the separator, `,` being the decimal place) it parses it on blur to be `10.000.012`. So instead of sending `100,000.12`, we’re now sending `10,000,012.00` (examples with English locale)

arikanmstf-tw:
> When pasting `100.000,12` into it when the locale is es (`.` being the separator, `,` being the decimal place) it parses it on blur to be `10.000.012`. So instead of sending `100,000.12`, we’re now sending `10,000,012.00` (examples with English locale)

found the issue `this.state.locale stays as `en-GB` even when props changed

fergusjordantw (approved):

helendurranttw (approved):
🎉

PR #920, opened 2021-04-28
fix(PhoneNumberInput): phone prefix detection based on locale

dozortsev:
## 🖼 Context

`getCountryFromLocale` provides incorrect country code which was leading to setting wrong phone code at PhoneNumberInput component

A brief demo of the bug:

https://user-images.githubusercontent.com/4015628/116429883-34ab1c80-a83e-11eb-9fae-c28fa51af854.mov

## 🚀 Changes

Changed `getCountryFromLocale` to be based on [`Intl.Locale.region`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/region) prop (note: we provide [polyfill](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/common/polyfills/intl.js#L5)).

Changed a bit logic for finding country object in `setDefaultPrefix` to cover cases when local code has explicit region and when this is not the case.

Here is how PhoneNumberInput works with these changes:

https://user-images.githubusercontent.com/4015628/116430509-cfa3f680-a83e-11eb-8977-36c3e4a8fba9.mov

## ✅ Checklist


packages/components/src/common/locale/index.js
```diff
@@ -69,16 +67,23 @@ export function getLangFromLocale(locale) {
...
  */
 export const getCountryFromLocale = (locale) => {
   const adjustedLocale = adjustLocale(locale);
   if (adjustedLocale === null) {
     return null;
   }
-  return adjustedLocale.slice(0, COUNTRY_ISO2_CODE_LENGTH);
+  try {
+    const { region } = new Intl.Locale(adjustedLocale);
+    return region ?? null;
```
dozortsev, line 83: Screen Shot 2021-04-28 at 4 30 23 PM

region can be `undefined` thus we have check this via `??` and return `null` in this case as stated in JSDoc

packages/components/src/common/locale/index.js
```diff
@@ -69,16 +67,23 @@ export function getLangFromLocale(locale) {
...
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    console.error(error);
```
helendurranttw, line 86: I don't think there is any benefit in logging an error here, do you?

helendurranttw (approved):

packages/components/src/phoneNumberInput/utils/setDefaultPrefix/index.js
```diff
@@ -1,17 +1,23 @@
...
-  }
-  return prefix;
+  const country =
+    findCountryByCode(countryCode) ||
+    // when `locale` code has explicit region: `en-GB`, `en-US`, `ar-AE`
+    findCountryByCode(getCountryFromLocale(locale)) ||
+    // when `locale` code is only two chars value: `fr`, `es`
+    findCountryByCode(locale);
+
+  return country?.phone ?? DEFAULT_PHONE_CODE;
```
dozortsev, line 22: this is a shorter version of 

```javascript
country && country.phone ? country.phone : DEFAULT_PHONE_CODE;
```

helendurranttw: I think here you should probably use ||. If phone was '' you'd want it to use the default.

dozortsev: we don't have empty phones at our [source](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/phoneNumberInput/data/countries.js) but okey, this make code more reliable 👍

packages/components/src/common/locale/index.js
dozortsev: it's considered as a bad practice to ignore errors at catch statement, it could be useful for avoiding/discovering bugs

PR #922, opened 2021-04-29
refactor: introduce Responsive Panel

andreapiras:
## 🖼 Context

1 Select refactor PR

## 🚀 Changes

 - Introduced ResponsivePanel

## ✅ Checklist


packages/components/src/select/Select.js
```diff
@@ -464,60 +424,42 @@ export default class Select extends Component {
...
-              )} ${s('bottom')} ${s('tw-select-chevron')}`}
-            />
-          </button>
-          {!shouldRenderWithPortal ? (
-            <Transition
-              in={open}
-              timeout={BOOTSTRAP_DROPDOWN_ANIMATION_TIME}
-              onEntering={() => {
-                if (this.dropdownMenuRef.current) {
-                  addClassAndTriggerReflow(this.dropdownMenuRef.current, openClass);
```
andreapiras, line 501: @Tankenstein Hey do you remember what was this about?

dozortsev (commented):
This PR closes https://github.com/transferwise/neptune-web/pull/817

packages/components/src/select/Select.js
```diff
@@ -36,17 +36,6 @@ function stopPropagation(event) {
...
 }
 
-function getShouldRenderWithPortal() {
-  return (
-    typeof document !== 'undefined' &&
-    typeof window !== 'undefined' &&
-    window.matchMedia &&
-    window.matchMedia(`(max-width: ${Breakpoint.SMALL}px)`).matches
-  );
-}
```
dozortsev, line 46: :feels_good:

you can also similar logic for `isTouchDevice` variable by using `isTouchDevice()` from `deviceDetection.js`

andreapiras: I will but also this will be replaced with useConditionalListener that will listen on touchstart, because we are interested in when a user touches not if the devices is touchable

packages/components/src/common/deviceDetection/isTouchDevice.js
```diff
@@ -6,7 +6,17 @@ function supportsTouchEvents() {
     window.DocumentTouch &&
     document instanceof window.DocumentTouch;
 
-  return !!(onTouchStartIsDefined || maxTouchPointsIsDefined || documentTouchIsDefined);
+  const windowMatchMedia =
```
andreapiras, line 9: This all thing will be replace with useConditionalListener and window.addEventListener('touchStart') as we are interested in if the user is touching the screen rather than knowing if the screen is touch

helendurranttw: Can you explain what you're doing here? Are there cases where `ontouchstart` is not defined but `pointer: coarse` is true?

packages/components/src/phoneNumberInput/PhoneNumberInput.spec.js
```diff
@@ -7,6 +7,22 @@ import { fakeEvent } from '../common/fakeEvents';
...
 
+jest.mock('../common/panel', () => {
+  const { forwardRef } = jest.requireActual('react');
+  const Position = jest.requireActual('../common');
+  return {
+    Position,
+    // eslint-disable-next-line react/prop-types
+    ...forwardRef(({ open, children }, ref) =>
+      open ? (
+        <div ref={ref} className="np-responsive-panel">
```
helendurranttw, line 18: `np-panel`?

packages/components/src/dateInput/DateInput.spec.js
```diff
@@ -6,6 +6,22 @@ import DateInput from '.';
...
 import { fakeEvent } from '../common/fakeEvents';
 
+jest.mock('../common/panel', () => {
+  const { forwardRef } = jest.requireActual('react');
+  const Position = jest.requireActual('../common');
+  return {
+    Position,
+    // eslint-disable-next-line react/prop-types
+    ...forwardRef(({ open, children }, ref) =>
+      open ? (
```
helendurranttw, line 16: This isn't how the panel works, so it doesn't seem right to implement the mock like this (`open ? '' : null`).

packages/components/src/phoneNumberInput/PhoneNumberInput.spec.js
helendurranttw: Since you're repeating it, do you want to pull it somewhere common?

packages/components/src/phoneNumberInput/PhoneNumberInput.spec.js
```diff
@@ -18,6 +34,14 @@ describe('Given a telephone number component', () => {
   const PREFIX_SELECT_SELECTOR = 'Select';
   const NUMBER_SELECTOR = 'input[name="phoneNumber"]';
 
+  beforeAll(() => {
+    jest.spyOn(window, 'requestAnimationFrame').mockImplementation((cb) => cb());
```
helendurranttw, line 38: Should you put this in test config? Seems unlikely we'd ever want the real thing in a unit test

packages/components/src/select/Select.js
```diff
@@ -449,60 +424,41 @@ export default class Select extends Component {
       s('dropdown-toggle'),
     );
 
-    const openClass = s('open');
```
helendurranttw, line 452: You've missed the functionality that adds this class to dropdownmenuref when the options list opens, and removes it when it closes

packages/components/src/select/Select.spec.js
```diff
@@ -92,12 +101,6 @@ describe('Select', () => {
     return elementIndex;
   };
 
-  const expectOpenClassToHaveBeenAdded = () =>
-    expect(addClassAndTriggerReflow).toHaveBeenCalledWith(container().getDOMNode(), 'open');
```
helendurranttw, line 96: The reflow doesn't need to happen, but open should still be added.

packages/components/src/select/Select.spec.js
```diff
@@ -150,13 +153,6 @@ describe('Select', () => {
     expectDropdownToBe().open();
   });
 
-  it('can be closed by clicking somewhere else', async () => {
-    openSelect();
```
helendurranttw, line 154: FocusBoundary ftw!

packages/components/src/common/deviceDetection/isTouchDevice.js
andreapiras: I moved this logic from Select here I would not change it for now, once the component is going to be functional we can use useConditionalListener to check the touch event

packages/components/src/dateInput/DateInput.spec.js
andreapiras: What do you mean? At high level Panel is either returning the children or not. How would you implement it?

packages/components/src/select/Select.spec.js
andreapiras: What is it used for?

andreapiras: This was happening onEntering so it was a temporary class added to the trigger button. Not sure what it was doing

packages/components/src/common/deviceDetection/isTouchDevice.js
dozortsev: I wouldn't copy this from `Select` here, `isTouchDevice()` is already sufficient and could be used in `Select`, no need to extend it with Select's code, if we trust this function in `Stepper` component for same thing why wouldn't we in `Select`?

packages/components/src/select/Select.js
dozortsev: good point, although I would try to figure out whether it's actually really necessary

if it does then I would consider expose "onEntering" prop at `ResponsivePanel` component in order to pass callback (which would add the `.open` class) this down to [`CSSTransition#onEntering` of `Panel`](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/common/panel/Panel.js#L69-L78) and reuse `onClose` to pass it into `CSSTransition#onExit` for removing the CSS class

PR #924, opened 2021-04-30
[dynamic-flows] OneOfSchema now partially matches schemas

JulianG-TW:
## 🖼 Context

In a OneOfSchema, the initially active child schema is determined by the `getActiveSchemaIndex` function, which depends on the schema and the model.

It currently returns the index of a child schema if the model is valid for it. But not if the model partially matches the schema.

Consider the following `oneOf` schema with three "options":
```
{
  "title": "Choose schema",
  "oneOf": [
    {
      "type": "object",
      "title": "Option Zero (0)",
      "properties": {
        "a": { "type": "string", "const": "AAAA" },
        "b": { "type": "string" }
      },
      "required": ["a", "b"]
    },
    {
      "type": "object",
      "title": "Option One (1)",
      "properties": {
        "b": { "type": "string" },
        "c": { "type": "string" }
      },
      "required": ["b", "c"]
    },
    {
      "type": "object",
      "title": "Option Two (2)",
      "properties": {
        "a": { "type": "string", "const": "BBBB" },
        "c": { "type": "string" }
      },
      "required": ["a", "c"]
    }
  ]
}
```
- "Option Zero" requires both "a" and "b" fields.
- "Option One" requires both "b" and "c" fields.
- "Option Two" requires both "a" and "c" fields.
- The "a" fields is `const` and can take two values: "AAAA" or "BBBB".

When the model is `{ a: "BBBB" }` it doesn't satisfy any of the 3 child schemas, but clearly "Option Two" should be active.

In the current implementation "Option Zero" (index: 0) is active.

## 🚀 Changes

This PR updates the `getActiveSchemaIndex` function to select/activate a schema when the model partially satisfies it.

## ✅ Checklist


packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.js
```diff
@@ -63,6 +83,11 @@ const OneOfSchema = (props) => {
       }
     }
 
+    if (model !== null && typeof model === 'object' && Object.keys(model).length >= 1) {
+      // Even if the model does not satisfy any of the schemas, it may be closer to one of them.
+      return getBestMatchingSchemaIndexForModel(schema, model);
```
felixxiatransferwise, line 88: if all schemas had a 0 point weighting then shouldn't we exit out of this if branch and return null instead?

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.spec.js
```diff
@@ -118,6 +118,58 @@ describe('Given a oneOfSchema component', () => {
...
+            title: 'Option Two (2)',
+            properties: {
+              a: { type: 'string', const: 'BBBB' },
+              c: { type: 'string' },
+            },
+            required: ['a', 'c'],
+          },
+        ],
+      };
+      describe('when the model partially matches one of the child schemas', () => {
```
felixxiatransferwise, line 154: shall we add 1 more test here to verify the points system implementation? 

ie.
```
model:
{ b: 'str', c: 'str1' }
should match Option One (1)
```

JulianG-TW: good idea! thanks! done @ bdd613a8a27b3e9ac8bce30adf3da8fb553093d5

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.spec.js
```diff
@@ -118,6 +118,77 @@ describe('Given a oneOfSchema component', () => {
...
+        it('should select the schema that is partially matched by the model', () => {
+          const localModel = { a: 'BBBB' };
+          component = shallow(<OneOfSchema {...props} schema={localSchema} model={localModel} />);
+          const control = component.find(SchemaFormControl);
+          expect(control.prop('value')).toBe(2);
+        });
+      });
+
+      describe('when the model partially matches a child schema without a const property', () => {
+        it('it should select the schema that is partially matched by the model', () => {
```
felixxiatransferwise, line 174: maybe better as `it should select the schema that is best matched by the model`?

felixxiatransferwise (approved):

PR #940, opened 2021-05-12
fix: remove overflow hidden

andreapiras:
## 🖼 Context

An overflow hidden has been added [here](https://github.com/transferwise/neptune-web/pull/593) to prevent overflow when Tabs were nested into Card.

https://user-images.githubusercontent.com/16908937/118004733-9fca1800-b341-11eb-81ae-3695dbd86f38.mov

As secondary effect this prevents other overflow to happen  for example DateInput
image

## 🚀 Changes

We decided to remove the overflow hidden and recommend to remove animation from tabs when they are used into Cards

## ✅ Checklist


helendurranttw (approved):
Should we add a note in the docs to tell people that if they're using tabs inside cards they shouldn't animate them?

PR #948, opened 2021-05-18
feat: add theme to accordion

nabaroa:
## 🖼 Context

1. !important usage in hover that creates difficult specificities in consumer apps.
3. Wrong colour with dark backgrounds. 

## 🚀 Changes

1. Remove the background colour without the !important.

3. Prepare the widget for dark backgrounds.

### Before

imagen

On hover state
imagen

### After

imagen

## ✅ Checklist


helendurranttw:
Hey @nabaroa - I think rather than introducing knowledge about the utility classes into the accordion less we could introduce a theme parameter the same way we do with [Avatar](https://transferwise.github.io/neptune-web/components/content/Avatar). In the future we will hopefully get rid of these, but for now we want to keep a consistent approach across components so we can easily remove them in one swoop in the future.

packages/components/src/accordion/Accordion.less
```diff
@@ -16,15 +16,21 @@
   width: 100%;
   padding: 10px 24px;
 
-  .btn {
-    border: none;
+  &.decision {
+    .bg-primary &,
+    .bg--dark {
+      border-color: #4c638a;
```
helendurranttw, line 22: Is there a variable we can use here?

matebek-tw (approved):

matebek-tw:
@nabaroa could you please check and fix the snapshot tests? Let me know if you need help with them.

PR #950, opened 2021-05-19
feat: allow multiple file types to be provided to Upload

matebek-tw:
## 🖼 Context

Currently the file types that can be provided to `Upload` are limited to `usAccept: PropTypes.oneOf(['*', 'image/*', 'application/*', 'text/csv'])`. We need to be more specific in certain situations what file types we are allowing to upload, and also allow multiple file types at the same time. This change is backward compatible.

https://transferwise.slack.com/archives/CJNBX9CP6/p1621326387004200

## 🚀 Changes

- Allow `usAccept` to accept all string values
- Extend `isTypeValid` to parse multiple types

## 🤔 Considerations

Though the HTML input is more permissive on the `accept` attribute, (e.g. it will allow you to browse `.jpg` files when the provided MIME type is `image/jpg`), our validator will fail in that case as the [valid MIME type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types) is `image/jpeg`. With the prop type being a string, we are losing some of this control and pushing this responsibility on developers.

Also, while the component would accept `.jpg` as the `usAccept` prop, and the input will filter files based on the extension, the `isTypeValid` validator is implemented working with MIME types only.

## ✅ Checklist


StevePole:
LGTM, would be good for a project owner to take a look as well.

packages/components/src/upload/utils/isTypeValid/isTypeValid.js
```diff
@@ -4,19 +4,23 @@ export const isTypeValid = (file, rule, file64) => {
...
+
+  const allowedTypes = rule.replace(/\s/g, '').split(',');
   const fileType = getFileType(file, file64);
 
-  if (rule === '*' || fileType === rule) {
+  if (rule === '*' || allowedTypes.includes(fileType)) {
     return true;
   }
 
-  const splittedRule = rule.split('/');
```
helendurranttw, line 13: You could use `allowedTypes.some` here and avoid the need to cast :)

packages/components/src/upload/Upload.js
```diff
@@ -371,7 +371,7 @@ Upload.propTypes = {
   psButtonText: PropTypes.string,
   psProcessingText: PropTypes.string,
   size: PropTypes.oneOf(['sm', 'md', 'lg']),
-  usAccept: PropTypes.oneOf(['*', 'image/*', 'application/*', 'text/csv']),
```
helendurranttw, line 374: Could use a JSDoc comment here to explain how to provide multiple types

helendurranttw (commented):
One suggestion and a docs request - once you address I'll merge. Thanks for the contribution!

helendurranttw (approved):
Thank you!

PR #953, opened 2021-05-24
Fix initial value for validation async schema

JulianG-TW:
## 🖼 Context

ValidationAsyncSchema was ignoring its `model` prop.

Because of this, our "update recipient" form in backoffice featured a bug where the "iban" field was not being populated since validation-async was enabled for it.

missing iban

## 🚀 Changes

ValidationAsyncSchema now
- initialises its model with props.model
- passes it down to BasicTypeSchema
- declares model in propTypes and defaultProps

## ✅ Checklist


packages/dynamic-flows/src/jsonSchemaForm/validationAsyncSchema/ValidationAsyncSchema.js
```diff
@@ -100,13 +101,15 @@ ValidationAsyncSchema.propTypes = {
       param: Types.string,
     }),
   }).isRequired,
+  model: Types.oneOfType([Types.string, Types.number, Types.bool]),
```
StevePole, line 104: I don't know if there's strictly any reason this couldn't be an object or array? 🤔   

It's perhaps not the intended use case, but it would probably work...

StevePole (approved):
Left a comment/question, but otherwise looks fine 👍

packages/dynamic-flows/src/jsonSchemaForm/validationAsyncSchema/ValidationAsyncSchema.js
JulianG-TW: Good point.

I used these three because it's in the propTypes for `BasicTypeSchema` which is what this `ValidationAsyncSchema` renders.

I think if ValidationAsyncSchema can be used with other types, we'd need more tests and probably out of scope for this bug fix.

Thanks for approving!

PR #962, opened 2021-05-27
feat (MoneyInput): RTL Support

joe-sayer:
## 🖼 Context

This PR is necessary in order for MoneyInput to support RTL locales.

## 🚀 Changes

 
- Adjusted styling

## ✅ Checklist


packages/components/src/moneyInput/MoneyInput.less
```diff
@@ -26,6 +26,45 @@
...
+      }
+
+      .input-group-btn {
+        .dropdown-menu {
+          right: auto;
+          left: 0;
+        }
+
+        .btn {
+          padding-right: calc(var(--size-8) * 2);
```
matebek-tw, line 53: Couldn't we use `var(--size-16)` here?

matebek-tw (approved):

packages/components/src/moneyInput/MoneyInput.less
joe-sayer: I guess we could but I kept it like that as originally the CSS was using `calc(var(--size-8) * 2)`

nabaroa: It's a good practice to replace these kind of values, `calc(var(--size-8) * 2)`, if there is a --size that suits, when we find them, in order to keep consistency eventually.

nabaroa (approved):

PR #964, opened 2021-05-27
fix: fix provider polyfill import for ie

helendurranttw:
## 🖼 Context

Provider is not working correctly on IE - I think because the polyfill is not being imported correctly. I believe this is an issue with the asynchronous loading of the polyfill (Provider attempts to load before the polyfills are active). We could look at a more refined solution but this is a quick fix that works.

## ✅ Checklist


helendurranttw:
Instead of bundling the polyfill this we're going to recommend consumers use polyfill.io to conditionally include it.

PR #965, opened 2021-05-28
feat: theming poc

helendurranttw:
🌞 Theming proof of concept 🌚

I'm proposing that to implement theming on web we:

- Add a theme prop to the Neptune Provider - leaving the consuming application to manage setting the theme
- Handling theme switching within Neptune
- Use custom properties to implement theming

Components would have to be refactored to use custom properties.

For the sake of simplicity in this proof of concept I've stored the themes inside `components`. Imagine these are imported from `neptune-tokens`.

I've made some changes to docs so you can see the effect easily (look for the theme switcher in the bottom left corner). **Note that the only values that are currently themed are the background and primary text colour,** and it doesn't persist across page refreshes.

[Play with it here!](https://transferwise.github.io/neptune-web/branch/theming/components/content/Accordion)


packages/components/src/provider/Provider.js
```diff
@@ -1,12 +1,18 @@
...
 import { DirectionProvider } from './direction';
+import { ThemeProvider } from './theme';
 import en from '../i18n/en.json';
 import { DEFAULT_LOCALE, adjustLocale } from '../common/locale';
 import '../common/polyfills/intl';
 
-function Provider({ i18n, children }) {
+async function loadTheme(theme) {
+  const json = await import(`@transferwise/components/build/themes/${theme}.json`);
+  return json;
```
sbrandwoo, line 12: Is this a temporary way to import the theme?

packages/components/src/provider/Provider.js
```diff
@@ -20,17 +26,27 @@ function Provider({ i18n, children }) {
     intlConfig = { locale: adjustedLocale, messages };
   }
 
+  useEffect(() => {
+    loadTheme(theme).then((themeDef) => {
+      Object.entries(themeDef).forEach(([cssProp, value]) => {
+        document.documentElement.style.setProperty(`--${cssProp}`, value);
```
sbrandwoo, line 32: This will overwrite the default values, which will still be supplied by neptune-css?

The use of a Provider here is actually a little misleading, as these styles aren't scoped to the children of the Provider. If you have any Neptune Provider on your page, the entire page will be themed?

sbrandwoo (commented):
Does the use of CSS variables here mean that we'll need to switch to CSS props in all of our product code, and thus that IE11 will be unsupported / not have theme support?

packages/components/src/themes/index.ts
```diff
@@ -0,0 +1,9 @@
+import dark from './dark.json';
+import light from './light.json';
+
+const themes: Record<string, Record<string, string>> = {
```
karanjthakkar, line 4: Perhaps:

```
// somewhere.ts
type SupportedThemes = 'dark' | 'light';
```

```suggestion
const themes: Record<SupportedThemes, Record<string, string>> = {
```

packages/components/src/provider/Provider.js
```diff
@@ -41,10 +57,12 @@ Provider.propTypes = {
     defaultRichTextElements: PropTypes.shape({}),
   }).isRequired,
   children: PropTypes.node,
+  theme: PropTypes.string,
 };
 
 Provider.defaultProps = {
   children: undefined,
+  theme: 'light',
```
karanjthakkar, line 65: is there a way to centralize these two things:
- default theme
- available themes in the app

asking since its repeated in ThemeProvider as well.

packages/components/src/provider/theme/ThemeProvider.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const SUPPORTED_THEMES = ['dark', 'light'];
+const DEFAULT_THEME = 'light';
+
+export const ThemeContext = React.createContext(DEFAULT_THEME);
+
+export const ThemeProvider = ({ theme, children }) => (
+  <ThemeContext.Provider value={SUPPORTED_THEMES.indexOf(theme) > -1 ? theme : DEFAULT_THEME}>
```
karanjthakkar, line 10: Can this not be enforced via TS? or the way we use TS in neptune doesnt allow for this?

packages/docs/pages/_app.js
```diff
@@ -49,16 +68,18 @@ class MyApp extends App {
     const { Component, pageProps, locale, messages } = this.props;
 
     return (
-      <Provider i18n={{ locale, messages }}>
-        <Head>
-          <title>Neptune Web - the Wise Design System on Web</title>
-          <link rel="icon" href={`${process.env.ASSET_PREFIX}/static/assets/favicon.ico`} />
-        </Head>
+      <ThemeContext.Provider value={this.state}>
```
joe-sayer, line 71: Hi @helendurranttw, when implementing DirectionProvider, it's included as part of the neptune [Provider](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/provider/Provider.js#L24), I think we should follow the same pattern for Theme so consumers won't have to use additional providers. Please let me know thoughts :)

packages/components/src/provider/Provider.js
dozortsev: I wouldn't bother about lazy loading file with dark vars, with first launch I would just bundle those vars as JS object, it's not a big deal and shouldn't slow down main thing - theming

packages/components/src/provider/Provider.js
dozortsev: if we want have scoped theming... first idea comes in my mind is storing CSS properties and their values in React Context

dozortsev, line 32: minor note: this logic should sit in dedicated provider component - `ThemeProvider`

dozortsev: I would apply those variable to children so that it has scopes

dozortsev:
Will be done in another PR

PR #979, opened 2021-06-03
feat!: move Field out of lab

andreapiras:
## 🖼 Context

This PR moves Field out of lab and make it available as public API.
[Field docs ](https://transferwise.github.io/neptune-web/branch/field-extraction/components/inputs/Field)

## 🚀 Changes

 
- Moved Field out of lab
- Refactored implementation to remove validation logic from within the Field
- Exposed default validation based on jsonSchema
- Add tests
- Add docs for field

## ✅ Checklist


packages/components/src/field/Field.messages.ts
```diff
@@ -0,0 +1,13 @@
+import { defineMessages } from 'react-intl';
```
dozortsev, line 1: note you will need tweak [messages build command](https://github.com/transferwise/neptune-web/blob/main/packages/components/package.json#L30) in order to pick up `.ts` files

packages/components/src/field/Field.story.js
```diff
@@ -0,0 +1,158 @@
...
+  title: 'Field',
+};
+
+export const TextField = () => {
+  const error = text('manual error', 'A manual');
+  const help = text('help text', 'Please insert a value between 3 and 6 character');
+  const [value, setValue] = useState('');
+  const [validations, setValidations] = useState([]);
+
+  const rules = {
```
StevePole, line 21: I don't really see the point of sticking to this format now that it's not part of the interface.  You end up having to write a new library function in neptune-validation to make this format compatible with json schema.  I think just using json schema is a better example to give people as a default approach.

packages/components/src/field/Field.tsx
```diff
@@ -0,0 +1,93 @@
...
+  };
+
+  return (
+    <div
+      className={classNames('form-group', {
+        'has-error': type === Sentiment.NEGATIVE && message,
+        'has-info': type === Sentiment.NEUTRAL,
+      })}
+    >
+      <label className="control-label d-block">
```
StevePole, line 74: what happens with checkboxes and radio button?  They contain their own labels.

packages/components/src/field/Field.tsx
```diff
@@ -0,0 +1,93 @@
...
+        {label}
```
dozortsev, line 75: I guess we don't need to render label always as some components have it's own label e.g Checkbox

packages/components/src/inlineAlert/InlineAlert.js
```diff
@@ -27,7 +28,7 @@ const InlineAlert = (props) => {
 InlineAlert.propTypes = {
   children: PropTypes.node.isRequired,
   className: PropTypes.string,
-  type: PropTypes.oneOf(['success', 'error', 'warning', 'info']),
+  type: PropTypes.oneOf(['success', 'error', 'warning', 'info', 'negative', 'neutral']),
```
dozortsev, line 31: why adding new types? what is diff between `'error'` and `'negative'`?

packages/components/src/withNormaliser/validation/event-utilities/event-utilities.spec.js
```diff
@@ -23,18 +23,6 @@ describe('when normalizing an event', () => {
     expect(normalizeEvent(event)).toEqual(event);
   });
 
-  it('returns a number when the type is number', () => {
```
StevePole, line 26: Is this because you've removed support for number types?  If so, I think this is needlessly purest, and doesn't recognise that the models we build with our forms are being submitted to APIs that (correctly) want values other than strings.

This is issue for dynamic forms for that very reason, they are building a model for submission to our APIs and our APIs want numbers & booleans in addition to strings.

packages/components/src/field/getAlertMessage/getAlertMessage.tsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import { Sentiment } from '../../common/propsValues/sentiment';
+import { useIntl } from 'react-intl';
+
+import defaultMessages from '../Field.messages';
+
+interface Validations {
+  message: string;
+  value: number;
+}
```
dozortsev, line 10: you can export this type from `Field.tsx` and import it here

packages/components/src/field/Field.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+interface Validations {
+  message: string;
+  value: number;
+}
+
+type FieldProps = {
+  children: React.ReactElement;
+  label?: string;
+  messages: { error: string; help: string; validations: Validations[] };
```
dozortsev, line 17: what is `validations`?

packages/validation/src/field-validators/index.js
```diff
@@ -0,0 +1,242 @@
...
+  }
+
+  const failures = [];
+  if (!isValidRequired(value, isRequired)) {
+    failures.push({
+      message: 'required',
+      ...rules?.required,
+    });
+  }
+  if (rules?.minLength && !isValidMinLength(value, rules.minLength.value)) {
```
StevePole, line 78: You can't change this like this.  This library is for json schema, it was (incorrectly) moved out of the json schema scope.  We should move it back there, and it should rely on json schema encoding, not your custom format. 

If you want to have this different format, which I really think we should avoid, you can implement a separate implementation on top of the json schema validators, but I really don't see the gain.

packages/validation/src/field-validators/index.js
```diff
@@ -0,0 +1,242 @@
+import {
```
StevePole, line 1: If there's any change to these validators would be good to see them separately, not alongside moving them, because it's impossible to see a diff.

packages/validation/src/field-validators/spec.js
```diff
@@ -0,0 +1,210 @@
+import { getFieldValidationFailures, TYPE_ERROR } from '.';
```
StevePole, line 1: This is all unneccesary if you just use json schema to validate

packages/validation/src/type-validators/index.js
```diff
@@ -3,11 +3,18 @@ const isNumber = (value) => typeof value === 'number' && !Number.isNaN(value);
 const isInteger = (value) => {
   return isNumber(value) && Math.floor(value) === value;
 };
+
+const isDate = (date) => validDateObject(date) || validDateString(date);
```
StevePole, line 7: This file is also incorrectly moved out of the json schema context.  Dates are not part of json schema, because they're not part of json.  We shouldn't add anything about dat objects in this file.

packages/components/src/field/Field.story.js
```diff
@@ -0,0 +1,158 @@
...
+    type: 'string',
+    required: { value: { value: true, message: 'Custom required' }, message: 'Custom required' },
+    minLength: { value: 3 },
+    maxLength: { value: 6, message: 'Custom maxLength error message' },
+  };
```
dozortsev, line 26: why this needs to be defined in parent component instead of passing into `Field` [like it's done atm there](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/lab/field/Field.js#L28-L32)?

packages/validation/src/field-validators/index.js
andreapiras: Dynamic form is still using its own JsonSchema validation, I duplicated this to not interfere with jsonSchema but as discusses I can create a function wrapper that translates my format into jsonSchema format. That will require jsonschema validation to be moved into validation package

packages/components/src/field/Field.story.js
andreapiras: I create a convert function that allows me to use the JsonSchema validation. I do feel that this format is more friendly than jsonSchema but happy to change

packages/components/src/field/Field.tsx
andreapiras: Is an array of Validations that contains the validations to be displayed

packages/components/src/field/Field.tsx
andreapiras: Good point, I moved to this format to avoid the id generation for hmtFor
How would you feel if we'll use soemthing like this https://github.com/transferwise/neptune-web/blob/main/packages/components/src/common/Option/Option.js#L26
Where Element can be type label if label is provided or span if not? I still don't love the id solution. Mostly because we would have to include a generateid library that adds weight

packages/components/src/inlineAlert/InlineAlert.js
andreapiras: I think these should be the new types according to sentiment no? @helendurranttw

packages/components/src/withNormaliser/validation/event-utilities/event-utilities.spec.js
andreapiras: Yeah but should be a concern of normalizer to change to number or should be the consumer that should handle the type and return the correct one?

packages/validation/src/field-validators/spec.js
andreapiras: I changed this to reuse the jsonSchema validation that I moved inside validation as discussed

packages/components/src/field/Field.story.js
andreapiras: We decided to keep the validation outside to make it more flexible. People can either use our default validation or they can use their custom ones

packages/components/src/field/Field.messages.ts
```diff
@@ -0,0 +1,13 @@
...
+
+let messages: {
+  [index: string]: { id: string; defaultMessage: string; description: string };
```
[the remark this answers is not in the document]
andreapiras: The reason why I used typescript here it's because in getAlertMessage  I was getting the following error
```
No index signature with a parameter of type 'string' was found on type '{ minLength: { id: string; defaultMessage: string; description: string; }; }'
```

import defaultMessages from '../Field.messages';

when I access the defaultMessages[message] somehow defaultMessages keys needs to be defined

packages/components/src/field/Field.tsx
andreapiras: I used an Element and render label or span depending on label being provided or not

packages/components/src/field/Field.tsx
```diff
@@ -0,0 +1,103 @@
...
+import { Sentiment } from '../common';
+
+export interface Validations {
+  message: string;
+  value: number;
+}
+
+type FieldProps = {
+  children: React.ReactElement;
+  /** If label is not provided the field will be wrapped in a span */
```
dozortsev, line 17: why?

aren't it would create redundant a `span` with unnecessary CSS classes on it

sbrandwoo:
> This PR moves Field out of lab and make it available as public API.

I wasn't actually aware it was in the lab - should we have been testing it in our apps? Which apps have tested it and verified the API is what we want?

packages/components/src/field/Field.tsx
dozortsev: but aren't `span` will be redundant? what it will be for?

packages/components/src/inlineAlert/InlineAlert.js
StevePole: Should probably deprecate the old one if we're doing this?  Also, I think this is missing `positive`?

packages/components/src/field/Field.story.js
```diff
@@ -0,0 +1,146 @@
...
+    if (val.length < minLength) {
+      failures.push('minLength');
+    }
+
+    if (val.length > maxLength) {
+      failures.push('maxLength');
+    }
+
+    setValue(val);
+    setValidations(failures);
```
StevePole, line 34: We need to pass in strings, not string keys, otherwise we're really limiting the options for custom validation.  You can still use the strings from the default messages though, it's just on the consumer.  You could create a utility function to turn an array of validation keys into an array of messages?

packages/components/src/field/Field.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  const child = (
+    <WithExtendedMethods
+      render={(overriddenMethods: OverriddenMethods) =>
+        cloneElement(children, {
+          className: 'form-control',
+          ...overriddenMethods,
+        })
+      }
+      methodsToOvverride={methodsToOvverride}
```
matebek-tw, line 78: Typo in `methodsToOverride`

packages/components/src/inlineAlert/InlineAlert.js
andreapiras: @helendurranttw

packages/components/src/field/Field.story.js
andreapiras: This now accepts an array of strings

packages/components/src/inlineAlert/InlineAlert.js
andreapiras: I used the deprecated Sentiment and removed and receverted inlineAlert

packages/components/src/field/Field.story.js
```diff
@@ -0,0 +1,146 @@
...
+  };
+
+  return (
+    <Field
+      label="Text Field"
+      messages={{
+        help,
+        error,
+        validations,
+        info: {
```
StevePole, line 44: Can you explain why this needs to be an object?

andreapiras: Both props are related to Info and I think it makes sense  to group them, also I don't think aria-label should be fixed so I think it makes sense to keep them both

andreapiras: With @StevePole we decided to refactor the props into two separate ones info and infoAriaLabel because we may deprecate the use of popover and inline info content, that would prevent breaking change

StevePole (approved):
👍

packages/components/src/field/getAlertMessage/getAlertMessage.tsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import { Sentiment } from '../../common/propsValues/sentiment';
+
+interface Args {
```
dozortsev, line 4: you don't need wrap all arguments of `getAlertMessage` function into one type, just list them out like this

```ts
const getAlertMessage = (blurred: boolean, changed: boolean, ....)
```

packages/components/src/field/getAlertMessage/getAlertMessage.tsx
```diff
@@ -0,0 +1,54 @@
...
+  blurred: boolean;
+  changed: boolean;
+  focused: boolean;
+  messages: { error?: string; help?: string; validations?: { message: string }[] };
+  submitted: boolean;
+}
+
+interface Return {
```
dozortsev, line 12: `AlertMessage` would sound a lot better and more useful

packages/components/src/field/getAlertMessage/getAlertMessage.tsx
```diff
@@ -0,0 +1,54 @@
...
+  message: undefined | string | React.ReactElement;
+  type: string | Sentiment.ERROR | Sentiment.INFO;
```
dozortsev, line 14: why it can be `string`?

packages/components/src/field/getAlertMessage/getAlertMessage.tsx
```diff
@@ -0,0 +1,54 @@
...
+  focused,
+  messages: { error, help, validations },
+  submitted,
+}: Args): Return => {
+  let type = '';
+  let message = undefined;
+
+  const isErrorVisible = !changed && error;
+  const isValidationVisible = (submitted || (changed && blurred)) && !!validations?.length;
+  const isHelpVisible = focused && help;
```
dozortsev, line 29: what type is this? is it `boolean` or at can also be object (`help`)

packages/docs/liveEditorCode/field.new.code.js
```diff
@@ -0,0 +1,38 @@
...
+    const maxLength = 6;
+    if (val.length < minLength) {
+      const message = intl.formatMessage({ id: 'neptune.field.minLength' }, { minLength });
+      failures.push(message);
+    }
+
+    if (val.length > maxLength) {
+      const message = 'A custom message for maxLength errors';
+      failures.push(message);
+    }
```
dozortsev, line 18: why would consumer implement something like this on their side? 

do we have/provide default functions for checking common things like min length?

cc @StevePole

packages/docs/liveEditorCode/field.new.code.js
```diff
@@ -0,0 +1,38 @@
+() => {
```
dozortsev, line 1: please remove `.new.` in file name

packages/docs/liveEditorCode/field.new.code.js
```diff
@@ -0,0 +1,38 @@
...
+  const [value, setValue] = React.useState('a value');
+  const [validations, setValidations] = React.useState([]);
+  const intl = useIntl();
+
+  const handleOnChange = (val) => {
+    const failures = [];
+    const minLength = 3;
+    const maxLength = 6;
+    if (val.length < minLength) {
+      const message = intl.formatMessage({ id: 'neptune.field.minLength' }, { minLength });
```
dozortsev, line 11: using here React Intl is redundant, consumers know that their messages needs to be translated

dozortsev: we should try keep docs demo examples as short as possible

this is too big example

Screen Shot 2021-07-01 at 1 13 52 PM

packages/docs/pages/components/inputs/Field.mdx
```diff
@@ -0,0 +1,23 @@
...
+Field is a component that wraps an input and returns it with an attached alert message.
+
+<LiveEditorBlock code={code} scope={{ Field, useIntl }} />
+<GeneratePropsTable componentName="Field" />
+
+## Validations
+
+The validation logic hasn't been integrated in Field on purpose so consumers can provide their own custom validations.
+
+Field accepts an array of validation messages that will be displayed in an attached InlineAlert component. A set of default translated messages are provided and can be access via id. See list of supported error messages [here](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/Field/Fiels.messages.ts)
```
dozortsev, line 17: "InlineAlert" should be a link to the component https://transferwise.github.io/neptune-web/components/alerts/InlineAlert

packages/docs/pages/components/inputs/Field.mdx
```diff
@@ -0,0 +1,23 @@
...
+
+Field is a component that wraps an input and returns it with an attached alert message.
+
+<LiveEditorBlock code={code} scope={{ Field, useIntl }} />
+<GeneratePropsTable componentName="Field" />
+
+## Validations
+
+The validation logic hasn't been integrated in Field on purpose so consumers can provide their own custom validations.
+
```
dozortsev, line 16: > ..A set of default translated messages are provided and can be access via id.

how consumers suppose to know about them? do you expect them to explore source code in order to figure things out?

packages/docs/liveEditorCode/field.new.code.js
StevePole: We can't anticipate all validation requirements that a consumer might have so we need to have custom validation.  

We had an earlier approach where you created a JSON structure to define your validation requirements, and then a library function to validate and extract messages.

Sharing what I wrote to Andrea:

I was suggesting adopting a declarative validation approach in most use cases simply isn’t necessary.
Rather than creating this validation structure so that we can do validation (pseudo code)
```
import { getFieldValidationFailures } from '...'
function() {
  const validationStructure = {
    minLength: {
      value: 6,
      message: "Must be at least 6 characters"
    }
  }
  const onChange = (value) = {
    setValidations(getFieldValidationFailures(validationStructure, value);
  }
  return (
    <Field ...>
      <input onChange={onChange} />
    </Field>
  );
}
```
You can just write your own validation function and it’s both simpler and easier
```
function() {
  const onChange = (value) = {
    if (value.length < 6)
      setValidations(["Must be at least 6 characters");
    }
  }
  return (
    <Field ...>
      <input onChange={onChange} />
    </Field>
  );
}
```
If you happen to be in a dynamic form where you already have a declarative json schema structure, you can just use the existing validation
```
import { getValidationFailures } from '/json-schema/validations'
function(schema) {
  const onChange = (value) = {
    setValidations(getValidationFailures(schema, value);
  }
  return (
    <Field ...>
      <input onChange={onChange} />
    </Field>
  );
}
```
But that’s a special case.  Anyone using it directly, I would expect just to write their own validation function, and not have to learn a special ‘language’ to manage validation.

packages/components/src/field/Field.tsx
```diff
@@ -0,0 +1,113 @@
...
+  children: React.ReactElement;
+  label?: string;
+  messages: {
+    error?: string;
+    help?: string;
+    info?: string;
+    infoAriaLabel?: string;
+    validations?: { message: string }[];
+  };
+  submitted?: boolean;
```
dozortsev, line 22: could you please add JSDoc to explain what role this prop plays, as a consumer me it's difficult to say what difference it would make

packages/docs/pages/components/inputs/Field.mdx
andreapiras: No there's an example in docs so consumers don't have to go through the source code to figure it out ☺️
https://transferwise.github.io/neptune-web/branch/field-extraction/components/inputs/Field

packages/components/src/field/getAlertMessage/getAlertMessage.tsx
```diff
@@ -0,0 +1,54 @@
...
+    type = Sentiment.ERROR;
+    message = error;
+  } else if (isValidationVisible) {
+    type = Sentiment.ERROR;
+    message = (
+      <ul className="list-unstyled">
+        {validations?.map((message, key) => (
+          <li key={key}>{message}</li>
+        ))}
+      </ul>
```
dozortsev, line 41: here I guess we should visually separate them better, e.g with dot or dash at the beginning

Screen Shot 2021-07-01 at 3 08 59 PM

packages/components/src/field/Field.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+type FieldProps = {
+  children: React.ReactElement;
+  label?: string;
+  messages: {
+    error?: string;
+    help?: string;
+    info?: string;
+    infoAriaLabel?: string;
+    validations?: { message: string }[];
```
dozortsev, line 20: I think `error` is redundant, consumers can use `validations`, or I'm missing something?

packages/components/src/field/Field.tsx
```diff
@@ -0,0 +1,113 @@
...
+      methodsToExtend={methodsToExtend}
+      {...children.props}
+    />
+  );
+
+  return (
+    <div
+      className={classNames('form-group', {
+        'has-error': type === Sentiment.ERROR && message,
+        'has-info': type === Sentiment.INFO,
```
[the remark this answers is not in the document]
dozortsev: I assume you use them here because `InlineAlert` relies on them but you can avoid increasing tech debt by adding little mapping

this would later help avoid us introducing breaking change for `Field`

packages/docs/pages/components/inputs/Field.mdx
dozortsev: I mean do you think it's sufficient and clear enough?

packages/components/src/field/Field.tsx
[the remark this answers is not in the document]
andreapiras: InlineAlert accepts these:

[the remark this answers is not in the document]
andreapiras: I know they are deprecated but we should not address them in this  PR

packages/components/src/field/Field.tsx
andreapiras: error is a manual error that has priority over validations

packages/docs/liveEditorCode/field.new.code.js
dozortsev: Right, my concern is that this component won't gain much of an adoption despite very big use case/demand (we have loads of forms in product). Because it doesn't fully provide end-to-end tooling (API) for form item validation. This leaves room for duplicate and inconsistent code across repos. I'm sure consumers who own a few apps won't be happy implementing similar validation several times, they would rather learn special ‘language’ to manage validation. It seems like this implementation built to mostly satisfy DF use case.

I'm wondering whether providing both solutions was considered (where Field internally validates value according to provided JSON validation object + having ability to write custom validation for custom things)? if so why not go this way?

StevePole: We already have a declarative structure for defining validity (JSON schema), and this repository already exposes a library to do that validation, and as you can see in my 3rd example it's trivial to put them together.  It would be good to demonstrate that in the docs though.

We also have a component that will eventually wrap this, effectively a "DynamicField" which will do exactly what you suggest in terms of working completely off the structure in a purely declarative manner.  

I'd rather have that layered approach than a complex interface supporting multiple validation styles.  

Even with custom validation I don't think a custom implementation of 'minLength' (or 'required' or 'minimum') can differ in behaviour, it's just a simple comparator.  In my examples, it's _less code_ than creating a JSON structure and importing/calling a library method to implement '<'.

packages/components/src/field/Field.messages.ts
```diff
@@ -0,0 +1,38 @@
+import { defineMessages, MessageDescriptor } from 'react-intl';
+
+const messages: {
+  [index: string]: MessageDescriptor;
+} = defineMessages({
+  minLength: {
+    id: 'neptune.field.minLength',
+    defaultMessage: 'Please enter at least {minLength} characters.',
```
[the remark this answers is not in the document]
andreapiras: No they have been provided by Katy D'Avella who is a copywriter

packages/components/src/field/Field.spec.js
```diff
@@ -0,0 +1,354 @@
...
+    expect(screen.queryByLabelText(ariaLabel)).not.toBeInTheDocument();
+  });
+  it('renders children', () => {
+    render(
+      <Field {...props}>
+        <input type="text" />
+      </Field>,
+    );
+
+    expect(screen.getByLabelText(props.label)).toBeInTheDocument();
```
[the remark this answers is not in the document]
andreapiras: this return the actual input associated with label
image

packages/components/src/field/Field.spec.js
```diff
@@ -0,0 +1,354 @@
...
+  });
+  it('renders info', () => {
+    render(
+      <Field {...props}>
+        <input type="text" />
+      </Field>,
+    );
+
+    expect(screen.getByLabelText(props.label)).toBeInTheDocument();
```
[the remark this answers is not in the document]
andreapiras: this return the actual input associated with label
image

packages/components/src/field/Field.tsx
```diff
@@ -0,0 +1,113 @@
...
+    <WithExtendedMethods
+      render={(extendedMethods: extendedMethods) =>
+        cloneElement(children, {
+          className: 'form-control',
+          ...extendedMethods,
+        })
+      }
+      methodsToExtend={methodsToExtend}
+      {...children.props}
+    />
```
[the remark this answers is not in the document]
andreapiras: https://github.com/transferwise/neptune-web/pull/979/files#diff-ce188bfe96344ff6bbbf9d2b65174694390b5ab7d4c434067b59e7a29991b96cR2

packages/components/src/field/getAlertMessage/getAlertMessage.tsx
andreapiras: why?

packages/components/src/field/getAlertMessage/getAlertMessage.tsx
andreapiras: It was because of initialization but I guess it can be undefined
```
let type = '';
```

packages/components/src/field/getAlertMessage/getAlertMessage.tsx
andreapiras: I personally prefer to pass objects as function so args not depend on the order they are passed in

packages/components/src/field/getAlertMessage/getAlertMessage.tsx
helendurranttw: @gusso what do you think?

helendurranttw: Should we even display all at once? Or just display the first error? @StevePole you may have opinions on this.

richardmccartney:
@andreapiras / @helendurranttw I see this was removed from a release milestone, is this component work still relevant or can we close this PR for the time being?

PR #982, opened 2021-06-09
fix: dynamic layout for actions

gabrieltaets-tw:
## 🖼 Context

The library doesn't define a layout called `action`, it defines `button`. The `button` contains an `action` object with `$ref` pointing to an `action.$id`.

## 🚀 Changes
Changes `action` type to `button` when parsing references.

## ✅ Checklist


StevePole:
I think this is going to conflict with this larger PR that addresses the same issue: https://github.com/transferwise/neptune-web/pull/987 may be better to focus on that one?

PR #983, opened 2021-06-09
Adjustments to Dynamic-flows

richardbata:
## 🖼 Context
Whilst integrating the DF component with the mitigator backend as well as send-flow UI flow, there are a couple adjustments necessary to the DF component.

## 🚀 Changes
- Fixing the initial construction of the model

- tweaks to the default httpClient to support cookie-based authentication

- slight change in behaviour of the callback events with the latest version being:
-- **onClose**: flow completion condition triggered without exceptions
-- **onStepChange**: successful response received from the getStep call
-- **onError**: in case of any http (get requirements, refresh, etc.) or JS exception, this callback is called, indicating the flow encountered a fatal error, and the consumer should act upon it

- adding parameters to callbacks for more context for consumers of the component
-- onClose() -> onClose(result)
-- onStepChange() -> onStepChange(nextStep, prevStep)
-- onError() -> onError(errors)

## ✅ Checklist


packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -20,7 +20,16 @@ import { Size } from '../common';
...
-  const { baseUrl, flowUrl, httpClient, onClose, onStepChange, onError } = props;
+  const {
+    baseUrl,
+    flowUrl,
+    onSuccessClose,
+    onFailureClose,
+    onStepChange,
+    onError,
+    httpClient: propsHttpClient,
+    oauthToken,
```
mikknarusk, line 31: as an aside - clients are generally strongly moving towards cookie based auth only. Meaning we shouldn't expect clients to be aware of any tokens, including `oauthToken`. The http client should just be able to send along all cookies and that should take care of authentication. Any reason this isn't working here?

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -31,72 +40,82 @@ const DynamicFlow = (props) => {
...
 
-    return getHttpClient()
+    return httpClient
       .request({ action, data })
       .then(checkForExitCondition)
       .then(({ data: json }) => {
         setStepSpecification(json);
 
-        onStepChange(json);
+        onStepChange({ nextStep: json });
```
StevePole, line 65: Why is this necessary?

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -175,11 +192,21 @@ const DynamicFlow = (props) => {
     return newModels;
   };
 
+  const buildInitialModels = (model, schemas) => {
```
StevePole, line 195: Why do you need to do this?  I believe each dynamic form will do this as needed?

packages/dynamic-flows/src/flow/DynamicFlow.js
richardbata: For consistency and being more explicit. {errors} and {result} also used for example

packages/dynamic-flows/src/flow/DynamicFlow.js
richardbata: Do cookies need to be passed explicitly, or does `@transferwise/api-call` need to be used to make authentication cookie based? Is there a good example in another repo I can follow?

mikknarusk: depends on what client this component uses but generally if you're using `axios` or plain `fetch` (the 2 most popular options) same-domain cookies get sent by default

richardbata: For cookie based authentication, does the `x-access-token` header need to be hardcoded to `Tr4n5f3rw153`?

mikknarusk: Possible but I'm not very sure actually. Best to test against actual APIs I guess, or check with public API folks

JulianG-TW: AFAIK in Ninjas, users are authenticated using `@transferwise/backoffice-auth`, which [stores auth tokens in localStorage](https://github.com/transferwise/backoffice-auth/blob/a5eefec43fd5c14415558beef9d24cf662fdcce5/src/http-client/features/login.js#L8-L9)

Is this a reason to keep the prop? I'm not sure the ninjas one is an `oauthtoken`.

packages/dynamic-flows/src/flow/DynamicFlow.js
richardbata: Primary reason I added this is because the hidden fields dont emit an onModelChange event, hence dont return back with the values to initialize the map of models. 

Just for recap, `models` is a map from schemaId to the model value for that schema. The way this map was initialized was listening to onModelChange events when initializing DynamicLayout, and constructing the models map that way. Hidden fields don't emit this initial onModelChange event, hence models won't contain them.

Also the way I wrote it with that `INITIALIZATION_SCHEMA_PROPERTY` property isn't the cleanest either

packages/dynamic-flows/src/flow/DynamicFlow.js
StevePole: When you say errors and result are used, do you mean in the same callback?  I think those would only appear in other callbacks?

I see people do this for purposes of extensibility i.e. ability to add return properties without changing signature, but this a bad trade off to me.  Readability is reduced as you need to interrogate the implementation to find out the prop that's used.  The signature has still changed, and it's also not really more extensible, as you could return additional props in exactly the same way.   So I prefer to use consistent naked props where possible, rather than interim objects.  Does that make sense?

If we're returning different things at different times, that's a different issue.  But I'd rather avoid that if we can.  Not sure this callback should fire on error for example.

packages/dynamic-flows/src/flow/DynamicFlow.js
StevePole: I see, would it be better to have hidden fields broadcast?  Not a huge fan of doing it twice in two different places, and I think it needs to be retained in the other spot.

packages/dynamic-flows/src/flow/DynamicFlow.js
richardbata: I noted the new parameters in the PR description. 
I'm happy with what you prefer, probably have more experience with it. Do you see any value in passing these values to the callbacks in the first place?
Could you elaborate on what onError shouldn't be fired in your example? Also in the description I'm mentioning all places the event are fired, lets agree on that first

packages/dynamic-flows/src/flow/DynamicFlow.js
richardbata: We can do that, but then this hack remains with `INITIALIZATION_SCHEMA_PROPERTY `. Note we're still using `getValidModelParts`

packages/dynamic-flows/src/flow/DynamicFlow.js
StevePole: Fine with the logic around what events fire when 👍

I tend to think about callbacks as "life cycle events", and to think about what makes sense to broadcast at that point, for example for `onStepChange` it makes a lot of sense to broadcast the new step, but might also make sense to pass the last one e.g. `onStepChange(newStep, oldStep)`.  Don't need to add that now, but generally if we're publishing all of the interesting meta data it gives the consumer a lot of options on how to use those callbacks.

For now would prefer to revert to non nested Paramus on the callbacks for consistency with the rest of the system.

richardbata: Updated the callbacks to use regular parameters 👍 . Added prevStep to onStepChange

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -161,25 +184,21 @@ const DynamicFlow = (props) => {
...
-
-    return newModels;
+  const buildInitialModels = (model, schemas) => {
+    return schemas?.reduce(
+      (accumulator, schema) => ({
+        ...accumulator,
+        [schema.$id]: getValidModelParts(stepSpecification.model, schema),
+      }),
+      {},
+    );
```
StevePole, line 194: I feel like the readability of this section has decreased because we're trying to be a bit too smart and do everything in a single, functional line.  Before we had sequential lines of code where each could be understood independently.  But now we have a function call, inside a spread operation, inside a reducer.  You need to understand how they all relate to each other to understand what's being returned.

This is far from the only way to do it, but as an example this is significantly easier to read, specifically because it isn't clever:
```
const schemaIdToModelMap = {};

schemas.forEach((schema) > {
  schemaIdToModelMap[schema.$id] = getValidModelParts(stepSpecification.model, schema);
});

return schemaIdToModelMap;
```
Because we set up variable with a descriptive name, you arguably don't even need to look at the implementation to understand what it's returning.

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -114,15 +130,22 @@ const DynamicFlow = (props) => {
   const onModelChange = (newModel, formSchema, triggerModel, triggerSchema) => {
     const { $id } = formSchema;
 
-    const updatedModels = updateModels($id, newModel);
+    setModels((prevModels) => {
```
StevePole, line 133: Why are we passing an arrow function the the setter?

richardbata: Because on initialisation it can happen that multiple state changes happen between react renders, resulting in only the last update reflecting in state... By passing a function we ensure the prevModel is the latest one, and not the one from the last react render cycle

packages/dynamic-flows/src/flow/DynamicFlow.js
richardbata: I appreciate the detailed feedback and agree your version is more readable. I'll update it

StevePole (approved):
👍

PR #987, opened 2021-06-10
fix: fix DynamicButton to use action.title instead of label

nmenego:
## 🖼 Context

Getting errors using the latest version of dynamic-flows 13.3.1 with DynamicButtons when using the button component. Seems that `label` is required, when it should be `action.title`

https://transferwise.slack.com/archives/C5FBXBX8W/p1622452763101900

## 🚀 Changes

 
1. Making DynamicButton use action.title
2. Modifying actionModel to have title instead of label
3. Fixing some minor errors on the storybook (missing params in payload)
4. 

## ✅ Checklist


packages/dynamic-flows/src/flow/layoutService/LayoutService.js
```diff
@@ -134,14 +134,14 @@ function convertStepImageToDynamicImage(image) {
...
     margin: 'lg',
   };
 }
 
 function convertStepActionToDynamicAction(action) {
-  const newAction = { ...action, label: action.title };
-  delete newAction.type;
-  delete newAction.title;
+  const newAction = { ...action, title: action.title };
+  delete newAction.label;
```
StevePole, line 144: Not sure this is necessary now?  This was basically a cast from what I thought was the 'old' action format to the 'new' one that was used internally.  However, now that we're using the original format everywhere, there shouldn't ever be a 'label' on an action.

packages/dynamic-flows/src/flow/layoutService/LayoutService.spec.js
```diff
@@ -119,7 +119,7 @@ describe('Given a utility service for handling dynamic layouts', () => {
             {
               type: 'button',
               context: exitAction.type,
-              action: { ...exitAction, label: exitAction.title, title: undefined, type: undefined },
+              action: { ...exitAction, title: exitAction.title, label: undefined, type: 'success' },
```
StevePole, line 122: Similar point here, I don't think we need to unset label as it shouldn't be there.

StevePole: Not sure we need `title: exitAction.title` either, as we already have that from the `...exitAction` part.

StevePole (approved):
Looks good except for those very minor cleanup points.

PR #1006, opened 2021-06-14
fix: add media class conditions

joe-sayer:
## 🖼 Context

Fix media class conditions for AccordionItem Chevron

## ✅ Checklist


packages/components/src/accordion/AccordionItem/AccordionItem.js
```diff
@@ -49,7 +49,12 @@ const AccordionItem = ({ id, title, content, onClick, open, icon }) => {
           >
             {isString(title) ? <span className="h5">{title}</span> : title}
           </div>
-          <div className="media-right d-flex align-items-center">
+          <div
+            className={classNames('d-flex align-items-center', {
+              'media-right': !isRTL,
```
dozortsev, line 54: why do we need to add class for LTR case?

joe-sayer: By default there is `media-right`, we can leave it in always but then there would be additional space

dozortsev: ah damn, didn't notice it, ignore this comment

dozortsev (approved):

PR #1013, opened 2021-06-17
feat: info onClick prop

joe-sayer:
## 🖼 Context

Add an `onClick` property to `Info` to allow consumers to track when Info Modals/Popovers are open

## ✅ Checklist


packages/components/src/info/Info.js
```diff
@@ -48,6 +65,7 @@ Info.propTypes = {
   className: PropTypes.string,
   /** Content displayed inside a Popover a Modal */
   content: PropTypes.node,
+  onClick: PropTypes.func,
```
dozortsev, line 68: could you please add a little test for this in prop in `Info.spec.js`?

dozortsev (approved):

packages/components/src/info/Info.js
```diff
@@ -30,12 +30,29 @@ const Info = (props) => {
     >
       {isModal ? (
         <>
-          <button type="button" onClick={() => setOpen(!open)} {...buttonProps} />
+          <button
+            type="button"
+            onClick={() => {
+              setOpen(!open);
+              if (onClick) {
+                onClick();
```
andreapiras, line 38: Do you think it would be a good idea to pass outside the value of open?

packages/components/src/info/Info.js
```diff
@@ -30,12 +30,29 @@ const Info = (props) => {
...
         </>
       ) : (
         <Popover content={content} preferredPlacement={Position.BOTTOM} title={title}>
-          <button type="button" {...buttonProps} />
+          <button
+            type="button"
+            {...buttonProps}
+            onClick={() => {
+              if (onClick) {
+                onClick();
```
andreapiras, line 52: Same here we should know when popover is open or not

packages/components/src/info/Info.js
dozortsev: good point, but I don't know... it looks very clumsy to return custom event `{ open: true  }`..

perhaps it would make sense to return just native event?

or extended version of the event? 🤔 

```javascript
{
  ...event,
  infoStatus: open,
  component: 'modal' | 'popover'
}
```

otherwise I wouldn't return anything, if consumer needs to know whether it's opened or closed they can define variable on their side

```javascript
var infoStatus = false;

<Info onClick={() => { infoStatus = true; ... }} />
```

andreapiras: I think this should be definitely managed component level not consumer level. I would not get into the mess of keeping the internal state and the consumer state in sync. Also I don't see the value of returning the full event as we are not interested in anything else. I get that API level we should be consistent with what we return but in this case event doesn't contain anything we need

dozortsev: > ..I would not get into the mess of keeping the internal state and the consumer state in sync.

what do you mean? there is no mess :) could you bring an example?

> ..I get that API level we should be consistent with what we return but in this case event doesn't contain anything we need

you never know, returning full event won't harm while not returning introduces inconsistency

---

I suggest we don't return anything at all for now, it feels like you bond to much to use case that was brought when they suggested this feature which now influence API design, it's not about "on open" it's about "on click" you never know what would people listen/check on this event

andreapiras: yeah sure go ahead :)

PR #1014, opened 2021-06-18
fix: remove scroll lock from panel and prevent focus boundary taking 100% height

helendurranttw:
## 🖼 Context

1. A previous change introduced a scroll lock when opening panel, even though the dimmer is transparent. We've decided this behaviour is undesirable, and have removed it.
2. Some positioning logic caused focus boundary to take 100% height of the page. We've updated the solution so this is no longer necessary.

## 🚀 Changes

1. We no longer lock scroll if the dimmer is transparent
2. Modals with content scroll will now have a max height of 100vh - this prevents the top of the modal being cut off when content is longer than the screen

## 🤔 Considerations

I found an existing bug in IE11 - if the content overflows the height of the screen, the modal won't expand to fit it when in viewport scroll mode. It was existing and it's an edge case so I haven't tackled it here.

## ✅ Checklist


packages/components/src/dimmer/Dimmer.js
```diff
@@ -81,19 +81,23 @@ const Dimmer = ({
           ref={dimmerRef}
           role="presentation"
         >
-          <FocusBoundary>{children}</FocusBoundary>
+          <FocusBoundary onClick={handleClick}>{children}</FocusBoundary>
```
andreapiras, line 84: is this astill needed? I though moving from alig-self-center would have get rid of this

packages/components/src/dimmer/Dimmer.js
```diff
@@ -106,6 +110,7 @@ Dimmer.propTypes = {
   disableClickToClose: PropTypes.bool,
   fadeContentOnEnter: PropTypes.bool,
   fadeContentOnExit: PropTypes.bool,
+  /** Sets the background to transparent and prevents scroll locking */
```
andreapiras, line 113: Sets the background to transparent and prevents body scroll locking. I'd specify because we also have Dimmer scrolling

packages/components/src/dimmer/Dimmer.js
helendurranttw: Oops good catch 👍

andreapiras (approved):

PR #1015, opened 2021-06-18
feat: add UploadInput component

czebe:
## 🖼 Context

https://transferwise.atlassian.net/browse/DS-1169

A new `FileUpload` component that supports multiple files and provides a list view of files already uploaded to the system.

**Demo:**

**File deletion:**

https://user-images.githubusercontent.com/71433023/125950766-83a895f0-3117-49b9-badc-8b14d8193b6a.mp4

**Hover effect on uploaded file/remove button:**

Jul-16-2021 14-13-55

**Rendered in RTL context:**

Screenshot 2021-07-16 at 14 50 12

### How is this different from the `Upload` component?

- it supports multiple files
- it has a list view to display existing files
- it doesn't have image preview (various file types can be uploaded - eg. pdf, csv, xls, doc)
- it's decoupled from the actual API request, so the host app can use any API library to perform the request

### Designs:

https://www.figma.com/file/1YNhPFlUDuOTQWjZQDESBn/File-upload--%2F-Component?node-id=201%3A15093

## 🚀 Changes

- implement new component
- add documentation
- improve build scripts to better support TS source files:
   - build `.tsx` story files
   - replace `formatter.js` with the built-in `simple` for generating translation files
   - process messages in `ts, tsx` files too

## 🤔 Considerations

The `FileUpload` component is intentionally implemented as a view component (like the rest of `components`) -- unlike the `Upload` component. It gives the host applications the freedom to use any library for API requests (eg. `fetch/axios/api-call`).

## ✅ Checklist


packages/components/src/fileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,183 @@
...
+   * @param id
+   */
+  onCreateDeleteRequest?: (id: string | number) => Promise<any>;
+} & Pick<UploadButtonProps, 'disabled' | 'multiple' | 'fileTypes' | 'sizeLimit'> &
+  CommonProps;
+
+const FileUpload: FC<FileUploadProps> = ({
+  files = [],
+  fileInputName = 'file',
+  className = '',
```
dozortsev, line 49: we tend to use `undefined`

packages/components/src/fileUpload/uploadButton/UploadButton.tsx
```diff
@@ -0,0 +1,167 @@
...
+              <div>{formatMessage(multiple ? MESSAGES.uploadFiles : MESSAGES.uploadFile)}</div>
+              <small className={`np-upload-description ${!disabled ? 'text-primary' : ''}`}>
+                {formatMessage(MESSAGES.instructions, {
+                  fileTypes: description,
+                  size: Math.round(sizeLimit / 1000),
+                })}
+              </small>
+            </div>
+            <div className={`np-upload-icon ${isRTL ? 'media-left' : 'media-right'}`} />
+          </div>
```
dozortsev, line 148: why do we need to render this empty div

Screen Shot 2021-06-21 at 4 09 49 PM

packages/components/src/fileUpload/uploadButton/UploadButton.tsx
```diff
@@ -0,0 +1,167 @@
...
+        </label>
+
+        {/* Drop area overlay */}
+        {isDropping && (
+          <div className="droppable-card droppable-dropping-card">
+            <div className="droppable-card-content">
+              <div className="text-info m-x-1">
+                <PlusIcon filled size={24} />
+              </div>
```
dozortsev, line 157: what if you apply those styles directly onto icon and remove `div`? will it work same way?

```javascript
<PlusIcon filled size={24} className="text-info m-x-1" />
```

dozortsev: if that would work you should be able to get rid of 2 `div`s, end result:

```jsx
          <div className="droppable-card droppable-dropping-card droppable-card-content">
            <PlusIcon filled size={24} className="text-info m-x-1" />
            <div className="text-info">{formatMessage(MESSAGES.dropFile)}</div>
          </div>
```

packages/components/src/fileUpload/uploadButton/UploadButton.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+  const fileTypesDescription = getAllowedFileTypes(fileTypes).join(', ');
+  const allowAllFileTypes = fileTypesDescription === '*';
+  const description = allowAllFileTypes
+    ? formatMessage(MESSAGES.allFileTypes)
+    : fileTypesDescription;
+
+  return (
+    <div
+      className={`np-upload-button-container droppable ${isDropping ? 'droppable-dropping' : ''}`}
```
dozortsev, line 110: please use `classNames` function:

```jsx
classNames('np-upload-button-container', 'droppable', { 'droppable-dropping' : isDropping })
```

packages/components/src/fileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,183 @@
...
+export type FileUploadProps = {
+  /**
+   * List of already existing, failed or in progress files
+   */
+  files?: UploadedFile[];
+
+  /**
+   * The key of the file in the returned FormData object (default: file)
+   */
+  fileInputName?: string;
```
dozortsev, line 29: struggle to understand what this for?

packages/components/src/fileUpload/uploadButton/UploadButton.tsx
```diff
@@ -0,0 +1,167 @@
...
+      {...(!disabled && { onDragEnter, onDragLeave, onDrop, onDragOver })}
+    >
+      <form method="POST">
```
dozortsev, line 113: wondering why you need form here? won't it break things when this components nested into another form?

czebe: good catch, it would make it invalid HTML. Removing `<form>`.

packages/components/src/fileUpload/FileUpload.tsx
czebe: This is the `key` in the resulting `FormData` object. It's not possible to change the key after it's been set, so we need to know what's the backend API expecting.

packages/components/src/fileUpload/uploadButton/UploadButton.tsx
czebe: ahh, well okay. I feel it's kind of useless with template literals.
(one more dependency to maintain, one more import line and it's not much shorter or easier to understand)

packages/components/src/fileUpload/uploadButton/UploadButton.tsx
czebe: It's not needed, removing!

packages/components/src/fileUpload/FileUpload.story.tsx
```diff
@@ -0,0 +1,113 @@
+import React, { useState } from 'react';
+import { Story, Meta } from '@storybook/react';
+
+import { Status } from '../common';
+import FileUpload, { FileUploadProps } from './FileUpload';
+import { UploadResponse } from './types';
+
+export default {
```
dozortsev, line 8: could you please stick to the current pattern of our storybook? 🙏 it now looks like FileUpload is some special one

Screen Shot 2021-06-22 at 2 11 08 PM

packages/components/src/fileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+import MESSAGES from './FileUpload.messages';
+
+import './FileUpload.css';
+
+export type FileUploadProps = {
+  /**
+   * List of already existing, failed or in progress files
+   */
+  files?: UploadedFile[];
```
dozortsev, line 25: I would call this prop `uploadFiles`, comment and type of this prop is hinting it be name this way

packages/components/src/fileUpload/uploadButton/UploadButton.tsx
```diff
@@ -0,0 +1,179 @@
...
+   * @param files
+   */
+  onChange: (files: FileList) => void;
+};
+
+export enum TEST_IDS {
+  uploadInputLabel = 'uploadInputLabel',
+  uploadInput = 'uploadInput',
+  mediaBody = 'mediaBody',
+}
```
dozortsev, line 47: could we please get rid of these `data-testid`? this is weird having hardcode them,

as far as I can see they defined to ease testing but this is a sign of poor a11y on in this component (see priority of queries https://testing-library.com/docs/queries/about/#priority)

e.g I can't jump on this component using <kbd>Tab</kbd> being in clean surface

Screen Shot 2021-06-22 at 2 20 56 PM

I think @helendurranttw and @andreapiras  could help us make it more accessible

packages/components/src/fileUpload/uploadButton/defaults.ts
```diff
@@ -0,0 +1,5 @@
+import { FileType } from '../types';
+
+export const DEFAULT_SIZE_LIMIT = 5000;
```
dozortsev, line 3: should we also have default max number of files allowed to be uploaded? e.g 2?

I guess this is most question to designers cc @gusso

packages/components/src/fileUpload/uploadItem/UploadItem.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const { isRTL } = useDirection();
+  const { formatMessage } = useIntl();
+  const { status, filename, error, url } = file;
+
+  const getIcon = () => {
+    if (error || status === Status.FAILED) {
+      return (
+        <span data-testid={TEST_IDS.errorIcon} className="text-danger">
+          <AlertCircle size={24} />
+        </span>
```
dozortsev, line 44: note, icon components have both `className` and `data-testid` props, so you could get rid of that `span`

packages/components/src/fileUpload/uploadItem/UploadItem.tsx
```diff
@@ -0,0 +1,140 @@
...
+        {title}
+      </a>
+    ) : (
+      title
+    );
+  };
+
+  return (
+    <div className="np-upload-button p-x-2 p-y-1" data-testid={TEST_IDS.uploadItem}>
+      <div className="media">
```
dozortsev, line 103: these two `div`s could be merged

packages/components/src/fileUpload/uploadItem/UploadItem.tsx
```diff
@@ -0,0 +1,140 @@
...
+}
+
+const PROGRESS_STATUSES = [Status.PENDING, Status.PROCESSING];
+
+const UploadItem: FC<UploadItemProps> = ({ file, deleteFile }: UploadItemProps) => {
+  const { isRTL } = useDirection();
+  const { formatMessage } = useIntl();
+  const { status, filename, error, url } = file;
+
+  const getIcon = () => {
```
dozortsev, line 39: please add type for this function as sometimes it return `void` sometimes JSX which is a bit confusing

packages/components/src/fileUpload/uploadItem/UploadItem.tsx
```diff
@@ -0,0 +1,140 @@
...
+  };
+
+  return (
+    <div className="np-upload-button p-x-2 p-y-1" data-testid={TEST_IDS.uploadItem}>
+      <div className="media">
+        <div
+          className={classNames('np-upload-icon', { 'media-right': isRTL, 'media-left': !isRTL })}
+        >
+          {getIcon()}
+        </div>
```
dozortsev, line 108: could all those classes be applied on icon instead?

packages/components/src/fileUpload/uploadItem/UploadItem.tsx
```diff
@@ -0,0 +1,140 @@
...
+        >
+          {deleteFile && (!status || (status && !PROGRESS_STATUSES.includes(status))) && (
+            <button
+              data-testid={TEST_IDS.deleteButton}
+              onClick={() => deleteFile(file)}
+              type="button"
+              className="np-upload-remove-button p-a-1 btn btn-unstyled"
+            >
+              <span className="sr-only">{formatMessage(MESSAGES.removeFile, { filename })}</span>
+              <CrossCircle filled size={16} />
```
dozortsev, line 131: I think here you could reuse `title` prop of icon component https://github.com/transferwise/icons#accessibility and delete this `span`

packages/components/src/fileUpload/uploadItem/UploadItem.tsx
czebe: Not at this time. `ProcessIndicator` is not accepting `className`.

packages/components/src/fileUpload/uploadItem/UploadItem.tsx
czebe: I don't think it returns `void` (tsc would complain if it couldn't infer the return type). But I've added it for clarity.

packages/components/src/fileUpload/uploadItem/UploadItem.tsx
czebe: This structure is needed to render the divider lines between list items. Merging them would make implementing separators more complicated.
image

packages/components/src/fileUpload/uploadButton/defaults.ts
czebe: This could be added later when any of the apps require it. At the moment we see APIs with one/multiple requirements only.

packages/components/src/fileUpload/FileUpload.tsx
czebe: I'm already using `uploadedFiles` const internally, so I kept the public prop name shorter on purpose.
(TS type makes it clear what's supposed to be set here anyways)

packages/components/src/fileUpload/FileUpload.story.tsx
czebe: Ok, removing the grouping. The issue with this is that internal/sub-components are not grouped together with the main component, and Storybook is not always alphabetical. So it's difficult to switch between components during development.

packages/components/package.json
```diff
@@ -27,23 +27,24 @@
...
     "test": "jest --env=jsdom --maxWorkers=4 --config=../../node_modules/@transferwise/test-config/jest.config.js",
     "test:watch": "jest --watch --env=jsdom --config=../../node_modules/@transferwise/test-config/jest.config.js",
     "lint": "jest --maxWorkers=4 --config ../../node_modules/@transferwise/test-config/test/jest.lint.js && yarn lintless",
     "test:visual": "build-storybook && percy-storybook --widths=1280",
-    "lint:fix": "eslint 'src/**/*.js' --fix",
+    "lint:fix": "eslint src --fix",
     "lintless": "stylelint --allow-empty-input './src/**/*.less'",
-    "lintless:fix": "yarn lintless --fix"
+    "lintless:fix": "yarn lintless --fix",
+    "check-types": "tsc --noEmit"
```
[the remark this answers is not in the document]
czebe: At the moment there was no way to lint TS files. So I added it there quickly.
But to make sense, I'm making a part of the main `lint` command, I forgot to do that.

Btw. What's the reason for using `jest` for linting instead of just `eslint`?

packages/components/package.json
```diff
@@ -27,23 +27,24 @@
...
+    "build:generate-types": "tsc --emitDeclarationOnly && node ./scripts/generate-type-declarations.js",
     "build:copy-files": "cpx 'src/**/!(db)/*.{css,json,svg}' build/es/polyfill & cpx 'src/**/!(db)/*.{css,json,svg}' build/es/no-polyfill",
     "build:copy-lang": "cpx 'src/i18n/*.json' build/i18n && cpx 'src/i18n/index.js' build/i18n",
     "docs": "yarn build",
     "test": "jest --env=jsdom --maxWorkers=4 --config=../../node_modules/@transferwise/test-config/jest.config.js",
     "test:watch": "jest --watch --env=jsdom --config=../../node_modules/@transferwise/test-config/jest.config.js",
     "lint": "jest --maxWorkers=4 --config ../../node_modules/@transferwise/test-config/test/jest.lint.js && yarn lintless",
     "test:visual": "build-storybook && percy-storybook --widths=1280",
-    "lint:fix": "eslint 'src/**/*.js' --fix",
+    "lint:fix": "eslint src --fix",
```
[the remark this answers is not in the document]
czebe: By default `eslint` only checks `js` files. Specifying the filetype is redundant here.

packages/docs/pages/components/inputs/FileUpload.mdx
```diff
@@ -0,0 +1,55 @@
+import { FileUpload, Status } from '@transferwise/components';
```
NewmanJWise, line 1: Are prop tables auto generated? Is it worth having one for things like file size?

czebe: PropTables generation fails for TS files, we're aware of the problem and hopefully will be fixed by `react-docgen` release.

packages/components/src/fileUpload/FileUpload.story.tsx
dozortsev: since `UploadButton` and `UploadItem` are internal components of `FileUpload` and not exposed from components package no need to create dedicated stories for them, I would just put them in `FileUpload`

Screen Shot 2021-06-25 at 1 39 58 PM

packages/components/src/fileUpload/FileUpload.tsx
dozortsev: ok

packages/components/src/fileUpload/uploadItem/UploadItem.tsx
dozortsev: ok

packages/components/src/fileUpload/uploadItem/UploadItem.tsx
dozortsev: well, it's totally fine to expose `className` in `ProcessIndicator` in order to archive this

dozortsev: re. `data-testid` prop I think you missed this https://github.com/transferwise/neptune-web/pull/1015#discussion_r656220072

packages/components/package.json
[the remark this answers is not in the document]
dozortsev: could you please still remove this from this PR?

I don't understand why we need `--emitDeclarationOnly` in `build:generate-types` command

packages/components/src/fileUpload/uploadItem/UploadItem.tsx
```diff
@@ -0,0 +1,139 @@
...
+            {getTitle()}
+          </div>
+          <small className="np-upload-description text-primary" data-testid={TEST_IDS.description}>
+            {getDescription()}
+          </small>
+        </div>
+        <div
+          className={classNames('np-upload-icon', { 'media-right': !isRTL, 'media-left': isRTL })}
+        >
+          {deleteFile && (!status || (status && !PROGRESS_STATUSES.includes(status))) && (
```
dozortsev, line 119: it's hard to understand this check, maybe doing like following would have same result?

```js
{deleteFile && !PROGRESS_STATUSES.includes(status)) && (
```

packages/docs/pages/components/inputs/FileUpload.mdx
```diff
@@ -0,0 +1,55 @@
...
+    .catch((e) => {
+      throw e;
+    });
+};
+```
+
+<GeneratePropsTable componentName="FileUpload" />
+
+export const meta = {
+    name: 'FileUpload',
```
dozortsev, line 54: please add badge that would indicate that this is new

```
badge: {
    type: 'new',
    expiryDate: '2021-07-31',
},
```

see example in [`CircularButton.mdx`](https://github.com/transferwise/neptune-web/blob/fileupload-component/packages/docs/pages/components/inputs/CircularButton.mdx)

packages/components/src/fileUpload/uploadItem/UploadItem.tsx
```diff
@@ -0,0 +1,139 @@
...
+            <button
+              data-testid={TEST_IDS.deleteButton}
+              onClick={() => deleteFile(file)}
+              type="button"
+              className="np-upload-remove-button p-a-1 btn btn-unstyled"
```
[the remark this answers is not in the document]
dozortsev: it should `--color-text-inactive` according to our [colors doc](https://transferwise.github.io/neptune-web/styles/tokens/Colors).

packages/components/src/fileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,184 @@
...
+   * @param formData
+   */
+  onCreateUploadRequest: (formData: FormData) => Promise<UploadResponse>;
+
+  /**
+   * Provide a callback if the file can be removed/deleted from the server
+   * Your app is responsible for reloading the uploaded files list and updating the component to ensure that the file has in fact been deleted successfully
+   * @param id
+   */
+  onCreateDeleteRequest?: (id: string | number) => Promise<any>;
```
dozortsev, line 43: maybe just `onFileUpload` and `onFileDelete`? I don't understand why we need "create" and "request" words here

packages/components/src/fileUpload/FileUpload.spec.tsx
```diff
@@ -0,0 +1,127 @@
...
+import FileUpload, { FileUploadProps } from './FileUpload';
+
+import { TEST_IDS as UPLOADBUTTON_TEST_IDS } from './uploadButton/UploadButton';
+import { TEST_IDS as UPLOADITEM_TEST_IDS } from './uploadItem/UploadItem';
+import { Status } from '../common';
+
+jest.mock('../common/hooks/useDirection/useDirection');
+
+jest.mock('react-intl', () => ({
+  useIntl: () => ({ formatMessage: ({ id }: MessageDescriptor) => `${id}` }),
```
NewmanJWise, line 17: This could just return id as we're not concatenating anything here?

packages/components/src/fileUpload/uploadItem/UploadItem.tsx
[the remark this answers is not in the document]
czebe: good catch, I didn't notice this one.

packages/components/src/fileUpload/uploadItem/UploadItem.tsx
czebe: Simplified to (the best we can do without TS errors):
`deleteFile && (!status || !PROGRESS_STATUSES.includes(status)) &&`

packages/components/src/fileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,184 @@
...
+} & Pick<UploadButtonProps, 'disabled' | 'multiple' | 'fileTypes' | 'sizeLimit'> &
+  CommonProps;
+
+const FileUpload: FC<FileUploadProps> = ({
```
dozortsev, line 47: we tend to avoid using `FC` ([see for more details](https://github.com/facebook/create-react-app/pull/8177)) this component don't need `children` prop

```ts
function UploadInput(): ReactElement {
```

packages/components/src/fileUpload/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default } from './FileUpload';
+export * from './types';
```
dozortsev, line 2:
```ts
export { default } from './UploadInput';
```

czebe (commented):
foo

packages/components/src/fileUpload/uploadItem/UploadItem.tsx
czebe: I've exposed it, but it falls apart without the extra wrapping container (due to `media-left` `padding-right` throwing off the process indicator elements)

packages/components/src/uploadInput/uploadItem/UploadItem.spec.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+import UploadItem, { UploadItemProps, TEST_IDS } from './UploadItem';
+import { Status } from '../../common';
+
+import MESSAGES from './UploadItem.messages';
+
+jest.mock('../../common/hooks/useDirection/useDirection');
+
+jest.mock('react-intl', () => ({
+  useIntl: () => ({ formatMessage: ({ id }: MessageDescriptor) => `${id}` }),
```
NewmanJWise, line 14: This one can also just be id as it is not concatenating

packages/components/src/uploadInput/uploadItem/UploadItem.tsx
```diff
@@ -0,0 +1,143 @@
...
+      case Status.DONE:
+      default:
+        return <span className="text-success">{formatMessage(MESSAGES.uploaded)}</span>;
+    }
+  };
+
+  const getTitle = () => {
+    const title = filename || formatMessage(MESSAGES.uploadedFile);
+    return url ? (
+      <a href={url} target="_blank">
```
NewmanJWise, line 93: From the convo on Slack it looks like this might not be needed?
https://transferwise.slack.com/archives/C020UEBLX3J/p1625587330075800?thread_ts=1625585519.075400&cid=C020UEBLX3J

packages/components/src/uploadInput/UploadInput.tsx
```diff
@@ -0,0 +1,186 @@
...
+      const formData = new FormData();
+
+      if (file) {
+        const { name, size } = file;
+        const id = `${name}_${size}`;
+
+        const allowedFileTypes = Array.isArray(fileTypes) ? fileTypes.join(',') : fileTypes;
+
+        // Check if file type is valid
+        if (!isTypeValid(file, allowedFileTypes)) {
```
matebek-tw, line 113: Not sure if it is intentional or not, but without passing down the base64 version of the file we can't actually tell if the mime type is matching the extension / file type determined by the browser. I am not a huge fan of doing these checks in the browser anyway, but the previous `Upload` component was implemented with that in mind if you check out the shape of the `isTypeValid` and `getFileType` functions (though because we are early returning if we have a `file.type` it doesn't really matter). I am happy to put this burden on the backend, but in that case we could simplify these two methods (can be done in a future commit).

packages/components/src/uploadInput/UploadInput.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+  /**
+   * The title of the confirmation modal on delete
+   */
+  deleteModalTitle?: string;
+
+  /**
+   * The body copy of the confirmation modal on delete
+   */
+  deleteModalBody?: string;
```
NewmanJWise, line 54: Is it possible to swap this from being a string to also accepting a HTML element. In our design we have a logo plus  the copy.

image

packages/components/src/uploadInput/UploadInput.less
```diff
@@ -0,0 +1,6 @@
+@import (reference) '@transferwise/neptune-tokens/tokens.less';
+
+.np-upload-input {
+  border: 1px solid @color-border-neutral;
+  border-radius: 4px;
+}
```
dozortsev, line 6: for neptune-tokens we should use CSS custom properties syntax (`var(--*)`) even at Less files

packages/components/src/uploadInput/uploadButton/UploadButton.less
```diff
@@ -0,0 +1,57 @@
+@import (reference) '@transferwise/neptune-tokens/tokens.less';
+
+.np-upload-button {
```
dozortsev, line 3: please use CSS custom properties syntax for tokens in this file

packages/components/src/uploadInput/uploadItem/UploadItem.tsx
```diff
@@ -0,0 +1,131 @@
...
+};
+
+export enum TEST_IDS {
+  uploadItem = 'uploadItem',
+  mediaBody = 'mediaBody',
+}
+
+const PROGRESS_STATUSES = [Status.PENDING, Status.PROCESSING];
+
+const UploadItem: FC<UploadItemProps> = ({ file, onDelete }: UploadItemProps) => {
```
dozortsev, line 29: do we need `children` prop for this component? let's not use `FC` here

packages/components/src/uploadInput/UploadInput.story.tsx
```diff
@@ -0,0 +1,130 @@
...
+    );
+  });
+
+const props = {
+  onUploadFile: (formData: FormData) => {
+    alert(`Uploading: ${(formData.get('file') as File).name}`);
+    return createDelayedPromise();
+  },
+  onDeleteFile: (id: string | number) => {
+    alert(`File with ${id} deleted`);
```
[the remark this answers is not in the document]
matebek-tw: I believe they are there to demonstrate you can tap into the delete API call, but I agree it can be removed.

packages/components/src/ssr.spec.js
```diff
@@ -37,6 +37,7 @@ const excluded = [
   'DecisionType',
   'AlertArrowPosition',
   'LogoType',
+  'FileType',
```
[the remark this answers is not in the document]
matebek-tw: Those are not exported anymore from `./uploadInput`

packages/components/src/uploadInput/types.ts
```diff
@@ -0,0 +1,65 @@
...
+   */
+  [key: string]: any;
+};
+
+export enum FileType {
+  PDF = '.pdf,application/pdf',
+  JPG = '.jpg,.jpeg,image/jpeg',
+  PNG = '.png,image/png',
+  CSV = '.csv,text/csv',
+}
```
dozortsev, line 65: this enum might be useful somewhere else as well, could you please move it in dedicated file at `src/common/fileType.ts`?

packages/components/src/uploadInput/uploadButton/UploadButton.spec.tsx
```diff
@@ -0,0 +1,106 @@
...
+import UploadButton, { UploadButtonProps, TEST_IDS } from './UploadButton';
+import MESSAGES from './UploadButton.messages';
+import { imageFileTypes } from './defaults';
+
+jest.mock('../../common/hooks/useDirection/useDirection');
+
+jest.mock('react-intl', () => ({
+  useIntl: () => ({ formatMessage: ({ id }: MessageDescriptor) => `${id}` }),
+  defineMessages: (translations: Record<string, string>[]) => translations,
+}));
```
dozortsev, line 16: no need to mock these, use custom render method from `test-utils/index.js`, see examples in any other RTL tests

packages/components/src/uploadInput/uploadItem/UploadItem.spec.tsx
```diff
@@ -0,0 +1,135 @@
...
+import { Status } from '../../common';
+
+import MESSAGES from './UploadItem.messages';
+
+jest.mock('../../common/hooks/useDirection/useDirection');
+
+jest.mock('react-intl', () => ({
+  useIntl: () => ({ formatMessage: ({ id }: MessageDescriptor) => id }),
+  defineMessages: (translations: Record<string, string>[]) => translations,
+}));
```
dozortsev, line 16: same here, no need to mock them

packages/docs/pages/components/inputs/UploadInput.mdx
```diff
@@ -0,0 +1,59 @@
+import { UploadInput, Status } from '@transferwise/components';
+import { LiveEditorBlock, GeneratePropsTable } from '../../../utils';
+import code from '../../../liveEditorCode/uploadInput.code';
+
+Upload one or more files and display the list of already uploaded file with optional link and delete functionality.
+
+Actual network request can be performed by the API library of your choice (`fetch/axios/api-call`). Perform the form submission for each file individually inside `onFileUpload(formData)` and return a `Promise` with the result that must contain an `id` and optionally `url`, `error` fields.
```
dozortsev, line 7: no need to mention here `api-call` as it's Wise private library

packages/docs/pages/components/inputs/UploadInput.mdx
```diff
@@ -0,0 +1,59 @@
+import { UploadInput, Status } from '@transferwise/components';
+import { LiveEditorBlock, GeneratePropsTable } from '../../../utils';
+import code from '../../../liveEditorCode/uploadInput.code';
+
+Upload one or more files and display the list of already uploaded file with optional link and delete functionality.
```
dozortsev, line 5:
```suggestion
An upload input lets users upload a single file, or multiple files, within a form. See [usage guidelines](https://transferwise.github.io/neptune/components/upload-input/) for more info.
```

dozortsev (approved):
Looks good, left some minor comments.

PR #1019, opened 2021-06-21
Refactor final step actions

JulianG-TW:
## 🖼 Context

The specification was updated. the code needs to change.

## 🚀 Changes

Before there was a single action object inside a step details:

```
{
  "type" : "final",
  "details: {
    ...,
    "action" : { ... }
  },
}
```

Now, final step actions are declared like this:

```
{
  "type" : "final",
  "details: { ...},
  "actions": [ ... ]
}
```

Also, `DynamicLayout` component does not convert a final step into layout components anymore. Consumers need to convert their final step into layout using the now exported `convertStepToLayout` function. An example was added to **DynamicLayout.story.js**.

These are both BREAKING changes.

## ✅ Checklist


packages/dynamic-flows/src/flow/layoutService/LayoutService.spec.js
```diff
@@ -69,22 +69,20 @@ describe('Given a utility service for handling dynamic layouts', () => {
         title: 'Exit',
         type: 'success',
         exit: true,
-        data: {
-          exitValue: 'value',
-        },
+        data: 'exit value',
```
JulianG-TW, line 72: This change is not really needed. I chose to remove the `exitValue` after seeing it being used as part of the spec.

packages/dynamic-flows/src/layout/DynamicLayout.story.js
```diff
@@ -2,11 +2,12 @@ import React from 'react';
...
-
-import allComponents from './layouts/all.json';
-import reviewStep from './layouts/review.json';
-import successStep from './layouts/success.json';
+import { convertStepToLayout } from '../flow/layoutService';
+import allComponentsLayout from './layouts/all.json';
+import reviewStepLayout from './layouts/review.json';
+import successStepLayout from './layouts/success.json';
 import finalStep from './layouts/finalStep.json';
+import finalStepLayout from './layouts/final-step-layout.json';
```
JulianG-TW, line 10: renamed some of these imports, to differentiate when the object is a step or a layout.

packages/dynamic-flows/src/layout/DynamicLayout.story.js
```diff
@@ -15,13 +16,14 @@ export default {
...
   const layouts = {
-    'All components': allComponents,
-    'Review step': reviewStep,
-    'Success step': successStep,
-    'Final step': finalStep,
+    'All components layout': allComponentsLayout,
+    'Review step layout': reviewStepLayout,
+    'Success step layout': successStepLayout,
+    'Final step layout': finalStepLayout,
+    'Final step (converted)': convertStepToLayout(finalStep),
```
JulianG-TW, line 23: Here's how consumers can convert step to layout.

packages/dynamic-flows/src/flow/layoutService/LayoutService.spec.js
StevePole: I would prefer to keep this as an object.  I believe it is an object in the spec.

packages/dynamic-flows/src/layout/layouts/finalStep.json
```diff
@@ -1,25 +1,23 @@
...
+    },
+    "title": "Details verified *",
+    "description": "We do this extra check to make sure your money goes to the right place...."
+  },
+  "actions": [
+    {
+      "title": "Continue",
+      "type": "success",
+      "exit": true,
+      "data": "exit value",
```
StevePole, line 18: Same thing, maybe better to keep as object?

StevePole (approved):
Couple minor thoughts, but not critical 👍

packages/dynamic-flows/src/layout/layouts/finalStep.json
JulianG-TW: Yes! I confused "data" with "result". "result" can be anything. "data" is always an object. I don't think we have examples with "result".

I changed it to:
```
"data": { "someKey": "some value" }
```

packages/dynamic-flows/src/flow/layoutService/LayoutService.spec.js
JulianG-TW: Yes.

PR #1022, opened 2021-06-22
Align DynamicButton props to Button props

JulianG-TW:
## 🖼 Context

`DynamicButton` is relying only on `type`, but the latest Neptune components `Button` [expects both `type` and `priority`](https://transferwise.github.io/neptune-web/components/inputs/Button).

This resulted in buttons rendering with `secondary` priority by default.

## 🚀 Changes

Align DynamicButton props to Button props.

## 🤔 Considerations

This is a breaking change.

## ✅ Checklist

### DIFF
--- a/packages/dynamic-flows/src/layout/button/index.js
+++ b/packages/dynamic-flows/src/layout/button/index.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import { Button } from '@transferwise/components';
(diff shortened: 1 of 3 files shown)


StevePole (changes requested):
This is incorrect, `priority` is not part of the dynamic layout spec, and is not part of our action model.  That's why we had the two mapping functions there.  

I don't think we should add it to the spec, it's a non obvious thing for partners, it's very specific decision in our design system.

PR #1023, opened 2021-06-23
feat(BottomSheet): design updates (for cross platform alignment)

dozortsev:
## 🖼 Context

We align BottomSheet component cross platform, in this PR we mostly add necessary changes for Web component.

<details>
  <summary>Summary of new design requirements</summary>

- Implement drag and scroll behaviour - The bottom sheet will hug the contents inside it and automatically grow in height as the contents grows. Once it’s reached its full height, if the content inside it overflows it will become scrollable. In this case, the bottom sheet can only be dragged down once the content is scrolled to the top. 

- Add a 64px safe zone at the top of the screen when the bottom sheet is fully expanded

- Add a drop shadow to elevate the top of the component when the content is scrolling. The drop shadow should only appear when the content is scrolling/has scrolled, not before. Use the same drop shadow as on the popover

- Add a handle — make it 40x4px

- Add an invisible close button screen reader users can tab to

- Make the dimmer black with 40% transparency

</details>

## 🚀 Changes

- Design updates for `BottomSheet`
- Fixed bug of disabling scroll for `Dimmer`
- Integrated BottomSheet inside `DateLookup` and `Select`
- Chores: migrated `BottomSheet` component to TypeScript, exposed the component from the `@transferwise/components` packages

## 🔍  Demo

You can play with the component on following pages/components, please open them on **mobile view 📱** 

- [BottomSheet](https://transferwise.github.io/neptune-web/branch/bottom-sheet/components/overlays/BottomSheet)
- See new component inside [DateLookup](https://transferwise.github.io/neptune-web/branch/bottom-sheet/components/inputs/DateLookup) vs [old one](https://transferwise.github.io/neptune-web/branch/bottom-sheet/components/inputs/DateLookup)
- [Popover](https://transferwise.github.io/neptune-web/branch/bottom-sheet/components/alerts/Popover) vs [old one](https://transferwise.github.io/neptune-web/components/alerts/Popover)

## ✅ Checklist


packages/components/src/common/DOMOperations/DOMOperations.js
```diff
@@ -1,9 +1,16 @@
 const NO_SCROLL_BODY_CLASS = 'no-scroll';
 
+/**
+ * Applying styles that disable scrolling for overlay elements
+ *
+ * See: https://stackoverflow.com/a/28411556/2650174
+ */
 export const addNoScrollBodyClass = () => {
```
dozortsev, line 8: we have a bug at Dimmer allowing to scroll background when overlay open (e.g Modal, BottomSheet) is opened, in order to fix that we should apply styles that disable scrolling to `html` and `body` tags

packages/components/src/common/bottomSheet/BottomSheet.less
```diff
@@ -1,4 +1,34 @@
...
+  &--handler {
+    width: 40px;
+    height: 4px;
+    background: var(--color-border-neutral);
+    border-radius: var(--radius-small);
+  }
+
+  &--close-btn {
+    position: absolute;
+    right: 8px;
```
matebek-tw, line 25: Shouldn't we use `var(--size-8)` here?

packages/components/src/common/bottomSheet/BottomSheet.spec.js
```diff
@@ -3,29 +3,18 @@ import { render } from '../../test-utils';
...
       </BottomSheet>,
     );
-
-    expect(container).toMatchSnapshot();
+    expect(baseElement).toMatchSnapshot();
   });
 
-  it("doesn't renders content when closed", async () => {
+  it('does not renders when closed', () => {
     const { container } = render(<BottomSheet onClose={jest.fn()}>content</BottomSheet>);
```
matebek-tw, line 17: Can we be explicit here and set `open={false}`? You could rename the file to `.ts` to highlight these errors.

packages/components/src/common/bottomSheet/BottomSheet.tsx
```diff
@@ -0,0 +1,185 @@
...
+   * @see {@link calculateOffsetAfterMove}
+   */
+  const moveOffset = useRef<number>(0);
+  const initialYCoordinate = useRef<number>(0);
+
+  // apply shadow to the bottom of top-bar when scroll over content
+  useConditionalListener({
+    attachListener: props.open && !isServerSideRandering(),
+    callback: () => {
+      if (topBarRef.current !== null) {
```
matebek-tw, line 57: You could remove one indentation layer by early returning from the function.

```javascript
      if (topBarRef.current === null) {
        return;
      }

      const { classList } = topBarRef.current;

      if (!isContentScrollPositionAtTop()) {
        classList.add('np-bottom-sheet--top-bar--shadow');
      } else {
        classList.remove('np-bottom-sheet--top-bar--shadow');
      }
```

packages/components/src/common/bottomSheet/BottomSheet.tsx
```diff
@@ -0,0 +1,185 @@
...
+   * and new position (when user moves component), it's get calculated on `onTouchMove` and `onMouseMove` events
+   *
+   * @see {@link calculateOffsetAfterMove}
+   */
+  const moveOffset = useRef<number>(0);
+  const initialYCoordinate = useRef<number>(0);
+
+  // apply shadow to the bottom of top-bar when scroll over content
+  useConditionalListener({
+    attachListener: props.open && !isServerSideRandering(),
```
matebek-tw, line 55: `isServerSideRendering`

packages/components/src/common/domHelpers/index.ts
```diff
@@ -1,3 +1,7 @@
+export function isServerSideRandering(): boolean {
```
matebek-tw, line 1:
```suggestion
export function isServerSideRendering(): boolean {
```

packages/components/src/common/bottomSheet/BottomSheet.tsx
```diff
@@ -0,0 +1,185 @@
...
+  });
+
+  function move(newHeight: number): void {
+    if (bottomSheetRef.current) {
+      bottomSheetRef.current.style.transform = `translateY(${newHeight}px)`;
+    }
+  }
+
+  function close(): void {
+    if (bottomSheetRef.current !== null) {
```
matebek-tw, line 77: Again, I think early returning and having less indentation levels simplifies the code.

```javascript
if (bottomSheetRef.current === null) {
  return;
}
```

packages/components/src/common/bottomSheet/BottomSheet.tsx
dozortsev: I tried, but I get warning from TS

> Property 'classList' does not exist on type 'null'.

Screen Shot 2021-07-27 at 2 40 40 PM

packages/components/src/common/bottomSheet/BottomSheet.tsx
dozortsev: we don't have multiple nested `if`s here,

I don't think following code is somehow better than current

```ts
    if (bottomSheetRef.current === null) {
      return;
    }
    bottomSheetRef.current.style.removeProperty('transform');
    if (props.onClose) {
      props.onClose();
    }
```

packages/components/src/common/bottomSheet/BottomSheet.tsx
matebek-tw: You should return when `topBarRef.current === null`, so in the lines after you can be sure that it is **not** `null` and TS will be happy.

matebek-tw (approved):
Looking good, few nitpicks only.

PR #1025, opened 2021-06-25
feat(Select): RTL support

joe-sayer:
## 🖼 Context

This PR is necessary in order for Select to support RTL locales.

## 🚀 Changes

 
- Adjusted styling

## ✅ Checklist


packages/css/src/less/select.less
```diff
@@ -62,6 +62,34 @@
...
+
+      &.btn-sm.dropdown-toggle {
+        padding-right: 12px;
+      }
+
+      &.btn-md {
+        padding-right: 16px;
+      }
+
+      &.btn-lg {
```
joe-sayer, line 83: What is the best way to target both `btn-md` and `btn-lg` as they both provide `padding-right: 16px;`?

packages/components/src/select/Select.js
```diff
@@ -430,10 +432,13 @@ export default class Select extends Component {
     const { open, shouldRenderWithPortal } = this.state;
     const s = this.style;
 
+    const isRTL = this.context === 'rtl';
```
dozortsev, line 435: I think we should have an enum for these values (in [`propsValues/`](https://github.com/transferwise/neptune-web/tree/main/packages/components/src/common/propsValues)), and use it instead of raw strings

```ts
enum Direction {
  LEFT_TO_RIGHT = 'ltr',
  RIGHT_TO_LEFT = 'rtl'
}
```

packages/css/src/less/select.less
dozortsev: I guess like this, no? 😄 

```less
&.btn-md, &.btn-lg {
```

dozortsev (approved):

packages/css/src/less/select.less
joe-sayer: Thank you :D

PR #1028, opened 2021-06-30
feat: add Link component (navigate away)

andreapiras:
## 🖼 Context

This PR introduces Link component
## 🚀 Changes

 
Add Link component

## ✅ Checklist


packages/components/src/link/Link.tsx
```diff
@@ -0,0 +1,77 @@
...
+              {
+                'm-l-05': isSmall,
+                'm-l-025': !isSmall,
+              },
+              className,
+            )}
+          />
+        )}
+      </a>
+      {fullStop && (
```
andreapiras, line 59: I added the logic to handle the fullStop. The reason why is handled by link is that the full stop needs to be after the icon for target blank.
When using the icon we need to adjust the margin for large version because the icon itself has an embedded margin
image
vs 
image

packages/components/src/link/Link.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { useIntl } from 'react-intl';
+import { NavigateAway as NavigateAwayIcon } from '@transferwise/icons';
+
+import messages from './Link.messages';
+import './Link.css';
+
+type LinkProps = {
+  'aria-label'?: string;
+  className?: string;
+  content: string;
```
dozortsev, line 13: hmm, I think this should be a `children` property and it should be `ReactNode` type of, to make this component API as similar to native link element `a`

packages/components/src/link/Link.tsx
```diff
@@ -0,0 +1,43 @@
...
+  href: string;
+  target?: '_blank';
```
dozortsev, line 15: what about others `_self` `_parent` [etc](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-target)?

packages/components/src/link/Link.tsx
```diff
@@ -0,0 +1,43 @@
...
+    ? intl.formatMessage(messages.ariaLabelNewTab)
+    : intl.formatMessage(messages.ariaLabel);
+
+  return (
+    <>
+      <a
+        href={href}
+        target={target || undefined}
+        className={classNames(`np-link d-inline-flex`, className)}
+        aria-label={props['aria-label'] || ariaLabelFallBack}
```
dozortsev, line 33: I would suggest extract this logic and logic of `ariaLabelFallBack` into one function and just invoke it here

```jsx
aria-label={getAriaLabel()}
```

packages/components/src/link/Link.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  const ariaLabelFallBack = isBlank
+    ? intl.formatMessage(messages.ariaLabelNewTab)
+    : intl.formatMessage(messages.ariaLabel);
+
+  return (
+    <>
+      <a
+        href={href}
+        target={target || undefined}
```
dozortsev, line 31: since `target` is optional prop you should specify default value for it, and just pass it here without `||`

packages/components/src/ssr.spec.js
```diff
@@ -154,6 +154,7 @@ describe('Server side rendering', () => {
       media: <img alt="img" />,
     },
     Modal: { position: 'top' },
+    Link: { size: 16 },
```
dozortsev, line 157: I guess this is redundant

packages/docs/pages/components/content/Link.mdx
```diff
@@ -0,0 +1,11 @@
+import { LiveEditorBlock, GeneratePropsTable } from '../../../utils';
+import { Link } from '@transferwise/components';
+
+import code from '../../../liveEditorCode/link.code';
+
+<LiveEditorBlock code={code} scope={{ Link }} />
+<GeneratePropsTable componentName="Link" />
+
+export const meta = {
+  name: 'Link',
```
dozortsev, line 10: please add badge

```js
  badge: {
    type: 'new',
    expiryDate: '2021-04-30',
  },
```

packages/components/src/link/Link.messages.js
```diff
@@ -0,0 +1,14 @@
...
+
+export default defineMessages({
+  ariaLabel: {
+    id: 'neptune.Link.ariaLabel',
+    defaultMessage: 'Open link',
+    description: 'Description for navigate away link',
+  },
+  ariaLabelNewTab: {
+    id: 'neptune.Link.ariaLabelNewTab',
+    defaultMessage: 'Open link in a new tab',
```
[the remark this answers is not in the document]
andreapiras: @StephanieTW

packages/components/src/link/Link.tsx
andreapiras: I agree with children, should we leave string for now, we don't want people to pass what they want here

packages/components/src/link/Link.tsx
andreapiras: I guess they can be added as needed.

packages/components/src/link/Link.tsx
andreapiras: I think this is more readable

packages/components/src/link/Link.messages.js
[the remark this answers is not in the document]
StephanieTW: Yes, ideally teams will replace the default copy with something more specific about where the link takes the user. But for default copy we talked about having 'Open link' for standard links, and 'Open link in a new tab' for those that navigate away to a new tab. 

Did you have a suggestion Anton?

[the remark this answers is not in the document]
dozortsev: I'm fine with these messages as long as you reviewed them

packages/components/src/link/Link.tsx
andreapiras: I removed the || and if target is not defined the attribute won't appear in the hmtl

packages/components/src/link/Link.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { useIntl } from 'react-intl';
+import { NavigateAway as NavigateAwayIcon } from '@transferwise/icons';
+import { CommonProps } from '../common';
+
+import messages from './Link.messages';
+import './Link.css';
+
+type LinkProps = {
+  'aria-label'?: string;
+  children: string;
```
dozortsev, line 13: could you please mark `children` as `ReactNode`? I'm sure there will be pretty legit cases where consumer would need pass node there

note: `children` can only be node type, even if set is to be `string` it still allows nodes, because of they way how we use it `<a>{children}</a>`

Screen Shot 2021-07-05 at 2 16 34 PM

packages/components/src/link/Link.tsx
dozortsev: okey

packages/components/src/link/Link.tsx
dozortsev: I mean currently aria label logic is in two places, while I'm suggesting put it into one:

```ts
  function getAriaLabel() {
    const ariaLabel = props['aria-label'];
    if (ariaLabel) {
      return ariaLabel;
    } else {
      return isBlank
      ? intl.formatMessage(messages.ariaLabelNewTab)
      : intl.formatMessage(messages.ariaLabel);
    }
  }
}
```

and keep JSX slightly cleaner

```jsx
aria-label={getAriaLabel()}
```

packages/components/src/link/Link.tsx
```diff
@@ -0,0 +1,41 @@
...
+  href: string;
+  target?: '_blank';
+} & CommonProps;
+
+export const Link = (props: LinkProps) => {
```
dozortsev, line 18: please add JSDoc comment with link to docs ([example](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/loader/Loader.tsx#L20-L25))

```
/**
 * 
 *
 * Docs link: https://transferwise.github.io/neptune-web/components/content/Link
 */
```

packages/components/src/link/Link.story.js
```diff
@@ -0,0 +1,38 @@
...
+
+export const basic = () => {
+  return (
+    <>
+      <p>
+        Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has
+        been the industry standard dummy text ever since the 1500s,&nbsp;
+        <Link
+          aria-label="Click here for more details"
+          content="content"
```
dozortsev, line 18: please delete redundant `content` prop

packages/components/src/link/Link.story.js
```diff
@@ -0,0 +1,38 @@
...
+        >
+          content
+        </Link>
+        . &nbsp;Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem
+        Ipsum has been the industry standard dummy text ever since the 1500s
+      </p>
+      <p>
+        Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has
+        been the industry standard dummy text ever since the 1500s,&nbsp;
+        <Link href="/test" content="content">
```
dozortsev, line 30: same here

dozortsev (approved):

PR #1038, opened 2021-07-06
chore: update storybook

matebek-tw:
## 🖼 Context

Using `react-docgen` to generate the documentation for components written in TypeScript has a known issue where it is not including imported types in the props table: https://gist.github.com/shilman/036313ffa3af52ca986b375d90ea46b0#short-term

Removing the custom `reactDocgen` configuration will let `storybook` to use their default `react-docgen-typescript` to extract props information correctly: https://gist.github.com/shilman/036313ffa3af52ca986b375d90ea46b0#react-docgen-missing-imported-type-support

## 🚀 Changes

- Upgrade `storybook` and its dependencies to the latest version
- Remove custom `reactDocgen` config

## 🤔 Considerations

Our docs are still using `react-docgen` to generate the documentation, which even on the latest alpha release is unable to extract props information from imported types. 😞 We might change the whole implementation of our doc generation to use [react-docgen-typescript](https://github.com/styleguidist/react-docgen-typescript) as well, or wait until they fix this issue on their end.

https://github.com/reactjs/react-docgen/blob/master/README.md#flow-and-typescript-support
> Note: react-docgen will not be able to grab the type definition if the type is imported or declared in a different file.

## ✅ Checklist


packages/components/.storybook/main.js
```diff
@@ -17,9 +17,6 @@ module.exports = {
     '@storybook/addon-viewport',
     '@pxblue/storybook-rtl-addon/register',
   ],
-  typescript: {
-    reactDocgen: 'react-docgen',
-  },
```
dozortsev, line 22: we should do same for Storybook setup up of dynamic-flows package

dozortsev (approved):

packages/components/package.json
```diff
@@ -64,12 +64,12 @@
...
-    "@storybook/addon-docs": "^6.2.9",
-    "@storybook/addon-knobs": "^6.2.9",
-    "@storybook/addon-viewport": "^6.2.9",
-    "@storybook/react": "^6.2.9",
+    "@storybook/addon-a11y": "^6.3.2",
+    "@storybook/addon-actions": "^6.3.2",
+    "@storybook/addon-docs": "^6.3.2",
+    "@storybook/addon-knobs": "^6.3.0",
+    "@storybook/addon-viewport": "^6.3.2",
+    "@storybook/react": "^6.3.2",
```
dozortsev, line 72: I think we should have these dependencies defined in one place, I would consider to move them into root `package.json`

PR #1041, opened 2021-07-09
fix: prevent DefinitionList from throwing error when null passed in

NewmanJWise:
## 🖼 Context

GPAID-118

This is a small proposal PR to make it so DefinitionList can destructure safely knowing that whatever it is destructing is not null. 

This is useful on pages where you might sometimes optionally not display some definitions. Currently we have to do something like this for optional definitions which just feels not awesome: (it looks a lot worse when there are multiple of them)

```
[
    ...(someObject.reference
      ? [
          {
            title: 'Reference',
            value: someObject.reference,
            key: 'reference',
          },
        ]
      : []),
]
```

## 🚀 Changes

 

Added a tweak filter to ensure that the definitions passed in are 

## 🤔 Considerations

As mentioned above this is just a proposal to make it easier for the consumer to use the component as I would normally except a component using an array to handle a null item in the array without exploding. 

I'm happy not bringing this forward but thought it was worth the discussion :) 

## ✅ Checklist


packages/components/src/definitionList/DefinitionList.spec.js
```diff
@@ -13,7 +13,7 @@ const {
 } = Layout;
 
 describe('DefinitionList', () => {
-  const someDefinitions = () => [{ title: 'First', value: 'first value', key: 'first' }];
+  const someDefinitions = () => [{ title: 'First', value: 'first value', key: 'first' }, null];
```
NewmanJWise, line 16: No need to add additional tests. “hasClass" would blow up in the below tests if a second node was found and previously the component threw an exception when null was passed in but now it works 🎉

helendurranttw (approved):
🚀

hendrikmolder (approved):
👏

oliverviljamaa (approved):
Lovely!

PR #1042, opened 2021-07-12
feat: Emphasis component

phildobsontw:
## 🖼 Context
Adding the 'Emphasis' component.

Docs:
image

Storybook:
image

## ✅ Checklist


packages/components/src/emphasis/Emphasis.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import createEmphasisHtmlTransformer from './createEmphasisHtmlTransformer';
+
+import './Emphasis.css';
+
+const SUPPORTED_TAGS = [
+  { name: 'br', replaceWithSpan: false },
+  { name: 'b', replaceWithSpan: false },
```
matebek-tw, line 9: I am not sure if we should allow HTML tags at all in this component, that raises the question why don't we allow `<strong />` or `<em />` or any other tag which can style text. That would even simplify the implementation as `replaceWithSpan` will always be `true`.

packages/components/scripts/generate-type-declarations.js
```diff
@@ -36,6 +36,7 @@ const jsFiles = [
   { filename: 'Upload', folder: 'upload' },
   { filename: 'DefinitionList', folder: 'definitionList' },
   { filename: 'DynamicFieldDefinitionList', folder: 'dynamicFieldDefinitionList' },
+  { filename: 'Emphasis', folder: 'emphasis' },
```
dozortsev, line 39: we can help write this component on TypeScript

packages/components/src/emphasis/Emphasis.less
```diff
@@ -0,0 +1,19 @@
+.emphasis {
+  font-weight: 600;
```
dozortsev, line 2: here we should use CSS custom property from neptune-tokens -> `var(--font-weight-semi-bold)`

packages/components/src/emphasis/Emphasis.js
phildobsontw: Thanks for pointing out - we are using `<b>` at the moment in our BFF, but I think we can swap it directly for `<important>`. I'm just double checking this - if so, i'll remove both html tags and we'll always replace with spans as you say! 👌

packages/components/src/emphasis/Emphasis.js
```diff
@@ -0,0 +1,35 @@
...
+  { name: 'important', replaceWithSpan: true },
+  { name: 'positive', replaceWithSpan: true },
+  { name: 'negative', replaceWithSpan: true },
+  { name: 'warning', replaceWithSpan: true },
+];
+const emphasisTransform = createEmphasisHtmlTransformer(SUPPORTED_TAGS);
+
+const Emphasis = ({ text }) => {
```
dozortsev, line 17: I think this component should also work with client side messages, could you please add a few examples

```
<Emphasis text={intl.formatMessage('some.key')} />

<p>Some very <Emphasis text={'<important>value</important>'} />.</p>
```

dozortsev: I just want to showcase to consumers that this component not only for 'rendering messages from backend' use case but could be use for local copies as well, make sense?

packages/components/src/emphasis/Emphasis.js
phildobsontw: Have removed it and it's simplified the implementation nicely, thanks for pointing out! 🙇

matebek-tw: Thanks, looking great!

packages/components/src/emphasis/Emphasis.js
phildobsontw: Thanks - have tried to make this clearer by adding a paragraph to the docs, and an extra example 👌.

packages/components/src/emphasis/EmphasisHtmlTransformer.ts
```diff
@@ -0,0 +1,58 @@
...
+          closing: new RegExp(`${ESCAPED_OPENING_CHEVRON}/${tag}${ESCAPED_CLOSING_CHEVRON}`, 'g'),
+        },
+      };
+    });
+  }
+
+  // Algorithm:
+  // 1) Escape all dangerous characters (<,>,&)
+  // 2) Replace all escaped, whitelisted tags with styled spans.
+  // 3) Transform new line characters to `<br />`s
```
dozortsev, line 36: could you please remove this feature, we think this component should just highlight copy not format it

phildobsontw: Although i do agree in principle, removing this would be a problem for us I'm afraid. 😞 
Just to explain the problem it solves - it means we avoid following this sort of pattern in our translated strings:
```
{
  xyzCopyLine1: "This is the first line"
  xyzCopyLine2: "This is the second line"
}
```
The problem with the above is that we become coupled to a specific number of lines of text. If we were to add a new third line, we'd have to make changes across all 3 platforms (Web/IOS + Android).
```
{
  xyzCopy: "This is the first line/nThis is the second line/nThis is a new third line added without having to make changes to IOS, Android and Web"
}
```
Hopefully that makes a bit of sense. This is already in the scope of the Android/IOS components and so we're looking for feature parity here.

For me - the actual problem here might be the name of the component. `FormattedText` for me might have been the better option, but I understand the other platforms have already standardised on `Emphasis`.

dozortsev (approved):

giljulio:
I see a few issues allowing `br` tag:

- if this is seen as a general text formatter, why not allow bullet points, numbered lists, italics, links. It quickly goes down a rabbit hole of basically supporting markdown
- AFAIK using br is less preferable than using p tags
- It forces mobile to use web tags, this is meant to be agnostic

dozortsev:
@giljulio it doesn't allow/support `br` tag in the messages, it won't transform it into a newline, 

it's support `\n`, which transforms into `<br>` to make a newline in HTML, so solution stays platform agnostic

aren't Android and iOS support `\n` as well?

giljulio:
Ah ok I thought we were still supporting br tag. Ok no worries ignore my comment 👌

PR #1044, opened 2021-07-16
Add SwitchOption component

mikknarusk:
## 🖼 Context

https://transferwise.atlassian.net/browse/SEND-1856 adding new stuff to settings which requires a list of switches
Designs here: https://www.figma.com/file/r0bTJPZ16amp33hNefBnEZ/Discoverability-opt-in?node-id=1575%3A37795

## 🚀 Changes

New feature:
- Added a new SwitchOption component

Some chores due to adding this component in TS:
- Adds Eslint support for ts/tsx files (and `lint:fix` existing files)
- Adds tsx support for stories
- Refactor Switch into TS
- Remove Switch from list of components that are supposed to get types automatically generate

## 🤔 Considerations

Option:

image

List:

image

## ✅ Checklist


fergusjordantw (approved):
cool

packages/components/src/switch/Switch.tsx
```diff
@@ -8,18 +7,35 @@ import { logActionRequiredIf } from '../utilities';
...
-const Switch = (props) => {
+type Props = {
+  /** Used to describe the purpose of the switch. To be used if there is no external label (i.e. aria-labelledby is null) */
+  'aria-label'?: string;
+  /** A reference to a label that describes the purpose of the switch. Ignored if aria-label is provided */
+  'aria-labelledby'?: string;
+  /** Whether the switch is checked or not */
+  checked?: boolean;
+  /** Classes to apply to the switch container */
+  className?: string;
```
dozortsev, line 18: please reuse this property from `CommonProps` type, example: [`Badge`](https://github.com/transferwise/neptune-web/blob/SEND-2637-add-switch-option/packages/components/src/badge/Badge.tsx#L21)

packages/components/src/switchOption/SwitchOption.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+type Props = {
+  checked?: boolean;
+  complex?: boolean;
+  content?: ReactChild;
+  disabled?: boolean;
+  id?: string;
+  media?: ReactChild;
+  onChange: (newValue: boolean) => void;
+  title: ReactChild;
```
dozortsev, line 14: we tend to use `ReactNode` for node props, is `ReactChild` somehow better?

packages/components/src/switchOption/SwitchOption.tsx
```diff
@@ -0,0 +1,57 @@
...
+};
+
+const SwitchOption: ComponentType<Props> = ({
```
dozortsev, line 17: just out of curiosity, why we need use `ComponentType` here? what's the benefits?

we tend to write components this way:

```ts
function ComponentName({
  className = undefined,
  size = Size.SMALL,
  children,
}: Props): ReactElement {
```

if there is not much value in `ComponentType` could you please change it to as shown on above example?

packages/components/src/switchOption/SwitchOption.tsx
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    onChange(!checked);
+  };
+
+  return (
+    <Option
+      {...sharedProps}
+      onClick={toggle}
+      button={<Switch id={id} checked={checked} onClick={toggle} disabled={disabled} />}
```
dozortsev, line 52: why here we need pass `onClick` in `Switch` and `Option` while we don't do same at very similar `CheckboxOption`?

packages/docs/pages/components/options/SwitchOption.mdx
```diff
@@ -0,0 +1,11 @@
+import { LiveEditorBlock, GeneratePropsTable } from '../../../utils';
+import { FastFlag as FastFlagIcon } from '@transferwise/icons';
+import { SwitchOption } from '@transferwise/components';
+import code from '../../../liveEditorCode/switchOption.code';
+
+<LiveEditorBlock code={code} scope={{ SwitchOption, FastFlagIcon }} />
+<GeneratePropsTable componentName="SwitchOption" />
+
+export const meta = {
+  name: 'SwitchOption',
```
dozortsev, line 10: please add badge to indicate that this is new component

```
badge: {
    type: 'new',
    expiryDate: '2021-09-30',
},
```

dozortsev (approved):

packages/components/src/switchOption/SwitchOption.tsx
mikknarusk: usually I use ReactNode as well but ReactNode actually allows some types that don't apply here like `ReactPortal, boolean, null, undefined`:
`type ReactNode = ReactChild | ReactFragment | ReactPortal | boolean | null | undefined;`

Maybe `media` could be `ReactNode` but imo `content` and `title` should be `ReactChild` as these specific props should be string-like.

packages/components/src/switchOption/SwitchOption.tsx
mikknarusk: I think typing the entire function is better as it provides a more informative type over just typing the props and return value:
Screenshot 2021-07-19 at 14 54 38

That said I should use `FunctionComponent` here instead of `ComponentType` in this case. WDYT?

packages/components/src/switchOption/SwitchOption.tsx
mikknarusk: `CheckboxOption` passes equvalent `onChange` prop to `CheckboxButton`.
As for the Option I think its just a question of do we want the entire Option to be clickable, or just the Switch/Checkbox inside it?

packages/docs/pages/components/options/SwitchOption.mdx
mikknarusk: cool, done

packages/components/src/switchOption/SwitchOption.tsx
dozortsev: okey, make sense

packages/components/src/switchOption/SwitchOption.tsx
dozortsev: `FC` or `FunctionComponent` added `children` prop by default, we don't want that

packages/components/src/switchOption/SwitchOption.tsx
dozortsev: > As for the Option I think its just a question of do we want the entire Option to be clickable, or just the Switch/Checkbox inside it?

I checked that in `CheckboxOption` it changes value (checked/unchecked) when you click on option area (e.g content or title) so `onClick` seems redundant 🤔

packages/components/src/switchOption/SwitchOption.tsx
dozortsev: re. `FC` https://github.com/facebook/create-react-app/pull/8177

mikknarusk: fair point, I'll make the change like you suggested

PR #1047, opened 2021-07-16
chore: upgrade to React 17

matebek-tw:
## 🖼 Context

To unblock React 17 usage with the component library.

## 🚀 Changes

- Upgrade React to version 17.0.2 across all packages
- Allow React ^16 || ^17 as peer dependency
- Update Enzyme adapter to [@wojtekmaj/enzyme-adapter-react-17](https://www.npmjs.com/package/@wojtekmaj/enzyme-adapter-react-17)
- Update `jest` and `testing-library` dependencies
- Update tests where old `jest` APIs are failing
- Fix `Select` component compatibility issue

### Select

I identified the problem that the `click` event opening the dropdown is delegated down to the `handleDocumentClick` event handler, triggering the `close()` method immediately. Deferring the `addEventListener` call in the `setState` callback while not a perfect solution, but it solves all of our problems for now.

Following the steps described on react blog regarding [event delegation](https://reactjs.org/blog/2020/08/10/react-v17-rc.html#changes-to-event-delegation) only partially solves our problems, some `testing-library` tests are failing by not rendering anything to the "page", and the behaviour itself still has weird bugs where sometimes you are not able to open the select dropdown even if it is closed (probably an issue with the event listeners).

## ✅ Checklist


packages/docs/package.json
```diff
@@ -68,15 +68,15 @@
...
     "babel-plugin-inline-react-svg": "^1.1.1",
     "deploy-to-github-pages": "1.0.0-beta.7",
     "typescript": "^4.1.5"
   },
   "peerDependencies": {
-    "react": "16.13.1",
-    "react-dom": "16.13.1"
+    "react": "17.0.2",
+    "react-dom": "17.0.2"
   },
```
sbrandwoo, line 80: Do these changes work on both React 16 and React 17? If so it'll make migration much easier.

For `peerDependencies` you should be as unspecific as possible. In this case you're saying React must be exactly `17.0.2`, ideally we'd say it could be any version with a major of 16 or 17:

```json
  "peerDependencies": {
    "react": "^16 || ^17",
    "react-dom": "^16 || ^17"
  },
```

sbrandwoo: Uh, I see I'm commenting on the docs package file - I think you need to update `packages/components/package.json` too.

packages/components/package.json
```diff
@@ -92,8 +92,8 @@
     "@transferwise/neptune-css": "^9.2.0",
     "currency-flags": "^2.1.1",
     "prop-types": "^15.7.2",
-    "react": "^16",
-    "react-dom": "^16",
+    "react": "^16 || ^17",
+    "react-dom": "^16 || ^17",
```
dozortsev, line 96: I guess we don't need define react here since we have it at root `package.json`

packages/components/src/select/Select.spec.js
```diff
@@ -151,14 +152,14 @@ describe('Select', () => {
   });
 
   it('can be closed by clicking somewhere else', async () => {
-    openSelect();
+    await await openSelect();
```
dozortsev, line 155:
```suggestion
    await openSelect();
```

packages/css/package.json
```diff
@@ -52,8 +52,8 @@
     "@zeit/next-css": "^1.0.1",
     "next": "^9.5.5",
     "prop-types": "^15.7.2",
-    "react": "16.13.1",
-    "react-dom": "16.13.1",
+    "react": "17.0.2",
+    "react-dom": "17.0.2",
```
dozortsev, line 56: do we even need React in css package?

packages/components/src/select/Select.js
```diff
@@ -256,18 +256,20 @@ export default class Select extends Component {
         if (!isTouchDevice && searchable && this.searchBoxRef.current) {
           this.searchBoxRef.current.focus();
         }
+
+        addClickClassToDocumentOnIos();
+        document.addEventListener('click', this.handleDocumentClick, false);
```
dozortsev, line 261: have you checked whether Select component continues works same way on React 16?

matebek-tw: Yeah, I checked the solution with React 16 as well and I haven't noticed any issues with it.

packages/css/package.json
matebek-tw: Removed, keeping it listed only in root `package.json`.

packages/components/package.json
dozortsev: make sense

packages/components/src/select/Select.js
dozortsev: ok

packages/components/package.json
```diff
@@ -72,8 +72,8 @@
     "@types/jest": "^26.0.20",
     "@types/jscodeshift": "^0.11.0",
     "@types/node": "^14.14.31",
-    "@types/react": "^17.0.2",
-    "@types/react-dom": "^17.0.1",
+    "@types/react": "^17.0.14",
+    "@types/react-dom": "^17.0.9",
```
dozortsev, line 76: it would be good to run visual test as well for this change

package.json
```diff
@@ -27,6 +27,7 @@
     ]
   },
   "devDependencies": {
+    "react-codemod": "^5.4.3",
```
dozortsev, line 30: this needs to be deleted

packages/components/.eslintrc
```diff
@@ -1,3 +1,7 @@
 {
-  "extends": ["@transferwise", "../../.eslintrc"]
+  "extends": ["@transferwise", "../../.eslintrc"],
+  "rules": {
+    "react/jsx-uses-react": "off",
+    "react/react-in-jsx-scope": "off"
+  }
```
dozortsev, line 6: https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#eslint

packages/components/src/test-utils/index.js
```diff
@@ -17,6 +17,17 @@ function customRender(ui, { locale = DEFAULT_LOCALE, messages = en, ...renderOpt
   return render(ui, { wrapper: Wrapper, ...renderOptions });
 }
 
+/**
+ * Custom `renderHook` function which wraps passed elements in Provider component
+ * For more info: https://react-hooks-testing-library.com/usage/advanced-hooks#context
+ */
+function customRenderHook(callback, { locale = DEFAULT_LOCALE, messages = en } = {}) {
```
matebek-tw, line 24: 🚀

dozortsev (approved):

PR #1049, opened 2021-07-19
feat!: remove CSS imports from components

czebe:
## 🖼 Context

BREAKING CHANGE: Since Next.js **9.5.4**, importing a CSS file from `node_modules` is permitted anywhere in your application. In order to upgrade `crab` and our apps to the latest `Next` version, we need to remove direct CSS imports from the DS components, and provide a `main.css` bundle that can be imported as a global stylesheet in the app root.

https://nextjs.org/docs/basic-features/built-in-css-support#import-styles-from-node_modules

https://transferwise.atlassian.net/browse/DS-1179

## 🚀 Changes

- remove CSS imports from components
- expose individual CSS files and `main.css` under `styles/` for easier importing
- update documentation

Tooling improvements:
- use the same storybook theme in `dynamic-flows` as in `components`

## 🤔 Considerations

Consumers need to explicitly include `@transferwise/components/build/main.css` in their app.
See [usage guidelines](https://github.com/transferwise/neptune-web/blob/8c344498a3926a527eca122fbb66df99784300a5/packages/docs/pages/about/Setup.mdx) for details.

## ✅ Checklist


packages/docs/pages/about/Setup.mdx
```diff
@@ -24,13 +24,27 @@ npm install react react-dom react-intl prop-types currency-flags # peer dependen
...
 import '@transferwise/neptune-css/dist/css/neptune.css';
 import '@transferwise/icons/lib/styles/main.min.css';
 
-// This is only necessary if you are using the UMD build
-import '@transferwise/components/build/main.css';
+// Include styles for all components (easiest and fail-safe solution)
+import '@transferwise/components/styles/main.css';
 
+// Or -- if concerned about bundle size -- list all the component styles that your application is using, including nested usages in other packages (@transferwise/calculator etc).
+// CAUTION: your app will break if you upgrade one of the libraries that starts using a new component and you forget to update this list.
```
matebek-tw, line 35: I might not even mention this option just because it can lead to many annoying issues without a breaking change in the component library itself (old components can start to use existing components as well).

czebe: Keeping it as an advanced option, because many projects will have visual regression tests, so they'd catch any layout bugs easily.

packages/components/src/main.less
```diff
@@ -0,0 +1,56 @@
+@import './accordion/Accordion.less';
+@import './avatar/Avatar.less';
+@import './badge/Badge.less';
+@import './button/Button.less';
+@import './card/Card.less';
+@import './checkboxButton/CheckboxButton.less';
+@import './chip/Chip.less';
+@import './circularButton/_button-label-states.less';
+@import './circularButton/CircularButton.less';
+@import './common/Option/Option.less';
```
dozortsev, line 10: I don't think we need `Option.less` here as it's internal component, it's styles should available via components that use `Option` component, e.g `NavigationOption` via `NavigationOption.less`, `CheckboxOption` via `CheckboxOption.less`

note: we don't have such file `CheckboxOption.less` but we should create it for this case and import there only `Option.less`

packages/components/src/main.less
```diff
@@ -0,0 +1,56 @@
...
+@import './common/RadioButton/RadioButton.less';
+@import './common/bottomSheet/BottomSheet.less';
+@import './common/closeButton/CloseButton.less';
+@import './common/panel/Panel.less';
+@import './dateInput/DateInput.less';
+@import './dateLookup/DateLookup.less';
+@import './dateLookup/dateTrigger/DateTrigger.less';
```
dozortsev, line 17: here is the same, `DateLookup.less` should bring `DateTrigger.less` for me as a consumer

packages/docs/pages/about/Setup.mdx
```diff
@@ -24,13 +24,27 @@ npm install react react-dom react-intl prop-types currency-flags # peer dependen
...
+import '@transferwise/components/styles/accordion/Accordion.css';
+import '@transferwise/components/styles/button/Button.css';
+import '@transferwise/components/styles/avatar/Avatar.css';
+// ...
+
+// It's also possible to import styles right where you're using the component:
+import { Accordion } from "@transferwise/components";
+import '@transferwise/components/styles/accordion/Accordion.css';
```
dozortsev, line 43: I think this is redundant bit, you already covered this use case above ('..Or -- if concerned about bundle size -- list all the component styles that your application is using..')

packages/docs/pages/about/Setup.mdx
```diff
@@ -24,13 +24,27 @@ npm install react react-dom react-intl prop-types currency-flags # peer dependen
...
+```
+
+```js
+// Usage in app components. Render <Provider /> at the root of your app.
```
dozortsev, line 47: what this for?

packages/dynamic-flows/.storybook/common/theme.js
```diff
@@ -1,35 +1,35 @@
...
-  appBorderRadius: 4,
+  appContentBg: '#f2f5f7',
+  appBorderColor: '#e2e6e8',
+  appBorderRadius: 3,
 
   // Typography
-  fontBase: '"Averta", "Avenir W02", "Avenir", Helvetica, Arial, sans-serif',
-  fontCode: 'monospace',
+  // fontBase: '"Averta", "Avenir W02", "Avenir", Helvetica, Arial, sans-serif',
+  // fontCode: 'monospace',
```
dozortsev, line 17: why is this commented out?

packages/dynamic-flows/README.md
```diff
@@ -21,10 +21,26 @@ npm install @transferwise/dynamic-flows react react-dom prop-types
 ```
 
 ```js
-// This line only needs to be imported once in your application
+// These CSS imports only need to be imported once in your application in 'pages/_app.{js,jsx,tsx}'
```
dozortsev, line 24: same here as above, it's purely NextJS thing

packages/dynamic-flows/README.md
```diff
@@ -21,10 +21,26 @@ npm install @transferwise/dynamic-flows react react-dom prop-types
...
+  schema={schema}
+  model={model}
+  errors={errors}
+  translations={translationsFromProps}
+  onChange={() => {}}
+  submitted={submitted}
+  disabled={disabled}
+  baseUrl=""
+  onPersistAsync={() => {}}
+/>
```
dozortsev, line 43: what this for?

dozortsev (commented):
it probably would be good to migrate docs app on NextJS 11 as part of this PR to properly test things but we have dependency on https://github.com/transferwise/neptune-web/pull/1047

packages/components/package.json
```diff
@@ -34,7 +35,8 @@
     "build:es": "NODE_ENV=es babel src --extensions '.tsx,.js,.ts' -d build/es/polyfill --ignore '**/*.spec.js','**/*.story.js','**/__mocks__','**/test-utils'",
     "build:es-nopolyfill": "NODE_ENV=es-nopolyfill babel src --extensions '.tsx,.js,.ts' -d build/es/no-polyfill --ignore '**/*.spec.js','**/*.story.js','**/__mocks__','**/test-utils'",
     "build:generate-types": "tsc && node ./scripts/generate-type-declarations.js",
-    "build:copy-files": "cpx 'src/**/!(db)/*.{css,json,svg}' build/es/polyfill & cpx 'src/**/!(db)/*.{css,json,svg}' build/es/no-polyfill",
+    "build:copy-files": "cpx 'src/**/!(db)/*.{json,svg}' build/es/polyfill & cpx 'src/**/!(db)/*.{json,svg}' build/es/no-polyfill",
+    "build:copy-css": "cpx 'src/main.css' build/ & cpx 'src/**/*.css' styles/",
```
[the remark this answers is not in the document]
czebe: yes, only to maintain backwards compatibility.
Some apps might have been importing `main.css` (like me) --> the flipside: I don't have to do anything to upgrade : DD

packages/components/rollup.config.js
```diff
@@ -43,11 +42,6 @@ const plugins = [
   }),
   // Convert CJ into ES6
   commonjs({ sourcemap: false, namedExports }),
-  postcss({
-    config: true,
-    extract: pkg.style,
-    extensions: ['.css'],
-  }),
```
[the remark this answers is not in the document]
czebe: This task extracts CSS that are imported in JS files. Since we don't import CSS any more, this step is not necessary.

packages/components/src/main.less
```diff
@@ -0,0 +1,56 @@
...
+@import './summary/Summary.less';
+@import './switch/Switch.less';
+@import './tabs/Tabs.less';
+@import './tile/Tile.less';
+@import './tooltip/Tooltip.less';
+@import './typeahead/Typeahead.less';
+@import './typeahead/typeaheadOption/TypeaheadOption.less';
+@import './upload/Upload.less';
+
+// List all less files in src in alphabetical order: find -s src -iname '*.less'
```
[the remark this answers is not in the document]
czebe: We shouldn't rely on a specific order. I listed in alphabetical order only to confirm that the command line found all `less` files.

packages/dynamic-flows/.storybook/common/theme.js
czebe: copied from the `components` theme file (it's commented out there as well).

packages/docs/pages/about/Setup.mdx
czebe: I think the end result is different in these three cases, so it's important to understand:
- `main.css` imported -> all styles are loaded with one line of import (largest bundle size)
- individual styles imported in `_app` -> optimised bundle size, but all CSS is loaded upfront
- individual styles imported at component level: optimised and chunked bundle files (eg. chunk is only loaded when navigating to a page that is using the component) -> best optimisation, but requires lot of changes and hard to maintain

packages/components/src/main.less
czebe: Many components seem to use `Option`. This is a substantial refactoring project, we can do it later.
(And I don't know if we would gain anything from it)

packages/components/src/main.less
czebe: ah, okay, I see what you mean, it's important when importing individual CSS files.
I don't have the complete understanding of inter-dependencies among the components. Also, I'll double-check if this can cause any code duplication.

packages/components/src/chevron/Chevron.js
```diff
@@ -1,4 +1,3 @@
-import '@transferwise/neptune-css/dist/css/chevron.css';
```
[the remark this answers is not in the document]
czebe: Why is this file different than the rest of `neptune-css`? Can we really remove, isn't it used in legacy apps?
I'd tackle it in a different PR, it's unclear why this was imported here.

packages/components/.storybook/preview.js
```diff
@@ -11,7 +11,7 @@ import { MINIMAL_VIEWPORTS } from '@storybook/addon-viewport';
 import { DEFAULT_LOCALE, getLangFromLocale } from '../src/common/locale';
 import supportedLangs from '../src/i18n';
 
-import './storybook.css';
```
helendurranttw, line 14: storybook.css still exists and is in use - it includes a fix for code blocks being the wrong colour. Have you ensured this was fixed?

packages/components/src/chevron/Chevron.js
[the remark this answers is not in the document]
helendurranttw: Neptune CSS is a prerequisite for components so it should be fine to remove it here. To move it from Neptune CSS we'd have to ensure it's not used in styleguide components.

packages/docs/pages/about/Setup.mdx
```diff
@@ -24,11 +24,9 @@ npm install react react-dom react-intl prop-types currency-flags # peer dependen
 ## Usage
 
 ```js
-// These CSS imports only need to be imported once in your application
+// These CSS files should be imported once in your application
 import '@transferwise/neptune-css/dist/css/neptune.css';
 import '@transferwise/icons/lib/styles/main.min.css';
-
-// This is only necessary if you are using the UMD build
```
helendurranttw, line 30: 👍

packages/docs/pages/about/Styles.mdx
```diff
@@ -12,8 +12,6 @@ Refer to the <DocLink href="/styles/core/Typography">documentation</DocLink> for
 
 We recommend using Neptune CSS with our [Neptune Web React components](https://www.npmjs.com/package/@transferwise/components), but it will also work with our [Angular components](https://www.npmjs.com/package/@transferwise/styleguide-components).
 
-**Note:** Do not use component CSS directly. It is considered an implementation detail and may change without warning.
```
helendurranttw, line 15: This is still true 😇

packages/dynamic-flows/.storybook/common/theme.js
```diff
@@ -1,35 +1,35 @@
...
 export default create({
-  base: 'dark',
+  base: 'light',
 
   colorPrimary: '#37517e',
   colorSecondary: '#00b9ff',
 
   // UI
   appBg: '#f2f5f7',
-  appContentBg: 'silver',
```
helendurranttw, line 11: Why did all of these change?

matebek-tw: `dynamic-flows` storybook theme/styles were pretty bad in some cases, so we decided to use the same theme as for the `components` storybook (basically copied over `theme.js`). We should extract the storybook config out and re-use it for both packages in a later PR.

packages/components/.storybook/preview.js
matebek-tw: I think this was removed by mistake.

helendurranttw (approved):

PR #1051, opened 2021-07-20
fix: pass aria-label from SwitchOption to Switch

mikknarusk:
## 🖼 Context
Switch requires `aria-label` but SwitchOption doesn't define it or pass it along.

## 🚀 Changes
Adds aria-label as a required prop to SwitchOption and passes it along to Switch.
 

## 🤔 Considerations
Technically a breaking change but 
- noone is using it yet
- doesn't actually break anything if you omit it
- technically also a bugfix as that's what it was supposed to be from the start

## ✅ Checklist


packages/components/src/switchOption/SwitchOption.tsx
```diff
@@ -49,7 +51,7 @@ const SwitchOption = ({
     <Option
       {...sharedProps}
       onClick={toggle}
-      button={<Switch id={id} checked={checked} onClick={toggle} disabled={disabled} />}
+      button={<Switch id={id} checked={checked} onClick={toggle} disabled={disabled} {...rest} />}
```
dozortsev, line 54: I would either make alias for `'aria-label'`

```ts
title,
'aria-label': ariaLabel,

// in jsx
aria-label={ariaLabel}
```

or explicitly get the prop from `rest`

```ts
aria-label={rest['aria-label']}
```

dozortsev (approved):

packages/components/src/switchOption/SwitchOption.tsx
mikknarusk: sounds good, didn't want to screw around with the kebab case variable but will update as you suggest

PR #1054, opened 2021-07-21
fix: accessibility issues with Select

joe-sayer:
## 🖼 Context

During an interview with a blind customer who uses JAWS screen reader, it was brought to our attention that the Select component didn't work as intended with JAWS (potentially other screen readers). The following issues were noted:

- JAWS reads out the word "menu" after every item
- When selecting an item with the arrow keys, the wrong item would be selected
- Loses focus after selection - returned focus back to top of the screen

Before changes
https://user-images.githubusercontent.com/44163070/126524156-8f0011bb-6dd3-4a94-8fd8-5720c3451dec.mov

After changes

https://user-images.githubusercontent.com/44163070/126524228-43c4d649-aa4e-4bef-a039-3e1e4a4ffb23.mov

## 🚀 Changes

 
- Change Select `ul` role from `menu` to `tablist`

## 🤔 Considerations
- Is this the most appropriate role for the Select component `ul`? 🤔 

## ✅ Checklist


packages/components/src/select/Select.js
```diff
@@ -6,6 +6,7 @@ import Transition from 'react-transition-group/Transition';
 import Option from './option';
 import Chevron from '../chevron';
 import KeyCodes from '../common/keyCodes';
+// eslint-disable-next-line
```
dozortsev, line 9: what this for?

PR #1063, opened 2021-07-28
feat: add showMediaAtAllSizes to all option components

fergusjordantw:
## 🖼 Context

Needed it for ActionOption and figured we'd update the rest of the Option components so they have a more standardised interface. 

## 🚀 Changes

Drills `showMediaAtAllSizes` down to `Option` for `RadioOption`, `SwitchOption` and `ActionOption`

## ✅ Checklist


mikknarusk (approved):
🆒

packages/components/src/actionOption/ActionOption.story.tsx
```diff
@@ -16,6 +16,7 @@ export const basic = () => {
   const content = text('content', 'Normally, the button and icon are vertically centered.');
   const actionText = text('action', 'Action');
   const disabled = boolean('disabled', false);
+  const showMediaAtAllSizes = boolean('showMediaAtAllSizes', false);
```
Tankenstein, line 19: How about we instead do `showMediaFromSize` which could be a size that defaults to current behaviour?

giljulio:
Out of curiosity where do we need this prop? Or an alternative question, where do we not show media at all sizes? 

Reason I ask is that on native we always show media at all sizes

fergusjordantw:
> Out of curiosity where do we need this prop? Or an alternative question, where do we not show media at all sizes?

By default, we're only showing the media on screens larger than 480px - I'm not sure what the reasoning is behind this, this component is re-using a bunch of CSS from the `Decision` component for controlling when to show it's media which makes it difficult to understand if this API was chosen by design or added as a hack on top of a previously existing component to override the re-used CSS.

packages/components/src/actionOption/ActionOption.story.tsx
fergusjordantw: Re [this comment](https://github.com/transferwise/neptune-web/pull/1063#issuecomment-888120565) I'm wondering if we should just do away with not showing it on all sizes altogether - mind you that would be a visually breaking change

dozortsev: https://github.com/transferwise/neptune-web/pull/1063#issuecomment-888120565

yeah, we might decide make it work same way as on mobile when we review and align option component this Q

dozortsev (approved):

giljulio:
> > Out of curiosity where do we need this prop? Or an alternative question, where do we not show media at all sizes?
> 
> By default, we're only showing the media on screens larger than 480px - I'm not sure what the reasoning is behind this, this component is re-using a bunch of CSS from the `Decision` component for controlling when to show it's media which makes it difficult to understand if this API was chosen by design or added as a hack on top of a previously existing component to override the re-used CSS.

alright thanks 👍  

@transferwise/design-system-designers might be interesting to see if we could remove this behaviour

PR #1072, opened 2021-08-02
feat(docs): upgrade Next.js and its dependencies

matebek-tw:
## 🖼 Context

https://transferwise.atlassian.net/browse/DS-1181

Upgrade `docs` to the latest Next.js.

## 🚀 Changes

- Fix breaking changes following the Next.js [upgrade guide](https://nextjs.org/docs/upgrading)
- Upgrade all the dependencies for the `docs` codebase
- Fix emotion `jsx` rendering issues with React 17

### Image optimization

Since version 10 Next.js has a [built-in image optimisation feature](https://nextjs.org/docs/basic-features/image-optimization), which causing the image source to be missing in production. I had to revert to the old `next-images` solution and disable the built-in image handling.

### Emotion `css` prop

There are [2 ways](https://emotion.sh/docs/css-prop) of using the `css` prop for components styled with emotion. We are using the [JSX Pragma](https://emotion.sh/docs/css-prop#jsx-pragma) version, but it has some issues with the [new JSX transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) introduced in React 17. The simplest fix for the issue is to replace the pragma with `/** @jsxImportSource @emotion/react */` on the top of these files.

The other solution would be to use the [Babel Preset](https://emotion.sh/docs/css-prop#babel-preset) provided by emotion, but it [cannot compile](https://github.com/mdx-js/mdx/issues/1441) together with `mdx-js` which we are using (through `@next/mdx`) to render markdown.

Another option would be to use the `styled` syntax instead of the `css` prop, and in case of a dynamic component (`Element`) we could [change the render tag using `withComponent`](https://emotion.sh/docs/styled#change-the-rendered-tag-using-withcomponent).

```javascript
// Flex.js
...
const StyledElement = styled.div(styles); // will create a styled `div`
const Element = StyledElement.withComponent(element); // changes the HTML element to be rendered keeping the applied styles in the previous step
...
```

## ✅ Checklist


packages/docs/components/layout/box/Box.js
```diff
@@ -1,6 +1,6 @@
-/** @jsx jsx */
+/** @jsxImportSource @emotion/react */
```
dozortsev, line 1: could you please add a comment (with some links to docs) explaining this?

packages/docs/next-env.d.ts
```diff
@@ -1,5 +1,2 @@
-// This file ensures Next.js types are picked up by the TypeScript compiler.
-// You cannot remove it, however, you can edit it (but you don't need to).
-// Source: https://nextjs.org/docs/basic-features/typescript
```
dozortsev, line 3: this comment is useful, why removing it? or it become irrelevant with a new version of Next.js?

dozortsev (approved):

packages/docs/next-env.d.ts
matebek-tw: The package installer removed it automatically, I can add it back.

packages/docs/components/layout/box/Box.js
matebek-tw: I can, but you have to provide this comment in every file you'd like to use the `css` property for emotion, as it is documented here: https://emotion.sh/docs/css-prop#jsx-pragma

I can add it to the `README.md` why we need this.

PR #1074, opened 2021-08-03
feat(components): add onValidationError prop to UploadInput

matebek-tw:
## 🖼 Context

The host application already has to provide the `onFileUpload` callback to handle file uploads where it can handle server errors, but in case the file validation fails before the upload callback is triggered, the host application has no way to know about the error at the moment.

## 🚀 Changes

- Add `onValidationError ` prop to `UploadInput`

## ✅ Checklist


packages/components/src/uploadInput/UploadInput.tsx
```diff
@@ -138,31 +147,37 @@ const UploadInput = ({
...
-            },
-          ]);
+          const failedUpload = {
+            id,
+            filename: name,
+            status: Status.FAILED,
+            error: formatMessage(MESSAGES.fileTypeNotSupported),
+          };
+
+          setUploadedFiles((currentFiles) => [...currentFiles, failedUpload]);
```
StevePole, line 157: Should we append files that file validation?

matebek-tw: We are appending them to the UI to display them in a failed state so the user will see the failed uploads and the error message.

packages/components/src/uploadInput/UploadInput.tsx
```diff
@@ -41,6 +41,12 @@ export type UploadInputProps = {
    */
   onDeleteFile?: (id: string | number) => Promise<any>;
 
+  /**
+   * Provide a callback to trigger on validation error
+   * @param file
+   */
+  onValidationError?: (file: UploadedFile) => void;
```
dozortsev, line 48: I would probably call this a bit differently, e.g `onFailedUpload`, I think this callback could be triggered not only for validation failure reasons

dozortsev (approved):

packages/components/src/uploadInput/UploadInput.tsx
matebek-tw: The `onUploadFile` callback is already provided by the consumer app, where they can handle failed uploads. But that callback wouldn't be called at all when any of the validations are failing, hence the `onValidationError` API.

dozortsev: I see, make sense 👍 

could you please mention that consumers can handle failed uploads in JSDoc for that prop?

PR #1075, opened 2021-08-03
feat: add safe fallback for useIntl and injectIntl

matebek-tw:
## 🖼 Context

https://transferwise.atlassian.net/browse/DS-1103

We would like to add a graceful fallback for the internationalisation in case the `react-intl` provider is not set.

## 🚀 Changes

- Add `defaultIntl` and re-implement the used `react-intl` APIs
- Add `useSafeIntl` custom hook which returns `defaultIntl` if the original `intl` object cannot be accessed
- Add `injectSafeIntl` which sets the `deafultIntl` in case the property is missing from the `IntlContext.Consumer`

TODO:
- [ ] Expose useSafeIntl from @transferwise/components
- [ ] Use useSafeIntl in dynamic-flows

## 🤔 Considerations

With this approach, we have to implement all the APIs used from `react-intl` and make sure that it works the same way as it is in the library. Testing this behaviour would require us to render the components with and without a provider, and compare the results.

### FormattedMessage
The `<FormattedMessage />` syntax will not work, as it is not relies on the `intl` prop itself.

```jsx
❌ <FormattedMessage id="neptune.MoneyInput.Select.placeholder" />
✅ {intl.formatMessage({ id: 'neptune.MoneyInput.Select.placeholder' })
```

## ✅ Checklist

### DIFF
--- a/packages/components/src/common/safeIntl/injectSafeIntl.tsx
+++ b/packages/components/src/common/safeIntl/injectSafeIntl.tsx
@@ -1,0 +1,60 @@
+import * as React from 'react';
+import hoistNonReactStatics from 'hoist-non-react-statics';
+import { IntlContext } from 'react-intl';
+import { Opts, WrappedComponentProps, WithIntlProps } from 'react-intl/src/components/injectIntl';
+import { SafeIntlShape } from './types';
+import { defaultIntl } from './intl';
+
+function getDisplayName(Component: React.ComponentType<any>): string {
+  return Component.displayName || Component.name || 'Component';
+}
+
+export function injectSafeIntl<
+  IntlPropName extends string = 'intl',
+  P extends WrappedComponentProps<IntlPropName> = WrappedComponentProps<any>,
+  ForwardRef extends boolean = false,
+  T extends React.ComponentType<P> = any
+>(
+  WrappedComponent: React.ComponentType<P>,
+  options?: Opts<IntlPropName, ForwardRef>,
+): React.ForwardRefExoticComponent<
+  React.PropsWithoutRef<WithIntlProps<P>> & React.RefAttributes<T>
+> & {
+  WrappedComponent: React.ComponentType<P>;
+} {
+  const { intlPropName = 'intl', forwardRef = false, enforceContext = true } = options || {};
+
+  const WithIntl: React.FC<P & { forwardedRef?: React.Ref<any> }> & {
+    WrappedComponent: React.ComponentType<P>;
+  } = (props) => (
+    <IntlContext.Consumer>
+      {(intl: SafeIntlShape): React.ReactNode => {
+        let safeIntl = intl;
+        if (enforceContext && !safeIntl) {
+          // eslint-disable-next-line no-console
+          console.warn(
+            '<IntlProvider> needs to exist in the component ancestry.',
+            `Falling back to "${defaultIntl.locale}".`,
+          );
+          safeIntl = defaultIntl;
+        }
+        const intlProp = { [intlPropName]: safeIntl };
+
+        return (
+          <WrappedComponent {...props} {...intlProp} ref={forwardRef ? props.forwardedRef : null} />
+        );
+      }}
+    </IntlContext.Consumer>
+  );
+  WithIntl.displayName = `injectIntl(${getDisplayName(WrappedComponent)})`;
+  WithIntl.WrappedComponent = WrappedComponent;
+
+  if (forwardRef) {
+    return hoistNonReactStatics(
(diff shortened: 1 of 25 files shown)


packages/components/src/common/safeIntl/injectSafeIntl.tsx
```diff
@@ -0,0 +1,60 @@
...
+  WithIntl.WrappedComponent = WrappedComponent;
+
+  if (forwardRef) {
+    return hoistNonReactStatics(
+      React.forwardRef<T, P>((props: P, ref) => <WithIntl {...props} forwardedRef={ref} />),
+      WrappedComponent,
+    ) as any;
+  }
+
+  return hoistNonReactStatics(WithIntl, WrappedComponent) as any;
```
dozortsev, line 59: I see this concept as redundant thing that brings very little value and loads of complexity

This new codebase [increase bundle size](https://bundlesize-store.vercel.app/build?info=%7B%22files%22%3A%5B%7B%22maxSize%22%3A146432%2C%22path%22%3A%22.%2Fpackages%2Fcomponents%2Fbuild%2Fumd%2Fpolyfill%2Fmain.js%22%2C%22size%22%3A149056%2C%22compression%22%3A%22gzip%22%7D%2C%7B%22maxSize%22%3A133120%2C%22path%22%3A%22.%2Fpackages%2Fcomponents%2Fbuild%2Fumd%2Fno-polyfill%2Fmain.js%22%2C%22size%22%3A129600%2C%22compression%22%3A%22gzip%22%7D%2C%7B%22maxSize%22%3A9216%2C%22path%22%3A%22.%2Fpackages%2Fcomponents%2Fbuild%2Fmain.css%22%2C%22size%22%3A8645%2C%22compression%22%3A%22gzip%22%7D%2C%7B%22maxSize%22%3A40960%2C%22path%22%3A%22.%2Fpackages%2Fcss%2Fdist%2Fcss%2Fneptune.css%22%2C%22size%22%3A39133%2C%22compression%22%3A%22gzip%22%7D%5D%2C%22repo%22%3A%22transferwise%2Fneptune-web%22%2C%22branch%22%3A%22DS-1103%22%2C%22commit_message%22%3A%22%22%2C%22sha%22%3A%22621ecdd1dc4b98ec02368478a63f9dc717c6b870%22%7D) like some new big component and at the same time it won't be necessary for 99% of our consumers yet they will bundle this code.

Think of maintaining this custom setup, atm "safe Intl" only provides `locale` and `formatMessage` API but what about [other features](https://formatjs.io/docs/react-intl/api/#intlshape)? I know we don't use them (yet) but what if our consumers do? Then it will still blow up with just another less obvious error message ("`undefined` is not a function" or something like that) unless we provide re-implement all react-intl functions.

Mostly component libraries that rely on `@transferwise/components` have it's own translations too, meaning they should provide it's own translations somehow at host application, this will remind consumer to setup a `Provider` component, some consumer libraries has it's own provider component that does translations setup and other context things if this is the case than that provider component should also setup Neptune `Provider` component

packages/components/src/common/safeIntl/intl.ts
```diff
@@ -0,0 +1,30 @@
...
+  if (!Object.keys(messages).includes(key)) {
+    return key;
+  }
+
+  return messages[key];
+}
+
+export const defaultIntl: SafeIntlShape = {
+  locale: 'en-GB',
+  formatMessage: (descriptor, values, opts) => {
```
dozortsev, line 17: but what about [other features](https://formatjs.io/docs/react-intl/api/#intlshape)? I know we don't use them (yet) but what if our consumers do? Then it will still blow up with just another less obvious error message ("`undefined` is not a function" or something like that) unless we provide re-implement all react-intl functions.

matebek-tw: We are not exposing this `intl` implementation yet, so the consumer of this API is our own component libraries at the moment, but if we decide that we would like to provide the `intl` object as part of our component library, we should re-implement all the functionalities with the fallbacks, that is correct.

StevePole:
Could you describe what the graceful fallback would do in these cases?  Presumably it would just fallback to English?  I don't if that's ideal.  I can imagine consumers (who mostly code in english) thinking it was properly set up, when actually it's not going to translate for them?

matebek-tw:
That is correct, we would only provide the English translations in this case. To be honest, I wouldn't recommend implementing a fallback as it probably would lead to more headaches and bugs even in 3rd party libraries than problems it solves.

Without the Provider:
- It would look like everything is set up correctly, but the translation strings are missing - failing early is always better in this case
- Some of our fallback APIs might work differently in certain conditions - we most likely don't want to take the burden of maintaining the whole `react-intl` API

PR #1079, opened 2021-08-04
Review fields in schema

JulianG-TW:
## 🖼 Context

Draft implementation of review fields as part of the schema.

## 🚀 Changes

Introduced `ReviewFormControl` which is used by `SchemaFormControl` when the schema is inferred to be a review field. 
this is when both `disabled` and `readOnly` properties are `true`.

## 🤔 Considerations

This is just a draft. Not all schemas/cases were contemplated. Testing is incomplete.

## ✅ Checklist

### DIFF
--- a/packages/dynamic-flows/src/jsonSchemaForm/schemaFormControl/SchemaFormControl.js
+++ b/packages/dynamic-flows/src/jsonSchemaForm/schemaFormControl/SchemaFormControl.js
@@ -3,6 +3,7 @@ import Types from 'prop-types';
 
 import { isNull, isUndefined } from '@transferwise/neptune-validation';
 import FormControl from '../../formControl';
+import ReviewFormControl from '../../formControl/ReviewFormControl';
 import { getValidModelParts } from '../../common/validation/valid-model';
 import { isOneOfSchema } from '../../common/schemaTypes/schemaTypes';
 import { FormControlType } from '../../common';
@@ -94,10 +95,17 @@ const SchemaFormControl = (props) => {
     placeholder: props.schema.placeholder,
     autoComplete: !props.schema.help,
     disabled: props.disabled || props.schema.disabled,
+    readOnly: props.readOnly || props.schema.readOnly,
     displayPattern: props.schema.displayFormat,
   };
 
-  return <FormControl type={controlType} value={safeValue} {...events} {...controlProps} />;
+  const isReview = controlProps.readOnly && controlProps.disabled;
+
+  return isReview ? (
+    <ReviewFormControl type={controlType} value={safeValue} {...controlProps} />
+  ) : (
+    <FormControl type={controlType} value={safeValue} {...events} {...controlProps} />
+  );
 };
 
 SchemaFormControl.propTypes = {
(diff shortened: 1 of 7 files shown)


packages/dynamic-flows/src/jsonSchemaForm/schemaFormControl/SchemaFormControl.js
```diff
@@ -94,10 +95,17 @@ const SchemaFormControl = (props) => {
     placeholder: props.schema.placeholder,
     autoComplete: !props.schema.help,
     disabled: props.disabled || props.schema.disabled,
+    readOnly: props.readOnly || props.schema.readOnly,
     displayPattern: props.schema.displayFormat,
   };
 
-  return <FormControl type={controlType} value={safeValue} {...events} {...controlProps} />;
+  const isReview = controlProps.readOnly && controlProps.disabled;
```
coolnirdh, line 102: Wouldn't this mean that if a control was either readOnly or disabled, but not both, we'd render it as a regular control that ignores readOnly?

There are 4 possible combinations, for 2 of them the rendering is quite clear
- readOnly: false / unspecified, disabled: false / unspecified -> renders as editable form control
- readOnly: false / unspecified, disabled: true -> renders as greyed out form control

For the other 2 combinations, we have a few options:

A: 
- readOnly: true, disabled: false / unspecified -> render it as review field
- readOnly: true, disabled: true -> render it as review field
So here it doesn't matter what is value of disabled, it will always render as review fields

B: 
- readOnly: true, disabled: false / unspecified -> render it as review field
- readOnly: true, disabled: true -> do not render at all, perhaps could be used by the client for its state but need not be shown to users

Do we have any way to specify hidden properties that need to be sent back to the server? Like CSRF tokens?

coolnirdh: Okay, saw that we also have a `hidden` property, which somewhat makes option B bad because it was relying on `disabled` to hide the readOnly field... 

To me, we should then settle with option A so that irrespective of whether a field is marked as disabled we render it as a readOnly field.... A field with `hidden: true` would ignore `disabled` property, likewise a field marked with `readOnly: true` could ignore whatever `disabled` is for the sake of rendering, wdyt?

StevePole:
I wouldn't approach it quite like this.  I would make it a `ReadOnlySchema` component, it should be in it's own folder, not a child of form control.  You shouldn't need to pass in individual props, you should pass in the `schema` and the model `value`.

We should render this with a definition list, not by inline styling.  That means integrating at a higher level, so that you can avoid rendering the `label` (which is semantically incorrect and invalid markup when not accompanied by a control).  I would think the place to integrate is as part of the main switch in `GenericSchema`.

Internally, it shouldn't work off of form control type, it should work off of data/schema type.  So rather than `if (formControlType === 'select' || formControlType === 'radio')` it should be `if (schema.oneOf)`.  Similarly for booleans it's `if(schema.type === 'boolean')` not `if (formControlType === 'checkbox')`.  This is because the list of controls may be extended, whereas the types of schema are fixed.

PR #1083, opened 2021-08-06
ReadOnlySchema implementation

JulianG-TW:
## 🖼 Context

We need to allow for read-only fields in schema forms. Mostly for PATCHing entities.

## 🚀 Changes

Introduce a new `ReadOnlySchema` which is rendered by `GenericSchemaForm` when required*.

Example of a Recipient update schema with some fields marked as `readOnly`:

image

## 🤔 Considerations

 - Boolean schema values are displayed using "yes/no" values that require translations.
 - Dates are printed as-is.
 - OneOf schemas of non-basic types avoid rendering tabs and only render the best matching schema. In the image above this is "Bank details" / "Local bank account".

- I didn't push the storybook I used to generate the image because it was messy and too specific to recipient updates.

## ✅ Checklist


packages/dynamic-flows/src/jsonSchemaForm/readOnlySchema/ReadOnlySchema.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import Types from 'prop-types';
+import { isBasicSchema } from '../../common/schemaTypes/schemaTypes';
+import ReadOnlyBasicTypeSchema from './ReadOnlyBasicTypeSchema';
+import ReadOnlyOneOfSchema from './ReadOnlyOneOfSchema';
+
+function ReadOnlySchema(props) {
+  const { schema } = props;
+
+  return isBasicSchema(schema) ? (
```
StevePole, line 10: Is this really an optimisation, I feel like you'd end up with less code and more readability if you just did it with one component?

JulianG-TW: I had it all in one component.
I felt like it was doing two different things, to the point that for a minute I considered branching off in GenericSchemaForm.
On the other hand, I'm not a fan of repeating propTypes all over the place, so I can bring it back to all in one component and then we can decide what's more readable.

JulianG-TW: Now it's all in one component. I think you're right. It's more readable.
Diff: 0c1a1cf00b1d61fb40ef242fb45b1c3b54eb4f79

packages/dynamic-flows/src/jsonSchemaForm/readOnlySchema/ReadOnlySchema.js
```diff
@@ -0,0 +1,78 @@
...
+
+function ReadOnlySchema(props) {
+  const { schema, model } = props;
+  const { title = '' } = schema;
+  const { formatMessage } = useIntl();
+
+  let value = null;
+  if (isBasicSchema(schema)) {
+    if (isOneOfSchema(schema)) {
+      const option = schema.oneOf.find((childSchema) => model === childSchema.const);
```
StevePole, line 20: Technically const is syntactic short hand for a schema with a single length `enum` e.g. `{ enum: ["GB"] }`, so maybe worth generalising to accommodate that formulation as well?  I would consider extracting to a function e.g. `getTitleForOneOfValue(value)`

packages/dynamic-flows/src/jsonSchemaForm/readOnlySchema/ReadOnlySchema.js
```diff
@@ -0,0 +1,78 @@
...
+      value = option ? getValueFromOption(option) : model;
+    } else {
+      switch (schema.type) {
```
StevePole, line 23: Why use a switch here?

packages/dynamic-flows/src/jsonSchemaForm/readOnlySchema/ReadOnlySchema.js
```diff
@@ -0,0 +1,78 @@
...
+        case 'boolean':
+          value = model ? formatMessage(messages.yes) : formatMessage(messages.no);
```
StevePole, line 25: Could the model be null?  Is it appropriate to show 'No' in that case?  You could use the isBoolean function from our library to ensure it's true or false.

packages/dynamic-flows/src/jsonSchemaForm/readOnlySchema/ReadOnlySchema.js
```diff
@@ -0,0 +1,78 @@
...
+          break;
+        default:
+          value = model && model.toString();
+      }
+    }
+  } else {
+    const index = getBestMatchingSchemaIndexForModel(schema, model);
+    if (index >= 0) {
+      value = <GenericSchemaForm {...props} schema={schema.oneOf[index]} />;
```
StevePole, line 34: I'm confused about what this is doing?

StevePole: Is this a clause we expect never to use?  I don't feel this component should be rendering child schemas, it could throw errors if given inappropriate schemas.

packages/dynamic-flows/src/jsonSchemaForm/readOnlySchema/ReadOnlySchema.js
JulianG-TW: I'm not happy about this. I used a switch because the alternative was:

```ts
    if (isOneOfSchema(schema)) {
      const option = schema.oneOf.find((childSchema) => model === childSchema.const);
      value = option ? getValueFromOption(option) : model;
    } else {
      if (schema.type === 'boolean') {
        value = model ? formatMessage(messages.yes) : formatMessage(messages.no);
      }
      value = model && model.toString();
    }
```
And I didn't like that it wasn't very when the `value` variable was being assigned to.

I can see the switch is no better, unless we come up with more cases. But we haven't. So, probably best to revert to conditionals.

packages/dynamic-flows/src/jsonSchemaForm/readOnlySchema/ReadOnlySchema.js
JulianG-TW: Yes. Thank you. I'll update.

packages/dynamic-flows/src/jsonSchemaForm/readOnlySchema/ReadOnlySchema.js
JulianG-TW: This basically "disables" the tabs, by:
- not rendering the tabs control
- rendering only the schema that best matches the current model

If a `oneOf` that is not a basic schema (for instance "local bank account" vs "IBAN" in GBP) is marked as readOnly and the model matches one of the schemas, then we render that schema as a `<GenericSchema>`.

The `getBestMatchingSchemaIndexForModel` function here is being reused, from the OneOfSchema, when we select the best matching schema to decide which tab to activate by default.

There's [a test for this](https://github.com/transferwise/neptune-web/blob/ef2cdd0d10362638e460b9438592c0aa2f00e563/packages/dynamic-flows/src/jsonSchemaForm/readOnlySchema/ReadOnlySchema.spec.js#L111-L143) which hopefully makes the intention more clear.

I'm happy to change it to something more readable, or perhaps it's not even what we want.

packages/dynamic-flows/src/jsonSchemaForm/readOnlySchema/ReadOnlySchema.js
JulianG-TW: I've pushed this commit c9f2e2adb23c4efc6d5e538aea827f9d83959556 for this

packages/dynamic-flows/src/jsonSchemaForm/readOnlySchema/ReadOnlySchema.js
JulianG-TW: The switch should be gone at ef2cdd0d10362638e460b9438592c0aa2f00e563

packages/dynamic-flows/src/jsonSchemaForm/readOnlySchema/ReadOnlySchema.js
JulianG-TW: This should be fixed at ef2cdd0d10362638e460b9438592c0aa2f00e563

packages/dynamic-flows/src/jsonSchemaForm/readOnlySchema/ReadOnlySchema.js
StevePole: When we discussed in the session, we agreed that we'd only support readOnly for basic type oneOfs.  Based off of what you've written it sounds like we need it for disabling the tabs?  Or is that a hypothetical need?  

We should probably revisit with the group if we want to allow disabling any schema, so that mobile are aware.

packages/dynamic-flows/src/jsonSchemaForm/readOnlySchema/ReadOnlySchema.js
```diff
@@ -0,0 +1,94 @@
+/* eslint-disable no-lonely-if */
+import React from 'react';
+import Types from 'prop-types';
+import { useIntl } from 'react-intl';
+import { DefinitionList, Layout } from '@transferwise/components';
+import { isBoolean } from '@transferwise/neptune-validation';
+import { isBasicSchema, isOneOfSchema } from '../../common/schemaTypes/schemaTypes';
+import { getCurrencyFlag } from '../schemaFormControl/availableCurrencyFlags';
+import { getBestMatchingSchemaIndexForModel } from '../oneOfSchema/OneOfSchemaModelMatcher';
```
StevePole, line 9: If we are going to need this for the generic schema stuff, I'd prefer to move it out of the oneOf schema into it's own location.

packages/dynamic-flows/src/jsonSchemaForm/readOnlySchema/ReadOnlySchema.js
JulianG-TW: **We need to disable the tabs.** This is already a problem in Ninjas update form, where OPS users can see the tabs, switch between tabs (which represent different recipient types), but we don't allow patching the recipient type.

**We don't need to disable tabs** if we decide the schema dedicated to updates should avoid a the oneOf altogether instead of marking it as readOnly. This is more work on the server side code, but may be the right thing to do.

packages/dynamic-flows/src/jsonSchemaForm/readOnlySchema/ReadOnlySchema.js
JulianG-TW: Good point. This depends on our decision about supporting readOnly oneOfs or not.

packages/dynamic-flows/src/jsonSchemaForm/readOnlySchema/ReadOnlySchema.js
JulianG-TW: I have now removed this functionality. We'll "disable tabs" by not providing the oneOf to begin with.

packages/dynamic-flows/src/jsonSchemaForm/readOnlySchema/ReadOnlySchema.js
JulianG-TW: This is now gone as well!

PR #1084, opened 2021-08-07
feat(select): incremental results when searching large Select lists for better performance 

sneakertack:
## 🖼 Context

When a `<Select />` is given a large list of options (e.g. 1000 options each with secondary text), attempting to search those options becomes very janky, with delays of a second or more upon entering the first few characters (the more valid results, the jankier - suggesting it's a rendering problem, and also explaining why the first few characters jank more (more hits) compared to when the search string is more fully formed).

In practical terms this makes the bank selection experience for a JPY recipient less than ideal, hence the motivation to fix this jank somehow.

## 🚀 Changes

This implements incremental loading of search results:

1. The 1st 50 results are shown first
2. If the user scrolls through 50 results, there's a button at the end to render the next 50, and so on

In practice, we expect that users won't even ever have to see this show-more button - they're hopefully searching properly, and scrolling through less than 50 matches instead (otherwise, they ought to be thinking of refining their search).

Regardless, the result is that due to having to render only max 50 results on every keychange, searching now performs more reasonably.

Video demo (slow before, fast after):

https://user-images.githubusercontent.com/2840821/128595613-58360d8e-d79c-4e74-b1fa-c9f679aef5ed.mp4

### Other solutions

There are other possible solutions to complement or replace this:

- debouncing key inputs for ~500ms so that a search only computes when the user has finished their input, rather than on every keychange
    - someone else could attempt this maybe
- making results beyond the 1st 50 automatically load (i.e. infinite-scroll), rather than require the user to explicitly click something
    - slightly nicer, but it'd be more complicated to implement and more things could go wrong, so I didn't because my time was limited. In any case, this PR is a good stepping stone towards infinite-scroll, so its incremental progress
    - and again, hopefully users are searching more efficiently in the 1st place, instead of opting to go through 50+ search results
 
### Your tests might break

In some cases tests depending on `<Select />` can now break due to limited results now being returned. For example, if your test happens to set itself up in the following way:

```
1. click on the country dropdown
2. find and click on the 'UK' option
3. ... do your actual test
```

It will fail when your country dropdown has more than 50 entries and is sorted alphabetically, because 'UK' wouldn't appear in the initial 50 results. To accomodate for this your test should search for the option of interest, or repeatedly click the show-more action until it can no longer do so.

(we faced this exact breakage in the past when we implemented a similar feature in another one of our frontend libraries)

## ✅ Checklist


packages/components/src/select/Select.js
```diff
@@ -344,14 +348,40 @@ export default class Select extends Component {
...
+        onClick={this.showMore.bind(this)}
+        onKeyPress={this.showMore.bind(this)}
+        className={classNames(
+          this.style('tw-dropdown-item--clickable'),
+          this.style('tw-dropdown-item--divider'),
+          this.style('show-more'),
+        )}
+      >
+        {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
+        <a>...</a>
```
dozortsev, line 375: why we need `a` tag here? have you tried `<li>...</li>`

dozortsev (approved):

PR #1098, opened 2021-08-13
Add manual file download handler to UploadInput

czebe:
## 🖼 Context

Sometimes it's not possible to directly link to a file (eg. when it's not publicly available, needs SCA controls before downloading). I'd like to introduce a callback handler that allows custom logic to override the link behaviour.

## 🚀 Changes

 - add the `onDownload` prop to `<UploadInput>`
 - update docs

## ✅ Checklist


packages/components/src/uploadInput/uploadItem/UploadItem.spec.tsx
```diff
@@ -105,4 +108,19 @@ describe('UploadItem', () => {
       expect(screen.queryByLabelText('Remove file ', { exact: false })).not.toBeInTheDocument();
     });
   });
+
+  describe('manual file download handler', () => {
+    it('should render the remove file button when deleteFile prop is supplied', () => {
```
matebek-tw, line 113: This test should have a different title, something like `it('calls onDownload handler provided through props')`.

packages/components/src/uploadInput/uploadItem/UploadItem.spec.tsx
```diff
@@ -105,4 +108,19 @@ describe('UploadItem', () => {
...
+      const onDownload = jest.fn();
+      renderComponent({ ...props, onDownload });
+      userEvent.click(screen.getAllByRole('link')[0]);
```
matebek-tw, line 116: Whenever it is possible, please try to query elements by text rendered on the UI (in this case probably the provided file name would work).

packages/components/src/uploadInput/uploadItem/UploadItem.spec.tsx
```diff
@@ -105,4 +108,19 @@ describe('UploadItem', () => {
...
+      expect(onDownload).toHaveBeenCalledTimes(1);
+      expect(onDownload).toHaveBeenCalledWith(props.file);
+    });
+
+    it('should not render the remove file button when canDelete prop is false', () => {
```
matebek-tw, line 121: This test probably should be outside of this `describe` block.

packages/components/src/uploadInput/uploadItem/UploadItemBody.tsx
```diff
@@ -1,16 +1,17 @@
...
 
-export const UploadItemBody = ({ children, url }: UploadItemLinkProps) => {
+export const UploadItemBody = ({ children, url, onDownload }: UploadItemLinkProps) => {
   if (!url) {
     return <div>{children}</div>;
   }
 
   return (
-    <a href={url} target="_blank" rel="noopener noreferrer">
+    <a href={url} target="_blank" rel="noopener noreferrer" onClick={onDownload}>
```
matebek-tw, line 14: I case the file is not publicly available, it can happen that it won't have an `url` as well, right? In that case, we might won't even render the link tag.

packages/components/src/uploadInput/uploadItem/UploadItem.spec.tsx
czebe: ignore, copy paste garbage

packages/components/src/uploadInput/uploadItem/UploadItemBody.tsx
czebe: If it doesn't have an `url`, it's not downloadable at all.
It's a different usecase, and `<UploadItemBody>` handles this well.

Here I'm dealing with cases where the file DOES have a URL, but you want to prevent the default behaviour. (eg. we need to download the file through SCA).

packages/components/src/uploadInput/uploadItem/UploadItem.spec.tsx
czebe: What's the reason for this? I had the impression that the official RTL recommendation was to use `getByRole` as this makes test far more stable and at the same time ensure accessibility. Doesn't make a big difference in this case, but it could break tests easily where copy is defined via `react-intl`.

matebek-tw: That is true in some cases, though one can argue that if the copy changes (from "save" to "delete") it worth a test to fail. :) Generally to try to access the UI how the users would see it rendered. 

But you are right, `getByRole` is the first in the query priority list https://testing-library.com/docs/queries/about/#priority. Can we access the link with `screen.getByRole('link', {name: file.name})` instead of using the array index?

matebek-tw (approved):

StevePole:
Could you elaborate on this?
> (eg. when it's not publicly available, needs SCA controls before downloading).
What do you mean by not publicly available?  And when would we need SCA controls?

I'm just thinking that this is not going to work in dynamic forms, as you have no way of specifying a special handler function.

czebe:
SCA is needed when the file contains PII. Otherwise you need to be logged in an have permissions to view the profile to download a file, but SCA is not required.

Initially we thought about generating a random URL (that expires after some time) to expose files for users without a Wise account (like an accountant). But this usecase seems the most risky so we'll go with the first two ones.

packages/components/src/uploadInput/uploadItem/UploadItem.spec.tsx
czebe: good idea!

czebe: Ok I tried this, but unfortunately partial text match (based on `name`) is not working. Putting the full text into the query selector beats the purpose of staying away from text-based selectors.

So I'd stick to the array access, which also ensures that the very first interactive item is the `download` button and not the `remove` one.

czebe: Regarding the text selectors: I see way too many copy fixes in our apps even after quite some time the app was released. Making minor copy tweaks should never break any tests, changes like the following happen so often:
`Uploaded` -> `Upload successful` -> `Upload successful.`
`Removing` -> `Removing...`
`Delete` -> `Remove` etc.

packages/components/src/uploadInput/uploadItem/UploadItem.tsx
```diff
@@ -1,29 +1,36 @@
-import React, { ReactElement } from 'react';
+import React, { ReactElement, MouseEvent } from 'react';
```
dozortsev, line 1: importing `React` is now redundant, could you please delete it from here?

dozortsev (approved):

packages/components/src/uploadInput/uploadItem/UploadItem.tsx
czebe: Why is it redundant? 
(I thought peerDependency is still at `"react": "^16.14 || ^17"`)

dozortsev: `16.14` also has this feature

czebe: ohh, I see 16.14 already behaves like 17...

PR #1089, opened 2021-08-10
feat!: new & accessible colours

dozortsev:
## 🖼 Context

We recently updated [Neptune colors](https://neptune.wise.com/design/colour/) to make them more accessible, plus introduced changes into our tokens naming to make them semantic (https://github.com/transferwise/neptune-tokens/pull/19).

## 🚀 Changes

- **BREAKING CHANGE**: updated `@transferwise/neptune-tokens` package to latest which brings different (new) set of CSS custom properties/LESS variables
- use new colors as CSS custom properties instead of old colors being used as LESS variables.
- updated docs:
      - [Working with styles](https://transferwise.github.io/neptune-web/branch/accessible-colours/about/Styles) - added more info about PostCSS setup.
      - [Tokens | Colors](https://transferwise.github.io/neptune-web/branch/accessible-colours/styles/tokens/Colors) - updated list of new color tokens
      - [Addons | Background utilities](https://transferwise.github.io/neptune-web/branch/accessible-colours/styles/addons/BackgroundUtilities) - updated list of `.bg-*` utility classes
- copied over content of some styles from Bootstrap to be able to adjust colors in them

## ✅ Checklist

- [ ] Make PR title meaningful and follow [the commit lint format](https://github.com/transferwise/neptune-web/blob/master/CONTRIBUTING.md#versioning-and-commit-lint) (it will appear in the changelog)
- [ ] Changes are tested and all tests pass
- [ ] Changes meet [accessibility standards](https://github.com/transferwise/neptune-web/blob/main/ACCESSIBILITY.md) and there are no violations in the console
- [ ] Changes work in all supported browsers (don't forget IE11)
- [ ] You've updated the documentation if necessary

PR #1099, opened 2021-08-15
feat: accessible colours

helendurranttw:
## 🖼 Context

Incorporating accessible colours into Neptune Web.

## 🤔 Considerations

Not sure if we should bring legacy variables into Neptune or fix them in Bootstrap.

## Still to do
- [ ] We're missing hover / active styles for links in alerts
- [ ] Need to refactor out usage of `color-text-control` and `color-base-white`
- [ ] Change `@color-border-focus` to use `--color-interactive-accent`
- [ ] We're missing `color-interactive-warning` - it's required for summary
- [ ] Fix borders on inputs like Select (they should be `color-interactive-secondary`)
- [ ] Fix the triangle on inline alerts
- [ ] Radio disabled states don't match Figma
- [ ] Update Switch to use `--color-interactive-secondary` instead of `@brand-smoke`
- [x] **Figma:** The grey in Switch uses `brand-smoke` - is there a semantic version?
- [x] **Figma:** Is the date input placeholder correct? It looks too dark
- [x] **Figma:** Should the radio button have a blue border on selection?
- [x] **Figma:** Date lookup - should the chevrons be interactive blue? What about the dates themselves? It's different from what we have

## ✅ Checklist


gusso:
> We're missing hover / active styles for links in alerts
We're missing color-interactive-warning - it's required for summary

Added in https://github.com/transferwise/neptune-tokens/pull/23.

> @color-border-focus was removed but we're still using it

We should use `interactive / accent` for this.

> Figma: The grey in Switch uses brand-smoke - is there a semantic version?
Figma: Is the date input placeholder correct? It looks too dark
Figma: Date lookup - should the chevrons be interactive blue? What about the dates themselves? It's different from what we have

Fixed on Figma, colours should be accurate now for these components.
  
> Figma: Should the radio button have a blue border on selection?

We could maaaybe, I tried but it looked a bit in focus, I don't think I would as a part of this.

helendurranttw:
Thank you!

> > Figma: Should the radio button have a blue border on selection?
> 
> We could maaaybe, I tried but it looked a bit in focus, I don't think I would as a part of this.

The reason I'm asking is it already does in [Neptune Web](https://transferwise.github.io/neptune-web/components/inputs/RadioGroup). I can remove it if that's wrong.

gusso:
> The reason I'm asking is it already does in [Neptune Web](https://transferwise.github.io/neptune-web/components/inputs/RadioGroup). I can remove it if that's wrong.

Ah! In that case, let's respect what's in the product, fixed it on Figma until we reconsider 👍

dozortsev:
@helendurranttw @gusso merged this branch into https://github.com/transferwise/neptune-web/pull/1089 (commit https://github.com/transferwise/neptune-web/pull/1089/commits/02d5e6276f4011110f57913a7003327fff128761)

PR #1104, opened 2021-08-17
chore: add browser compatibility eslint and stylelint plugin

joe-sayer:
## 🖼 Context

ESLint can flag browser compatibility problems in code and warn developer. The same can be done for CSS/LESS using stylelint plugin.

## ✅ Checklist


.eslintrc.js
```diff
@@ -1,5 +1,11 @@
 module.exports = {
-  extends: ['@transferwise', 'plugin:jsdoc/recommended'],
+  extends: ['@transferwise', 'plugin:jsdoc/recommended', 'plugin:compat/recommended'],
+  env: {
+    browser: true,
+  },
+  settings: {
+    polyfills: ['Promise', 'fetch', 'navigator', 'IntersectionObserver'],
```
matebek-tw, line 7: `eslint-plugin-compat` should read the `.babelrc` and shouldn't check for ES APIs transpiled by `babel`, but maybe because of the monorepo setup it couldn't figure it out for itself. Adding these polyfills solves the problem, though it would be nicer to make it work automatically.

stylelint.config.js
```diff
@@ -1,13 +1,22 @@
...
     'at-rule-no-vendor-prefix': true,
     'media-feature-name-no-vendor-prefix': true,
     'property-no-vendor-prefix': true,
     'selector-no-vendor-prefix': true,
     'value-no-vendor-prefix': true,
     'no-descending-specificity': null,
+    'plugin/no-unsupported-browser-features': [
+      true,
+      {
+        severity: 'warning',
```
matebek-tw, line 17: Setting the severity to `warning` to pass the CI step, we are relying on some partially supported browser features for IE11 like `lfexbox` or `calc`.

.eslintrc.js
```diff
@@ -1,5 +1,11 @@
 module.exports = {
-  extends: ['@transferwise', 'plugin:jsdoc/recommended'],
+  extends: ['@transferwise', 'plugin:jsdoc/recommended', 'plugin:compat/recommended'],
+  env: {
+    browser: true,
```
dozortsev, line 4: what this for?

cc @matebek-tw

.eslintrc.js
```diff
@@ -1,5 +1,11 @@
...
+  },
+  settings: {
+    polyfills: ['AbortController', 'IntersectionObserver', 'Promise', 'fetch', 'navigator'],
```
dozortsev, line 7: I'm wondering how we can avoid having this list as this means that we need maintain it

could you please point out to some doc where I could learn more about this?

.eslintrc.js
```diff
@@ -101,6 +107,7 @@ module.exports = {
     {
       files: ['*{spec,test}.{js,jsx,ts,tsx}'],
       rules: {
+        'compat/compat': 0,
```
dozortsev, line 110: what this rule about?

stylelint.config.js
dozortsev: `calc` seems supported by IE11 https://developer.mozilla.org/en-US/docs/Web/CSS/calc()#browser_compatibility

.eslintrc.js
dozortsev: strange that we have `Promise` here as it's belongs to Standard API https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise

stylelint.config.js
matebek-tw: It is partially supported, yes.

.eslintrc.js
matebek-tw: It is in the plugin [setup guide](https://www.npmjs.com/package/eslint-plugin-compat), guessing this is telling the plugin to check for browser compatibility issues.

.eslintrc.js
matebek-tw: It is in the standard documentation as well: https://github.com/amilajack/eslint-plugin-compat#adding-polyfills

matebek-tw: It is in the plugins documentation as well https://github.com/amilajack/eslint-plugin-compat#adding-polyfills

.eslintrc.js
matebek-tw: `0` is the error level, it means the rule is disabled for tests. https://eslint.org/docs/user-guide/getting-started#configuration

.eslintrc.js
matebek-tw: Yes I saw those issue threads when searching for a solution, and adding `Promise` to the polyfills is the most common workaround. As we probably won't introduce support for older browsers, we won't have to spend much time on maintaining this list.

.eslintrc.js
dozortsev, line 7: we should be able safely delete `navigator` from here as the only place we get eslint warning in [`deviceDetection`](https://github.com/transferwise/neptune-web/blob/DS-1197-browser-lint/packages/components/src/stepper/deviceDetection.js#L3) for `maxTouchPoints` prop, incase the props is not defined it would default to `false`

.eslintrc.js
dozortsev, line 4: this PR is great occasion to align our browserslist config, could you please create `.browserslistrc` at the root folder and delete same file from DF package and remove `"browserslist"` from root package.json to have single source of truth

.eslintrc.js
dozortsev, line 7: ideally this list should match the one we have one on [docs](https://transferwise.github.io/neptune-web/about/Setup#polyfills-for-web-browser-api) so it will be good to have single source for both spaces

I would probably ditch linting DF package for now and remove `AbortController` from the list

dozortsev, line 7: I don't know why but when I remove `IntersectionObserver` I don't get warning in places where we use it 🤔 (tried restarting VS Code and checked in different IDE (InteliJ))

dozortsev: no, we still have it

https://github.com/transferwise/neptune-web/blob/main/packages/components/src/common/hooks/useHasIntersected/useHasIntersected.js#L42

matebek-tw: Yeah I have no idea, but actually this is the expected behaviour as we are transpiling these APIs with `babel`.

.eslintrc.js
```diff
@@ -1,5 +1,11 @@
 module.exports = {
-  extends: ['@transferwise', 'plugin:jsdoc/recommended'],
+  extends: ['@transferwise', 'plugin:jsdoc/recommended', 'plugin:compat/recommended'],
```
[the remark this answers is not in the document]
matebek-tw: Could be same why `IntersectionObserver` is not raising an error anymore, maybe some library updates fixed bugs and now it can correctly detect our babel configuration.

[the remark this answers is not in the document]
dozortsev: both are Web API meaning that babel doesn't polyfill them in anycase

.eslintrc.js
dozortsev: https://github.com/transferwise/neptune-web/pull/1104#discussion_r701029526

dozortsev (approved):
re. `eslint-plugin-compat`

I checked it's repo and it looks like this packing is lacking support/maintenance, all the weird issues that we encountered are already known and just needs to be fixed

- https://github.com/amilajack/eslint-plugin-compat/issues/445
- https://github.com/amilajack/eslint-plugin-compat/issues/258
- https://github.com/amilajack/eslint-plugin-compat/issues/168

so I suggest we give it a try and merge what we currently have,

in future maybe if we find time try contribute and improve plugin

PR #1108, opened 2021-08-20
feat: add onClick to the Nudge

helendurranttw:
## 🖼 Context

We want to add some tracking to measure how many people click the nudge action. To do this we need an onClick handler in the Nudge.

## 🚀 Changes

Added an onClick handler to the nudge.

## 🤔 Considerations

Ideally we would rework the API to match that of the Alert. We should have consistent APIs for all actions, for example:

```
// Link based actions
 action: PropTypes.shape({
    'aria-label': PropTypes.string,
    href: PropTypes.string.isRequired,
    target: PropTypes.string,
    text: PropTypes.node.isRequired,
    onClick: PropTypes.func,
  }),

// Button based actions
 action: PropTypes.shape({
    text: PropTypes.node.isRequired,
    onClick: PropTypes.func,
  }),
```

This would of course be a breaking change that would require a codemod. As such, I've raised this PR to do this initial change and add the prop, and if you decide you want to do the above, we can address it as part of the next major release.

## ✅ Checklist


packages/components/src/nudge/Nudge.js
```diff
@@ -38,6 +39,7 @@ Nudge.propTypes = {
 Nudge.defaultProps = {
   id: null,
   className: null,
+  onClick: null,
```
andreapiras, line 42: should this be undefined?

matebek-tw (approved):

dozortsev (approved):

PR #1109, opened 2021-08-23
feat(docs): drop less support and migrate to css variables

matebek-tw:
## 🖼 Context

We are not using any features of less for the docs page, so with dropping less compilation step we can simplify things a bit.

Without running `yarn build` first, the compiled `main.css` file was missing (imported in `docs/_app.js`). This error is solved by these changes as well.

## 🚀 Changes

- Remove less compilation step from `docs/package.json`
- Migrate the styles from `main.less` to `main.css`

## ✅ Checklist


packages/docs/public/static/assets/main.css
```diff
@@ -1,11 +1,8 @@
-@import (reference) '@transferwise/neptune-css/src/less/variables/_typography.less';
-@import (reference) '@transferwise/neptune-css/src/variables/neptune-tokens.less';
-
 /* Typography */
 
 h1[id] + p,
 h1.colored-dot + p {
-  font-size: @font-size-large;
+  font-size: var(--font-size-20);
```
dozortsev, line 5: but this CSS custom props needs to be preprocessed by some tool, does Nextjs provides PostCSS out of the box?

dozortsev (approved):
👍

packages/docs/public/static/assets/main.css
matebek-tw: Yes it does: https://nextjs.org/docs/advanced-features/customizing-postcss-config

Though the docs says it is not possible to compile variables, the compiled `.css` files looks good to me:

```css
h1.colored-dot+p,h1[id]+p{font-size:1.25rem;font-size:var(--font-size-20,1.25rem)}
```

> CSS variables are not compiled because it is [not possible to safely do so](not possible to safely do so). If you must use variables, consider using something like Sass variables which are compiled away by Sass.

matebek-tw: Because we have a `postcss.config.js` file in the root of the project, Next.js reads that configuration and applies the same rules to the css we have defined there. https://nextjs.org/docs/advanced-features/customizing-postcss-config#customizing-plugins

Though this is an important warning:
> Warning: When you define a custom PostCSS configuration file, Next.js completely disables the default behavior.

dozortsev: okey, but are you 100% sure re. this

> Because we have a postcss.config.js file in the root of the project..

can you try edit our `postcss.config.js` config (e.g comment out `@arshaw/postcss-custom-properties` plugin) and see how it would handle CSS custom properties

matebek-tw: Yes, I tried, removing the plugin will drop the fallbacks as intended.

dozortsev: at some point we will need to sort of level up our PostCSS config as it's missing loads of plugins (transformers) for new CSS features e.g ([nesting](https://www.npmjs.com/package/postcss-nesting) or [media-minmax](https://www.npmjs.com/package/postcss-media-minmax))

ideally we should have been using [postcss-preset-env](https://www.npmjs.com/package/postcss-preset-env) but official [postcss-custom-properties](https://www.npmjs.com/package/postcss-custom-properties) doesn't provide proper fallback https://github.com/postcss/postcss-custom-properties/issues/169 😞

nabaroa (commented):
+1 to dropping less variables in favor of custom properties.
Out of curiosity: Why changing main.less -> to main.css? Just for simplicity?

matebek-tw:
Because `.less` requires an additional compilation step, while we can use `.css` without that, while Next.js still applies all our postcss plugins during the build to our styles. I don't feel it adds any value to such a simple codebase like our documentation, and we indeed haven't used any of the features of less.

PR #1057, opened 2021-07-22
chore!: remove deprecated props

ghost:
## 🖼 Context

Removing deprecated props for the next major release.

### Alert

- Remove `deprecatedTypeMap` mapping deprecated sentiments to new ones
- Remove info, error, success types from `type` prop
- Remove `arrow`, `children` and `dismissable` props
- Make `message` prop required

### Button

- Remove `typeClassMap` to map deprecated `primary`, `pay`, `secondary`, `danger`, `link` types
- Remove `xs` size
- Remove `legacyUtils`

### CircularButton

- remove `typeClassMap` and compose button class with `type`

### Sentiment

- Remove `info`, `error` and `success` sentiments

### InlineAlert

- Remove `typeClassMap` to map `ERROR` sentiment to `danger`

### Markdown

- Remove log how to use a specific prop - this should be documented with codeblock

### Popover

- Remove `left-top`, `right-top`, `left-bottom` and `right- bottom` `preferredPlacement` values
- Remove `deprecatedPlacements` mapping

### Summary

- Remove `content`, `help` and `illustration` props

### Switch

- Remove warning about making a prop required in the indefinite future

### Typeahead

- Replace `error` sentiment with `negative`

### utilities

- Remove `deprecatedProp` utility functions `deprecatedMessage` and `deprecate`
- Remove `logActionRequired` utility functions `logActionRequired` and `logActionRequiredIf`
- Remove bootstrap styled `alerts.html` test component

## 🤔 Considerations

- Should we update `InlineAlert` component `type` prop to match with the updated Sentiments? (wasn't marked as deprecated)
- Displaying a warning how to use `Markdown` component props feels inappropriate, should live in documentation/code block comment instead, but it has a real value so probably we should keep it for now
- `Typeahead` component `error` sentiment wasn't marked as deprecated

## ✅ Checklist

- [x] Make PR title meaningful and follow [the commit lint format](https://github.com/transferwise/neptune-web/blob/master/CONTRIBUTING.md#versioning-and-commit-lint) (it will appear in the changelog)
- [x] Changes are tested and all tests pass
- [ ] Changes meet [accessibility standards](https://github.com/transferwise/neptune-web/blob/main/ACCESSIBILITY.md) and there are no violations in the console
- [ ] Changes work in all supported browsers (don't forget IE11)
- [ ] You've updated the documentation if necessary

PR #1113, opened 2021-08-26
feat: new `Chips` component

jacklesliewise:
## 🖼 Context

New `Chip` component

## Demo

- [`Chip`](https://transferwise.github.io/neptune-web/branch/BIZGRO-1831-chip/components/internal/Chip)
- `Chip` in [`Chips`](https://transferwise.github.io/neptune-web/branch/BIZGRO-1831-chip/components/inputs/Chips)
- `Chip` in [`Typeahead`](https://transferwise.github.io/neptune-web/branch/BIZGRO-1831-chip/components/inputs/Typeahead)

## 🚀 Changes

- New `Chips` component
- New `Chip` component
  - Used in `Chip`
  - Used in `Typeahead`
- Updated `CloseButton` to be able to take additional props for the cross icon
- Added to docs

## 🤔 Considerations

See https://transferwise.slack.com/archives/CJNBX9CP6/p1630573091018000

## ✅ Checklist


sbrandwoo:
My initially feeling is there are two pieces here that have been mixed together: a visual Chip component, and a stateful form component.

We're currently using a visually similar component with investments:
Selection_398

Because our chips act more like radio buttons, and don't allow multi-select, we would be unable to use this Neptune Chips component.

I'd rather see Neptune focus on the visual aspect and provide a `<Chip selected={true} label="Payroll" onClick={...} />` and let my app handle which chips are selected or not. This has the bonus that all state is kept inside my app, so if the Chips are part of a larger form all of the state is held together inside Formik.

Neptune could perhaps provide both, a `Chip` and a `MultiSelectChips`.

sbrandwoo:
It's possible that I've missed something since there's a lot of code about "choices" and "filters". Please can you add the various styles to https://transferwise.github.io/neptune-web/branch/BIZGRO-1831-chip/components/inputs/Chips ?

packages/components/src/chips/Chips.tsx
```diff
@@ -0,0 +1,155 @@
...
+    clearedValue,
+  }: {
+    prevState: Set<ChipValue>;
+    updatedState: Set<ChipValue>;
+    clearedValue: ChipValue;
+  }) => void;
+  /** Optional list of pills which are selected by default for filter chips. */
+  initialSelectedFilters?: ChipValue[];
+  /** Optional default selected pill for choice chips. */
+  initialSelectedChoice?: ChipValue;
```
sbrandwoo, line 43: Because the Chips contains the state, rather than taking a `selected` prop, there might be some situations where it's difficult to manage this state. For example if you wanted a "Clear filters" button next to the chips, you'd be unable to tell the Chips to deselect. There's no way to externally adjust which chips are selected after the component has been rendered.

sbrandwoo:
Demo page and new API are looking good :+1:

packages/components/src/chips/Chips.spec.js
```diff
@@ -0,0 +1,93 @@
...
+    it(`will pass previous state, updated state, and cleared value on clear`, () => {
+      render(<Chips {...filterProps} />);
+      fireEvent.click(screen.getAllByRole('button')[3]);
+      expect(filterProps.onChange).toHaveBeenLastCalledWith({
+        selectedValue: 'payments',
+        isEnabled: false,
+      });
+    });
+  });
+});
```
matebek-tw, line 93: Could you please rename the test file to `.ts` as well?

packages/components/src/chips/Chips.tsx
```diff
@@ -0,0 +1,103 @@
...
+  /** Callback which is invoked when a chip is selected or deselected */
+  onChange: ({
+    isEnabled,
+    selectedValue,
+  }: {
+    isEnabled: boolean;
+    selectedValue: ChipValue;
+  }) => void;
+  /** Used to describe the purpose of the chips. */
+  'aria-label': string;
```
matebek-tw, line 29: We have a `commonProps/AriaLabelProperty` for this.

packages/components/src/chips/Chips.less
```diff
@@ -0,0 +1,65 @@
...
+
+.np-chip {
+  color: var(--color-text-secondary);
+  background-color: transparent;
+  height: var(--size-32);
+  margin-right: var(--size-8);
+  padding: 0 var(--size-12);
+  border-radius: var(--size-16);
+  border: 1px solid var(--color-border-neutral);
+  display: flex;
```
andreapiras, line 15: You can also use utility classes: d-flex etc
Benefits:
- less css generated
- they'll appear in snapshot so they have some sort of testing

packages/components/src/chips/Chips.less
```diff
@@ -0,0 +1,65 @@
...
+  align-items: center;
+  justify-content: space-between;
+  transition: color 0.15s ease-in-out, border-color 0.15s ease-in-out, background-color 0.15s ease-in-out;
```
andreapiras, line 18: This is repeated all over the codebase, I think it should be in neptune-css and standardized.

packages/components/src/chips/Chips.tsx
```diff
@@ -0,0 +1,101 @@
...
+  multiple,
+}: ChipsProps): ReactElement => {
+  const intl = useIntl();
+
+  const isSelected = (value: ChipValue) => {
+    if (Array.isArray(selected)) {
+      return selected.includes(value);
+    }
+    return selected === value;
+  };
```
andreapiras, line 49: const isSelected = (value: ChipValue) => Array.isArray(selected) ? selected.includes(value)  :  selected === value

also you can reuse https://github.com/transferwise/neptune-web/blob/main/packages/validation/src/type-validators/index.js#L9

packages/components/src/chips/Chips.spec.tsx
```diff
@@ -0,0 +1,100 @@
...
+describe('Chips', () => {
+  describe('Choice Chips', () => {
+    it(`renders as expected`, () => {
+      expect(render(<Chips {...choiceProps} />).container).toMatchSnapshot();
+    });
+
+    it(`will pass previous state, updated state, and selected value on select`, () => {
+      render(<Chips {...choiceProps} />);
+      const accounting = screen.queryByText('Accounting');
+      if (accounting) {
```
andreapiras, line 64: Do we need this if? I guess test will fail if screen.queryByText won't find anything

packages/components/src/chips/Chips.spec.tsx
```diff
@@ -0,0 +1,100 @@
...
+      expect(filterProps.onChange).toHaveBeenLastCalledWith({
+        selectedValue: 'accounting',
+        isEnabled: true,
+      });
+    });
+
+    it(`will pass previous state, updated state, and cleared value on clear`, () => {
+      render(<Chips {...filterProps} />);
+      fireEvent.click(screen.getAllByRole('button')[3]);
+      expect(filterProps.onChange).toHaveBeenLastCalledWith({
```
andreapiras, line 94: Should we use toHaveBeenCalled only as I would expect nothing else gets called

packages/components/src/chips/Chips.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  const handleOnChange = (selectedValue: ChipValue, isCurrentlyEnabled: boolean) => {
+    onChange({ isEnabled: !isCurrentlyEnabled, selectedValue });
```
andreapiras, line 52: What is isEnabled used for? Can you not guess that from selectedValue? 🤔

packages/components/src/chips/Chips.tsx
```diff
@@ -0,0 +1,101 @@
...
+          >
+            <span className="np-chip-label">{chip.label}</span>
+            {multiple && chipSelected ? (
+              <button
+                className="btn-unstyled m-l-1 np-chip-close-button"
+                type="button"
+                aria-label={intl.formatMessage(messages.ariaLabel, { choice: chip.label })}
+                onClick={() => handleOnChange(chip.value, chipSelected)}
+              >
+                <CrossCircleIcon filled size="16" />
```
andreapiras, line 91: I think 16 is the default

packages/components/src/chips/Chips.story.tsx
```diff
@@ -0,0 +1,109 @@
...
+      selected={selected}
+      multiple
+      onChange={({ selectedValue, isEnabled }) => {
+        if (isEnabled) {
+          setSelected([...selected, selectedValue]);
+        } else {
+          const updatedSelected = selected.filter((value) => selectedValue !== value);
+          setSelected(updatedSelected);
+        }
+      }}
```
andreapiras, line 27: Optional:
onChange={({ selectedValue, isEnabled }) =>
        isEnabled
          ? setSelected([...selected, selectedValue])
          : setSelected(selected.filter((value) => selectedValue !== value))
      }

packages/components/src/chips/Chips.tsx
```diff
@@ -0,0 +1,101 @@
...
+            key={chip.value}
+            className={classNames('np-chip', {
+              'np-chip--selected': chipSelected,
+              'p-r-1': multiple && chipSelected,
+            })}
+            {...wrapperProps}
+          >
+            <span className="np-chip-label">{chip.label}</span>
+            {multiple && chipSelected ? (
+              <button
```
andreapiras, line 85: Can we not reuse Button here?

packages/components/src/chips/Chips.tsx
andreapiras: We do have a similar close button used in other places. Should this be consistent everywhere?
https://github.com/transferwise/neptune-web/blob/main/packages/components/src/common/closeButton/CloseButton.js

andreapiras: Or maybe this should be a variant of the existing one

packages/components/src/chips/Chips.spec.tsx
matebek-tw: `queryByText` returns `null` if no elements match. https://testing-library.com/docs/queries/about/#types-of-queries You could rewrite it using `getByText`

```javascript
fireEvent.click(screen.getByText('Accounting'))
```

packages/components/src/chips/Chips.less
dozortsev: you mean values of `transition` prop (like we've done in [`Tabs.less`](https://github.com/transferwise/neptune-web/blob/BIZGRO-1831-chip/packages/components/src/tabs/Tabs.less#L5-L6)) or some utility CSS class?

jacklesliewise: I'm not 100% sure about where it would go in the codebase if it was standardised

dozortsev: yeah, I would probably create e.g `transition.css` in `packages/css/src/props` where would define those as CSS custom properties

packages/components/src/chips/Chips.tsx
jacklesliewise: Since it's JS I'd need to add a declaration for the module

packages/components/src/chips/Chips.less
```diff
@@ -0,0 +1,65 @@
...
+  white-space: nowrap;
+
+  .np-chip-label {
+    // TODO: we need this to offset text alignment vertically, revisit if this gets fixed
+    padding-top: 2px;
```
dozortsev: I think you could try apply one of Neptune CSS utility classes instead of this, I think `.p-r-2` would do same

https://transferwise.github.io/neptune-web/styles/addons/SpacingUtilities

packages/components/src/chips/Chips.tsx
```diff
@@ -0,0 +1,101 @@
...
+  multiple?: boolean;
+}
+
+const Chips = ({
+  chips,
+  onChange,
+  selected,
+  'aria-label': ariaLabel,
+  className,
+  multiple,
```
dozortsev, line 40: this prop is optional, could you please specify default value for it

packages/components/src/index.ts
```diff
@@ -46,7 +47,6 @@ export { default as Dimmer } from './dimmer';
 export { default as SlidingPanel } from './slidingPanel';
 export { default as Drawer } from './drawer';
 export { default as Sticky } from './sticky';
-export { default as Chip } from './chip';
```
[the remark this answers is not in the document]
jacklesliewise: How do we do the deprecation warning?

packages/components/src/chips/Chips.tsx
```diff
@@ -0,0 +1,101 @@
...
+import messages from './Chips.messages';
+
+export type ChipValue = string | number;
+
+export type Chip = {
+  value: ChipValue;
+  label: string;
+};
+
+export interface ChipsProps extends CommonProps, AriaLabelProperty {
```
dozortsev, line 17: we usually use `type` for component prop types in this repo, any specific reason to use `interface` here? 🙂

packages/components/src/chips/Chips.tsx
jacklesliewise: Yeah when it's a choice chip you don't need `isEnabled`, but since for filter chips you can select and deselect any chip we'll need the state

packages/components/src/chips/Chips.story.tsx
jacklesliewise: I think the `if` `else` is a bit more readable, but that's my personal preference and happy to update if the ternary is preferred

packages/components/src/chips/Chips.tsx
jacklesliewise: I considered this but I think the component is just too different from what `Button` is normally used for and would be overwriting most of the styles either way

packages/components/src/chips/Chips.less
andreapiras: yeah I think a custom prop would be good

packages/components/src/index.ts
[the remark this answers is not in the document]
andreapiras: Here's an example https://github.com/transferwise/neptune-web/blob/main/packages/components/src/summary/Summary.js#L148

packages/components/src/chips/Chips.tsx
andreapiras: I guess you can still manage that from parent no?

packages/components/src/chips/Chips.story.tsx
andreapiras: it s ok as I said it was optional :)

packages/components/src/chips/Chips.tsx
andreapiras: As I said somewhere above maybe this should be a variant of closeButton

packages/components/src/chips/Chips.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const isSelected = (value: ChipValue) =>
+    Array.isArray(selected) ? selected.includes(value) : selected === value;
+
+  const handleOnChange = (selectedValue: ChipValue, isCurrentlyEnabled: boolean) => {
+    onChange({ isEnabled: !isCurrentlyEnabled, selectedValue });
+  };
+
+  return (
+    <div className={classNames('np-chips d-flex', className)} aria-label={ariaLabel}>
+      {chips.map((chip) => {
```
dozortsev, line 53: I wonder whether we need to render parent `div` if `chips` array is empty 🤔

packages/components/src/index.ts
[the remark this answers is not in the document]
dozortsev: Ideally we should use new Chips in Typeahead and delete 'typeaheadChip', have we considered this?

packages/components/src/typeahead/typeaheadChip/TypeaheadChip.less
```diff
@@ -7,6 +7,11 @@
   height: var(--size-32);
   border-radius: 100px;
 
+  .chip-label {
+    // TODO: we need this to offset text alignment vertically, revisit if this gets fixed
+    padding-top: 2px;
```
dozortsev, line 12: [`.p-t-2`](https://transferwise.github.io/neptune-web/branch/BIZGRO-1831-chip/styles/addons/SpacingUtilities)

packages/components/src/chips/Chips.tsx
```diff
@@ -0,0 +1,100 @@
+import { CrossCircle as CrossCircleIcon } from '@transferwise/icons';
+import classNames from 'classnames';
+import React, { ReactElement } from 'react';
```
dozortsev, line 3: importing `React` is now redundant in this repo 🙂

packages/components/src/chips/Chips.tsx
dozortsev: yeah, I would try reuse `CloseButton` component here, but you would need adjust it for this case, e.g expose `aria-label` and `filled` (for icon design) props

packages/components/src/index.ts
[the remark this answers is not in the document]
dozortsev: @andreapiras we agreed we don't use this anymore, this approach doesn't really work, this neither push consumer to migrate away from deprecated stuff not for us, we end up having loads of legacy with backward compatibility stuff which very painful to remove e.g https://github.com/transferwise/neptune-web/pull/1057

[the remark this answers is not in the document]
dozortsev: I'm not 100% sure but I think we better introduce breaking change

packages/components/src/typeahead/typeaheadChip/TypeaheadChip.less
jacklesliewise: Isn't that 16px?

dozortsev: yeah, my bad, sorry

packages/components/src/chips/Chips.tsx
jacklesliewise: Do we need to cover this edge case?

packages/components/src/chips/Chips.tsx
jacklesliewise: Nope just a force of habit, I can change to `type` if that's the preference 👍

packages/components/src/chips/Chips.tsx
jacklesliewise: Is `undefined` not valid as a default value?

packages/components/src/index.ts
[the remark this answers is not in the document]
jacklesliewise: I don't think there are any consumers of `Chip` so perhaps it could just be a minor release?

packages/components/src/chips/Chips.tsx
jacklesliewise: Yeah true, but I think there's no harm in having a bit of extra data in the payload, especially in an object where it can be ignored - wdyt?

packages/components/src/chips/Chips.tsx
dozortsev: okey, this is something that we can fix later if necessary

packages/components/src/index.ts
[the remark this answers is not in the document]
dozortsev: re. https://github.com/transferwise/neptune-web/pull/1113#discussion_r713812164

here is Chip for both components

https://github.com/transferwise/neptune-web/pull/1153

with this change it won't be a breaking change

dozortsev (approved):

matebek-tw:
When using Chips with a screenreader, there are some weird behaviour when the chip is active, it recognised as a `none` element, not as a `button`.

Screenshot 2021-09-28 at 19 32 06
Screenshot 2021-09-28 at 19 32 20
Screenshot 2021-09-28 at 19 33 03

matebek-tw (approved):

PR #1124, opened 2021-09-01
fix: Switch disabled state

joe-sayer:
## 🖼 Context

When Switch component is disabled there is no styling changes and the onClick handler still functions. Ideally Switch should have no onClick when it is `disabled`.

## 🚀 Changes

 
- Added `--disabled` class when `disabled`
- Check `disabled` before passing onClick handler
- Added a new Switch story for disabled state

## ✅ Checklist


packages/components/src/switch/Switch.tsx
```diff
@@ -59,7 +60,7 @@ const Switch = (props: Props): ReactElement => {
       aria-labelledby={ariaLabelledby}
       id={id}
       aria-disabled={disabled}
-      onClick={onClick}
+      onClick={!disabled ? onClick : undefined}
       onKeyDown={handleKeyDown}
```
dozortsev, line 64: `onClick` function can also we invoked by pressing <kbd>Space</kbd> on focused component

packages/components/src/switch/Switch.story.tsx
```diff
@@ -32,3 +32,28 @@ export const Basic = () => {
...
+      />
+
+      <Switch
+        checked
+        disabled
+        className="a-class-name"
+        aria-labelledby="labelID"
+        id="switchId"
+        onClick={() => setCheck(!checked)}
+      />
```
dozortsev, line 56: why having two same Switches here?

matebek-tw: One of them is `checked` and `disabled`.

dozortsev (approved):

PR #1128, opened 2021-09-02
fix: convertFinalStepToDynamicLayout to conform to specification

JulianG-TW:
## 🖼 Context

The [final step specification](https://transferwise.atlassian.net/wiki/spaces/DF/pages/1182042085/Full+V3+Json+Contract#Final-step) (and the dynamic-interfaces library) use just a string for images. (instead of an object with "url" and "text" properties)

The `convertStepToLayout` function in LayoutServices needs to properly convert images in final steps.

## 🚀 Changes

 Updates `convertFinalStepToDynamicLayout` to use a new `convertFinalStepImageToDynamicImage` function.

**Before:**  the function expected a `description.image` as an object:
```json
{
  "type": "final",
  "details": {
    "title": "Details verified *",
    "description": "We do this extra check to make sure your money goes to the right place....",
    "image": {
      "url": "https://wise.com/public-resources/assets/spot-illustrations/dedicated_fraud_team.png"
    }
  },
  "actions": []
}
```

**After:** the function also accepts `description.image` to be a string, which is aligned with the [specification(https://transferwise.atlassian.net/wiki/spaces/DF/pages/1182042085/Full+V3+Json+Contract#Final-step):
```json
{
  "type": "final",
  "details": {
    "title": "Details verified *",
    "description": "We do this extra check to make sure your money goes to the right place....",
    "image": "https://wise.com/public-resources/assets/spot-illustrations/dedicated_fraud_team.png"
  },
  "actions": []
}
```

## 🤔 Considerations

~This may be a breaking change if clients were relying on this (potentially wrong) implementation.~
Not a breaking change, because the function now accepts both a string and an object. (since eb346c03d18de83a52c46a6b53e995f4b854c774)

## ✅ Checklist


packages/dynamic-flows/src/flow/layoutService/LayoutService.js
```diff
@@ -52,7 +52,23 @@ function convertFormStepToDynamicLayout(step) {
...
+
+  if (step.details) {
+    if (step.details.title) {
+      layout.push(convertStepTitleToDynamicHeading(step.details.title));
+    }
+    if (step.details.image) {
+      const image =
+        typeof step.details.image === 'string'
+          ? convertFinalStepImageToDynamicImage(step.details.image)
+          : convertStepImageToDynamicImage(step.details.image);
```
StevePole, line 65: Why is this necessary?

JulianG-TW: It's not strictly necessary. It only ensures backwards compatibility (which is arguably outside of the spec)
If we only use `convertFinalStepImageToDynamicImage` then I think we should mark this PR as BRAKING CHANGE.
**I'm happy with either option**, only slightly prefer avoiding the breaking change and major version bump as a result.
Please advice.

JulianG-TW: I've removed this

StevePole (approved):
👍

PR #1131, opened 2021-09-06
[VUX-950] Update persist async to work with `blob` type

richardbata:
## 🖼 Context
We introduced a schema type of `blob` only for persist async where a file upload component is shown, and the file gets uploaded using as a multipart file.

Note, later the camera upload will be added to peristAsync as well

## 🚀 Changes
- Custom component for `blob` which directly uses the `Upload` component
- Tweak to validation async
- Tweak to the `Upload` component within `@transferwise/components`

## 🤔 Considerations
Happy flow

https://user-images.githubusercontent.com/51439827/132242327-c4130c55-c8b6-42fe-b668-88b20f4bac05.mov

Sad flow

https://user-images.githubusercontent.com/51439827/132242119-7d89cc82-0290-458a-9fd9-bb9959130dc8.mov

## ✅ Checklist


packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/blob/PersistAsyncBlobSchema.js
```diff
@@ -0,0 +1,123 @@
...
+        animationDelay={700}
+        maxSize={props.schema.persistAsync.schema.maxSize}
+        usLabel={props.schema.title}
+        usPlaceholder={props.schema.description}
+        httpOptions={{
+          url: baseUrl + url,
+          method,
+          fileInputName: props.schema.persistAsync.param,
+          headers: {
+            'X-Access-Token': 'Tr4n5f3rw153',
```
StevePole, line 66: ?

richardbata: For cookie based authentication

packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/PersistAsyncSchema.js
```diff
@@ -4,35 +4,62 @@ import Types from 'prop-types';
...
+        schema={props.schema}
+        model={props.model}
+        errors={persistAsyncError || props.errors}
+        onPersistAsync={props.onPersistAsync}
+        onChange={props.onChange}
+      />
+    );
+  }
+
+  if (!supportedControlTypes.has(getControlType(props.schema.persistAsync.schema))) {
```
StevePole, line 62: Why do we need to do this?  

We're going to use BasicTypeSchema anyway.  BasicTypeSchema can only return strings, numbers (& integers), and booleans.  I don't see why it matters which control generates those basic pieces of data?

packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/blob/PersistAsyncBlobSchema.js
StevePole: How would we do this if it was open source?  We're perhaps not at this point yet, but we'd like to provide a reference implementation for partners.  With that in mind, it feels a bit wrong to have this in here.

(also it is open source already :-) )

richardbata: Next to URL in the schema we could have HEADERS: {}, right? 
Its needed everywhere where there is URL though, so persistAsync, validationAsync...

packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/PersistAsyncSchema.js
richardbata: The onBlur behaviour messes things up for the 'not supported' components

StevePole: What onBlur behaviour is problematic?

Ideally I'd avoid, but if we need it, could we switch to a deny-list rather than allow-list.  All controls should ideally work.

packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/PersistAsyncSchema.js
```diff
@@ -4,35 +4,62 @@ import Types from 'prop-types';
...
+        required={props.required}
+        submitted={props.submitted || fieldSubmitted}
+        schema={props.schema}
+        model={props.model}
+        errors={persistAsyncError || props.errors}
+        onPersistAsync={props.onPersistAsync}
+        onChange={props.onChange}
+      />
+    );
+  }
```
JulianG-TW, line 60: I love that we now have a separate component (schema) for Blobs.

However, I don't like such an "early return" in a react component. Have you considered having a "root" `PersistAsyncSchema` component that either renders `PersistAsyncBlobSchema` or `PersistAsyncWhateverSchema` ?

Actually, it could also guard against unsupported control types.
This way, `PersistAsyncSchema` could be a component that only makes a decision to render A, B, or throw.

There may be a reason this is not possible/desirable?

JulianG-TW (commented):
Looks good to me. Just left a comment.

packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/PersistAsyncSchema.js
richardbata: These are the main culprits

// RADIO: 'radio', - no onBlur event
// SELECT: 'select', - onBlur event happens too late, bad UX
// DATELOOKUP: 'date-lookup', - no onChange event
// UPLOAD: 'upload', - no onBlur event

packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/PersistAsyncSchema.js
richardbata: Makes sense to me :)

packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/PersistAsyncSchema.js
StevePole: What is the effect though?  What goes wrong?

richardbata: No call in most cases

StevePole: Ok, I took a look at the onBlur implementation and I get it now: we use onBlur as the trigger for the persistAsync.  For the controls that don't have onBlur, it would make logical sense to trigger persist using onChange. e.g. that would make sense for all of radio, datelookup and upload.  Select is an interesting one, onChange could also be used for that, it might end up triggering several times if the customer was using keyboard controls, but that's probably ok. 

Not sure why Date lookup doesn't have an onChange?!  Presumably it has something equivalent with a different name that just needs to be mapped correctly?  How do we update the model otherwise?

richardbata: Let me fix it for those components then

packages/dynamic-flows/src/formControl/FormControl.js
```diff
@@ -6,14 +6,15 @@ import {
   PhoneNumberInput,
   RadioGroup,
   Select,
+  Tabs,
```
StevePole, line 9: Tabs?

packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/blob/PersistAsyncBlobSchema.js
StevePole: That's a good question.  I was thinking it would be something we'd expose to the consumer, either via props, or a service they can interact with to set them.  But it could also be in the schema.

packages/dynamic-flows/src/formControl/FormControl.js
richardbata: Just a reorder 👇

packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/basic/PersistAsyncBasicSchema.js
```diff
@@ -0,0 +1,175 @@
...
+        headers: {
+          'Content-Type': 'application/json',
+          'X-Access-Token': 'Tr4n5f3rw153',
+        },
+        body: JSON.stringify(requestBody),
+        signal,
+      }),
+    );
+    props.onPersistAsync(persistAsyncFetch);
+    const response = await persistAsyncFetch;
```
JulianG-TW, line 70: I know this comes as-is from the previous component, but

I'm quite sure we don't need a `QueryablePromise` in this case. 
We seem to be only "awaiting" the promise, without using any of the "queryable" functions (`isPending`, `isRejected`, `isFulfilled`)

packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/basic/PersistAsyncBasicSchema.js
```diff
@@ -0,0 +1,175 @@
...
+export const getErrorFromResponse = (errorProperty, response) => {
+  return response.validation?.[errorProperty];
+};
+
+const controlTypesWithPersistOnChange = new Set([
+  FormControlType.RADIO,
+  FormControlType.SELECT,
+  FormControlType.DATELOOKUP,
+  FormControlType.FILE,
+  FormControlType.UPLOAD,
```
JulianG-TW, line 30: Aren't control types `FILE` and `UPLOAD` always of schema type `blob`?
Or is this for text files? (Just curious)

JulianG-TW (approved):
Looks good! I left a comment just in case.

packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/basic/PersistAsyncBasicSchema.js
richardbata: Before they were just 
`{ type: string, format: base64url }`
now they can also be a blob type but only within a persistAsync

packages/dynamic-flows/src/common/requirements.js
```diff
@@ -360,6 +362,7 @@ const prepValidationMessages = (field, validationMessages) => {
   }
 };
 
+// TODO: refactor to align with SchemaFormControl.getControlType
```
richardbata, line 365: @StevePole Can I refactor these two functions to a common one?

packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/PersistAsyncSchema.js
richardbata: Should now work for all pending on the getControlType refactor

packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/basic/PersistAsyncBasicSchema.js
richardbata: Its returned on the `onPersistAsync` callback and could potentially be used then

packages/dynamic-flows/src/common/requirements.js
```diff
@@ -360,28 +365,37 @@ const prepValidationMessages = (field, validationMessages) => {
   }
 };
 
-const getControlType = (field) => {
-  if (field.control) {
-    return field.control.toLowerCase();
+const getControlType = (schema) => {
+  if (schema.control) {
+    if (isOneOfSchema(schema) && schema.oneOf.length > 3) {
```
StevePole, line 370: Why do we need this one?

packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/basic/PersistAsyncBasicSchema.js
StevePole: What is the difference between FILE and UPLOAD?  Aren't these the same, is this just because we've incorporated the (unnecessary) backwards compatibility I mentioned in slack?

packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/basic/PersistAsyncBasicSchema.js
StevePole: Feels like premature optimisation, do we actually have a need for it?  It should be simple enough for someone to implement this with a regular promise if they need it, right?

packages/dynamic-flows/src/jsonSchemaForm/schemaFormControl/SchemaFormControl.js
```diff
@@ -1,55 +1,16 @@
...
 import FormControl from '../../formControl';
 
-import { mapConstSchemaToOption } from './optionMapper';
+import { mapConstSchemaToOption, mapSchemaToUploadOptions } from './optionMapper';
 
 const isNativeInput = (propsSchemaType) => {
   return propsSchemaType === 'string' || propsSchemaType === 'number';
 };
 
-const getControlType = (schema) => {
```
StevePole, line 15: As mentioned in slack.  This function is for mapping from a schema to a control type.  The function in the requirements.js file, was for backwards compatibility with older dynamic forms specifications, which we don't support in React.  It was copied wholesale from the Angular implementation.  I don't think it should be used at all.

packages/dynamic-flows/src/common/requirements.js
StevePole: Explained below, this file should not be used in this implementation, it was designed for supporting old dynamic form specs in the angular implementation and at some point has been copied to this repo.

packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/basic/PersistAsyncBasicSchema.js
richardbata: Probably yes. FILE is used in practice, upload only if its explicitly set as the control type

packages/dynamic-flows/src/jsonSchemaForm/schemaFormControl/SchemaFormControl.js
richardbata: The question remains as to whether Field and SchemaFormControl should share the same getControlType?

packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/basic/PersistAsyncBasicSchema.js
JulianG-TW: oh! I completely missed the `props.onPersistAsync(persistAsyncFetch);` line

JulianG-TW: in recipient-form we don't use queryable functions either, our `onPersistAsync` handler expects a `Promise` and looks like this:
```ts
  const onPersistAsync = (promise: Promise<unknown>) => {
    setCount(c => c + 1);
    promise.finally(() => {
      delayUntilNextCycle(() => setCount(c => Math.max(0, c - 1)));
    });
  };
```

packages/dynamic-flows/src/jsonSchemaForm/schemaFormControl/SchemaFormControl.spec.js
```diff
@@ -2,8 +2,42 @@ import { shallow } from 'enzyme';
...
+    ['tel', { type: 'string', format: 'phone' }],
+    ['number', { type: 'number' }],
+    ['number', { type: 'integer' }],
+    ['password', { type: 'string', format: 'password' }],
+    ['textarea', { type: 'string', control: 'textarea' }],
+    ['tab', { oneOf: [], control: 'tab' }],
+  ];
+
+  it.each(data)('should be type %s for %o', (controlType, schema) => {
+    expect(getControlType(schema)).toBe(controlType);
```
StevePole, line 37: getControlType should really be an internal detail, not exposed publicly.  The way to test is to see what the FormControl receives to it's props.  This is a better test because it is less coupled to the implementation (doesn't care if getControlType is renamed), and it tests that the logic is used correctly to actually pass that value to the child component.  You could change this line to do that, but that's effectively what the tests below already do, so this seems duplicative (although might contain more cases).

packages/dynamic-flows/src/jsonSchemaForm/schemaFormControl/SchemaFormControl.spec.js
```diff
@@ -4,6 +4,73 @@ import FormControl from '../../formControl';
...
+      sharedProps = { id, locale, translations };
+
+      props = { ...events, ...sharedProps, value, schema };
+
+      component = shallow(<SchemaFormControl {...props} />);
+      formControlComponent = component.find(FormControl);
+    });
+
+    it('should pass down correct control type', () => {
+      expect(formControlComponent.props().type).toBe(controlType);
```
StevePole, line 69: These tests feel very repetitive of the ones below, do we need both?

StevePole (approved):
Please add a ticket to change the way we supply the headers for the async calls.  We shouldn't hard code these values.  My proposal is to create a singleton service where we can set default headers from the consumer, without passing down the prop hierarchy chain.  WE would need to integrate that service with the different async calls.

I left a note about some repetitive tests, would be nice to tidy, but there's nothing wrong there.

PR #1134, opened 2021-09-06
Align web implementation to iOS implementation

JulianG-TW:
## 🖼 Context

ActionTypes in the library [are](https://github.com/transferwise/dynamic-interfaces/blob/master/v3/src/main/java/com/transferwise/df/v3/steps/internal/ActionType.java):
```java
public enum ActionType {
    @JsonProperty("primary")
    PRIMARY,
    @JsonProperty("secondary")
    SECONDARY,
    @JsonProperty("negative")
    NEGATIVE,
    @JsonProperty("link")
    LINK;
}
```

The iOS implementation determines a button's `type` and `priority` from the `action.type` like this:

```swift
let (style, priority) = buttonStyleMapper?(type) ?? {
    switch type {
    case .link:
        return (.accent, .tertiary)
    case .primary:
        return (.accent, .primary)
    case .secondary:
        return (.accent, .secondary)
    case .negative:
        return (.negative, .primary)
    }
}()
```
[Here's the source](https://github.com/transferwise/transferwise-ios/blob/dev/DynamicFlow/DynamicFlowUI/Step/BasicFormStepViewController.swift#L174-L185
). And there’s a [special case](https://github.com/transferwise/transferwise-ios/blob/dev/Wise/Wise/Recipients/RecipientCreationFlow/RecipientCreationFlowImpl.swift#L108-L119
) for recipient creation flow.

However, the current web implementation is different:

```js
const getButtonType = (action) => {
  switch (action.type) {
    case 'primary': // not in the spec
      return 'accent';
    case 'success': // not in the spec
      return 'positive';
    case 'failure': // not in the spec
    case 'warning': // not in the spec
      return 'negative';
    default:
      return 'accent';
  }
};

const getButtonPriority = (action) => {
  switch (action.type) {
    case 'primary': // not in the spec
    case 'success': // not in the spec
      return 'primary';
    default:
      return 'secondary';
  }
}; // tertiary is missing
```

## 🚀 Changes

Align the web implementation to the iOS one

## 🤔 Considerations

This would be a breaking change for any apps expecting the previous behaviour. But considering iOS behaviour is different, we don't think these are being used.

You probably want to switch to "split" view for this one:
image

## ✅ Checklist


StevePole (approved):
👍

PR #1136, opened 2021-09-06
[VUX-976] Missing title fix

richardbata:
## 🖼 Context
The layout property is not yet fully supported on mobile platforms, leading top level title properties not within the layout to be displayed as page headings. This issue will be solved correctly, but until then I'm indroducing a fix on web to make is consistent with the behaviour on mobile.

Essentially the JSON below result in double titles on mobile, but only a single one on web. Web should be correct, as specifying a `layout` property should override all default layouts. Until the issue is solved on mobile, web is introduction this fix.
image

## ✅ Checklist


packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -40,7 +40,16 @@ const getComponents = (step) => {
     return [];
   }
 
-  const layout = step.layout ? step.layout : convertStepToLayout(step);
+  let layout = step.layout ? step.layout : convertStepToLayout(step);
+
+  // This condition is meant as a temporary hack until mobile platforms will be refactored to fully support the layout property
+  if (step.layout && step.title) {
+    const topStep = {
```
StevePole, line 47: We should do this within the convertStepToLayout function, that's effectively designed for backwards compatibility with steps that don't have layouts.

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -40,7 +40,16 @@ const getComponents = (step) => {
...
+      type: step.type,
```
StevePole, line 48: Bit confused by this, the step type is not going to be a valid layout component type.

packages/dynamic-flows/src/flow/layoutService/LayoutService.js
```diff
@@ -3,6 +3,10 @@ function convertStepToLayout(step) {
     throw new Error('Missing step type');
   }
 
+  if (step.layout) {
```
StevePole, line 6: Could you just add back the note about why this exists?

richardbata: Done

StevePole (approved):

PR #1137, opened 2021-09-07
refactor(Link): remove default aria-label

matebek-tw:
## 🖼 Context

Removing the default `aria-label` attribute from the `Link` component, as it doesn't provide much value over the default accessibility label generated for the link based on its content, e.g.: `<Link href="https://wise.com">https://wise.com</Link>` will read "link, https://wise.com" which provides more context than "link, Open link".

Providing an `aria-label` is essential to provide more detail, but defaulting to reducing context is not a good practice in my humble opinion.

## 🚀 Changes

Remove default `aria-label` from `Link`

## ✅ Checklist


packages/components/src/link/Link.story.tsx
```diff
@@ -0,0 +1,27 @@
...
+  component: Link,
+  title: 'Link',
+};
+
+export const Basic = () => {
+  return (
+    <>
+      <p>
+        Lorem Ipsum is simply dummy text of the printing and typesetting industry{' '}
+        <Link aria-label="Click here for more details" href="#">
```
andreapiras, line 13: remove here too?

andreapiras (approved):

dozortsev (approved):

PR #1140, opened 2021-09-09
fix: dimmer color for Angular and fix .on-scroll on unmount

dozortsev:
## ✅ Checklist


packages/css/package.json
```diff
@@ -42,7 +42,7 @@
   },
   "dependencies": {
     "@transferwise/neptune-tokens": "^1.0.0",
-    "bootstrap": "github:transferwise/bootstrap#semver:^5.20.0",
+    "bootstrap": "github:transferwise/bootstrap#7.0.0",
```
dozortsev, line 45: just noticed that our BT dependency is slightly off for a few versions

packages/css/src/less/modals.less
```diff
@@ -3,4 +3,184 @@
 @import (reference) '../variables/legacy-variables.less';
 @import 'bootstrap/less/mixins/vendor-prefixes.less';
 @import 'bootstrap/less/mixins/opacity.less';
-@import 'bootstrap/less/modals.less';
+
```
dozortsev, line 6: bringing content of [`modals.less`](https://github.com/transferwise/bootstrap/blob/master/less/modals.less) into neptune-css with update color for dimmer

packages/docs/package.json
```diff
@@ -43,7 +43,6 @@
     "@transferwise/icons": "^2.14.0",
     "@transferwise/neptune-css": "^10.0.4",
     "@transferwise/neptune-tokens": "^4.1.0",
-    "bootstrap": "github:transferwise/bootstrap#semver:^5.20.0",
```
dozortsev, line 46: seems redundant in this package

packages/css/src/less/modals.less
fergusjordantw: Are we sure this is the best idea? Not sure we should be bringing styles into a package that isn't actually connected to the component the styles are referencing

dozortsev: > ..Are we sure this is the best idea?.

yes, this is according our Neptune CSS plan

> ..Not sure we should be bringing styles into a package that isn't actually connected to the component the styles are referencing

Neptune CSS is a css for styleguide-components as well

eventually all necessary BT styles will migrate into Neptune CSS, this is better than keep making changes and keep supporting Bootstrap

fergusjordantw: Huh, okay, didn't know that

fergusjordantw (approved):

PR #1142, opened 2021-09-13
[VUX-1068] Fix exit condition logic for Dynamic-flows

richardbata:
## 🖼 Context
The exit condition when the server responds with `X-DF-Exit` header was not fully working for the DF component. The issue was in the HttpClient, where Response.json() would fail because of an empty response body throwing an exception, and the exit header is never checked for.

## 🚀 Changes
Tweaking the exit code logic in order to stay consistent with the [proposal](
[Proposal](https://transferwise.atlassian.net/wiki/spaces/DF/pages/1946032207/Flow+Completion+Revisited)
).

## 🤔 Considerations
The httpClient response type has changed, hence consumers of DynamicFlows using custom clients need to adjust for it. Luckily now it is only verification doing so.

## ✅ Checklist


StevePole (approved):
👍

JulianG-TW (approved):

PR #1144, opened 2021-09-14
fix: fix wrong peerdependency version, again

fergusjordantw:
## 🖼 Context

Made a mistake in my last PR which kept the version stuck on 9 🤦 

## 🚀 Changes

Actually fixed the peer dependencies this time.

## 🤔 Considerations

I have run it through a validator to confirm this time. Sorry.

## ✅ Checklist


matebek-tw (approved):

packages/dynamic-flows/package.json
```diff
@@ -58,7 +58,7 @@
   "peerDependencies": {
     "@transferwise/components": ">=36.6.2 <=37",
     "@transferwise/icons": "^2.6.0",
-    "@transferwise/neptune-css": "^9.2.0 <=10",
+    "@transferwise/neptune-css": ">=9.2.0 <=10",
```
karanjthakkar, line 61: shouldnt this be <10? not <=10? or are you saying we want to allow neptune-css v10? https://jubianchi.github.io/semver-check/#/%3E%3D9%20%3C%3D10/10.1

fergusjordantw: Yes, this is adding support for neptune-css 10

karanjthakkar: would it make sense to follow a consistent format when we want 2 versions? Like we do for react?

karanjthakkar: ie ^9 || ^10? i know there's no guidance for it yet but maybe this is the right time to do it?

fergusjordantw: Potentially since we're within neptune, yes, although I'll let @matebek-tw and @dozortsev answer that from their perspective. For most other packages, I've been changing to support a range for neptune-css since they're all going to be taking whatever the consumer has anyway, so it makes a bit more sense for most packages to define pretty loose ranges for it.

fergusjordantw: Slept on it and agree. Makes sense to have a more readable `||` range if we're limiting it to 2 major versions instead of a range of major versions

PR #1145, opened 2021-09-17
fix: height in IOS safari and word-break in modal

nabaroa:
## 🖼 Context

Related tickets: https://transferwise.atlassian.net/browse/OGE-234 & https://transferwise.atlassian.net/browse/OGE-235

Modal is showing layout issues in mobile devices. This PR solves 2 problems:
Page to test: https://wise.com/campaign/ownwise
| Hidden buttons in IOS devices and improves the display in Android. | Long copy strings, for example a long url, makes the modal wider than the viewport |
|---|---|
| image | image |

### fix: modal height in mobile devices e049875

There is a common issue in browsers with `height: 100vh` and the mobile device implementation of it. Basically the **footer is hiding below Safari’s menu bar**, missing all the CTA buttons. [Here an article](https://allthingssmitty.com/2020/05/11/css-fix-for-100vh-in-mobile-webkit/) explains the issue and gives a solution we used in this PR. It is kind of non standard using a vendor prefix to solve a particular layout problem, but this Modal components is going to be rebuilt soon, so I take the "hacky" solution for temporary and to solve the situation to tons of frustrated users who can't go through the modal until we get the refactor.

- Fixed height to Android
- Fixed 100vh issue for IOS safari

| Before | After|
|---|---|
| Iphone 12 Pro M||
| image | image|
| Iphone X||
| image| image|
|Android Pixel 3a Local modal component in Neptune Docs||
|image|image|

### fix: allow work breaking for long strings in modals 504fda6 

Long url/copy strings can break the modal layout in small devices. With this simple CSS property `word-break`we can control the behavior of these strings and force a new line before they push the layout off.

| Before | After|
|---|---|
| image | image |

## 🤔 Considerations

With this height solution for mobile we change the default behavior and now the modal is going to display in its total length in mobile. Is this something acceptable from design perfective?

| Before | After|
|---|---|
| image | !image |

## ✅ Checklist


helendurranttw:
Yes! They should display full height on mobile. I think them not doing so is a bug. Nice fixes!

packages/components/src/modal/Modal.less
```diff
@@ -18,12 +18,18 @@
   }
 
   .tw-modal-content {
-    max-height: 100vh;
+    height: calc(100vh - 80px);
```
dozortsev, line 21: what is 80px in this case? a hight of footer?

could you please extract this value into LESS variables so that it would easier to understand what this 80px about

helendurranttw (approved):

dozortsev (approved):

dozortsev (commented):
> ..With this height solution for mobile we change the default behavior and now the modal is going to display in its total length in mobile. Is this something acceptable from design perfective?

yes, in fact we plan to refactor Modal to use Drawer for mobile view which will cover whole screen

packages/components/src/modal/Modal.less
```diff
@@ -18,17 +18,24 @@
   }
 
   .tw-modal-content {
-    max-height: 100vh;
+    height: calc(100vh - 88px); // 88 = Android bottom nav menu height (56px) + viewport top padding (16px) + viewport bottom padding (16px)
```
nabaroa, line 21: Commented the 80px magic number. In fact it was 88px 😬 

Looks perfect now:
image

packages/components/src/modal/Modal.less
nabaroa: Updated ✅

packages/components/src/modal/Modal.less
dozortsev: I wonder whether Android browsers have and support some environment variables, like we use at [BottomSheet to uplift content over iOS home indicator](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/common/bottomSheet/BottomSheet.less#L30-L32)

PR #1149, opened 2021-09-21
Add support for step errors

JulianG-TW:
## 🖼 Context

We want to be able to display validation errors **before** submitting a form.
For instance, when a form is first loaded, it may contain validation errors prompting the user to update certain fields.

image

## 🚀 Changes

These changes update the validation messages when a step is loaded.

## 🤔 Considerations

Errors in the step override any existing errors that may be present as a result of a failed submission, or previous step.

## ✅ Checklist


packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -83,6 +93,11 @@ const DynamicFlow = (props) => {
     if (stepSpecification?.model) {
```
StevePole, line 93: Wondering if you could just move this to a named function now it's grown in complexity a bit e.g.
```
useEffect(onNewStep, [stepSpecification]);

const onNewStep = () => { ... };
```

StevePole (approved):
Glad that was easy :-) 

I left one minor readability suggestion.

packages/dynamic-flows/src/flow/DynamicFlow.js
JulianG-TW: https://github.com/transferwise/neptune-web/pull/1151

PR #1151, opened 2021-09-21
Refactor DynamicFlow component, eliminating one useEffect

JulianG-TW:
## 🖼 Context

We had one effect setup like this:
```js
useEffect(onNewStep, [stepSpecification]);
```
where `stepSpecification` is the value returned from a `React.useState` hook.

The corresponding `setStepSpecification` function is used 2 or 3 times in callbacks.

This meant that when one of those callbacks were triggered we had to "wait" until the next render, where the value of `stepSpecification` was updated, for the `onNewStep` function to execute.

## 🚀 Changes

Instead, we can wrap `setStepSpecification` into a new function that also executes what `onNewStep` does. And use it directly as callback.

Instead of:
```js
const [a, setA] = useState();
useEffect(() => doStuff(a), [a]);

return <Thing onChange={setA} />
```

We can do:
```js
const [a, setA] = useState();

const updateA = (newA) => {
  setA(newA);
  doStuff(newA);
};

return <Thing onChange={updateA} />
```
### 🤔 Considerations

I think it's a good thing to avoid using `useEffect` as much as possible, but I can't think of a good reason, other than making it easier to follow the code flow when reading.
Perhaps it avoids unnecessary re-renders, but I'm not too worried about that.

## ✅ Checklist


JulianG-TW:
I don't think this is such a good idea any more.

PR #1154, opened 2021-09-22
Fix: Logo on small screen in overlayHeader should only show fast flag

johannacarlberg:
## 🖼 Context

https://transferwise.atlassian.net/browse/AT-2916

## ✅ Checklist


packages/components/src/logo/Logo.js
```diff
@@ -1,5 +1,8 @@
 import classNames from 'classnames';
 import PropTypes from 'prop-types';
+import { useWindowSize } from 'react-use';
```
dozortsev, line 3: we have own equivalent of this, called `useClientWidth`

```ts
const [clientWidth] = useClientWidth({ ref: isServerSideRendering() ? undefined : window })
```

packages/components/src/logo/Logo.js
```diff
@@ -10,10 +13,13 @@ const logoPaths = {
   WISE_BUSINESS: 'brand_logo_business.svg',
   WISE_INVERSE: 'brand_logo_inverse.svg',
   WISE_BUSINESS_INVERSE: 'brand_logo_business_inverse.svg',
+  WISE_FLAG: 'brand_flag.svg',
```
dozortsev, line 16: great,

but now we have duplicate logic in FlowNavigation which we need to delete from there and reuse logic of Logo component

I'm talking about these styles https://github.com/transferwise/neptune-web/blob/main/packages/components/src/flowNavigation/FlowNavigation.js#L46-L50 they load `brand_flag.svg` via CSS https://github.com/transferwise/neptune-web/blob/main/packages/components/src/flowNavigation/FlowNavigation.less#L74-L94

packages/components/src/logo/Logo.js
johannacarlberg: ah nice will use that instead 👍

dozortsev (commented):
thanks for contributing, it looks great 🙇

packages/components/src/logo/Logo.js
johannacarlberg: 👍  will address

packages/components/src/flowNavigation/FlowNavigation.js
```diff
@@ -43,12 +43,7 @@ const FlowNavigation = ({ activeStep, avatar, logo, onClose, onGoBack, done, ste
             onClick={onGoBack}
           />
         )}
-        <div
-          className={classNames('np-flow-navigation--flag', {
-            'np-flow-navigation--flag__hidden': displayGoBack,
-            'np-flow-navigation--flag__display': !displayGoBack,
-          })}
-        />
+        {!displayGoBack && <>{logo}</>}
```
dozortsev, line 46: could use ternary operator here

```ts
{displayGoBack ? (<BackButton...) : logo}
```

dozortsev (approved):
thanks a lot!

johannacarlberg:
thanks @dozortsev , changed it to a ternary!

PR #1155, opened 2021-09-23
Avoid styling the title of OneOf with "has-error"

JulianG-TW:
## 🖼 Context

This edge case was first identified when working on CoP. In the image below you can see that the refreshed step containing errors causes the title of the oneOf: "Bank details" to be red.

Screenshot 2021-09-23 at 09 27 43

## 🚀 Changes

We avoid styling the title of OneOf with "has-error" when schema is oneOf non-consts.

We want to keep rendering "red" titles when it's a oneOf const values, like the example below:

Screenshot 2021-09-23 at 09 21 38

But we don't want to render the title "red" when it's above tabs:

Screenshot 2021-09-23 at 09 21 40

## 🤔 Considerations

I haven't written any tests for this just yet, because I want to validate the functionality first.

## ✅ Checklist


packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.js
```diff
@@ -221,6 +211,27 @@ const OneOfSchema = (props) => {
   );
 };
 
+function getTitleAndHelp(schema, id) {
+  const titleElement = isConstSchema(schema.oneOf[0]) ? (
+    <label className="control-label d-inline" htmlFor={id}>
+      {schema.title}
+    </label>
+  ) : (
+    <h4>{schema.title}</h4>
```
StevePole, line 220: The spacing is too tight with the following content.  Can you try adding a class `.m-b-2` to the `h4` and see how that looks?

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.js
```diff
@@ -221,6 +211,27 @@ const OneOfSchema = (props) => {
...
+  );
+
+  const helpElement = schema.help ? <Help help={schema.help} /> : null;
+
+  return schema.title ? (
+    <div className="d-inline-block">
```
StevePole, line 226: I know this was there already but not sure why we need these `.d-inline-block` classes, and also the `.d-inline` on `label` above?  We shouldn't really need modifiers for standard form styling.

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.js
JulianG-TW: image

Better, but perhaps too much space (to my taste). I'm happy to add it.

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.js
JulianG-TW: It seems to make no difference. I'll remove it.

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.js
JulianG-TW: done @81f472885e4a81e8fc29d4af74bff641a1f1050a

StevePole (approved):
👍

PR #1163, opened 2021-09-30
refactor: get rid of SizeSwapper in ResponsivePanel

dozortsev:
## 🖼 Context

`Info` component uses `ResponsivePanel`, while `Info` is used quite often next to some content in `<p>` elements which causes error:

Screenshot 2021-09-30 at 11 40 15

## 🚀 Changes

Removed `SizeSwapper` from `ResponsivePanel` as it just redundant there, and instead use hook that `SizeSwapper` actually based one.

## ✅ Checklist


packages/components/src/common/responsivePanel/ResponsivePanel.js
```diff
@@ -1,39 +1,36 @@
...
-          {children}
-        </Panel>,
-      ],
-      breakpoint: Breakpoint.SMALL,
-    },
-  ];
-  return <SizeSwapper ref={windowReference} items={items} inline />;
+  const [clientWidth] = useClientWidth({ ref: windowReference });
+
+  if (clientWidth >= Breakpoint.SMALL) {
```
dozortsev, line 13: using same Breakpoint as was passed into `SizeSwapper` (see line 36)

packages/components/src/common/responsivePanel/ResponsivePanel.js
```diff
@@ -1,39 +1,36 @@
...
+      >
+        {children}
+      </Panel>
+    );
+  } else {
+    return (
+      <BottomSheet key="bottomSheet" open={open} className={className} onClose={onClose}>
+        {children}
+      </BottomSheet>
+    );
```
matebek-tw, line 32: You don't need the `else {}` block when you have an early return.

```suggestion
  }

  return (
      <BottomSheet key="bottomSheet" open={open} className={className} onClose={onClose}>
        {children}
      </BottomSheet>
    );
```

matebek-tw (approved):

PR #1169, opened 2021-10-01
Neptune Web Release

dozortsev:
## 🚀 Changes

Deployment: https://transferwise.github.io/neptune-web/branch/next/about/Home

 https://github.com/transferwise/neptune-web/milestone/4

## ✅ Checklist


packages/components/src/common/Option/Option.js
```diff
@@ -51,7 +51,7 @@ const Option = ({
         </div>
       )}
       <div className="media-body">
-        <div className="h5">{title}</div>
+        <div className="title-4">{title}</div>
```
helendurranttw, line 54: Is this a mistake?

dozortsev: I don't think so, cc @nabaroa

helendurranttw:
I noticed some issues:

- [ ] **Hover styles in select are using accent**

image

image

- [ ] **Some inputs have lost their border focus styles**

image

- [ ] **Options avatars background should be white on hover**

image

- [ ] **The nudge link is now too small and the layout looks off (probably because of this)**

Screenshot 2021-11-30 at 10 51 44

- [ ] **The tooltip appears broken, and the content inside is the wrong font size**

Screenshot 2021-11-30 at 10 52 20

Screenshot 2021-11-30 at 10 53 02

- [ ] **Wrapping for secondary text in selects is broken**

Screenshot 2021-11-30 at 10 56 05

packages/components/src/avatar/Avatar.less
```diff
@@ -25,17 +27,17 @@
...
-      border-color: var(--color-control-accent-hover);
+      border-color: var(--color-interactive-accent-hover);
     }
   }
 
+  // theming: this becomes redundant when we implement theming as we will just change value of CSS custom props
   &--dark &__content {
-    background: @color-base-navy-mid;
-    color: @color-text-control;
+    background: var(--color-background-neutral);
```
helendurranttw, line 40: This will break dark themes for accordions right now

packages/components/src/common/panel/Panel.less
```diff
@@ -1,3 +1,5 @@
+@import (reference) '@transferwise/neptune-css/src/less/mixins/_logical-properties.less';
```
helendurranttw, line 1: Doesn't seem like we need this import

packages/components/src/avatar/Avatar.less
dozortsev: could you please show screenshot of what you mean?

packages/components/src/common/panel/Panel.less
dozortsev: fixed in https://github.com/transferwise/neptune-web/pull/1234

packages/components/src/avatar/Avatar.less
helendurranttw: Sorry, I meant Avatar, not accordions. This is main:
Screenshot 2021-11-30 at 13 51 11

This is next:

Screenshot 2021-11-30 at 13 52 47

dozortsev: fixed in https://github.com/transferwise/neptune-web/pull/1234

matebek-tw (approved):

PR #1188, opened 2021-10-14
Scaffold elements no rtl mixin

nabaroa:
## 🖼 Context

There are specificity collisions with the rtl mixin with the generic ul element.

image

## 🚀 Changes

- Get free of the `html:not([dir="rtl"])` selector to margin and padding, as can be quite common to overwrite them in the project custom CSS, (Excluded utilities), avoiding specificity collisions.

The logic change consist in:

**Previous**

html:not([dir="rtl"])

[dir="rtl"]

**Now**

Default (can be without dir attribute of dir=ltr)

[dir="rtl"]

## 🤔 Considerations

This change have to make the rtl code transparent when bumping the version, not needed to update anything. As counterpart when migrating to rtl layout, we have to take into account a possible second variable in padding/margins to reset the opposite value.

image

☝️ this means:

**Default**: 
padding-right: 40px;

**Rtl**:
padding-left: 40px;
padding-right: 16px;

## ✅ Checklist


packages/css/src/less/mixins/_logical-properties-IE-friendly.less
```diff
@@ -62,7 +62,27 @@
 
 // CSS Logical Properties: https://caniuse.com/css-logical-props
 
-.padding(@direction, @value) {
+.padding(@direction, @value, @value2: ~'') {
```
dozortsev, line 65: I think we should give more self explanatory names to last two arguments

if I understand correctly how this works I would call them `@fallbackDirectionPaddingValue` (instead of `@value`) and `@currentDirectionPaddingValue` (instead of `@value2`) or vise-versa 🤔

dozortsev: same applicable to `.margin` mixin

dozortsev (approved):

PR #1195, opened 2021-10-25
feat(components): Add Global Currency support to MoneyInput

DeeHechEss:
## 🖼 Context

https://transferwise.atlassian.net/browse/GC-337
https://www.figma.com/file/0yH9P2LWXXygC6LjsJYOrE/Global-USD-calc-experiment?node-id=0%3A1
https://docs.google.com/document/d/1_oCAr9otb4qFXzoymqG67lq-0bczswRqZ6lmoLC0KNs/edit

We are going to show a "USD - Outside of the US" target currency option in the calculator to increase Swift transfer visibility. The design for this will show a globe icon as opposed to a country flag.

## 🚀 Changes

Added an optional `global` property to currencies accepted by MoneyInput.
Display a globe icon instead of a flag for currencies with `global` set to `true`.

## ✅ Checklist


packages/components/src/select/option/Option.js
```diff
@@ -11,14 +12,22 @@ const Option = ({ currency, label, note, secondary, icon, classNames, selected }
         `${selected ? 'hidden-xs' : ''}`,
       ])}`
     : null;
+  const currencyIcon = global ? (
```
okanoktartw, line 15: Just for a curiosity, can't we have global-usd flag under [flags](https://wise.com/public-resources/assets/flags/rectangle/inr.png) ? (sth like usd-global ?) so it can be rendered in the same with other flags ?

okanoktartw (approved):
lgtm 👍

packages/components/src/select/option/Option.js
DeeHechEss: I believe the currency code is used to find the flag, so we would need to introduce a new currency code.

PR #1200, opened 2021-10-28
feat: add theming support with themed color tokens

matebek-tw:
## 🖼 Context

We decided to add theming support as part of our next major release to be able to resolve some of the stylistic issues resulted by the new accessible color tokens on product pages.

## 🚀 Changes

- Add `ThemeProvider`
- Add `useTheme` hook and `withTheme` HOC to be able to access the theme variable
- Use the themed color tokens instead of the base tokens and assign them to the `:root` element

## 🤔 Considerations

- Theming just won't work on IE11 and below, no matter that if we use the `postcss-custom-properties` plugin, because the fallback values will be added to the stylesheet in build time, and in case the browser does not support variables, it will use the fallback value, which won't be re-evaluated in run-time.
- Injecting the `className` only works if the `children` of the `ThemeProvider` is a valid React element, as we have to use `React.cloneElement` to achieve this - adding a wapping `<div />` would break layouts for the component, finding the first valid element is not as simple as it sounds, I have no better idea at the moment, than just show a warning to the developers.

## ✅ Checklist


packages/components/src/common/hocs/theme/withTheme.tsx
```diff
@@ -0,0 +1,30 @@
+import { ThemeType } from '../..';
```
dozortsev, line 1: this file seems doesn't have any JSX so file could have `.ts` extension

packages/components/src/test-utils/index.js
```diff
@@ -13,12 +13,12 @@ import en from '../i18n/en.json';
...
 function customRender(
   ui,
-  { locale = DEFAULT_LOCALE, messages = en, direction, ...renderOptions } = {},
+  { locale = DEFAULT_LOCALE, messages = en, theme = Theme.LIGHT, direction, ...renderOptions } = {},
 ) {
   // eslint-disable-next-line react/prop-types
   var Wrapper = ({ children }) => {
     return (
-      <Provider i18n={{ locale, messages }} direction={direction}>
+      <Provider i18n={{ locale, messages }} theme={theme} direction={direction}>
```
dozortsev, line 21: but atm there no `theme` prop in `Provider` component

packages/components/.storybook/preview.js
```diff
@@ -84,10 +85,15 @@ const DirectionDecorator = (storyFn) => {
 
 const ProviderDecorator = (storyFn) => {
   const locale = select('locale', severalExamplesOfSupportedLocales, DEFAULT_LOCALE);
+  const theme = select('theme', Object.values(Theme));
```
dozortsev, line 88: we don't need this as we don't launch theming yet

packages/components/src/utilities/styleInjector/StyleInjector.tsx
```diff
@@ -0,0 +1,24 @@
+import { Children, cloneElement, isValidElement } from 'react';
+
+export const StyleInjector: React.FC<{ className?: string }> = ({ children, className }) => {
```
dozortsev, line 3: I would actually keep this logic inside ThemeProvider rather than abstract this in separate component, wdyt?

packages/components/src/utilities/styleInjector/StyleInjector.tsx
```diff
@@ -0,0 +1,24 @@
...
+  const styledChildren = Children.map(children, (child) => {
+    if (!isValidElement(child)) {
```
dozortsev, line 5: just discovered one more bug (but perhaps you were already aware about that), it doesn't add CSS class to React component that have "legit" (e.g `div` or `button` or `span`) HTML element at root

packages/components/src/provider/theme/ThemeProvider.spec.tsx
```diff
@@ -0,0 +1,47 @@
+import { render, screen } from '@testing-library/react';
+
+import { Theme } from '../../common/theme';
+
+import { ThemeProvider } from './ThemeProvider';
+
+describe('ThemeProvider', () => {
```
dozortsev, line 7: we need add one more case here, when children is React component with real HTML element, you could create custom component here or use one of existing (e.g Button, Modal, Accordion ...)

packages/components/src/utilities/styleInjector/StyleInjector.tsx
matebek-tw: Added a test to cover this case in https://github.com/transferwise/neptune-web/pull/1200/commits/fe19f15d33919e6630145bb5797eea148ed8ce0f. This should work, so probably something else causing this problem.

packages/components/src/snackbar/SnackbarProvider.js
```diff
@@ -45,7 +45,7 @@ SnackbarProvider.propTypes = {
 };
 
 SnackbarProvider.defaultProps = {
-  timeout: 4500,
+  timeout: 450000,
```
helendurranttw, line 48: Why was this changed?

packages/components/src/tabs/Tabs.less
```diff
@@ -74,7 +74,6 @@
   &__panel {
     &--selected {
       padding: var(--size-8);
-      border: 1px solid @color-base-navy-light;
```
helendurranttw, line 77: Why was this removed?

dozortsev: I recall that it wasn't making any affect so I removed since we not long ship base colors

you can try compare [main](https://transferwise.github.io/neptune-web/components/content/Tabs) with [this one](https://transferwise.github.io/neptune-web/branch/add-theme-provider/components/content/Tabs)

packages/components/src/snackbar/SnackbarProvider.js
dozortsev: opps, this needs to reverted

```suggestion
  timeout: 4500,
```

.github/workflows/ci-cd.yml
```diff
@@ -289,7 +289,7 @@ jobs:
     name: Publish beta packages
     if: ${{ github.ref != 'refs/heads/main' }}
     runs-on: ubuntu-latest
-    needs: test
+    needs: setup
```
dozortsev, line 292:
```suggestion
    needs: test
```

packages/css/src/less/buttons.less
```diff
@@ -93,21 +93,22 @@
...
       var(--color-background-screen);
-      var(--color-base-white-light);
+      #fff;
     );
   }
 }
 
 // @DEPRECATED:.btn-link (use Button component with type Accent / priority Tertiary)
 .btn-link,
+.btn-info,
```
helendurranttw, line 111: ?

packages/css/src/less/buttons.less
```diff
@@ -143,36 +144,6 @@
   }
 }
 
-// @DEPRECATED:.btn-success (use Button component with type Positive)
```
helendurranttw, line 146: Why did this move?

packages/css/src/less/buttons.less
```diff
@@ -220,19 +221,37 @@
...
       var(--color-interactive-negative-hover);
       var(--color-interactive-negative-active);
       var(--color-background-screen);
-      var(--color-base-white-light);
+      #fff;
     );
   }
 }
 
+// @TODO Deprecate / Remove
```
helendurranttw, line 237: Why did this move?

packages/css/src/less/core/_scaffolding.less
```diff
@@ -54,6 +54,19 @@ a,
   text-decoration: underline;
   background-color: transparent; // Normalize 8.0.1
 
+  .bg-primary & {
+    color: @color-navy-content-accent;
```
helendurranttw, line 58: Should this not be interactive?

packages/css/src/less/core/_scaffolding.less
```diff
@@ -70,7 +83,7 @@ a,
   }
 
   &.secondary {
-    color: @color-base-navy-light;
+    color: var(--color-content-positive);
```
helendurranttw, line 86: Positive?

packages/css/src/less/core/_typography.less
```diff
@@ -225,7 +226,7 @@ blockquote {
 
   footer {
     display: block;
-    color: @color-base-grey-light;
+    color: #fff;
```
helendurranttw, line 229: Grey to white?

packages/css/src/less/footer.less
```diff
@@ -60,50 +60,51 @@
     text-decoration: none;
 
     svg {
-      fill: var(--color-interactive-secondary);
+      fill: var(--color-content-primary);
```
helendurranttw, line 63: Completely different?

packages/docs/components/Layout.js
```diff
@@ -54,6 +54,7 @@ const Layout = ({ children, router: { pathname } }) => {
               </Link>
             </li>
           ))}
+        <li className="Nav__Group">Configs</li>
```
helendurranttw, line 57:
```suggestion
        <li className="Nav__Group">Config</li>
```

packages/css/src/less/neptune.bundle.less
```diff
@@ -32,7 +31,6 @@
 @import './loaders.less';
 @import './media.less';
 @import './modals.less';
-@import './panels.less';
```
helendurranttw, line 35: Why was this removed?

packages/css/src/props/legacy-custom-props.css
```diff
@@ -1,65 +0,0 @@
-/*
```
helendurranttw, line 1: Why was this removed? This needs to be put back. We can tell people we'll be removing them, but unless you provide a simple migration strategy or give them time to update, you cannot simply remove props they may be depending on. At a minimum you need to let them know what you're removing. A simple search of codesearch shows 119 usages of `--spacer` alone.

packages/css/src/variables/neptune-tokens.less
```diff
@@ -1,2 +1 @@
 @import '@transferwise/neptune-tokens/tokens.less';
-@import '@transferwise/neptune-tokens/colors-base.less';
```
helendurranttw, line 2: This is another breaking change. I realise the base colours in tokens may not be what they were before, but we should do our best to alias them or give people an upgrade guide. From what I see the tokens are mostly the same, except for @color-base-white, which we could add a manual alias for to @color-base-white-light.

packages/docs/public/static/assets/main.css
```diff
@@ -1,3 +1,5 @@
+@import "@transferwise/neptune-tokens/themes/navy/tokens.css";
```
helendurranttw, line 1: I don't think this is necessary

helendurranttw:
This PR is way too big. Like the release PR you have open to be merged into main, it is very difficult to review, because of the breadth of changes. Many of the changes are unrelated to the original scope of the PR, and there are several important breaking changes in here which are not explained or called out in the PR description.

Please try and keep unrelated changes out of PRs, and try and break the changes up across PRs to make reviewing easier. Thank you!

packages/css/src/less/buttons.less
dozortsev: fixed in https://github.com/transferwise/neptune-web/pull/1231

packages/css/src/less/buttons.less
dozortsev: fixed in https://github.com/transferwise/neptune-web/pull/1231

packages/css/src/less/buttons.less
dozortsev: fixed in https://github.com/transferwise/neptune-web/pull/1231

packages/css/src/less/core/_typography.less
dozortsev: fixed in https://github.com/transferwise/neptune-web/pull/1231

packages/css/src/less/footer.less
dozortsev: fixed in https://github.com/transferwise/neptune-web/pull/1231

PR #1214, opened 2021-11-15
Problem refreshing step with new model

JulianG-TW:
## 🖼 Context

There's an edge case with `refreshFormOnChange`.

When the refreshed step contains a new schema and a new model, `DynamicFlow` first re-renders `DynamicLayout` with a new schema (the `components` prop, to be exact). This results in a re-render of `DynamicFlow` with a new `key` prop, because it's based on a serialised version of components. Which in turn results in a new instance of `JsonSchemaForm`.

Then a `useEffect` hook is triggered and the `models` state is updated. This causes a re-render, but this time with the same `key` because the schema is the same.

`JsonSchemaForm` has a limitation at the moment, and its `model` prop is actually `initialModel` since subsequent updates to that prop have no effect. (Mostly because the `AllOfSchema` and the `OneOfSchema` components keep their model in local state (`useState`) ignoring model prop updates.

The result is that in some cases (see GIFs below) the visual representation of the form doesn't reflect the `model(s)` in `DynamicFlow` which is what eventually gets submitted.

| Recipients Page | DynamicFlow |
| ---------------- | ------------- |
| email-match-recipients-page | email-match-dynamic-flow |
| ✅   the correct option is selected | ❌  the wrong option is selected |

## 🚀 Changes

Without going to the trouble of attempting to change `JsonSchemaForm` we can easily avoid this edge case by grouping to local `useState` hooks.

Instead of:

```js
  const [stepSpecification, setStepSpecification] = useState({});
  const [models, setModels] = useState({});
```

We can have:

```js
```

We also eliminate the `useEffect` hook that updates the `models` when the `stepSpecification` has changed, and instead do it all in one go in the new `updateStepSpecification` function. **This is the key to avoid an early re-render with updated schema but  stale models.**

| ❌ Before | ✅  After |
| ---------------- | ------------- |
| before | after |

## 🤔 Considerations

I wrote a new DynamicFlow.end-to-end.spec.js test suite, because the DynamicFlow.spec.js test suite uses Enzyme, and makes assertions on the props that DF passes down to its `DynamicLayout` child.
This makes it very difficult (or impossible) to properly reproduce this issue in a test and then get the test to pass with the fix.
The props that DF passes down to DL are always correct in the test, even without the fix.

The new test suite contains only one test, which fails in the commit prior to the fix commit, and then passes.

## ✅ Checklist


packages/test-config/config/setupTests.js
```diff
@@ -1,15 +1,13 @@
-/* eslint-disable */
-const Enzyme = require('enzyme');
 const Adapter = require('@wojtekmaj/enzyme-adapter-react-17');
+const Enzyme = require('enzyme');
 
 global.fetch = require('jest-fetch-mock');
 Enzyme.configure({ adapter: new Adapter() });
 
-global.requestAnimationFrame = (cb) => cb();
```
JulianG-TW, line 8: I removed this line to prevent an infinite loop:

```
console.error
  Error: Uncaught [RangeError: Maximum call stack size exceeded]
```

packages/dynamic-flows/src/flow/DynamicFlow.end-to-end.spec.js
```diff
@@ -0,0 +1,59 @@
...
+import DynamicFlow from './DynamicFlow';
+import recipientDetailsJson from './examples/recipient_details.json';
+import recipientDetailsRefreshJson from './examples/recipient_details_refresh.json';
+
+const routes = {
+  '/recipient-details': recipientDetailsJson,
+  '/recipient-details-refresh': recipientDetailsRefreshJson,
+};
+
+const wait = (t) => new Promise((resolve) => setTimeout(resolve, t));
```
richardbata, line 15: You can use wait from `import { wait } from '../test-utils';`

packages/dynamic-flows/src/flow/DynamicFlow.end-to-end.spec.js
```diff
@@ -0,0 +1,59 @@
...
+  describe('when the step is refreshed and the new step contains data', () => {
+    it('displays the new data in the form', async () => {
+      renderDF();
+
+      const emailField = await screen.findByLabelText('Their email');
+
+      act(() => fireEvent.change(emailField, { target: { value: 'anything' } }));
+
+      await waitFor(async () => {
+        expect(screen.getByDisplayValue('bobloblaw@law.blog')).toBeInTheDocument();
```
richardbata, line 55: Did you not want to test for if the correct checkbox was selected based on whats in the model of the onRefresh?

richardbata (approved):

packages/test-config/config/setupTests.js
```diff
@@ -1,15 +1,15 @@
...
+global.requestAnimationFrame = async (callback) => callback();
```
JulianG-TW, line 7: I changed this line to prevent an infinite loop in the new test.

```
console.error
  Error: Uncaught [RangeError: Maximum call stack size exceeded]
```

packages/dynamic-flows/src/flow/DynamicFlow.end-to-end.spec.js
JulianG-TW: Yes. But I think it's much easier to assert the value of the email field.

In this test:
- We change the value of the email field to "anything"
- That causes a refresh.
- The refreshed step contains `"email": "bobloblaw@law.blog"`
- and we should expect the email field to change as a result

Also, in this PR, we're not testing nor fixing the function that decides which of the `oneOf` schemas should be selected. The reason for the issue was that anything in the new model was being ignored.

I picked the easiest thing to assert on.

packages/dynamic-flows/src/flow/DynamicFlow.end-to-end.spec.js
JulianG-TW: Thanks! done!

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -1,7 +1,7 @@
 import { Loader } from '@transferwise/components';
 import { isEmpty, isObject } from '@transferwise/neptune-validation';
 import Types from 'prop-types';
-import { useEffect, useState } from 'react';
+import { useEffect, useState, useMemo } from 'react';
```
StevePole, line 4: Where did you use 'useMemo'?

JulianG-TW: In line 113.

I noticed that 
```js
const [modelIsValid, setModelIsValid] = useState(false);
```
and
```js
  useEffect(() => {
    setModelIsValid(areModelsValid(models, stepSpecification.schemas));
  }, [models]);
```
were being used to memoise the `areModelsValid` calculation.

So I changed them to a memoized, computed value, rather than local state:

```js
  const modelIsValid = useMemo(
    () => areModelsValid(models, stepSpecification.schemas),
    [models, stepSpecification.schemas],
  );
```

PR #1216, opened 2021-11-18
feat: add dropfade animation

helendurranttw:
## 🖼 Context

We use this little animation to show an asynchronously loaded Nudge. I thought I'd componentise it so we could make use of it in other places in the product.

## 🚀 Changes

Added a new DropFade component.

https://user-images.githubusercontent.com/61203573/142447171-bc87992c-c4e6-48d2-a5fd-6c3740b6268e.mov

## ✅ Checklist


packages/components/src/dropFade/DropFade.tsx
```diff
@@ -0,0 +1,31 @@
...
+      onEntering={() => {
+        // Set height so we can animate to the correct height
+        const height = reference?.current?.scrollHeight || 0;
+        setHeight(`${height}px`);
+      }}
+      // Un-set height, as we don't want to presume it should be that height forever
+      onEntered={() => setHeight('')}
+    >
+      {/* eslint-disable-next-line react/forbid-dom-props */}
+      <div style={{ height }} className="np-dropfade">
```
matebek-tw, line 24: I believe `height: 10` is the same as `height: '10px'` in React world, which could simplify some of the code a bit.

package.json
```diff
@@ -44,6 +44,7 @@
     "@transferwise/neptune-tokens": "^1.0.0",
     "@types/react": "^17.0.14",
     "@types/react-dom": "^17.0.9",
+    "@types/react-transition-group": "4.4.4",
```
matebek-tw, line 47: Shouldn't this be added to `packages/components/package.json` instead?

packages/components/src/dropFade/DropFade.tsx
```diff
@@ -0,0 +1,36 @@
+import React, { useRef, useState, ReactElement } from 'react';
+import CSSTransition from 'react-transition-group/CSSTransition';
+
+const DropFade = ({ children, show }: { children: ReactElement; show: boolean }) => {
+  const reference = useRef<HTMLDivElement>(null);
+  const [height, setHeight] = useState<number | undefined>(0);
```
matebek-tw, line 6: Will this be ever `undefined`? We are calling `setHeight` with `|| 0` everywhere, so I believe this can be a `number` all the time.

matebek-tw (approved):

PR #1217, opened 2021-11-18
fix(PhoneNumberInput): do not change prefix on phone number input change

matebek-tw:
## 🖼 Context

https://transferwise.atlassian.net/browse/DS-1453

As `internalValue` is stored as a singe value in the local state, entering into the phone number input an otherwise matching prefix number triggering a prefix change. Fixed it by storing the `prefix` and the `suffix` numbers as different state keys, and only updating one or the other.

## 🚀 Changes

- Refactoring `PhoneNumberInput` to handle `prefix` and `suffix` as different keys on `internalValue` state

## 🤔 Considerations

On paste, we still do not differentiate between +20-11223344 and +2011-223344, it will always be evaluated as the latter.

## ✅ Checklist


packages/components/src/phoneNumberInput/PhoneNumberInput.js
```diff
@@ -39,26 +39,35 @@ const PhoneNumberInput = (props) => {
...
+
+    return {
+      prefix: initialPrefix,
+      suffix: initialSuffix,
+    };
   };
 
   const [internalValue, setInternalValue] = useState(getInitialValue());
-  const [broadcastValue, setBroadcastValue] = useState(getInitialValue());
+  const [broadcastedValue, setBroadcastedValue] = useState(
```
helendurranttw, line 61: Since you've got checking inside getInitialValue to ensure the value is clean, can we not still use getInitialValue here?

matebek-tw: The problem is (that I created for myself) that the shape of `getInitialValue` is `{prefix: string, suffix: string}`, which makes it more complicated to compare values before we trigger `props.onChange`.

packages/components/src/phoneNumberInput/PhoneNumberInput.js
```diff
@@ -39,26 +39,26 @@ const PhoneNumberInput = (props) => {
...
+        suffix: '',
+      };
+    }
+
+    return explodeNumberModel(cleanValue);
   };
 
   const [internalValue, setInternalValue] = useState(getInitialValue());
-  const [broadcastValue, setBroadcastValue] = useState(getInitialValue());
+  const [broadcastedValue, setBroadcastedValue] = useState(null);
```
helendurranttw, line 56: 👍

packages/components/src/phoneNumberInput/PhoneNumberInput.js
```diff
@@ -81,60 +81,68 @@ const PhoneNumberInput = (props) => {
...
     }
   };
 
   useEffect(() => {
-    const newbroadcastValue = isValidPhoneNumber(internalValue) ? cleanNumber(internalValue) : null;
-    if (newbroadcastValue !== broadcastValue) {
-      onChange(newbroadcastValue);
-      setBroadcastValue({ newbroadcastValue });
+    if (broadcastedValue === null) {
+      return setBroadcastedValue(internalValue);
```
helendurranttw, line 115: Don't think you need this `return`?

helendurranttw (approved):
Nice and clear 😌

packages/components/src/phoneNumberInput/PhoneNumberInput.js
matebek-tw: Have to early return otherwise the effect will continue to run and fail as the `broadcastedValue` is `null` at the moment of initialisation. We are using its initial `null` value to bail out from the `props.onChange` call on initial render.

PR #1219, opened 2021-11-19
fix: OneOfSchema. Form controls not properly reset when changing selected schema

JulianG-TW:
## 🖼 Context

In the video we  can see a `oneOfSchema` with two objects. The current model is printed below the form in a `<pre>` at all times.
After completing the fields in the `SortCode` Tab, when switching to the `Iban` tab, the `sortcode` and `account` properties are removed from the model because they don't match the schema.
When switching back to the `SortCode` tab, the input fields for sort code and account number continue to display the user-entered values, even though they're missing from the model.

In the past this has caused some problems in the recipient creation flow, where the forms looks correct, but the underlying model is different, and the form cannot be submitted.

See video below

https://user-images.githubusercontent.com/67635620/142675187-ad92c5ea-cf99-4c97-a7e4-e794c2fcbafa.mov

## 🚀 Changes

One way of fixing this is to add a `useEffect` hook in `OneOfSchema` to update its local state when the model prop changes, just like when it's first initialised.

- The first commit in this PR contains a new failing test.
- The second commit fixes the issue and gets the test to pass.

## 🤔 Considerations

- I haven't looked into avoiding local state completely for `OneOfSchema`.
- The `AllOfSchema` could probably benefit from a similar `useEffect` hook, but it wouldn't fix similar problem, since there's no switching between schemas in an `allOf`.

## ✅ Checklist


packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.js
```diff
@@ -101,6 +101,11 @@ const OneOfSchema = (props) => {
   const [schemaIndex, setSchemaIndex] = useState(getActiveSchemaIndex(props.schema, props.model));
   const [models, setModels] = useState(getModelPartsForSchemas(props.model, props.schema.oneOf));
 
+  // When the model or the schema.oneOf from props changes
+  useEffect(() => {
+    setModels(getModelPartsForSchemas(props.model, props.schema.oneOf));
```
StevePole, line 106: Is there a reason this can't be done in the next useEffect?

StevePole: This bit
image

My instinct was that this should happen whichever one changes

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.end-to-end.spec.js
```diff
@@ -0,0 +1,144 @@
...
+import { useState } from 'react';
+import { act } from 'react-dom/test-utils';
+
+import { wait } from '../../test-utils';
+
+import OneOfSchema from './OneOfSchema';
+
+const schema = {
+  title: 'Recipient Bank Details',
+  oneOf: [
```
StevePole, line 12: I would also add an allOf to check how that behaves.  
```
allOf: [
  { 
    title: "...",
    oneOf: [
      { ... }
    ]
  }
]
```

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.js
JulianG-TW: I can try merging the two effects. I didn't want to cause extra executions of the existing effect, though.

JulianG-TW: > I was expecting to see something similar in the AllOfSchema
I'd prefer to do that in the next PR, since it's not related to the inconsistency happening with oneOfs.

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.end-to-end.spec.js
JulianG-TW: I'd need to render a `GenericSchema` in that case, right?

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.js
JulianG-TW: > Is there a reason this can't be done in the next useEffect?

Are we sure we want to generate a new id every time the model changes?

https://user-images.githubusercontent.com/67635620/143049715-c6a07c60-04b2-4731-8439-4386cd320c18.mov

(the number of re-renders seem to be unaffected, though)

StevePole: You're right, I don't want to generate new ids, but I wasn't sure about the other things.  Thinking about it, default index should be first time only.  Using the model to figure out the valid index would only be useful if the refreshOnChange was attempting to change the value of the oneOf.  I think that's _probably_ not something we want to do.

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.end-to-end.spec.js
StevePole: I didn't realise this was a test only for oneOf,  Wouldn't it better to use dynamic form level end to end tests for this behaviour?  I'm not really sure what an "end to end test" for oneOf Schema really means, given it's recursive, it's could be testing any type of schema.

JulianG-TW: Agree. I don't like the end-to-end suffix either.

**I can rename the test file to something else... not sure what.**

The difference between this test and the OneOfSchema.spec.js suite, is that this one makes assertions on the DOM, rather than on props passed to children, and callback functions being called.

I wrote my test in a new file because I had to use React Testing Library instead of Enzyme for this.
Actually, **I** don't know how to properly mount and assert on the DOM using Enzyme, and Enzyme is being discouraged in this repo, so RTL sounded like a better option.

How about:
- [ ] OneOfSchema.dom.spec.js
- [ ] OneOfSchema.rtl.spec.js
- [ ] OneOfSchema.test.js (needs a config change AFAIK)

StevePole: The philosophy of the testing suite thus far has been that unit tests mock child components (i.e. shallow in enzyme terms) and test what is passed down, because they are unit tests.  That doesn't mean they can't interrogate the DOM, in fact, they should interrogate the DOM if they actually render things.  It's just that dynamic forms is somewhat unusual in that most of the components actually don't render much, they typically render Neptune components (which are mocked), or nothing, because they're just logical containers.  

The end to end tests (you might also call them integration, but personally I consider them e2e due to the DOM) on the other hand, do not mock child components (mount in enzyme, or default in RTL), and do not test what is passed down to child components, they only test the direct impact on the DOM.

There aren't any end to end tests for the schema components in the hierarchy because they're recursive.  If you were to test ObjectSchema with RTL, you wouldn't only be testing ObjectSchema, but also GenericSchema, BasicTypeSchema, Field, FormControl, ControlValidation, and potentially others.  This is also true of other schemas.  So writing end to end tests at this level would end up retesting much of the same code, and lead to a total loss of test independence.  i.e. A breaking change to Form Control would break the tests of every other schema component.  With the setup the way it is, it would break two tests.  The FormControl unit tests and the JsonSchemaForm end to end tests. 

If you look at the JsonSchemaForm.e2e.spec.js, these are actually already RTL, and could easily be extended to test what happens when the model changes.

JulianG-TW: Ok, I have now moved the new test to JsonSchemaForm.end-to-end.spec.js, of course rendering a `JsonSchemaForm` instead of a `OneOfSchema`; and I also wrapped the schema in an `allOf`.

StevePole (approved):

PR #1223, opened 2021-11-23
fix(Dimmer): fade background color on open/close

matebek-tw:
## 🖼 Context

https://transferwise.atlassian.net/jira/software/projects/DS/boards/277?label=Web&selectedIssue=DS-1212

## 🚀 Changes

- Migrate `Dimmer` to TypeScript
- Move background color on `Dimmer` to the `.dimmer--enter-done` class

## ✅ Checklist


packages/components/src/dimmer/Dimmer.tsx
```diff
@@ -14,8 +13,19 @@ export const EXIT_ANIMATION = 350;
 
 const dimmerManager = new DimmerManager();
 
-export const handleTouchMove = (event) => {
-  const isTouchedElementDimmer = event.target.classList.contains('dimmer');
+type DimmerProps = PropsWithChildren<{
+  className?: string;
```
dozortsev, line 17: `className` should come from `CommonProps`

packages/components/src/dimmer/Dimmer.tsx
```diff
@@ -118,36 +148,11 @@ export const DimmerContentWrapper = ({ children, scrollBody }) => {
...
-  disableClickToClose: false,
-  fadeContentOnEnter: false,
-  fadeContentOnExit: false,
-  transparent: false,
-  onClose: undefined,
-  open: false,
-  scrollable: false,
-};
-
 export { Dimmer };
```
dozortsev, line 156: I think you can now delete this statement ([for context here is similar example](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/moneyInput/MoneyInput.js#L381-L383))

packages/components/src/dimmer/Dimmer.tsx
```diff
@@ -118,36 +148,11 @@ export const DimmerContentWrapper = ({ children, scrollBody }) => {
...
 
 export default withNextPortal(Dimmer);
```
dozortsev, line 158: I think in order for tsc to properly generate correct types [this function](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/withNextPortal/withNextPortal.js) needs written also on TS

dozortsev (commented):
Looks good 👍 

But please make changes to improve types.

atm it generates following

```js
// at the end of Dimmer.d.ts file
..
export { Dimmer };
declare const _default: (props: any) => import('react').ReactPortal | null;
export default _default;
```

this doesn't look right 🤔 I think consumers will get following error/warning on their side, but I could be wrong

> JSX element type 'Dimmer' does not have any construct or call signatures

dozortsev (commented):
I would also delete Dimmer from here https://transferwise.github.io/neptune-web/blog/2021-04-30-april-release#missing-types

dozortsev (approved):

packages/components/src/dimmer/Dimmer.tsx
```diff
@@ -14,8 +13,19 @@ export const EXIT_ANIMATION = 350;
 
 const dimmerManager = new DimmerManager();
 
-export const handleTouchMove = (event) => {
-  const isTouchedElementDimmer = event.target.classList.contains('dimmer');
+type DimmerProps = CommonProps & {
+  children?: ReactNode;
```
dozortsev, line 17: maybe `React.PropsWithChidren`?

dozortsev (approved):

packages/components/src/withNextPortal/withNextPortal.tsx
```diff
@@ -1,8 +1,8 @@
-import { useState, useEffect } from 'react';
+import { useState, useEffect, ComponentType } from 'react';
 import { createPortal } from 'react-dom';
 
-export default function withNextPortalWrapper(Component) {
-  return function (props) {
+export default function withNextPortalWrapper<T>(Component: ComponentType<T>) {
```
dozortsev, line 4: what type of data this function returns? shouldn't we explicitly define type here? 🤔

matebek-tw: It either returns `ComponentType<T>` or `null`, TypeScript can assert the type based on the return value.

packages/components/src/dimmer/Dimmer.tsx
dozortsev: ok

packages/components/src/withNextPortal/withNextPortal.tsx
dozortsev: ok

PR #1229, opened 2021-11-25
fix: remove dir selector from position properties in rtl mixin

nabaroa:
## 🚀 Changes

- In the rtl mixing for `left`& `right` position properties we had `html:not([dir="rtl"]) & ` selector to target default direction or LRT. This was causing specificity issues if the consumer is not going to be using the rtl mixin when bumping the version. We decided to do this rtl change transparent for the consumer so I removed that selector, as we did with `margin` & `padding` in previews PR.
- Change the second variable logic in the mixin, to set the original direction value to `initial` by default for RTL. This is happening in the most cases, so I believe that when adopting is going to be easier and it is also more easy to explain.
- Adjust some values: Removed some cases that the mixin was not needed and update some second value cases.

## ✅ Checklist


packages/docs/pages/styles/addons/RtlLayout.mdx
```diff
@@ -50,23 +50,32 @@ padding: 0 var(--foo) 0;
...
 ```
 margin-left: var(--foo); → .margin(left, var(--foo));
 ```
-There is a second optional value for the mixin `.margin(direction, value1, value2)`, if you need to specify/reset the oposite direction value.
+The output of above code is: 
+```
+Default or LTR → margin-left: var(--foo);
+RTL → margin-right: var(--foo); margin-left: 0;
+```
+The original margin resets to initial value. If that value is not 0 or initial, there is a second optional value (value2 in the next example) for the mixin `.margin(direction, value1, value2)` which you can use if you need to specify another oposite direction value.
```
helendurranttw, line 64:
```suggestion
The original margin resets to the initial value. If that value is not 0 or initial, there is a second optional value (`value2` in the next example) for the mixin `.margin(direction, value1, value2)` which you can use if you need to specify a different value for the opposite direction.
```

packages/docs/pages/styles/addons/RtlLayout.mdx
```diff
@@ -50,23 +50,32 @@ padding: 0 var(--foo) 0;
...
 
-By instance, let's imagin that we have a `margin-left: 8px`. The mixin for it would be `.margin(left, 8px)` and on RTL cases would be transformed to `margin-right: 8px`.
-By default, it also maintains the property `margin-left`, because the mixin doesn't have the information of its value. We may want to reset it to `0` or add another value depending the case. We have this `value2` option to do that.
+By instance, let's imagin that you have a `margin-left: 8px` in your code that you want to transform to RTL friendy property. The mixin for it would be `.margin(left, 8px)` and in RTL cases would be transformed to `margin-right: 8px; margin-left: 0;`.
```
helendurranttw, line 66:
```suggestion
For instance, let's imagine that you have a `margin-left: 8px` in your code that you want to transform into a RTL friendly property. The mixin for it would be `.margin(left, 8px)`, and in RTL cases it would be transformed into `margin-right: 8px; margin-left: 0;`.
```

packages/docs/pages/styles/addons/RtlLayout.mdx
```diff
@@ -50,23 +50,32 @@ padding: 0 var(--foo) 0;
...
+You might want to change the `0` to another value depending the case, you have this optional `value2` variable do that:
```
helendurranttw, line 67:
```suggestion
You might want to change the `0` to a different value. If you do, you can provide the optional second value (the third parameter below):
```

helendurranttw (approved):
Left a few docs suggestions, otherwise looks great :)

PR #1232, opened 2021-11-29
fix: add back tokens removed in neptune-tokens updates

helendurranttw:
## 🖼 Context

We will eventually want to remove these tokens, but to do so in the next release would mean a big breaking change. Instead we will deprecate and advise people what to use instead.

## 🚀 Changes

 
Add back removed tokens.

## ✅ Checklist


matebek-tw (approved):

packages/css/src/props/neptune-tokens.css
```diff
@@ -1,2 +1,12 @@
 @import '@transferwise/neptune-tokens/colors-base.css';
 @import '@transferwise/neptune-tokens/tokens.css';
+
+/* DEPRECATED */
```
dozortsev, line 4: when deprecate something always mention what to use instead

dozortsev, line 4:
```suggestion
/* DEPRECATED, use neptune-tokens instead: https://transferwise.github.io/neptune-web/styles/tokens/Colors */
```

packages/css/src/variables/neptune-tokens.less
```diff
@@ -1,2 +1,12 @@
 @import '@transferwise/neptune-tokens/tokens.less';
 @import '@transferwise/neptune-tokens/colors-base.less';
+
+// DEPRECATED
```
dozortsev, line 4:
```suggestion
// DEPRECATED, use neptune-tokens instead: https://transferwise.github.io/neptune-web/styles/tokens/Colors
```

PR #1236, opened 2021-11-30
[VUX-1226] Adding Tracking to DF + Camera Tracking Events

richardbata:
## 🖼 Context
Lets add tracking to Dynamic-flows. Its consumer will have control over how the events are sent, and internally a React.Context is used to bubble events up to the library's consumer.

## 🚀 Changes
- React.Context internal event bubbling
- onTrackingEvent prop added to DynamicFlow which will allow consumer of the lib to control sending tracking events

## ✅ Checklist


packages/dynamic-flows/src/step/cameraStep/cameraCapture/CameraCapture.js
```diff
@@ -72,6 +80,21 @@ const tryAcquireFullScreenAndLockOrientation = () => {
   return Promise.resolve();
 };
 
+const getTrackingProperties = (props, videoStream, error) => {
+  const videoTrack = (videoStream?.getVideoTracks() || [])[0];
+  return {
+    ...(videoTrack && {
+      'Camera Capabilities': videoTrack?.getCapabilities(),
+      'Camera Settings': videoTrack?.getSettings(),
```
StevePole, line 88: I'm slightly wary of capturing this data, as I presume it is not mentioned in our privacy policy.  If you do want/need it, I'd consult the policy to see if you feel this is covered.

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -79,6 +81,11 @@ const DynamicFlow = (props) => {
 
   const httpClient = propsHttpClient || defaultHttpClient.init({ baseUrl });
 
+  useEffect(
+    () => trackingEventBus.on(({ eventName, properties }) => trackEvent(eventName, properties)),
```
StevePole, line 85: Can this be `trackingEventBus.on(trackEvent)`, or is there a scoping issue?

packages/dynamic-flows/src/step/cameraStep/CameraStep.js
```diff
@@ -61,6 +66,7 @@ const CameraStep = (props) => {
         instructions={instructions}
         showPreview={false}
         direction={direction}
+        trackEvent={handleTrackEvent}
```
StevePole, line 69: Is this because you're anticipating moving this to Neptune?

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -255,6 +262,7 @@ DynamicFlow.propTypes = {
   httpClient: Types.shape({
     request: Types.func,
   }),
+  trackEvent: Types.func,
```
StevePole, line 265: We have a convention in the project that callback are prefixed 'on', perhaps you could change this to `onTrackableEvent`?

packages/dynamic-flows/src/common/tracking/index.js
```diff
@@ -0,0 +1,33 @@
+const EVENT_NAME_PREFIX = 'Dynamic Flow - ';
+const TRACKING_EVENT = 'TRACKING_EVENT';
+
+const getGeneralEventProperties = () => {};
+
+const eventBus = {
```
StevePole, line 6: Can we write tests for this file?

StevePole (commented):
It looks good 👍  

Minor comments/requests.  Tests are most important :-)

packages/dynamic-flows/src/flow/DynamicFlow.js
richardbata: I'd like to separate eventName and properties to stay with the mixpanel API convention. 
Ie. onTrackableEvent should have 2 regular parameters, but in the bus its not ideal sending 2 parameters, rather just the object

packages/dynamic-flows/src/step/cameraStep/cameraCapture/CameraCapture.js
richardbata: image

It doesn't explicitly say anything about IO devices, or hardware information. What do you think? I'd love to atleast see:
- resolution
- direction of camera

JulianG-TW (commented):
Would you consider changing the PR title to avoid mentioning Mixpanel?
I think it sounds like we're adding a dependency on mixpanel, but we aren't, right?

As I understand, consumers can use anything to handle the trackable events.

JulianG-TW (approved):
Looks good. I left two in-line comments, and perhaps we should update the PR description, since it still mentions the event bus.

packages/dynamic-flows/src/step/cameraStep/CameraStep.js
```diff
@@ -61,6 +64,7 @@ const CameraStep = (props) => {
         instructions={instructions}
         showPreview={false}
         direction={direction}
+        onTrackableEvent={onTrackableEvent}
```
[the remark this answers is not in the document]
richardbata: Yeah, thats the idea

PR #1242, opened 2021-12-01
Update the convertFinalStepToDynamicLayout utility function to match mobile clients implementation

JulianG-TW:
## 🖼 Context

The web implementation of DynamicFlows seems to differ slightly from its mobile counterparts:

| Web | Android | iOS |
|-----|-----|-----|
| web | Android | iOS |

- The title is above the image instead of below.
- The call-to-action button is not solid blue (primary).

## 🚀 Changes

When converting a final step to a dynamic layout:
- Place the image above the title
- Make actions default to type "primary"

| Before | After |
| --- | --- |
| before | after |

## ✅ Checklist


StevePole (approved):
I think I did this to match the implementation of V2 on web, but if mobile does it differently, we should align.

PR #1244, opened 2021-12-02
feat: add onFilesChange callback to UploadInput

NewmanJWise:
## 🖼 Context

In our flow we would like to be able to block the user from carrying on in our flow if there's a validation error in our optional `UploadInput` field until they resolve it.

The current implementation of the component does not tell us when an internal component validation error is resolved such as wrong file type. This means that we do not know when the user resolves this error when they delete the file as `onDeleteFile` is only ever called when it's a server request which it is not when you are deleting an internal upload.

I spoke with @matebek-tw about the best quick solution and we decided that a non breaking quick fix would be to just expose the internal files array whenever it is changed. This adds a lot more power and flexibility to the component for all consumers without breaking any existing implementations already working with the prior assumptions of when `onDeleteFile` is called.

## 🚀 Changes

 

Added `onFilesChange` callback

## 🤔 Considerations

N/A

## ✅ Checklist


packages/components/src/uploadInput/UploadInput.tsx
```diff
@@ -209,6 +218,16 @@ const UploadInput = ({
     }
   };
 
+  useEffect(() => {
+    setMounted(true);
```
NewmanJWise, line 222: Used to prevent the callback being called instantly when the component is mounted with `files` passed in as a prop.

packages/components/src/uploadInput/UploadInput.tsx
```diff
@@ -209,6 +218,16 @@ const UploadInput = ({
...
+  }, []);
+
+  useEffect(() => {
+    if (onFilesChange && mounted) {
+      onFilesChange(uploadedFiles);
+    }
+  }, [onFilesChange, uploadedFiles]); // eslint-disable-line react-hooks/exhaustive-deps
```
matebek-tw, line 229: Why do we need to disable this eslint rule? Couldn't we list all the dependencies used inside the hook?

packages/components/src/uploadInput/UploadInput.story.tsx
```diff
@@ -133,3 +133,12 @@ withManualDownloadHandler.args = {
     alert(`Manual download handler triggered for: ${JSON.stringify(file)}`);
   },
 };
+
+export const withFilesChangeHandler = Template.bind({});
+withFilesChangeHandler.args = {
+  ...props,
+  files,
+  onFilesChange: (files: UploadedFile[]) => {
+    alert(`Files change handler triggered for: ${JSON.stringify(files)}`);
```
matebek-tw, line 142: Would be better to use storybook `action` or `console.log` for this I think.

NewmanJWise: Agreed. I was following the pattern in the file but I prefer `action` for this too, I'll make the change.

packages/components/src/uploadInput/UploadInput.tsx
NewmanJWise: If `mounted` is added to the deps then we go back to firing the callback as soon as we are mounted with files :(

matebek-tw (approved):

PR #1247, opened 2021-12-03
feat(dynamic-flows): accept size property for step and layout images

JulianG-TW:
## 🖼 Context

`DynamicImage` component does not specify its size. Until now we had to wrap them in boxes of different widths to get it to render at a specific size.

Also in `xs` (narrow viewport) the images were always full-width.

## 🚀 Changes

We now accept a `size` property for images, with possible values of: `xs`, `sm`, `md`, `lg`, `xl`.

These map to some css utility classes which are different than those used in `DynamicBox`:

| size | classes |
| ---- | -------- |
| xs | col-xs-2 col-xs-offset-5 |
| sm | col-xs-4 col-xs-offset-4 |
| md | col-xs-6 col-xs-offset-3 col-lg-4 col-lg-offset-4 |
| lg | col-xs-10 col-xs-offset-1 col-md-8 col-md-offset-2 |
| xl | col-xs-12 |

| viewport | image size: xs                                                                                                                   | image size: sm                                                                                                                   | image size: md                                                                                                                   | image size: lg                                                                                                                   | image size: xl                                                                                                                   |
| -------- | -------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------- |
| xs       | xs | sm | md| lg | xl |
| md       | xs | sm| md | lg |xl |
| lg       | xs | sm | md | lg | xl |

## 🤔 Considerations

Images now have a default margin-bottom with a value of `lg` that translates to `m-b-3`.

LayoutService no longer wraps each image in a box, instead it gives images a default size of `md`.

## ✅ Checklist


packages/dynamic-flows/src/step/layoutStep/layoutService/LayoutService.js
```diff
@@ -155,12 +155,14 @@ function convertStepImageToDynamicImage(image) {
...
-    margin: 'lg',
-  };
+  return typeof image === 'string'
+    ? {
+        type: 'image',
+        url: image,
+        text: undefined,
+        margin: 'lg',
+      }
+    : convertStepImageToDynamicImage(image);
```
JulianG-TW, line 165: in order to provide an image size, the `image` property in the final step cannot be just a string. it needs to be an object, in that case we can use the more generic `convertStepImageToDynamicImage`.
I think we should keep supporting strings for some time to avoid immediate breaking changes.

packages/dynamic-flows/src/layout/image/index.js
```diff
@@ -9,6 +9,23 @@ const getImageClasses = (component) => {
...
 
+const getWrapperClasses = (component) => {
+  const margin = getMarginBottom(component.margin ? 'lg' : 'xs');
+  switch (component.size) {
+    case 'xs':
+      return `${margin} col-xs-2 col-xs-offset-5`;
+    case 'sm':
+      return `${margin} col-xs-4 col-xs-offset-4`;
+    case 'md':
+      return `${margin} col-xs-6 col-xs-offset-3 col-lg-6 col-lg-offset-3`;
```
StevePole, line 20: You don't need to specify classes for wider grids unless you're overriding something from the smaller grid, so the lg grid classes are redundant here.

StevePole:
I don't understand what you're saying here:

> the screenshots above are for different image sizes wrapped in an "xl" box which is full-width. However, when converting final step details to layout, the utility function uses "sm" as default width for the wrapping box, and this results in smaller image sizes.

I wouldn't want to visually optimise the image rendering for a case where it's inside another box, as that will not be the norm in layouts.  It would be better to think about what the image should do independently.

StevePole:
With the images in the description, it would be really helpful to use an image without a white background.  It's not possible to tell how wide the image actually is due to the whitespace.

packages/dynamic-flows/src/layout/image/index.js
StevePole: However, I have typically used slightly narrower grids on wider screens, as to retain the same feel across screen sizes, it is often necessary to use less of the available space.  Hard to tell because the screenshot use an image with a white background, and I'm not sure how wide the "wide" page was.  Ideally should look at a grid at least 1440px wide (from memory I think that's our widest container width).

packages/dynamic-flows/src/layout/image/index.js
```diff
@@ -19,14 +36,25 @@ const DynamicImage = (props) => {
     shrink: true,
   };
 
-  return <Image className={getImageClasses(image)} {...imageProps} />;
+  if (!image.size || image.size === 'xl') {
+    return <Image className={getImageClasses(image)} {...imageProps} />;
+  }
+
+  return (
+    <div className="row">
```
StevePole, line 44: I believe rows and columns both add margin bottom, at least on some screen widths, so I think we need to cancel that on both components with an `.m-b-0`.

packages/dynamic-flows/src/layout/image/index.js
JulianG-TW: I've updated the screenshots to:
- use a squared, non-transparent placeholder image
- red dotted and dashed borders for the DynamicBox component
- blue dotted and dashed borders for the DynamicImage component

We can see a margin-bottom, like you predicted. I'll try to fix it as suggested.

packages/dynamic-flows/src/layout/image/index.js
JulianG-TW: Done in 2daffdb92c2e4e093c546e431053af3ae55a7cf2

But then I had to add a margin to the box that surrounds the image in LayoutServices in 03680f05377c0fb49b36211cd65a84c7f3750e99

| Before                                                                                                           | After                                                                                                           |
| ---------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------- |
| before | after |

JulianG-TW:
@StevePole 
> I don't understand what you're saying here:
I've now expanded on that topic. Hopefully it now becomes clear. Let me know if it's not, and we can talk.

packages/dynamic-flows/src/layout/image/index.js
StevePole: Yeah, makes sense.  Adding in the layout service is preferable, because it means those using layouts still have full control.  those who didn't specify a layout get something that looks decent automatically.

StevePole (approved):
👍

PR #1254, opened 2021-12-09
feat: remove embedded paddings from Header and Option components

matebek-tw:
## 🖼 Context

https://transferwise.atlassian.net/browse/DS-1535

https://www.figma.com/file/x00ITfRqd3XzKeo1jhdF2x/Embedded-paddings%2C-semantic-spacing-tokens%2C-new-Section-components%2C-changes-in-Option-and-Card?node-id=1147%3A35120

Remove the embedded paddings from Option components we added in 92e91aa87596f9de1b767d8c7ffa5352be6ddda3 as part of the Section implementation. This is currently causing a lot of confusion, and on most product pages the expected behaviour is not really achievable without targeting implementation details in our stylesheets.

image

## 🚀 Changes

- Remove embedded padding from `Header`
- Remove the embedded padding from `Option`
- Do not remove bottom border from Option when used inside a Section
- Add `withHorizontalPadding` property to `Section`
- Reduce `.decision.media` horizontal paddings to `12px`

## ✅ Checklist


dozortsev (commented):
~do we use `space-content-horizontal` token? asking because it was create as part of this project, wondering whether we still need it?~

packages/css/src/less/decision.less
```diff
@@ -57,12 +56,6 @@
     }
   }
 
-  // Style overrides for Angular components
-  &.media {
-    margin: 0;
-    padding: var(--size-24) var(--size-16);
-  }
```
dozortsev, line 64: why removing this? it might be necessary for styleguide-components

dozortsev (approved):

PR #1255, opened 2021-12-10
feat(UploadInput): Accept an optional description for the upload button

dhananjaykha:
## 🖼 Context

The JSON Schema based DFv3 spec specifies a description for the Multiple file upload component. However, currently the UploadInput does not accept a description. 

It is computed from the fileTypes provided.

## 🚀 Changes

 
 UploadInput component accepts a prop description which, if present, is shown below the UploadButton, else is calculated from the filetypes(earlier logic).

## ✅ Checklist

- [x] Tested Edge, Firefox, Safari, and Chrome


packages/components/src/uploadInput/uploadButton/UploadButton.tsx
```diff
@@ -24,13 +25,18 @@ export type UploadButtonProps = {
...
+  fileTypes?: typeForAllowedFileTypes;
 
   /**
    * Size limit in KBs 1000 KB = 1 MB (default: 5000 KB)
    */
   sizeLimit?: number;
 
+  /**
+   * Description for the upload button
+   */
```
smallosaurus, line 37: Not convinced that this comment adds extra value to the variable name. WDYT?

packages/components/src/uploadInput/uploadButton/UploadButton.tsx
```diff
@@ -105,23 +112,44 @@ const UploadButton = ({
...
+      fileTypes: derivedFileDescription,
+      size: Math.round(sizeLimit / 1000),
+    });
+  }
+
+  function getAcceptedTypes() {
+    const areAllFilesAllowed = getFileTypesDescription() === '*';
+
+    if (areAllFilesAllowed) {
+      return null; //file input by default allows all files
```
smallosaurus, line 144: why not return an empty array? that way if something is iterating, the code doesn't need to do a null-check
something like this - `{ accept: [] };`

smallosaurus: possible advantage of the above is then this function can just populate that list as it goes, and then we get a single return statement

OFFLlNE (approved):
Very clear and easy to read, LGTM!
Otherwise I think Adam has good comments brought up :)

packages/components/src/uploadInput/uploadButton/UploadButton.tsx
dhananjaykha: I was thinking to rename the property as it may clash with the uploaded file's description.
https://neptune.wise.com/components/upload-input/#anatomy

Someone from the design system can suggest if there is a terminology we use already.

packages/components/src/uploadInput/uploadButton/UploadButton.tsx
dhananjaykha: 1. Returning empty array would set an `accept` attribute on the file with `""` value. 
Hence I felt better to give it types as `accept:"something"` or `null` (not add the attribute to the input itself)

Also, I attempted to not break the existing contract there. I just rewrote the existing logic extracting in a function.

- - - - 

2. Populating a variable is good idea but it makes things less traceable, IMO. Also, more imperative. Early returns are helpful to reduce the Cyclomatic complexity.

sviat-volkov (approved):

packages/components/src/uploadInput/uploadButton/UploadButton.tsx
```diff
@@ -10,6 +10,7 @@ import MESSAGES from './UploadButton.messages';
 import { imageFileTypes, DEFAULT_SIZE_LIMIT } from './defaults';
 import getAllowedFileTypes from './getAllowedFileTypes';
 
+type typeForAllowedFileTypes = string | string[] | FileType[];
```
matebek-tw, line 13: We usually write custom types capitalised, could you please rename this to `AllowedFileTypes`?

matebek-tw (approved):

packages/components/src/uploadInput/uploadButton/UploadButton.tsx
dhananjaykha: Done ✅

PR #1256, opened 2021-12-14
fix: RTL layout bugs

nabaroa:
## 🖼 Context

Founded some layout issues when bumping the neptune-css v.12 in public pages.

## 🚀 Changes

-fix: float utility directions rtl: Wrong use of the RTL mixin in float utilities. This is a bug that we didn't catch before because depending of the ordering of the CSS it was hidden. 

-fix: remove RTL mixin when both values are equal, specially in the column grid system. In the beta running (https://wise.com/ar/currency-converter/) was stable but with the last changes (removing the breaking change need) appeared an inconsistency in the directions. Most easy and clean solution is to use the mixin only when it's needed.

| Before | After|
|---|---|
| image|image |

## ✅ Checklist


nabaroa:
LTR
image

RTL
image

dozortsev (approved):

PR #1260, opened 2021-12-16
fix: OneOfSchema. Form controls not properly reset when changing selected schema (using radio buttons)

JulianG-TW:
## 🖼 Context

We partially fixed this in #1219, but we only tested the fix with **tab** controls. 

It turns out tabs trigger their `onChange` callback on mount, and that was helping clean up the model. Radio buttons, on the other hand, don't trigger `onChange` on mount. Because of this, the model was not being properly cleaned, and would not match the form controls.

Here's a video using radio buttons and an explanation of the problem:

https://user-images.githubusercontent.com/67635620/146406672-331c684e-e985-4cd1-93c0-4a8c9db1022e.mov

## 🚀 Changes

- The first commit in this PR adds some tests, and one of them fails as expected.
- The second commit fixes the issue and gets the test to pass.

The way we fix this is by always updating the model in the `useEffect` hook that runs on mount. The model is updated to contain only the properties in the selected one-of schema.

| ❌ . Before                                                                                                             | ✅ . After                                                                                                             |
| ----------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------- |
| one-of-before | one-of-after |
| The text fields in the second option contain values that are not really in the model.                                   | The text fields in the second option are empty, matching the underlying model.                                         |

## 🤔 Considerations

When submitting the form without switching one-ofs (depending on the control type) the original model is never updated so we end up submitting the entire model, including the parts that "don't belong" to the selected one-of.

[See my in-line comment in a test](https://github.com/transferwise/neptune-web/pull/1260/files#r771817939).

## ✅ Checklist


packages/dynamic-flows/src/flow/DynamicFlow.end-to-end.spec.js
```diff
@@ -47,11 +54,109 @@ describe('E2E: Given a DynamicFlow component to render', () => {
...
+          expect(screen.getByLabelText('BIC code')).toHaveValue('');
+
+          act(() => {
+            fireEvent.click(screen.getByLabelText('Inside Europe'));
+          });
+
+          expect(screen.getByLabelText('IBAN')).toHaveValue('');
+        });
+      });
+    });
```
JulianG-TW, line 106: This is the new test that fails before the fix commit.

packages/dynamic-flows/src/flow/DynamicFlow.end-to-end.spec.js
```diff
@@ -47,11 +54,173 @@ describe('E2E: Given a DynamicFlow component to render', () => {
...
+        </Provider>,
+      );
+    }
+
+    const expectedAction = { method: 'POST', title: 'Submit', type: 'primary', url: '/submit' };
+    const expectedData = {
+      name: { fullName: 'Bob Loblaw' },
+      details: expect.objectContaining({ iban: 'NL54ABNA9397935380' }),
+      type: 'Iban',
+    };
```
JulianG-TW, line 87: In three tests we assert on this `expectedData`.

The initial data is the model included in the initial step. And it looks like this:

```json
{
    "name": { "fullName": "Bob Loblaw" },
    "details": {
      "bic": "BUKBGB22",
      "iban": "NL54ABNA9397935380",
      "accountNumber": "123456789012"
    },
    "type": "Iban"
  }
```

Note that in `expectedData`, the `details` is expected to contain (at least) an `iban: 'NL54ABNA9397935380'` but it could contain more properties.

The intial one-of specifies only `iban`, so we could argue that when submitting with that one- of selected we don't want `bic` nor `accountNumber` to be submitted. But they are.

If we asserted on an `expectedData` with exactly `details: { iban: 'NL54ABNA9397935380' }` the 2 tests would fail.

This may be acceptable or not. I'm not sure. See the Considerations section in the PR description.

JulianG-TW:
Close in favour of https://github.com/transferwise/neptune-web/pull/1273

PR #1270, opened 2022-01-06
[VUX-1303] Fetch image from same domain using authenticated fetch

richardbata:
## 🖼 Context
In the IDWL V3 flow, we want dynamic images to be served by the backend through a `/files` endpoint rather than include the image contents in the JSON response in Base64, in order to improve efficiency, and solve an issue on Android where a buffer is overflows by large Base64 images in the JSON response.

## 🚀 Changes
Web implementation is such that when the image.url is a relative path in the JSON response, or a full URL, when the domain of the image matches that of the page its called from, authentication headers are sent in a manual fetch request, the response of which is manually populated in the IMG tag.

Specifically, the difference between the src of img and fetch is that fetch sends authentication cookies + `X-Access-Token` as headers.

Note, mobile clients have a slightly different approach, so best to sync all around before proceeding

## ✅ Checklist


JulianG-TW:
I'm curious why we can't let the browser `img` tag load the image url by default.

Is it because we need to add the `X-Access-Token` header?

packages/dynamic-flows/src/layout/image/index.js
```diff
@@ -25,12 +26,50 @@ const getImageClasses = (image) => {
...
+      .then((response) => response.blob())
+      .then(
+        (imageBlob) =>
+          new Promise((resolve) => {
+            const reader = new FileReader();
+            reader.onloadend = function () {
+              resolve(reader.result);
+            };
+            reader.readAsDataURL(imageBlob);
+          }),
```
JulianG-TW, line 54: Should we not reject the promise on error? e.g.
```js
reader.onerror = () => reject(reader.error)
```
https://developer.mozilla.org/en-US/docs/Web/API/FileReader/onerror

JulianG-TW: Suggestion: Perhaps we could even extract this into a `readImageBlobAsDataURL` module scoped function that returns a promise.

Then the promise chain would read something like this:

```js
return fetch(url, { ... })
  .then((response) => response.blob())
  .then(readImageBlobAsDataURL);
```

packages/dynamic-flows/src/layout/image/index.js
```diff
@@ -25,12 +26,50 @@ const getImageClasses = (image) => {
...
 
+const getImageSource = async (image) => {
+  try {
+    const url = new URL(image.url, window?.location?.origin);
+
+    if (url.origin !== window.location.origin) {
+      return image.url;
+    }
+
+    return await fetch(url, {
```
JulianG-TW, line 37: I don't think we need to `await` here.

packages/dynamic-flows/src/layout/image/spec.js
```diff
@@ -43,4 +61,22 @@ describe('Given a component for dynamically rendering image', () => {
...
+      };
+      component = mount(<DynamicImage component={spec} />);
+    });
+
+    it('should render an image component using by fetching its content manually', async () => {
+      component.update();
+      expect(fetch).toHaveBeenCalledTimes(1);
+      expect(component.find(Image).prop('src')).toBe(IMAGE_BLOB_URL);
+    });
+  });
```
JulianG-TW, line 81: Should we add tests for when the image is in a different domain, and one for when the fetch call fails?

packages/dynamic-flows/src/layout/image/index.js
```diff
@@ -25,12 +26,51 @@ const getImageClasses = (image) => {
...
+};
+
 const DynamicImage = (props) => {
   const image = props.component;
 
+  const [imageSource, setImageSource] = useState();
+
+  useEffect(() => {
+    getImageSource(image).then(setImageSource);
+  }, [image.url]);
```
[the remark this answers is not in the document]
richardbata: Makes sense, but if the origin doesn't match we don't fetch anyway (if filters it out). Only difference is that as is, src is initalized as undefined then later set to image.url (without fetch), rather than your approach, where we already initialize with the image.url, but have this added tertiary statement...
You prefer having the extra tertiary still?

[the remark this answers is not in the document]
richardbata: I'll merge, because I anyways have another PR about this

PR #1273, opened 2022-01-10
fix: OneOfSchema now persists the model of each of the child schemas.

JulianG-TW:
## 🖼 Context

When the form contains a `OneOfSchema` of `ObjectSchemas`, we want them we want the user to be able to switch between schemas (tabs, radio buttons, select) and NOT “lose” the model they entered in each of the options.

But this is not what happened until recently. Before this PR, the behaviour was like this:

v2-non-persisted-inconsistent

The name was kept between schemas, but the properties that were not present in both schemas would be removed from the model. What’s worse, the UI text fields would still show the old values, but when submitting the values wouldn’t be sent. **This sometimes resulted in the form being invalid and would make it impossible to submit. Leaving users clueless.**

Recently we attempted to fix this similar problem by adding a `useEffect` that would update the local models state when the `props.model` changed:

```js
// When the model or the schema.oneOf from props changes
useEffect(() => {
  // re-initialise models
  setModels(getModelPartsForSchemas(props.model, props.schema.oneOf));
}, [props.model, props.schema.oneOf]);
```

v2-consistent-but-no-persisted

This seemed to work just fine, but it was a shame that the values entered in one of the tabs would be lost when switching tabs. **And it’s also inconsistent with mobile clients, which keep the values of each “tab”.**

This PR brings the web implementation in line with the mobile clients, by keeping separate components for the different `ObjectSchema`s in a `oneOf`.

## 🚀 Changes

This allows users to switch between tabs (or radio btn or select) without losing what they typed.

We achieve this in GenericSchema, by adding a key when rendering ObjectSchema. The key is based on... you guessed it: the schema.

In order for this to work we had to remove a recently added useEffect that would re-initialise the model when props.model changed. See details in PR description.

v2-consisten-persisted

Finally, we added some tests and updated the existing ones because the expectations are now different.

## 🤔 Considerations

By removing the `useEffect` in `OneOfSchema` we are ignoring prop changes to the model. But it turns out that that effect alone was not addressing that concern. There are other components in DF that also ignore model props updates.

**Ignoring prop updates is generally not a good idea.** For instance we could request a step refresh, get a step with an identical schema but different model and when we re-render `JsonSchemaForm` with the same key (because the key is function of the schema) the new model would be ignored.

If/when we address this issue, we can still make `OneOfSchema` react to `prop.model` changes in a way that persists its internal models:

```js
// When the model or the schema.oneOf from props changes
// we only update the model undet the current schemaIndex
useEffect(() => {
  const modelsFromProps = getModelPartsForSchemas(props.model, props.schema.oneOf);

  setModels((previousModels) => {
    const newModels = [...previousModels];
    newModels[schemaIndex] = modelsFromProps[schemaIndex];
    return newModels;
  });
}, [props.model, props.schema.oneOf]);
```

## ✅ Checklist


StevePole (approved):
Nice 👏

PR #1274, opened 2022-01-10
feat(dynamic-flows): Multiple file upload component support for DFv3

dhananjaykha:
## 🖼 Context

Some of the flows like CDD verification require a multiple file upload component. 
This PR adds the capability to parse the schema and render the component. 

PR introduced a new schema type `ArraySchema` to the DF. 
- Distinguishing schema as `Array Schema` from other types
- Classifying the array schema into ArrayListSchema and TupleSchema based on array `items`
- The ArrayListSchema implements the File type repeating items as `MultipleFileUpload`.

Multiple File Upload Schema implementation
- The Schema directly uses the `neptune-web`'s `UploadInput` component since `Repeating ArraySchema item` is the only way the MultiUpload is used. There is no reusability to the indiviudual component.
- Uses the recent `fetcher` prop implementation for `persistAsync` behaviour
- The component supports both `{type: blob}` and `{type:string, format: base64url}`, `blob` uses the `multipart/formdata` approach. While the `base64url` uses the old json value string approach.

References used:
[Array Schema documentation](https://stunning-bassoon-72932d23.pages.github.io/specification/schemas/array/ )
[Multipart/formdata over base64Urls proposal](https://transferwise.atlassian.net/wiki/spaces/DF/pages/2218559991/Make+persistAsync+required+for+uploading+of+large+assets+and+disallow+synchronous+upload+Operation+codename+Kill+base64)

## 🚀 Changes
https://transferwise.atlassian.net/browse/MC-2121 
 
 

## ✅ Checklist


packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/ArrayTypeSchema.js
```diff
@@ -0,0 +1,59 @@
...
+
+import {
+  ArraySchemaVariants,
+  getArrayTypeSchemaVariant,
+} from '../../common/schemaTypes/arraySchemaTypes';
+
+import MultipleFileUploadSchema from './multipleFileUploadSchema';
+
+//Responsibility: decision-maker what array type to render, repeating or static
+const ArrayTypeSchema = (props) => {
```
StevePole, line 11: Could we just call this file/class "arraySchema", it would be more inline with what we do for Objects.  For the BasicTypeSchema, the "type" was required for clarity.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/ArrayTypeSchema.js
```diff
@@ -0,0 +1,59 @@
...
+  const { schema } = props;
+  const { STATIC, MULTIPLE_FILE_UPLOAD, REPEATABLE } = ArraySchemaVariants;
+
+  const arraySchemaType = getArrayTypeSchemaVariant(schema);
+
+  switch (arraySchemaType) {
+    case MULTIPLE_FILE_UPLOAD:
+      return <MultipleFileUploadSchema {...props} />;
+
+    case STATIC:
```
StevePole, line 21: What would static mean?

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/ArrayTypeSchema.js
```diff
@@ -0,0 +1,59 @@
...
+    case REPEATABLE:
```
StevePole, line 22: All arrays are "repeatable", let's maybe reconsider these enums.  The array types I think we care about are:
- FILES
- OBJECTS
- ENUMS
- BASIC_TYPES (I would distinguish this from enums, because enums inherently have a limited set of values, so we'd likely use a different control approach).

Conceptually, we can also have ONE_OF and ALL_OF.  At some layer of implementation we probably need to rely on the `<GenericSchema ...>` which will enable us to handle any type of array, but with potentially poor UX.  I think we will be able to use it for objects, all_of, and oneOf.

StevePole: It looks like we actually need to handle two different things:
- what type of 'items' we have i.e. is it one schema which all entries in the array need to satisfy, or is it an array of items, which need to be matched to the entries in the array.  Perhaps we refer to this as a 'single-schema array' or a 'multi-schema array'?
- what _schema type_ that item, or items, have.

My above message is focused on schema type.  I think we want to use the multi file upload when we have a multi-schema array, where every schema type is `file`?

packages/dynamic-flows/src/jsonSchemaForm/schemas/multipleFileUpload.json
```diff
@@ -0,0 +1,37 @@
+{
+  "type": "object",
+  "properties": {
+    "multiUpload": {
```
StevePole, line 4: Unless I misunderstood, I think this schema is incorrect.  If we're using persist Async, I would assume we're getting an array of IDs back? So more like this
```
multiUpload: {
  type: "array",
  minItems: 1,
  maxItems: 5,
  items: {
    title: "upload all the important pages of your passport",
    type: "string" // int?
    persistAsync: { 
      schema: {
        type: "string"
        format: "base64url"
      }
    }
  }
}
```
Or if we're dealing with a multiple schema array, it would be:
```
multiUpload: {
  type: "array",
  minItems: 1,
  maxItems: 5,
  items: [{
    title: "First page of passport",
    type: "string" // int?
    persistAsync: { 
      schema: {
        type: "string"
        format: "base64url"
      }
    }
  },{
    title: "Second page of passport",
    type: "string" // int?
    persistAsync: { 
      schema: {
        type: "string"
        format: "base64url"
      }
    }
  }]
}
```
That does lead to duplication of the persistAsync data (if it was meant to be the same), but it feels like the correct schema to describe what's going on...

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/ArrayTypeSchema.js
dhananjaykha: Static, as in, one where the number of array items is fixed. (e.g front and back pages of passport)
I picked this term it from [one of the wiki documents](https://transferwise.atlassian.net/wiki/spaces/DF/pages/1984534526/Array+Schema+Documentation). Might not be a correct place to look, maybe.

In short, with the enums, the idea was to distinguish the array schemas based on data structure of `items` - array or object.

* added more details in my replies to below comments *

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/ArrayTypeSchema.js
dhananjaykha: Static, as in, one where the number of array items is fixed. (e.g front and back pages of passport)
I picked this term it from [one of the wiki documents](https://transferwise.atlassian.net/wiki/spaces/DF/pages/1984534526/Array+Schema+Documentation). Might not be a correct place to look, maybe.

In short, with the enums, the idea was to distinguish the array schemas based on data structure of `items` - array or object.

_ _added more details in my replies to below comments_ _

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/ArrayTypeSchema.js
dhananjaykha: About -
> It looks like we actually need to handle two different things:
> 
> * what type of 'items' we have ...
> * what schema type ...


My idea, behind those enums, was to separate the handling of different types kinds of array schemas first  i.e. repeatable/fixed items.   And then, those components (repeatable/fixedItems) will handle their specific internals.

1. Repeating item (if `items` is object)
    1. figuring out item type and rendering logic
    2. add/remove item logic: state maintenance
    3. aggregating the values 
2. fixed items (if `items` is array)
    1. rendering mapping on the array, figuring out each item’s type
    2. aggregating the values

This very much aligns with your above comment. I just named the enums differently.

dhananjaykha: About the first comment -

>All arrays are "repeatable", let's maybe reconsider these enums. The array types I think we care about are:
>
> * FILE 
> * <OTHER TYPES ...>

I completely agree to the comment. I initially started with something just like that, since Multiple upload is a case of repeatable `file`.

  However, I found repeatable schema handling to be a little different (in implementation) from the MultipleFileUpload, particularly. Based on these points -
- Logic of adding/removing items, state management etc.
- Aggregating the values of the dynamic addition/deletion of files

These things are out of the box available (_and has an opinionated implementation_) in the Neptune component.

It very well belongs inside a wrapper of Repeatable schema, I agree. It would just be a very special case from others, and hence pulled out until we start on other repeatable types.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/ArrayTypeSchema.js
dhananjaykha: UPDATE: Reconsidered the schemas as discussed below.

packages/dynamic-flows/src/jsonSchemaForm/schemas/multipleFileUpload.json
dhananjaykha: UPDATE: Incorporated the schema structure as discussed above.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/ArrayTypeSchema.js
dhananjaykha: Introduced array schema variants (based in `items` structure), Tuple or List, referring to [the documentation here](https://stunning-bassoon-72932d23.pages.github.io/specification/schemas/array/).

packages/dynamic-flows/src/common/schemaTypes/arraySchemaTypes.js
```diff
@@ -0,0 +1,49 @@
...
+export const ArraySchemaVariants = {
+  TUPLE: 'STATIC',
+  LIST: 'REPEATABLE',
+};
+
+export const ArrayItemVariants = {
+  FILE: 'FILE',
+};
+
+export const isMultipleFileUploadSchema_deprecated = (schema) => {
```
StevePole, line 12: any reason to retain this?

packages/dynamic-flows/src/common/schemaTypes/arraySchemaTypes.js
```diff
@@ -0,0 +1,49 @@
...
+  const { TUPLE, LIST } = ArraySchemaVariants;
+
+  if (Array.isArray(schemaItems)) {
+    return TUPLE;
+  }
+
+  return LIST;
+};
+
+function isFileItemSchema(items) {
```
StevePole, line 37: Doesn't this function need to check if it's dealing with a TUPLE or LIST?

packages/dynamic-flows/src/formControl/FormControl.js
```diff
@@ -339,6 +340,18 @@ export default class FormControl extends PureComponent {
         );
       }
 
+      case FormControlType.MULTIPLE_FILE_UPLOAD: {
```
StevePole, line 343: I'm not entirely sure about putting multi-file upload inside FormControl 🤔  Might be ok, but I think up to this point FormControl only deals with basic types, and it might be better to keep it that way and just use the multi file upload directly from the arraySchema.  I think might just keep things a bit simpler.  You wouldn't need to add all the new props to this component, for example.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/listSchema/ListSchema.js
```diff
@@ -0,0 +1,46 @@
+import Types from 'prop-types';
+
+import { getArrayItemType, ArrayItemVariants } from '../../../common/schemaTypes/arraySchemaTypes';
+
+import MultipleFileUploadSchema from './multipleFileUploadSchema';
+
+const ListSchema = (props) => {
```
StevePole, line 7: I also wonder if you could just do this with the array schema?  I'm fine with a sub component, but there's not really not very much happening in here.  There's some logic about which array types are supported, but that could also belong to the array component.  If you do want to keep it separate, I think we need a different name, ListSchema is not very clear.  Aren't all arrays lists? I think it's only clear what this means in the context of the Items collection.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/listSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js
```diff
@@ -0,0 +1,161 @@
...
+const MultipleFileUploadSchema = (props) => {
+  const [inputChanged, setInputChanged] = useState(false);
+
+  const [files, setFiles] = useState(props.model);
+  const validationErrorsReference = useRef([]);
+  const [validationErrors, setValidationErrors] = useState([]);
+  const performPersistAsync = usePersistAsync(props.schema.items.persistAsync);
+
+  //TODO: figure this out
+  const fileInputName = 'inputNameWhereToGetThis?';
```
StevePole, line 19: Usually would be the prop name of the array, but with indexes
```
myProp: {
  type: "array",
  ...
}
```
So 
```
myProp[0]
myProp[1]
```
Not sure how easy that is to do?

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/listSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js
```diff
@@ -0,0 +1,161 @@
...
+  const showError =
+    Boolean(props.errors) ||
+    ((inputChanged || props.submitted) && Boolean(validationErrorsReference.current.length));
+
+  return (
+    <div className={classNames('form-group', { 'has-error': showError })}>
+      {/* eslint-disable jsx-a11y/label-has-for */}
+      <label className="d-block control-label">
+        {props.schema.items.title}
+        <SchemaFormControl
```
StevePole, line 53: Yeah, I think you could use the mulit upload directly here.  We're in a scope that only supports multi-upload, so introducing a switch that can render other things just adds unnecessary complexity.

packages/dynamic-flows/src/common/schemaTypes/arraySchemaTypes.js
dhananjaykha: No. I created another utility and went on replacing usages one by one as touched the files. Missed removing it, in the end.

packages/dynamic-flows/src/formControl/FormControl.js
dhananjaykha: I agree. I initially started with the same (directly using the UploadInput in the component), but then moved inside behind a consistent interface of SchemaFormControl. However, considering the component to be only used under a specific schema structure only, always, I will get it replaced directly.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/listSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js
dhananjaykha: Agree. I will do that.

packages/dynamic-flows/src/common/schemaTypes/arraySchemaTypes.js
dhananjaykha: Removed now.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/listSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js
dhananjaykha: Ideally, a form would have the names strictly required for the form to submit correct value. But with complex custom components it becomes more programmatic handling of it.

Currently, in the context of UploadInput component, the `fileInputName` is only consumed in order to return the uploaded file. There is no semantic use of it (e.g. on the hidden html file input as `name` attribute).

So generating a random name should also not be a problem. Even keeping a same name would not make any issues as all the components (if more than one UploadComponents) will handle their file handling in their closures.

So maybe going with a random id per component should be fine, I am thinking. Your thoughts?

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/listSchema/ListSchema.js
dhananjaykha: **Separate sub component -**
I was thinking to separate them into sub types based on these responsibilities -

LIST
- figuring out item type and rendering logic
- add/remove item logic: state maintenance
- aggregating the values 

TUPLE
- rendering mapping on the array, figuring out each item’s type
- aggregating the values

---
**About the name -**
I have got these terms, List and Tuple, from [this documentation microsite](https://stunning-bassoon-72932d23.pages.github.io/specification/schemas/array/). So felt to name it in the same context.

Alternatively, we can use the names you mentioned in some of your previous comments - 'single-schema array' & 'multi-schema array'.

---
I have no strong opinions on both these points. Open to your thoughts, if any different.

packages/dynamic-flows/src/common/schemaTypes/arraySchemaTypes.js
dhananjaykha: Added the checks in the function invoking it. Kept this as a reusable module-scoped function.

packages/dynamic-flows/src/jsonSchemaForm/JsonSchemaForm.story.js
```diff
@@ -55,6 +71,7 @@ export const basic = () => {
         disabled={disabled}
         baseUrl=""
         onChange={(newModel) => {
+          console.log(newModel);
```
StevePole, line 74: log

packages/dynamic-flows/src/common/hooks/usePersistAsync.js
```diff
@@ -0,0 +1,104 @@
...
+
+  async function performPersistAsync(persistAsyncModel) {
+    const signal = abortCurrentRequestAndGetNewAbortSignal();
+
+    const persistAsyncSpec = persistAsyncSchema;
+    const requestBody = { [persistAsyncSpec.param]: persistAsyncModel };
+
+    try {
+      const persistAsyncFetch = new QueryablePromise(
+        fetch(getAsyncUrl(persistAsyncSpec.url, baseUrl), {
```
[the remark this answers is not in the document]
dhananjaykha: Yeah, I will do that before making this PR 'Ready for review'.

packages/dynamic-flows/src/jsonSchemaForm/JsonSchemaForm.story.js
dhananjaykha: I'll remove it. Now that model is shown on the screen. Added earlier to debug and felt it is harmless since it is a `.story.js` file.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/listSchema/ListSchema.js
StevePole: Ah, I just commented above.  It's not that 'list' is wrong, it's just not specific enough.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/listSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js
StevePole: Random is better than fixed, as we typically should also use these in a label to provide accessibility support to jump from the label to the control.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/listSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js
```diff
@@ -0,0 +1,162 @@
...
+
+  const showError =
+    Boolean(props.errors) ||
+    ((inputChanged || props.submitted) && Boolean(validationErrors.length));
+
+  const fileSchemaDescriptor = props.schema.items.persistAsync.schema;
+  return (
+    <div className={classNames('form-group', { 'has-error': showError })}>
+      {/* eslint-disable jsx-a11y/label-has-for */}
+      <label className="d-block control-label">
```
StevePole, line 50: Yeah, there should really be a `for` attribute in the label, otherwise it will fail accessibility validation.  If this is not actually the label, because the labels are in the control, then I would just style it as a label.  However, I think these are the instructions

Do you know if anyone has thought about the accessibility of this control?

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/listSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js
dhananjaykha: Yes. I agree. I have a commit already for a random id. Will get that in. Perfect!

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/listSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js
```diff
@@ -0,0 +1,162 @@
...
+        {props.schema.items.title}
+        <UploadInput
+          multiple
+          // className="form-control" //TODO: Neptune PR to fix the component breaking on adding 'form-control' class
```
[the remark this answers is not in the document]
dhananjaykha: There is a rule in neptune-css where, the `.form-control` gets outlines (focus, warning, error), out of the box depending on the parent wrapper classes (has-error, has-warning etc.) which we are applying.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/listSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js
dhananjaykha: The only reason I wrapped it in label was because the neptune component is not very considerate of it.
So I tried [implicitly adding a label](https://www.w3.org/WAI/tutorials/forms/labels/#associating-labels-implicitly).

I felt that component accessibility work to be outside the PR scope, however, I was thinking to take a look on it - raise in neptune channel, at least.

StevePole: Yeah, I would agree it's outside scope.

packages/dynamic-flows/src/jsonSchemaForm/JsonSchemaForm.story.js
StevePole: Ah, yes, sorry.  Fine to leave in a story.

packages/dynamic-flows/src/common/hooks/usePersistAsync.js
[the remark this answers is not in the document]
dhananjaykha: Done!

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/arrayListSchema/multipleFileUploadSchema/MultipleFileUploadSchema.spec.js
```diff
@@ -0,0 +1,164 @@
+import { UploadInput } from '@transferwise/components';
+import { shallow } from 'enzyme';
+
+import { usePersistAsync } from '../../../../common/hooks/usePersistAsync';
+
+import MultipleFileUploadSchema from './MultipleFileUploadSchema';
+
+jest.mock('../../../../common/hooks/usePersistAsync');
```
JulianG-TW, line 8: I'm not sure mocking `usePersistAsync` is a good idea.
For what I see, this custom hook contains most of the logic used by the subject of this test suite (`MultipleFileUploadSchema`), yet we're mocking it.

The `MultipleFileUploadSchema` uses the `useFetcher` hook. This means we can wrap it in a `FetcherProvider` while mock the fetcher function.

I think having an integration test suite in which we mount the component (instead of shallow rendering) and only mock the server responses would give us more confidence that this component as a whole works as expected.

What do you think?

JulianG-TW (commented):
Hi! I left a few comments and questions.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/arrayListSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js
```diff
@@ -0,0 +1,175 @@
...
+        {props.schema.items.title}
+        <UploadInput
+          multiple
+          className="form-control"
+          files={files}
+          fileInputName={fileInputName}
+          fileTypes={fileSchemaDescriptor.accepts}
+          sizeLimit={fileSchemaDescriptor.maxSize}
+          maxFiles={props.schema.maxItems}
+          maxFilesErrorMessage={props.schema?.validationMessages?.maxItems}
```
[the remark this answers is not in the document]
dhananjaykha: You're right. It should be `props.schema.validationMessages?.maxItems`.
Validation messages is optional.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/arrayListSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js
```diff
@@ -0,0 +1,175 @@
...
+    setInputChanged(true);
+
+    const inputValue = isBlobSchema ? userUploadedFile : await toBase64(userUploadedFile);
+    try {
+      const responsePostFileProcessing = await performPersistAsync(inputValue);
+      return constructFileObject(responsePostFileProcessing);
+    } catch (error) {
+      return Promise.reject(constructFileObject(error));
+    }
+  };
```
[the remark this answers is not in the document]
dhananjaykha: I could gauge the reasons it would break.

However, right now, it will still be handled. The file item on UploadInput component will show the Error.message and the page won't break. 

However, I agree. I can move it in since that adds a generic error handling and a safety net.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/arrayListSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js
```diff
@@ -0,0 +1,175 @@
...
+
+  const [files, setFiles] = useState(props.model);
+  const [validationErrors, setValidationErrors] = useState([]);
+  const performPersistAsync = usePersistAsync(props.schema.items.persistAsync);
+  const fileInputName = useUniqueId('multi-upload-input');
+
+  const { onChange, schema } = props;
+
+  const fileSchemaDescriptor = props.schema.items.persistAsync.schema;
+  const isBlobSchema = fileSchemaDescriptor.type === 'blob';
```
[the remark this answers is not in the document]
dhananjaykha: Sure. I must have got rid of the other usages while refactoring. I'll do that.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/arrayListSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js
```diff
@@ -0,0 +1,175 @@
...
+
+  useEffect(() => {
+    const validFiles = getSuccessfullyProcessedFiles(files);
+    const arrayValidationFailures = getArrayValidationFailures(validFiles, schema);
+    setValidationErrors(arrayValidationFailures);
+
+    const data = validFiles.map((file) => file.id);
+    onChange(data);
+  }, [files, schema]);
```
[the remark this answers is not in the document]
dhananjaykha: I don't recall the reason I had done this way. 
However, the component is simplified a lot now. The one you've shown is 'the' way this should be done. I agree.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/arrayListSchema/multipleFileUploadSchema/MultipleFileUploadSchema.spec.js
```diff
@@ -0,0 +1,164 @@
...
+    validationMessages: {
+      minItems: 'minimum items not met error (from schema)',
+      maxItems: 'maximum items uploaded already (from schema)',
+    },
+  };
+  const onChangeSpy = jest.fn();
+  const randomNumber = 0.123456789;
+  function getRandomGeneratedInputName() {
+    return `multi-upload-input-${Math.floor(randomNumber * 10000000)}`;
+  }
```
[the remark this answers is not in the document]
dhananjaykha: Yes. Started with the random number approach first. But it did not work.
I'll get this done.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/arrayListSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js
```diff
@@ -0,0 +1,175 @@
...
+
+import { usePersistAsync } from '../../../../common/hooks/usePersistAsync';
+import { getArrayValidationFailures } from '../../../../common/validation/validation-failures';
+import ControlFeedback from '../../../controlFeedback';
+
+const useUniqueId = (prefix) => {
+  const prefixString = prefix ? `${prefix}-` : '';
+  const [id] = useState(() => `${prefixString}${generateRandomValue()}`);
+  return id;
+};
```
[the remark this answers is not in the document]
dhananjaykha: Only reason to do it this way was to preserve the id across re-renders.

Only intention was to - generate a random value, persist it across re-renders, and keep this logic even if there are multiple instances of the same component in a page. 

- - - 
The best way would be to use the schema key and then use indexes. 
e.g. `allThePassportPages-1`, `allThePassportPages-2`
```
allThePassportPages: {
    type: 'array',                   // multiple file upload schema
}
```
But that's not something we have done so far. 

---
Suggestions, if you have any, please let me know. We may need this when working more on ArraySchemas.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/arrayListSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js
```diff
@@ -0,0 +1,175 @@
...
+
+  return {
+    id: !isError ? response.data : generateRandomValue(),
+    ...(isError ? { message: response.message } : response),
+  };
+}
+
+function getSuccessfullyProcessedFiles(allFiles) {
+  return allFiles.filter((file) => !file.error && file.status === 'succeeded');
+}
```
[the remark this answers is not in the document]
dhananjaykha: This logic is very much specific to the Neptune's component, how it structures errors/statuses etc. Hence kept it out of the component but in the same file. 

Let me know if you have any opinions.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/arrayListSchema/multipleFileUploadSchema/MultipleFileUploadSchema.spec.js
dhananjaykha: Fair point. 

The hook `usePersistAsync` has its own tests with all possible scenarios. And hence felt it not bad to just mock it out.
In the integration way, like mentioned, it gives end-to-end view and thus confidence.

I have pushed the change.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/arrayListSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js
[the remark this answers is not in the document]
dhananjaykha: Refactored this.

Initially component went into infinite re-renders scenario. On debugging found `validationErrors` state to be the culprit. It is always a derived state from `files`. 

Got rid of that state variable too.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/arrayListSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js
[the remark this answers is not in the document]
dhananjaykha: There is a comment from Steve. That also talks about the same thing.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/arrayListSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js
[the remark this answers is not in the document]
JulianG-TW: awesome!

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/arrayListSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js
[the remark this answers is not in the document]
JulianG-TW: I'm happy either way :)

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/arrayListSchema/multipleFileUploadSchema/MultipleFileUploadSchema.spec.js
```diff
@@ -0,0 +1,171 @@
...
+    },
+    validationMessages: {
+      minItems: 'minimum items not met error (from schema)',
+      maxItems: 'maximum items uploaded already (from schema)',
+    },
+  };
+
+  const onChangeSpy = jest.fn();
+  const randomNumber = 0.123456789;
+  const randomInputName = `multi-upload-input-${Math.floor(randomNumber * 10000000)}`;
```
JulianG-TW, line 46: I'm quite sure we can eliminate all references to "random" in the test suite, if we do something like this:
```js
    const uploadedFile = new File([], 'test-file.pdf');
    const fileUploadData = new FormData();
    const uploadInputComponent = component.find(UploadInput);
    const { fileInputName } = uploadInputComponent.props(); // // this is how we get the filename
    fileUploadData.append(fileInputName, uploadedFile);
```
This way, if the implementation changes and the filenames are not generated using `Math.random()` our tests will continue to pass.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/arrayListSchema/multipleFileUploadSchema/MultipleFileUploadSchema.spec.js
```diff
@@ -0,0 +1,171 @@
...
+    );
+
+    const uploadInputComponent = component.find(UploadInput);
+    expect(uploadInputComponent).toBeTruthy();
+    expect(uploadInputComponent.props()).toStrictEqual({
+      files: [],
+      maxFiles: 2,
+      maxFilesErrorMessage: multipleFileUploadSchemaObject.validationMessages.maxItems,
+      className: 'form-control',
+      fileInputName: randomInputName,
```
JulianG-TW, line 86: If we eliminate random from the test, we'd need to change this to `fileInputName: expect.any(String),`

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/arrayListSchema/multipleFileUploadSchema/MultipleFileUploadSchema.spec.js
```diff
@@ -0,0 +1,171 @@
...
+
+    expect(response).toStrictEqual(expect.any(Promise));
+
+    //eslint-disable-next-line jest/no-test-return-statement
+    return response.catch((error) => {
+      expect(error).toStrictEqual({
+        message: errorMessage,
+        id: expect.any(Number),
+      });
+    });
```
JulianG-TW, line 167: I believe all these lines can be replaced by:
```js
    await expect(response).rejects.toStrictEqual({
      message: errorMessage,
      id: expect.any(Number),
    });
```

JulianG-TW (approved):
Nice work! Thank you!

packages/components/CHANGELOG.md
```diff
@@ -122,7 +122,6 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 
 
 
-
```
dozortsev, line 125: we shouldn't touch CHANGELOG.md files, they maintained by Lerna

dhananjaykha: Yes, you're right. 

This is not something I've done. (I checked `git blame`) it is a autogenerated change.
Since this PR had gotten out of sync with the `main` branch, I `git pull rebase`d and those changes did come in.
I probably should have done a `merge`.

All in all, this is in sync with `main` anyways. And not a human change. Hope that's okay. Will take care of it the next time.

dozortsev (approved):

PR #1279, opened 2022-01-11
fix: Avoid HttpClient singleton

JulianG-TW:
## 🖼 Context

Currently, a single instance of `HttpClient` was exported form the module.

This could cause a conflict if two different `DynamicFlow` instances were to be simultaneously rendered with different `baseUrl` props and no `httpClient` prop.

This is possible because `DynamicFlow` imports the instance and re-initialises it in the absence of an `httpClient` from the props.

```js
const httpClient = propsHttpClient || defaultHttpClient.init({ baseUrl });
```
([source in context](https://github.com/transferwise/neptune-web/blob/32fa0b766d6e851f97a3f2a32521547a27c28a29/packages/dynamic-flows/src/flow/DynamicFlow.js#L82))

## 🚀 Changes

With these changes, `HttpClient` class is exported form the module and each instance of `DynamicFlow` creates their own instance without conflicting with other instances.

## ✅ Checklist


packages/dynamic-flows/src/flow/client/HttpClient.spec.js
```diff
@@ -0,0 +1,20 @@
+import { HttpClient } from './HttpClient';
+
+describe('HttpClient tests', () => {
+  describe('when instantiating two HttpClients with different baseUrls', () => {
+    it('each instances makes requests using their own baseUrl', () => {
+      const clientA = new HttpClient({ baseUrl: 'https://foo' });
+      const clientB = new HttpClient({ baseUrl: 'https://test' });
+
+      clientA.request({ action: { url: '/some-path-in-foo' } });
+      expect(lastCallToFetch()[0]).toBe('https://foo/some-path-in-foo');
```
fergusjordantw, line 10: Reading off the array should be done in `lastCallToFetch` imo

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -79,7 +79,7 @@ const DynamicFlow = (props) => {
   const [stepAndModels, setStepAndModels] = useState({ stepSpecification: {}, models: {} });
   const { stepSpecification, models } = stepAndModels;
 
-  const httpClient = propsHttpClient || defaultHttpClient.init({ baseUrl });
+  const httpClient = propsHttpClient || new HttpClient({ baseUrl });
```
fergusjordantw, line 82: This will create a new HttpClient every render if one isn't passed in. I'd extract this to a hook to memoize the returned httpClient which can get re-evaluated on baseUrl or props instance changing

fergusjordantw (commented):
I think it's fine, but what's the main actual use-case for passing in a custom HttpClient?

packages/dynamic-flows/src/flow/client/HttpClient.spec.js
JulianG-TW: In this case `[0]` is the first argument in the last call.

fergusjordantw: Hahaha gotcha

JulianG-TW:
> I think it's fine, but what's the main actual use-case for passing in a custom HttpClient?

**It was already possible to pass a custom http client.**

We have at least one team with a use case where they need to pass some custom headers, and will use the `httpClient` prop in `DynamicFlow`.

I also think my ninjas pages could use a custom httpClient, since they use `backoffice-auth`.

packages/dynamic-flows/src/flow/DynamicFlow.js
JulianG-TW: done @ 8ee8777eabb52d8e9dd7c24d8d3a5fa5da165e52

Thanks for reviewing!

packages/dynamic-flows/src/flow/client/HttpClient.spec.js
JulianG-TW: Each call to fetch in this test is made with two arguments, `resource` and `init`. In this test we only care about the first argument. It doesn't read very nice, I know.

JulianG-TW: Changing it to
```js
expect(firstAgumentIn(lastCallTo(fetch))).toBe('https://foo/some-path-in-foo');
```
in bd8b4c6145b244059a11fed5f6fdf33a02c2551c

fergusjordantw:
> I also think my ninjas pages could use a custom httpClient, since they use `backoffice-auth`.

Got it, backoffice-auth is a perfect example where this is super useful

packages/dynamic-flows/src/flow/client/HttpClient.spec.js
```diff
@@ -0,0 +1,24 @@
...
+      clientB.request({ action: { url: '/some-path-in-test' } });
+      expect(firstAgumentIn(lastCallTo(fetch))).toBe('https://test/some-path-in-test');
+    });
+  });
+
+  function lastCallTo(mockFn) {
+    return mockFn.mock.calls[mockFn.mock.calls.length - 1];
+  }
+
+  function firstAgumentIn(callArgs) {
```
fergusjordantw, line 21: Would maybe make it `getRequestUrl` so it's a bit clearer, the confusion I had initially was that I thought it was reading from a nested array that needed to be flattened.

JulianG-TW: I'm not it would read better:

```js
expect(firstAgumentIn(lastCallTo(fetch))).toBe('https://foo/some-path-in-foo');
// reads "expect first argument in last call to fetch to be 'https://...'"
```
vs.
```js
expect(getRequestUrl(lastCallTo(fetch))).toBe('https://foo/some-path-in-foo');
// reads "expect get request URL, last call to fetch to be 'https://...'"
```

fergusjordantw: Personally I do think it reads better, the fact that it's the first argument in an array of arguments is an implementation detail of how jest displays mocked arguments from a call, the thing we care about is the fact that it's the requestUrl for the last call to fetch

JulianG-TW: I see. I agree.
I've changed the line to be
```js
expect(requestUrlInLastCallToFetch()).toBe('https://foo/some-path-in-foo');
```
in 15f1bf1b68f1d49ac026cdb99f4acc6644798dad

richardbata (approved):

PR #1280, opened 2022-01-11
update Sentiment values accepted by InlineAlert

obasemolaTW:
## 🖼 Context

Since the enum values for `Sentiment` have been changed, this PR makes corresponding changes in `InlineAlert` as already been done in `Alert`.

## 🚀 Changes

- Update `Sentiment` enums from the deprecated values to the new ones and updated relevant tests.

## ✅ Checklist


packages/components/src/inlineAlert/InlineAlert.js
```diff
@@ -27,12 +27,12 @@ const InlineAlert = (props) => {
 InlineAlert.propTypes = {
   children: PropTypes.node.isRequired,
   className: PropTypes.string,
-  type: PropTypes.oneOf(['success', 'error', 'warning', 'info']),
+  type: PropTypes.oneOf(['positive', 'negative', 'warning', 'neutral']),
```
dozortsev, line 30: this is breaking change, please for just add new types into the array

packages/components/src/inlineAlert/InlineAlert.js
```diff
@@ -5,7 +5,7 @@ import withArrow, { AlertArrowPosition } from '../alert/withArrow';
 import { Sentiment } from '../common';
 
 const typeClassMap = {
-  [Sentiment.ERROR]: 'danger',
+  [Sentiment.NEGATIVE]: 'negative',
```
dozortsev, line 8: here you would need keep both old (danger) and new one (negative)

packages/components/src/inlineAlert/InlineAlert.js
```diff
@@ -6,6 +6,7 @@ import { Sentiment } from '../common';
 
 const typeClassMap = {
   [Sentiment.ERROR]: 'danger',
+  [Sentiment.NEGATIVE]: 'negative',
```
dozortsev, line 9: I think this map should same as in [Alert component](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/alert/Alert.js#L14-L18)

obasemolaTW: I see the object highlighted is for deprecated types. Do you mean I should include all the types, both deprecated and otherwise or just the deprecated types?

dozortsev: I double checked this and we don't need here any other mapping except for `Sentiment.ERROR`

dozortsev:
```suggestion
```

dozortsev: please make this change and merge PR

dozortsev (approved):

PR #1281, opened 2022-01-11
Use BaseUrlProvider and Image component fixes

richardbata:
## 🖼 Context
- Lets use Context to make baseUrl available to all children on DynamicFlow
- Image.js needs to add baseUrl to images with relative path

## 🚀 Changes
- Use BaseUrlProvder around DynamicFlow and remove old props
- Add baseUrl value to image URLs that are relative

## ✅ Checklist


JulianG-TW (commented):
I left one comment

richardbata:
Yes! I tried to solve that issue in BaseUrlProvider
```  const context = useContext(BaseUrlContext);
  if (context.baseUrl !== null && context.baseUrl !== undefined) {
    return children;
  }```

File: packages/dynamic-flows/src/common/contexts/baseUrlContext/BaseUrlContext.js [hunk 1]
```diff
@@ -1,9 +1,27 @@
...
-export const BaseUrlContext = createContext({
+const BaseUrlContext = createContext({
   baseUrl: null,
 });
 
+export const BaseUrlContextProvider = ({ children, value }) => {
+  const context = useContext(BaseUrlContext);
+  if (context.baseUrl !== null && context.baseUrl !== undefined) {
+    return children;
+  }
```
JulianG-TW, line 12: Ah! I now understand the reason for this.

**This is not a required change. Just a suggestion.**

I think I would prefer this to be done within `JsonSchemaForm`.

...something like:

```js
const JsonSchemaForm = (props) => {
  const { baseUrl: propsBaseUrl } = props;
  const contextualBaseUrl = useBaseUrl()
  return contextualBaseUrl ? (
    <GenericSchema {...props} />
    ) : (
    <BaseUrlContext.Provider value={{ baseUrl }}>
      <GenericSchema {...props} />
    </BaseUrlContext.Provider>
  );
};
```

The reason I would prefer this is because having this "check" inside `BaseUrlContextProvider` is analogous to a singleton pattern. There's no way of overriding the value baseUrl down the tree even intentionally. Whereas, by explicitly checking in `JsonSchemaForm` we can decide when to do it.

**This is not a required change. Just a suggestion.**

JulianG-TW (approved):

PR #1282, opened 2022-01-13
fix: Avoid rendering broken img tag in DynamicImage

JulianG-TW:
## 🖼 Context

Since https://github.com/transferwise/neptune-web/pull/1270 we support authenticated images which are fetched in JS, then loaded as a blob and applied to an `Image` component as a data Url.

But while the image url is being fetched, the `Image` component has a "falsy" (empty string) `src` which results in a broken image.

image

## 🚀 Changes

We render `null` instead of `Image` until fetching resolves.

**Before:**

img-null-before

**After**

img-null-after

## 🤔 Considerations

This PR doesn't address the layout-shift occurring when the image loads.

## ✅ Checklist


packages/dynamic-flows/src/layout/image/index.js
```diff
@@ -81,15 +81,16 @@ const DynamicImage = (props) => {
...
 
   return (
     <div className="row m-b-0">
-      <div className={getWrapperClasses(image)}>
-        <Image className={getImageClasses(image)} {...imageProps} />
-      </div>
+      <div className={getWrapperClasses(image)}>{getImage()}</div>
     </div>
   );
 };
```
JulianG-TW, line 96: here

richardbata (approved):

packages/dynamic-flows/src/layout/image/index.js
richardbata: Do we want to render a placeholder while images load to reduce jittering also, or that not too bad?

JulianG-TW: Yes, we discussed this. But I'll be in a future PR.

PR #1284, opened 2022-01-14
fix: Add baseUrl context provider to DynamicLayout

JulianG-TW:
## 🖼 Context

Some consumers are using the `DynamicLayout` component directly (without `DynamicFlow`).

Since version 16.1.13 the `baseUrl` prop was removed form many component in favour of a React Context.

Notably, `JsonSchemaForm` had a special treatment to allow it to benefit from the context if present, but also use the `baseUrl` prop when rendered in isolation.

`DynamicLayout` didn't get this special treatment, and the `baseUrl` prop was removed. In hindsight, that should have been a breaking change.

## 🚀 Changes

This PR restores `DynamicLayout` ability to accept a `baseUrl` prop  and, just like `JsonSchemaForm`, conditionally render a `BaseUrlProvider` depending on the case.

The context provider and related hooks were refactored to avoid a warning when using the context directly:
```
The object passed as the value prop to the Context provider (at line 53) changes every render.
To fix this consider wrapping it in a useMemo hook.eslint[react/jsx-no-constructed-context-values]
(https://github.com/yannickcr/eslint-plugin-react/tree/master/docs/rules/jsx-no-constructed-context-values.md)
```

## ✅ Checklist


packages/dynamic-flows/src/jsonSchemaForm/JsonSchemaForm.js
```diff
@@ -1,21 +1,20 @@
...
     return <GenericSchema {...props} />;
   }
-
-  const { baseUrl } = props;
   return (
-    <BaseUrlContext.Provider value={{ baseUrl }}>
-      <GenericSchema {...props} />
-    </BaseUrlContext.Provider>
+    <BaseUrlProvider baseUrl={props.baseUrl}>
+      <GenericSchema {...props} />{' '}
```
richardbata, line 16: Do we need this empty string?

JulianG-TW: oh!

richardbata (approved):

packages/dynamic-flows/src/common/contexts/baseUrlContext/BaseUrlContext.js
```diff
@@ -1,16 +1,30 @@
...
+import Types from 'prop-types';
+import { createContext, useContext, useMemo } from 'react';
 
-export const BaseUrlContext = createContext({
-  baseUrl: null,
-});
+const BaseUrlContext = createContext();
+
+export const BaseUrlProvider = ({ baseUrl, children }) => {
+  const value = useMemo(() => ({ baseUrl }), [baseUrl]);
```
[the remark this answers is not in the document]
JulianG-TW: it's because there's a linter warning otherwise, when we pass an object literal to the `value` prop.

```
The object passed as the value prop to the Context provider (at line 53) changes every render.
To fix this consider wrapping it in a useMemo hook.eslint[react/jsx-no-constructed-context-values]
(https://github.com/yannickcr/eslint-plugin-react/tree/master/docs/rules/jsx-no-constructed-context-values.md)
```

PR #1289, opened 2022-01-17
feat(UploadInput): optionally restrict the maximum number of files to upload

dhananjaykha:
## 🖼 Context

[MC-2122](https://transferwise.atlassian.net/jira/software/projects/MC/boards/1146?assignee=615b036b72f6970069490f66&selectedIssue=MC-2122
)

The UploadInput component needed to have a feature to support `maxItems` validation from DFv3. The component, as of this PR, would accept an optional prop that restricts the number of files to upload and shows passed error message.

Also, when used in a form (`.form-control` class applied), does not break the height of the component.

BEFORE (with the .form-control class)
image

AFTER (with the .form-control class)
image

---
The component behaviour:
LimitFileUploads

## 🚀 Changes

 
On upload, the component already applies some validations and then hands it over to, `onFileUpload` callback. Added one more validation in between.
The files used to be updated in state, though. And that resulted in making a stale upload count for the validation calculations. Hence, along with the state, introduced a `ref` that also gets updated with the state and reflects latest state.

Also considering `upload time` in the fileId to make a file uploaded unique (react `key`) if uploaded twice. This was resulting into weird file items on screen on upload.

## ✅ Checklist

- [x] Changes work in all supported browsers (don't forget IE11) Did MS Edge instead of IE11


StevePole:
I'm wondering if we should hide the upload button once you exceed the maximum?  You'd presumably still be able to exceed the max by using a mutli-select in the native dialog, but not sure we need to allow it through our UX?

dhananjaykha:
> 

I had ran through it with the designer from our team. 
The feedback on "hiding the button" was that the button not being visible may not be a good UX, he suggested.
Also the max number is also mentioned in the description on the button, so keeping it there was something we agreed.

packages/components/src/uploadInput/uploadButton/UploadButton.tsx
```diff
@@ -37,6 +37,11 @@ export type UploadButtonProps = {
    */
   description?: string | undefined;
 
+  /**
+   * Description for the upload button
```
matebek-tw, line 41: This docstring should be updated to describe the actual prop.

dhananjaykha: My bad. Will do.

matebek-tw (approved):

PR #1292, opened 2022-01-19
Use HttpClient for all network requests

JulianG-TW:
## 🖼 Context

The `DynamicFlow` component accepts an optional `httpClient` prop with a default/fallback implementation that uses fetch.

The default implementation is pretty much the same as `window.fetch` except that it adds two headers to every request:
```
'Content-Type': 'application/json'
'X-Access-Token': 'Tr4n5f3rw153'
```
and it also rejects when the response status code is 4xx or 5xx.

Consumers of `DynamicFlow` can implement and pass a custom `httpClient` which will be used to **fetch steps, refresh steps and submit forms**.

Internally, the most common use case is back-office applications. 

But this `httpClient`, is not used for everything within `DynamicFlows`:

feature | uses httpClient?
-- | --
fetch/refresh step | ✅
submit form | ✅
validation-async | ❌
dynamic image | ❌
persist-async-basic | ❌
persist-async-blob | ❌ (uses the Upload component from and that uses fetch by default)

## 🚀 Changes

We now use HttpClient for most network requests:

feature | uses httpClient?
-- | --
fetch/refresh step | ✅
submit form | ✅
validation-async | ✅
dynamic image | ✅
persist-async-basic | ✅
persist-async-blob | ❌ (uses the Upload component from and that uses fetch by default)

`HttpClient` was also refactored to accept `url` and `method` rather than an `action` object, because not all requests were  initiated by actions. This resulted in a lot of test changes, unfortunately.

## 🤔 Considerations

The method signature for `HttpClient#request` is growing in complexity:

- we had to add a `signal` for AbortController.
- we had to add `contentType` because now not all requests expect JSON
- we had to add `credentials` because one request uses `same-origin` instead of `include`.

Considering the change in method signature already means this PR has **BREAKING CHANGES**,  perhaps we should make it to match the signature of `fetch`.

Another **BREAKING CHANGE** is that persist-async-blob now requires absolute URLs to be provided in the schema, since we don't always have access to the `baseUrl` that was being used.

## ✅ Checklist


packages/dynamic-flows/src/flow/client/HttpClient.js
```diff
@@ -4,22 +4,21 @@ export class HttpClient {
...
   }
 
-  async request({ action, data }) {
-    const { url, method } = action;
-
+  async request({ method, url, data, signal, contentType = 'application/json' }) {
     const endpoint = startsWithHTTP(url) ? url : `${this.baseUrl}${url}`;
 
-    const body = method === 'GET' ? undefined : JSON.stringify(data);
+    const body = data !== undefined ? JSON.stringify(data) : undefined;
```
JulianG-TW, line 10: 🤔

packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/blob/PersistAsyncBlobSchema.js
```diff
@@ -63,7 +60,7 @@ const PersistAsyncBlobSchema = (props) => {
         usLabel={props.schema.title}
         usPlaceholder={props.schema.description}
         httpOptions={{
-          url: baseUrl + url,
+          url,
```
JulianG-TW, line 63: We're not using baseUrl for persist async blob any more.
This means the schemas will need to provide absolute URLs.

JulianG-TW:
Close in favour of https://github.com/transferwise/neptune-web/pull/1294

PR #1293, opened 2022-01-19
fix: MoneyInput label alignments

matebek-tw:
## 🖼 Context

https://transferwise.atlassian.net/browse/DS-1596

## 🚀 Changes

image

## ✅ Checklist


packages/components/src/moneyInput/MoneyInput.less
```diff
@@ -39,7 +39,7 @@
     min-width: 2.6em; // fits a 3 letter string, ZMW being the longest one
 
     @media (--screen-sm) {
-      min-width: var(--size-32);
+      min-width: calc(2.6em + var(--size-32));
```
helendurranttw, line 42: This seems a little bit arbitrary - do we have any guidance from the designers? To me it looks a bit wide.

helendurranttw:
Is there any chance we could tackle the vertical alignment of the text / flag / chevron as part of this?

packages/components/src/moneyInput/MoneyInput.less
mikknarusk: Ultimately it'll be arbitrary as it will just be a number that looks good visually. Although I agree it seems a bit too wide on the screenshot. Before components 39, this was targeting the span that contained both the flag and text, hence the 32px (flag + margin). Now the flag is a separate element, so this is only targeting the text - probably should just be 2.6em now
Screenshot 2022-01-20 at 09 59 49

dozortsev: @mikknarusk if `2.6em` is fine then we even don't need this style became redundant as above we have style that sets `2.6em`

Screen Shot 2022-01-24 at 3 37 31 PM

mikknarusk: @dozortsev yes, I think that rule was originally just meant for mobile (as we're not showing the flag on mobile), but it can always apply now

dozortsev (approved):

PR #1294, opened 2022-01-20
feat: DynamicFlow - Remove httpClient prop in favour of a fetcher prop

JulianG-TW:
## 🖼 Context

The `DynamicFlow` component accepts an optional `httpClient` prop with a default/fallback implementation that uses fetch.

The default implementation is pretty much the same as `window.fetch` except that it adds two headers to every request:
```
'Content-Type': 'application/json'
'X-Access-Token': 'Tr4n5f3rw153'
```
and it also rejects when the response status code is 4xx or 5xx.

Consumers of `DynamicFlow` can implement and pass a custom `httpClient` which will be used to **fetch steps, refresh steps and submit forms**.

Internally, the most common use case is back-office applications. 

But this `httpClient`, is not used for everything within `DynamicFlows`:

feature | uses httpClient?
-- | --
fetch/refresh step | ✅
submit form | ✅
validation-async | ❌
dynamic image | ❌
persist-async-basic | ❌
persist-async-blob | ❌ (uses the Upload component and that uses fetch by default)

## 🚀 Changes

In DynamicFlow component, we remove the ability to pass a custom `httpClient` as a prop. Instead we allow consumers to pass a new `fetcher` prop, which is a function with the same signature as global fetch.

We replace `BaseUrlContext` with a `FetcherContext` and is used everywhere we were previously using fetch.

When only `baseUrl` is passed, a default `fetcher` is created which prepends this base url to all requests made to non-absolute URL. (see the new `makeFetcher` function)

⚠️ `HttpClient` class was removed and its `request` method was moved into `DynamicFlow` under a new name: `requestStep` since it's used for fetching and refreshing steps and  submitting forms.

feature | uses custom fetcher?
-- | --
fetch/refresh step | ✅
submit form | ✅
validation-async | ✅
dynamic image | ✅
persist-async-basic | ✅
persist-async-blob | ✅ (uses the `Upload` component that **NOW** supports a `fetcher` prop)

## 🤔 Considerations

This is a **BREAKING CHANGE** for any consumers passing `httpClient` to `DynamicFlow`.

### Before 
```jsx
<DynamicFlow
  flowUrl="/my-flow"
  httpClient={myCustomHttpClient}
  ...
/>
```

### After
```jsx
<DynamicFlow
  flowUrl="/my-flow"
  fetcher={myCustomFetchFunction}
  ...
/>
```

## ✅ Checklist


richardbata:
I'm not a huge fan persist-async losing baseUrl. Both blob and regular peristAsync would lose it

JulianG-TW:
@richardbata 
> I'm not a huge fan persist-async losing baseUrl. Both blob and regular peristAsync would lose it

My intention is that, in using fetcher, the regular persist async would benefit form the baseUrl used within it. Isn't that what's happening?

Regarding Blob persist-async, yes. it's true. it would lose the baseUrl. 
But only NOW I realise that, the current code using BaseUrlProvider allows consumers to have both a custom `httpClient` and a  separate `baseUrl` for persist-async, validation-async and dynamic-image.

I can see two solutions for this:

Option A | Option B
-- | --
We could have both a `fetcher` and a `baseUrl` that would be added to each request regardless of whether the request is made with `fetcher` or not (the only case being persist-async-blob because of Upload) | We could discuss with Design System team and see what they think of adding a `fetcher` prop to the `Upload` component

I think I prefer option B.

JulianG-TW:
@richardbata 
> I'm not a huge fan persist-async losing baseUrl. Both blob and regular peristAsync would lose it

With the latest commit, now `Upload` component also supports a `fetcher` prop, so both persist-async schemas should can be used with relative urls. (AFAIK).

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -80,10 +80,8 @@ const DynamicFlow = (props) => {
   const [stepAndModels, setStepAndModels] = useState({ stepSpecification: {}, models: {} });
   const { stepSpecification, models } = stepAndModels;
 
-  const httpClient = useMemo(
-    () => propsHttpClient || new HttpClient({ baseUrl }),
-    [baseUrl, propsHttpClient],
-  );
+  const fetcher = useMemo(() => propsFetcher || makeFetcher(baseUrl), [propsFetcher, baseUrl]);
+  const httpClient = useMemo(() => new HttpClient(fetcher), [fetcher]);
```
richardbata, line 84: Just a thought, but don't we loose the need for an HttpClient. We could achieve all it does in the default fetcher...

packages/dynamic-flows/src/layout/DynamicLayout.js
```diff
@@ -92,10 +92,14 @@ const DynamicLayout = (props) => {
     }
   };
 
-  if (useHasBaseUrlProvider() || baseUrl == null) {
+  if (useHasFetcherProvider() || baseUrl == null) {
     return <>{components.map(renderComponent)}</>;
```
richardbata, line 96: No need for <> here

richardbata:
Changes make sense 👍 
Do you plan to add fetcher to Upload before this PR?

packages/dynamic-flows/src/flow/DynamicFlow.js
JulianG-TW: I suppose we don't need it.
It could be a wrapping function on top of the fetcher, since it still needs to add those headers and stringify the data for the body.
I'd do that in a separate PR, to avoid a huge PR.

JulianG-TW:
> Do you plan to add fetcher to Upload before this PR?

It's already in this PR.

packages/dynamic-flows/src/layout/DynamicLayout.js
JulianG-TW: Thanks! Done!

packages/dynamic-flows/src/flow/DynamicFlow.js
JulianG-TW: I changed my mind. The `HttpClient.spec.js` test suite was failing to run on CI, so I got rid of `HttpClient` entirely in this PR. See b60a2c346595ed59f670e492e129aed16bc55195

richardbata (approved):

StevePole (approved):

StevePole:
Let's be conscious to communicate with existing consumers.  Perhaps do code search to find them.

PR #1299, opened 2022-01-25
fix(dynamic-flows): Fix input focus after a refresh

JulianG-TW:
## Context

When working on adding support for ETAGs, I added some delay to the mock fetcher in storybook and realised the loading animation appears after every keystroke (in a schema marked as refresh-on-change).

This causes the input field to lose focus after each refresh:

Even if we soon add debouncing, I think the user should be able to continue typing on the same text field while we refresh requirements. Losing focus after each refresh creates a bad user experience.

too-much-loading

## Changes

We avoid toggling the `loading` state when refreshing; instead we toggle a new `refreshing` reference.

This avoids rendering the loading animation, which in turn means the Layout component with the DynamicForm instance is the same after a refresh containing the same schema, which means the text input component is the same instance and focus is maintained.

too-much-loading-fixed

We're also now disabling action buttons when the DF component is refreshing:

button-disabled

We achieve this by using a Context Provider and a custom hook, to avoid prop-drilling through a lot of components that are not concerned with this.

## Considerations

If we want to additionally render the action buttons as `disabled` we'll need to pass down a property to `DynamicLayout` that would be propagated down the render tree. Also if we do that, we probably want to change `refreshingReference` to be a React state rather than a reference.

## Checklist


packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -128,15 +129,16 @@ const DynamicFlow = (props) => {
...
       .then(async (response) => {
         const step = await response.json();
         updateStepSpecification(step);
       })
       .catch(handleFetchError)
-      .finally(() => setLoading(false));
+      .finally(() => {
+        refreshingReference.current = false;
+      });
   };
```
JulianG-TW, line 142: Here's where we toggle the `refreshingReference` instead of the `loading` state.

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -72,6 +72,7 @@ const DynamicFlow = (props) => {
     loaderSize,
   } = props;
 
+  const refreshingReference = useRef(false);
```
richardbata, line 75: Any reason to use ref over state

JulianG-TW: It's because this boolean flag is only used to early-return in `onAction` so nothing that is rendered depends on this.
If we used a state, then we'd get a re-render when we set it.

If we wanted to render the action buttons as disabled based on this flag, then yes, I think it should be in state.

richardbata: Nice. Disabled buttons could make sense, no?

JulianG-TW: I can give that a try.

JulianG-TW: Disabling action buttons during refresh:
- using props: https://github.com/transferwise/neptune-web/pull/1302/files
- using context: https://github.com/transferwise/neptune-web/pull/1301/files

JulianG-TW: After reading about [disabled state in Neptune](https://transferwise.github.io/neptune-web/branch/apply-new-disabled-states-approach/styles/addons/DisabledState) I think we should go with the React Context approach.

It reads:
> Note: avoid applying disabled state down the DOM tree after you already applied to some parent element as it would have double visual effect.

The props approach (#1302) involves passing a `disabled` prop to `DynamicLayout` but we don't mean to disable the entire UI. It only means "disable actions". So I think we either rename the prop to `disableActions` or go with the context approach (#1301), in which the `DynamicButton` component decides to disable itself when it knows that the DF container is `refreshing`. I prefer this.

richardbata (approved):

packages/dynamic-flows/src/flow/DynamicFlow.js
richardbata: Nice research! I personally would prefer context as well after what you've mentioned. Setting everything to disabled in the tree just to pass down the value definitely seems wrong. And its such a pain to have to pass down X layers of props :)

JulianG-TW: **UPDATE** We're now disabling buttons using a context provider 96ae26280ef5375c5b04d8b3d6932f2c9f2baede

PR #1301, opened 2022-01-26
Disable action buttons during refresh using React Context

JulianG-TW:
## ✅ Checklist


packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -183,7 +182,7 @@ const DynamicFlow = (props) => {
   };
 
   const onAction = async (action) => {
-    if (refreshingReference.current) {
+    if (refreshing) {
```
richardbata, line 185: This is now a double level of safety, as buttons should be disabled anyways

richardbata (approved):

richardbata:
The only thing that comes to mind, is that visually it could look a bit weird if all buttons on the screen get disabled at the same time, like they looked weird in V2 when all were loading at the same time... I do thing disabled is less weird though than  loading 👍 Also, this is something we can later change if we notice its an issue

PR #1303, opened 2022-01-27
feat(dynamic-flows): Add support for ETags when refreshing steps

JulianG-TW:
## 🖼 Context

Some endpoints support [ETags](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) when handling step requests.
Mobile clients already support this, by avoiding replacing the step when the endpoint responds with status 304 and an empty body.

## 🚀 Changes

Whenever a step is loaded, its `etag` header is also stored (even if missing or `undefined`).
When making a request to refresh steps we now add the `If-None-Match` header with the last `etag` value.
If the service responds with a status 304 we don't replace the current step.

## ✅ Checklist


packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -78,16 +78,21 @@ const DynamicFlow = (props) => {
   const [submitted, setSubmitted] = useState(false);
   const [validations, setValidations] = useState();
 
-  const [stepAndModels, setStepAndModels] = useState({ stepSpecification: {}, models: {} });
-  const { stepSpecification, models } = stepAndModels;
+  const [stepAndModels, setStepAndModels] = useState({
+    stepSpecification: {},
+    models: {},
+    etag: undefined,
```
richardbata, line 84: Maybe use another state for eTag?

JulianG-TW: I'm not sure.

I actually want `etag` and `stepSpecification` in the same state because they should be updated simultaneously.

If we keep them in separate states, when we call their respective set-state functions after a promise resolve in an async callback, [React <18 won't batch the updates](https://github.com/reactwg/react-18/discussions/21). This means there could be renders that contain a stale `etag` or a stale `stepSpecification`.

Arguably, `loading` and `refreshing` should be in the same state as well.

For example in `fetchStep` (line 115) we call `requestStep` and when it resolves we call:
- `updateStepSpecification` (in the `then` clause)
- `setSubmitted`  (in the same `then` clause)
- `setLoading` (in the `finally` clause)
this causes 3 re-renders, and only after the last one we have the state we want

[I read React 18 will batch these in the same way that other set-state calls are batched within sync callbacks](https://github.com/reactwg/react-18/discussions/21).

mikknarusk: From the description it sounds like `etag` could be stored in a `useRef` instead as it's not used in the UI, only in API calls and you always want to have the latest value regardless of the lifecycle

JulianG-TW: Yes. I think a `useRef` would be certainly better than two separate `useState`. However, I think there would be still room for a race condition.
If the `etag` is a `useRef` it will be updated immediately, even before the `stepSpecification`. 
This means that an action triggered in the middle of those operations would be based on the **old** step, but the request would be made using the **new** etag.

Just like with the 2 states approach, I don't think this race condition is likely to happen, I just think that having `etag` in the same state as `stepSpecification` and `model` would eliminate the possibility of a race condition.

We already keep `stepSpecification` and `models` in the same state for a similar reason. 

And `etag` could be considered a property of the `stateSpecification`, so I'm still convinced that they belong in the same state.

mikknarusk: if the `etag` and `stateSpecification` are tightly coupled then yeah its probably fair to include them in the same state

StevePole: I don't _think_ there's a race condition, but there might still be a problem.  Being single threaded JS is only going to switch call stack when current stack is completed or exited.  I'm not an expert on react internals, but I don't believe it has anything equivalent to Angular's digest cycle (which makes it's own decisions about when to update the view based on other events/state), so either the first setState/setRef call will trigger the repaint, _always_ leading to a pass with invalid values.  Or, React defers processing and it will always have access to both latest values.  My guess is setRef will not trigger paint, so if that's first it's probably safe?  I don't have a problem with it being in state though.

JulianG-TW: > if the etag and stateSpecification are tightly coupled then yeah its probably fair to include them in the same state

**Exactly!** In most implementations the etag values is a hash of the content.

JulianG-TW: > I don't think there's a race condition

It's true, we haven't identified any. Buy for the reasons above, I still think the `etag` should be stored along with the `step`.

I don't mind changing it to be a React reference, I'm more interested in getting this merged that getting my way.

But if we all agree it can stay in the same `useState`, then we could leave it as it is and approve it.

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -298,3 +310,7 @@ function makeRequestStep(fetcher) {
     });
   };
 }
+
+function getETag(headers) {
```
StevePole, line 314: Tiny point, but I find this capitalisation odd.  If I saw `getETag`, I would expect the associated variable to be `eTag`, not `etag`.

packages/dynamic-flows/src/flow/DynamicFlow.spec.js
```diff
@@ -157,15 +157,21 @@ describe('Given a component for rendering a dynamic flow', () => {
   const resolve = (response, init = {}) => {
     return new Promise((promiseResolve) => {
       setTimeout(async () => {
-        promiseResolve(new Response(JSON.stringify(response), init));
+        promiseResolve(
+          new Response(JSON.stringify(response), {
+            ...init,
+            headers: { ...init.headers, etag: JSON.stringify(response).slice(0, 100) },
```
StevePole, line 163: Can we extract to a `generateEtag` function?

packages/dynamic-flows/src/flow/DynamicFlow.spec.js
```diff
@@ -157,15 +157,21 @@ describe('Given a component for rendering a dynamic flow', () => {
...
+        );
       }, 0);
     });
   };
 
   const rejectWithJsonResponse = (response, init = {}) => {
     return new Promise((_, promiseReject) => {
       setTimeout(() => {
-        promiseReject(new Response(JSON.stringify(response), init));
+        const body = response != null ? JSON.stringify(response) : null;
```
StevePole, line 173: This whole method feels a bit awkward, do you know why we need the timeout?  Would be nicer to do `Promise.reject(...)` and simplify.

StevePole: Actually, I didn't realise these were tests initially.  A bit less concerned in the test, that probably explains the timeout.

JulianG-TW:
> Please forgive my naiveness on this, but don't browsers already send etags automatically, without us doing anything in our code?

AFAIK they only do it for GET requests, not POST.
In our use case, we're using POST to get/refresh requirements only because the payload contains PII, but we still want the benefit of "304 not modified" responses.

packages/dynamic-flows/src/flow/DynamicFlow.js
JulianG-TW: Yes. You're right.
They're called `ETags` (as in entity tags), so I agree the variable should be called `eTag`.

But the linter thinks it's smarter than us and throws an error:

```
Please rename the variable `eTag`. Suggested names are: `errorTag`, `eventTag`. A more descriptive name will do too.
```

I don't think we should rename it to `entityTag` only to appease the linter. They're really known as [ETags](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) everywhere.

I think we can keep `etag`.

packages/dynamic-flows/src/flow/DynamicFlow.spec.js
JulianG-TW: Yes. The `setTimeout` was there since the beginning.
I think it's used to settle the promise in the next even loop cycle.
AFAIK promises can resolve in the same "thread" under some conditions.
This `setTimeout` with a delay of zero is probably here to make sure the promise is rejected after the thread is done.

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -131,10 +136,15 @@ const DynamicFlow = (props) => {
...
   const fetchRefresh = (action, data) => {
     setRefreshing(true);
-    return requestStep({ action, data })
+    return requestStep({ action, data, etag })
       .then(async (response) => {
         const step = await response.json();
-        updateStepSpecification(step);
+        updateStepSpecification(step, getETag(response.headers));
+      })
+      .catch(async (error) => {
```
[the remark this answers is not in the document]
JulianG-TW: Done @ b2ece50020bf5fbf2fcc841120b4d03253bba357

packages/dynamic-flows/src/flow/DynamicFlow.spec.js
JulianG-TW: Done @ 80af5c6e0e86db5742ffa38946df83756648f161

StevePole (approved):

PR #1307, opened 2022-01-28
feat: render Modal content in Drawer on mobile

ghost:
## 🖼 Context

https://transferwise.atlassian.net/browse/DS-1528

Render modal content as a full-screen panel on mobile form factor.

## 🚀 Changes

- Add `useLayout` hook to return layout breakpoints as `boolean`
- Migrate `Modal` to TypeScript
- Render `Modal` inside `Drawer` when layout `isMobile`
- Deprecate `scroll` property on `Modal`

## ✅ Checklist


packages/components/src/modal/Modal.tsx
```diff
@@ -0,0 +1,143 @@
...
+import Dimmer from '../dimmer';
+import SlidingPanel from '../slidingPanel';
+
+const TRANSITION_DURATION_IN_MILLISECONDS = 150;
+
+type ModalProps = CommonProps & {
+  title?: ReactNode;
+  body: ReactNode;
+  footer?: ReactNode;
+  size?: Size.SMALL | Size.MEDIUM | Size.LARGE | Size.EXTRA_LARGE;
```
dozortsev, line 17: we shouldn't use enums for defining type, it makes prop to only accept that enum while we want allow consumers pass strings, we have[ all size types defined](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/common/propsValues/size.ts#L9-L13) which you can reuse here.

See similar example e.g in [Loader](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/loader/Loader.tsx#L6).

packages/components/src/modal/Modal.tsx
```diff
@@ -0,0 +1,143 @@
...
+  onClose: () => void;
+  open: boolean;
+  scroll?: Scroll.CONTENT | Scroll.VIEWPORT;
+  position?: Position.TOP | Position.CENTER;
```
dozortsev, line 21: same, shouldn't use enums here

packages/components/src/modal/Modal.tsx
```diff
@@ -0,0 +1,143 @@
...
+          </div>
+        )}
+      </div>
+    );
+  };
+
+  return isMobile ? (
+    <Dimmer open={open} onClose={onClose}>
+      {/* @ts-expect-error TODO: Remove this once SlidingPanel is migrated to TypeScript */}
+      <SlidingPanel open={open} position={Position.RIGHT}>
```
dozortsev, line 91: just wondering why not using Drawer component here?

packages/components/src/modal/Modal.tsx
```diff
@@ -0,0 +1,143 @@
...
+      className={classNames('d-flex', 'justify-content-center')}
+      onClose={onClose}
+    >
+      <CSSTransition
+        appear
+        in={open}
+        classNames={{ enterDone: 'in' }}
+        timeout={TRANSITION_DURATION_IN_MILLISECONDS}
+        unmountOnExit
+      >
```
dozortsev, line 113: I think here we could reuse Panel component, have you tried it?

packages/components/src/modal/Modal.tsx
matebek-tw: Possible, but then we wouldn't be able to reuse the `ModalContent` and it has a slightly different design. I am happy to refactor if that makes more sense though.

packages/components/src/modal/Modal.tsx
matebek-tw: Tried it, wouldn't really simplify things in this case I think.

packages/components/src/modal/Modal.tsx
dozortsev: I would still refactor to use Drawer because Drawer is basically a combination of Dimmer & SlidingPanel

I see that Drawer's `headerTitle` is `string` type, I would change that to `ReactNode` similar to Modal

dozortsev: I would try create three separate bits `<Title>`, `<Body>` and `<Footer>`

```js
<Drawer
      open={open}
      headerTitle={<Title>}
      footerContent={<Footer>}
      position={Position.RIGHT}
      onClose={onClose}
    >
      <Body>
</Drawer>
```

packages/components/src/modal/Modal.tsx
dozortsev: 👌

packages/components/src/modal/Modal.tsx
dozortsev: here it's still the case for `scroll` prop

packages/components/src/common/hooks/useLayout/useLayout.ts
```diff
@@ -0,0 +1,51 @@
...
+  const [breakpoint, setBreakpoint] = useState<keyof typeof Breakpoint>();
+  const [clientWidth] = useClientWidth({ ref: windowReference }) as [number | null];
+
+  useEffect(() => {
+    if (!clientWidth) {
+      return;
+    }
+
+    if (clientWidth <= Breakpoint.EXTRA_SMALL) {
+      setBreakpoint('EXTRA_SMALL');
```
dozortsev, line 18: can we pass there enum `Breakpoint.EXTRA_SMALL` rather than raw string and use the enum in `return` statement

packages/components/src/common/hooks/useLayout/useLayout.ts
```diff
@@ -0,0 +1,51 @@
+import { useEffect, useState } from 'react';
+
+import { Breakpoint } from '../../propsValues/breakpoint';
+import useClientWidth from '../useClientWidth';
+
+export const useLayout = () => {
```
dozortsev, line 6: shouldn't we explicitly define type of this function?

packages/components/src/common/hooks/useLayout/useLayout.ts
matebek-tw: We could, but that is equal to the resolution rather than an expressive name, that's why I typed the state `keyof typeof Breakpoint` which gives ` useState<"EXTRA_SMALL" | "SMALL" | "MEDIUM" | "LARGE" | "EXTRA_LARGE">` so it is type safe to pass the string value.

packages/components/src/common/hooks/useLayout/useLayout.ts
matebek-tw: That's possible and could be helpful to know we are returning the expected values.

packages/components/src/modal/Modal.tsx
```diff
@@ -1,35 +1,74 @@
...
   const isCompact = checkSpecialClasses('compact');
   const noDivider = checkSpecialClasses('no-divider');
 
-  return (
+  return isMobile ? (
+    <Drawer
+      open={open}
+      headerTitle={title}
+      footerContent={footer}
+      position={Position.BOTTOM}
```
dozortsev, line 63: for mobile web (small screen) version of Modal underthehood we reuse Drawer component, because Modal and Drawer are very similar components (both have header, body and footer, and in many cases serve same purpose) I would suggest we make Modal show up from left to right just like we do for Drawer, because otherwise we sort of introducing inconsistency (Drawer on mobile web shows up from left to right, while Drawer which is inside Modal for mobile web shop up from bottom to top)


https://user-images.githubusercontent.com/4015628/152157828-d7ff0743-023c-4b94-b890-f198d48d4e32.mov



if you still disagree then I would suggest we make Drawer also appear from bottom to top on mobile web

cc @javier-cuello

javier-cuello: The reason for this from showing from the bottom is aligning to a more natural behaviour on mobile for modals, like it happens on iOS and Android for full-screen content, that is usually appearing from the bottom. Panels like the drawer that cover one side of the screen, would make sense to show from the closes side (appearing from the left).

> I would suggest we make Drawer also appear from bottom to top on mobile web

Yes. If we had to keep the same behaviour on mobile only, I'd prefer to have it bottom to top. @dozortsev

dozortsev: okey, it should be easy change in Drawer

```js
const { isMobile } = useLayout();

<SlidingPanel open={open} position={isMobile ? Position.RIGHT : position}>
```

packages/components/src/common/hooks/useLayout/useLayout.ts
```diff
@@ -0,0 +1,57 @@
+import { useEffect, useState } from 'react';
+
+import { Breakpoint } from '../../propsValues/breakpoint';
+import useClientWidth from '../useClientWidth';
+
+export const useLayout = (): {
```
dozortsev, line 6: I would probably expose this hook from `@transferwise/components` package, it could be pretty useful for our consumers, and beneficial for consistency

packages/components/src/modal/Modal.tsx
matebek-tw: It might be confusing for consumers, who specifically set the position and it is overridden on mobile without them know why.

dozortsev: I think for such things it's totally ok for us to enforce some patterns / behaviours

packages/components/src/drawer/Drawer.story.js
```diff
@@ -14,7 +14,11 @@ export default {
 export const Basic = () => {
   const [openDrawer, setOpenDrawer] = useState(false);
   const [openModal, setOpenModal] = useState(false);
-  const position = select('position', Object.values(Position), Position.RIGHT);
+  const position = select(
+    'position',
+    [Position.LEFT, Position.RIGHT, Position.BOTTOM],
```
dozortsev, line 19: I probably wouldn't add here `Position.BOTTOM` option, as we don't want people to suggest using bottom position mostly because it's look buggy, it might confuse consumers...

Screen Shot 2022-02-03 at 12 43 18 PM

packages/components/src/modal/Modal.rtl.spec.js
```diff
@@ -1,4 +1,4 @@
-import { Position, Scroll } from '../common';
+import { Position } from '../common';
```
dozortsev, line 1: strange that none of Enzyme tests failed because of refactoring of Modal component 🤔

I've though that as part of this refactoring it's nice occasion to get rid of [Enzyme tests](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/modal/Modal.spec.js) and if necessary improve React Testing library one, this is not really mandatory so up to you

dozortsev (approved):

packages/components/src/modal/Modal.rtl.spec.js
dozortsev: I would actually double check them, they might doesn't check things actually https://github.com/transferwise/neptune-web/runs/5041675409?check_suite_focus=true#step:5:1100

PR #1321, opened 2022-02-07
feat: adding basic ActionButton component implementation

helendurranttw:
## 🖼 Context

We want ActionButtons!

## 🚀 Changes

 
Added the new ActionButton component.

Normal:

Screenshot 2022-02-07 at 16 57 41

Hover:

Screenshot 2022-02-07 at 16 59 11

Focus:

Screenshot 2022-02-07 at 16 58 01

Active:

Screenshot 2022-02-08 at 09 37 02

Focus & Active:

Screenshot 2022-02-08 at 09 37 13

## 🤔 Considerations

I haven't implemented icons as part of this - can add in a follow up PR.

## ✅ Checklist


packages/components/src/actionButton/ActionButton.less
```diff
@@ -0,0 +1,30 @@
+.np-action-btn {
```
helendurranttw, line 1: I took the decision not to use existing button styles, as there were many we didn't need for this button and I ended up having to overwrite a lot of the existing styles. Open to this being challenged though.

nabaroa: Loving how neat these styles look like ❤️ 
It totally makes sense creating them apart from existing buttons classes (too bloated and complex).

packages/components/src/actionButton/ActionButton.less
```diff
@@ -0,0 +1,30 @@
...
+  transition: all 0.15s ease-in-out;
+  transition-property: color, border-color, background-color, box-shadow;
+
+  &:not(:disabled) {
+    &:hover {
+      background-color: var(--color-background-accent);
+    }
+
+    &:active {
+      background-color: rgba(0, 162, 221, 0.15);
```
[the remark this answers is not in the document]
helendurranttw: Decision was made to not implement this for now. If we want to progress with arbitrary opacity on tokens we will need a different solution.

packages/components/src/actionButton/ActionButton.less
```diff
@@ -0,0 +1,30 @@
+.np-action-btn {
+  display: inline-flex;
+  height: var(--size-32);
+  padding: 7px var(--size-12) var(--size-8);
```
dozortsev, line 4: why we have custom `7px` padding here?

packages/components/src/actionButton/index.ts
```diff
@@ -0,0 +1 @@
+export { default } from './ActionButton';
```
dozortsev, line 1: you seems forget add page in neptune web docs for this component?

I would also add link to ActionButton page in [Button one](https://transferwise.github.io/neptune-web/branch/action-button/components/inputs/Button), e.g mention some key differences and when to use one and another.

packages/components/src/actionButton/ActionButton.tsx
```diff
@@ -0,0 +1,29 @@
+import classNames from 'classnames';
+import { MouseEvent, ReactElement } from 'react';
+
+type ActionButtonProps = {
+  className?: string;
+  disabled?: boolean;
+  text: string;
```
dozortsev, line 7: I think this prop should be called `children` and be type of `ReactNode`

packages/components/src/actionButton/ActionButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+  text,
+}: ActionButtonProps): ReactElement => {
+  return (
+    <button
+      type="button"
+      className={classNames(`np-action-btn`, className)}
+      disabled={disabled}
+      onClick={onClick}
+    >
+      <span>{text}</span>
```
dozortsev, line 24: why we need `span` here?

helendurranttw: We don't - I was using it when I started adding icon support, but as I've stashed that for now I'll remove it 👍

packages/components/src/actionButton/ActionButton.tsx
dozortsev: > ..we would want to add support for it ourselves, rather than letting people add their own

they should have know that they rather should contribute icon feature in the component for consistency, otherwise it's their fault

and I think people would actually consider to contribute

> ..Also this matches the API for Android and iOS :)

we should respect platform specific things

---

```js
<Button>Copy</Button>
<Link>Copy</Link>
<CircularButton>Copy</CircularButton>
<ActionButton text="Copy" />
```

doesn't this look odd? it's clearly inconsistency in API

helendurranttw: Right, but the end goal is that we'll have an API that looks like this: 

```<ActionButton text="Copy" icon={<Chevron/>} iconPosition="trailing" />```

With this API it would be weird to allow passing of icons in two ways (one as part of children, and via the API). Our old buttons have no support for icons, so it's a bit different. We have been moving towards this way of doing things (e.g. the Alert component which also takes a message). Children should only be used when the contents is unknown: but here we know it should only ever be a string, because we're going to add handling of icons ourselves.

Keep in mind - we will be adding icon support at some stage - Javier has specified it in the docs. iOS already support an optional trailing / leading icon. I just didn't do it as part of this PR because I didn't have the time to get into it today.

Comment (helendurranttw) [reply packages/components/src/actionButton/ActionButton.tsx:7]: I guess we could also use children and just say in the docs that the behaviour is unspecified if they pass in anything other than text 🤷 Otherwise styling the icons would become problematic.

Would you be happy with that?

Comment (dozortsev) [reply packages/components/src/actionButton/ActionButton.tsx:7]: [yep, okey](https://github.com/transferwise/neptune-web/pull/1321#discussion_r801677929)

or we could keep it `string` but it call it `children` [same as we have in CircularButton](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/circularButton/CircularButton.js#L38) 😄 

later if we need we can change it to `ReactNode` without breaking changes.

Comment (richardmccartney) [reply packages/components/src/actionButton/ActionButton.tsx:7]: Keeping it as `text` with a type of `string` would be an ideal approach for the API of this component in its current state. Looking at the designs we wouldn't need to reference `children` as @helendurranttw mentioned and if the action button doesn't require Icon support allowing `children` with a type of `React.Node` feels a bit redundant.

We should also stay away from confusing the expectations of props, `children` would be expected to take multiple child elements and think it would be best to avoid typing it with what potentially other developers wouldn't expect 😄

Comment (richardmccartney) [reply packages/components/src/actionButton/ActionButton.tsx:7]: For now, let's keep it with the `text` prop and we can look at how we organise some of our wider prop API in the future. We certainly should try to keep it as close to the native style but only when it warrants the need for that component 👍 

Keen to have a run-through on how we're managing props and the switch to typescript within the DS in the coming weeks.

File: packages/components/src/actionButton/ActionButton.less (hunk 2 above)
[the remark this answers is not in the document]
Comment (richardmccartney) [reply packages/components/src/actionButton/ActionButton.less]: I would love to know more about this one @helendurranttw when you have time 😄

File: packages/docs/pages/components/inputs/ActionButton.mdx [hunk 7]
```diff
@@ -0,0 +1,15 @@
...
+
+Use action buttons for non-primary standalone actions or when there are multiple actions a user can choose from.
+
+For primary actions, use a normal <DocumentLink href="components/inputs/Button">Button</DocumentLink>. Refer to the <Link href="https://neptune.wise.com/components/action-button/" target="_blank">action button usage guidelines</Link> for more information on the differences between the two components.
+
+<LiveEditorBlock code={code} scope={{ ActionButton }} />
+<GeneratePropsTable componentName="ActionButton" />
+
+export const meta = {
+  name: 'ActionButton',
```
dozortsev, line 14:
```suggestion
  name: 'ActionButton',
  badge: {
    expiryDate: '2022-03-04',
    type: 'new',
  },
```

packages/components/src/actionButton/ActionButton.tsx
dozortsev: note, if allowing props spread (you mentioned about this in https://github.com/transferwise/neptune-web/pull/1321#issuecomment-1033719993) you should make props works as expected, including `children`

Screen Shot 2022-02-10 at 11 42 58 AM

packages/components/src/actionButton/ActionButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+    { disabled = false, className = undefined, onClick, text, ...rest }: ActionButtonProps,
+    reference,
+  ) => (
+    <button
+      ref={reference}
+      type="button"
+      className={classNames(`np-action-btn`, className)}
+      disabled={disabled}
+      onClick={onClick}
+      {...rest}
```
dozortsev, line 22: `disabled` and `onClick` are now redundant here and a props because can be passed via `...rest`

packages/components/src/actionButton/ActionButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+const ActionButton = forwardRef<HTMLButtonElement, ActionButtonProps>(
+  (
+    { disabled = false, className = undefined, onClick, text, ...rest }: ActionButtonProps,
+    reference,
+  ) => (
+    <button
+      ref={reference}
+      type="button"
+      className={classNames(`np-action-btn`, className)}
```
dozortsev, line 19: same for `className`
```suggestion
      className={classNames(`np-action-btn`, rest['className'])}
```

packages/components/src/actionButton/ActionButton.tsx
helendurranttw: Because the props being spread are specifically HTML Button attributes :)

packages/components/src/actionButton/ActionButton.tsx
helendurranttw: I think we should leave these as is. For most use cases people will only need to provide these props, and this makes the API easier to understand.

dozortsev: it will work same way if we delete (redundant) code, 

we can add note in docs mentioning that it accepts all native button's attrs/props

dozortsev (approved):

richardmccartney (approved):

richardmccartney:
Approved ✅ little rebase and this should be good to go @helendurranttw

PR #1323, opened 2022-02-08
fix(Button): transition for disabled state

dozortsev:
## 🖼 Context

When we introduce new disabled state we didn't took into account transition in some components.

## 🚀 Changes

Add `transition` for `filter` and `opacity` props of disabled state.

## ✅ Checklist


packages/css/src/less/core/_scaffolding.less
```diff
@@ -380,6 +380,8 @@ video {
   filter: grayscale(1);
   opacity: 0.45;
   cursor: not-allowed !important;
+  transition: all 0.15s ease-in-out;
```
mikknarusk, line 383: actually doesn't moving it here only cause it to be transitioned in a single direction, from disabled to active? I think active to disabled will not be transitioned

dozortsev: yep, you are right

I actually just need delete `transition-property` which overrides `transition: all [duration] [ease function]`

here is it works now (increased duration for 1 sec just for this demo)


https://user-images.githubusercontent.com/4015628/153030160-214bad18-abcc-4585-8413-a7211c861e5e.mov

packages/css/src/less/mixins/_buttons.less
```diff
@@ -14,7 +14,6 @@
   background-color: @background;
   border-color: @border;
   transition: all 0.15s ease-in-out;
-  transition-property: color, border-color, background-color;
```
dozortsev, line 17: this prop was just overriding setting in `transition` prop which wouldn't allow apply `transition` for `filter` and `opacity` of disable state

Screen Shot 2022-02-08 at 4 25 48 PM

nabaroa (approved):

PR #1328, opened 2022-02-14
DynamicFlow component: new initialStep prop

JulianG-TW:
## 🖼 Context

Currently the only way to initialise a `DynamicFlow` component is using a `flowUrl` which will be requested using the `GET` method.

In some cases, applications need to send some data to the initial request, when said data contains PII, we can't use the `GET` method, and need to use `POST` instead.

## 🚀 Changes

Now, the `DynamicFlow` component can be initialised with a `initialStep` instead of a `flowUrl`.
If both props are passed, `flowUrl` is ignored.

**Before**

```jsx
return (
  <DynamicFlow
    flowUrl="/some/flow"
    ...
  />
);
```

**After** (optionally)

```jsx
return (
  <DynamicFlow
    initialStep={myInitialStep}
    ...
  />
);
```

## 🤔 Considerations

And alternative would be to initialise it with an "action" where some data and an http method can be specified.

## ✅ Checklist


packages/dynamic-flows/src/flow/DynamicFlow.end-to-end.spec.js
```diff
@@ -88,36 +88,62 @@ const mockFetcher = jest.fn((input, init) => {
...
-        </Provider>,
-      );
-    }
+  const initialStepResponse = new Response(JSON.stringify(getFormStep()));
+
+  function renderComponent() {
+    render(
+      <Provider i18n={i18n}>
+        <DynamicFlow
+          initialStepResponse={initialStepResponse}
```
StevePole, line 101: Could we rename to `initialStep`?  I think the response bit is kind of redundant, it might not even be a response, maybe the page just knows what it wants to render 🤷‍♂️

JulianG-TW: As discussed, I've changed the implementation to accept/expect an `initialStep` rather than a Response object; and named the prop `initialStep` of course.

richardbata (approved):

PR #1336, opened 2022-02-18
[VUX-1440] OverlayHeader maxWidth value added

richardbata:
## 🖼 Context
OverlayHeader does not have a max with like other navigation components do in Neptune, and this leads to a bad layout on large viewport devices. It is agreed that adding a max width value improves the UX of the component

Screenshot 2022-02-17 at 15 50 32

Screenshot 2022-02-17 at 15 50 52

Slack: https://transferwise.slack.com/archives/C2XG7R3EZ/p1645109609993739

## 🚀 Changes
Surrounded the existing component with a flex div, and set the max width of the OverlayHeader the same as is for FlowNavigation, 1164px (+24px padding each side for both).

## ✅ Checklist


packages/components/src/overlayHeader/OverlayHeader.less
```diff
@@ -1,14 +1,19 @@
 @import (reference) '@transferwise/neptune-css/src/less/addons/_spacing-utilities.less';
 
 .np-overlay-header {
-  // This is the same height of FlowNavigation.
-  min-height: 97px;
+  .np-overlay-header__content {
+    // This is the same height of FlowNavigation.
```
dozortsev, line 5:
```suggestion
    // Same height and width as in FlowNavigation
```

dozortsev (approved):

PR #1338, opened 2022-02-22
feat(UploadInput): Add additional props to UploadInput to make it customisable with DF/other clients

dhananjaykha:
## 🖼 Context

The PR adds three props to the UploadInput Component to make it more in sync with DF's customisability as that in mobile implementations.

1. `id` : (Optional) To allow selecting input from labels. Earlier had a hardcoded label which would make it buggy on having multiple components in the same page.

2. `uploadButtonTitle`: (Optional) To allow customising the upload button text. If not provided, the existing logic is used.

3. `sizeLimitErrorMessage`: (Optional) To allow custom error message. If not provided, falls back to the Neptune error message.

## 🚀 Changes

 
 Adds the props, needed tests, and the stories.

## ✅ Checklist


packages/components/src/uploadInput/UploadInput.spec.tsx
```diff
@@ -281,4 +281,34 @@ describe('UploadInput', () => {
       });
     });
   });
+
+  describe('Misc', () => {
```
smallosaurus, line 285: What is `Misc`? IMO we need to describe that properly, the description is not yet useful for others that later read this code ;P

packages/components/src/uploadInput/UploadInput.spec.tsx
```diff
@@ -281,4 +281,34 @@ describe('UploadInput', () => {
...
+          <label htmlFor={inputId}>{labelText}</label>
+          <UploadInput {...props} inputId={inputId} />
+        </>,
+      );
+      expect(screen.getByLabelText(labelText)).toHaveAttribute('id', inputId);
+    });
+
+    it('should show the file size error when provided in props', async () => {
+      const oneKbSize = 1;
+      const twoKbInBytes = 2048;
```
smallosaurus, line 300: are there any repeat occurrences of these constants? Feels like these would be good candidates for a static constants file

packages/components/src/uploadInput/UploadInput.spec.tsx
```diff
@@ -281,4 +281,34 @@ describe('UploadInput', () => {
...
+        <>
+          <label htmlFor={inputId}>{labelText}</label>
+          <UploadInput {...props} inputId={inputId} />
+        </>,
+      );
+      expect(screen.getByLabelText(labelText)).toHaveAttribute('id', inputId);
+    });
+
+    it('should show the file size error when provided in props', async () => {
+      const oneKbSize = 1;
```
smallosaurus, line 299: I think this might be a typo... the line below is `2048` whereas this is `1`.

Also, why not consider defining a constant somewhere `const KB = 1024` and reusing?

packages/components/src/uploadInput/UploadInput.story.tsx
```diff
@@ -150,3 +150,16 @@ withMaxFilesToUploadLimit.args = {
...
+withFileSizeErrorMessage.args = {
+  ...props,
+  sizeLimit: 1,
+  sizeLimitErrorMessage: 'The file is oversized',
+};
+
+export const withCustomUploadButtonTitle = Template.bind({});
+withCustomUploadButtonTitle.args = {
+  ...props,
+  uploadButtonTitle: 'Upload all files you want',
```
smallosaurus, line 164: `Upload all the files you want`?

packages/components/src/uploadInput/UploadInput.spec.tsx
dhananjaykha: Not occurring anywhere else. Also this is a test file, and single test using it, so just kept it there.

packages/components/src/uploadInput/UploadInput.story.tsx
dhananjaykha: This is a playground like environment. Not source code/tests. So it does not matter what we pass in.

packages/components/src/uploadInput/UploadInput.spec.tsx
dhananjaykha: This is more of a group of tests. 
I did not feel like keeping a group for just this one-off test. And felt that all such one-off tests can be grouped under a group than a describe block for just a test of a type. Hence named - `miscellaneous`.

Later, someone adding the tests can refactor this grouping as they feel the need.

packages/components/src/uploadInput/UploadInput.story.tsx
richardmccartney: Could we add title text that fit's with our content guidelines? https://neptune.wise.com/content/product-voice/

packages/components/src/uploadInput/uploadButton/UploadButton.tsx
```diff
@@ -182,7 +201,7 @@ const UploadButton = ({
     >
       <input
         ref={inputReference}
-        id="np-upload-button"
+        id={inputId || 'np-upload-button'}
```
richardmccartney, line 204: Do we think it might be better to set `inputId` to have a default value of `np-upload-button` rather than placing a logical expression within the component output?

packages/components/src/uploadInput/UploadInput.spec.tsx
dhananjaykha: Fair point. It was used just here. But if it confuses you, it may confuse others as well. I'll take rewrite it better.

packages/components/src/uploadInput/uploadButton/UploadButton.tsx
dhananjaykha: Yeah, you're right. That should be the way.

packages/components/src/uploadInput/UploadInput.story.tsx
dhananjaykha: Replaced with a better one.

dhananjaykha:
@smallosaurus @richardmccartney changed the needful/commented on the comments.

richardmccartney (approved):

packages/components/src/uploadInput/uploadButton/UploadButton.tsx
```diff
@@ -48,6 +48,16 @@ export type UploadButtonProps = {
    * @param files
    */
   onChange: (files: FileList) => void;
+
+  /**
+   * Id for the upload input
+   */
+  inputId?: string;
```
dozortsev, line 55: please rename this prop to be just an `id` as in [HTML](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id) we tend keep API as similar to natives as possible

PR #1345, opened 2022-03-01
feat(Typeahead): specify where arbitrary data can live on Typeahead options

joebartels-tw:
## 🖼 Context
**A quick overview of some Typeahead behavior:**
The `option` objects that are passed to `<Typeahead options={options}/>` are purely for UI (e.g. the fields are: `label`, `note`, `secondary`). And when options are selected in the UI - the selected `option` objects are passed to the `onChange` callback that was supplied by a parent component

**Motivation for this change:**
Typically a parent component is probably remapping some search results, or other list that contains UIDs, into a list of `option` objects. And so, when a specific `option` is selected (and the `onChange` callback is invoked) - the parent component needs to figure out, from this specialized `option` object - which of the actual items it represents. Also the existing approach is not TypeScript friendly, whereas if `option.data` can hold "arbitrary" data then we can create a definitions like so (my syntax may be off 😬 ):
```ts
type onChange<T>(res: TypeaheadOption<T>) => void

type TypeaheadOption<T> {
  label: string;
  note: string;
  secondary: string;
  data: T
}
```

**But, this already works... doesn't it?**
It turns out, that Typeahead does not mutate these `option` objects, so a parent component can actually overload it with w/e they want. However, this is undocumented, and open for interpretation on how to implement basic usage. And there are no tests on this behavior to ensure it won't break in the future... It's also not TypeScript friendly when this component is migrated

🥁 
🥁 
🥁 

The idea with this PR is to formalize exactly where unique data can go. This way, when the `onChange` callback is invoked, the Consumer can inspect `option.data` and find whatever they put there (e.g. an `id` or entire domain objects)

## 🚀 Changes

I've added a PropType definition for this `data` field + a test to ensure that the value(s) in `data` are still present when the `onChange` callback is invoked

## 🤔 Considerations
Can this be further documented some how? I'm unfamiliar with how the docs are generated, but an example in the right place would be enough I'm sure...

## ✅ Checklist


packages/components/src/typeahead/Typeahead.js
```diff
@@ -435,6 +435,7 @@ Typeahead.propTypes = {
       label: PropTypes.string.isRequired,
       note: PropTypes.string,
       secondary: PropTypes.string,
+      data: PropTypes.object,
```
dozortsev, line 438: I see you point, make sense,

but I would call this property `value` same as in HTML and in [Select component for exactly similar thing](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/select/Select.js#L567).

joebartels-tw: consistency! awesome! I'll push an update

dozortsev (approved):

PR #1346, opened 2022-03-01
feat!: TypeScript setup and types generation for `@transferwise/dynamic-flows`

dozortsev:
## 🖼 Context

Enabling TypeScript in `@transferwise/dynamic-flows` package and generate types based on PropTypes for JS exposed components via `react-to-typescript-definitions` library.

## 🚀 Changes

For context, the setup is almost same as in `@transferwise/components` https://github.com/transferwise/neptune-web/pull/846.

- Enable TypeScript in Babel via `@babel/preset-typescript` in dynamic-flows package
- Add `tsconfig.json` for TypeScript files (`.ts`, `.tsx`)
- Create `generate-type-declarations.js` script that would generate types for JS components

## ✅ Checklist


packages/dynamic-flows/src/field/Field.js
```diff
@@ -228,8 +228,24 @@ Field.propTypes = {
     autoComplete: PropTypes.bool,
     placeholder: PropTypes.string,
     searchPlaceholder: PropTypes.string,
-    control: PropTypes.oneOf(Object.values(FormControlType)),
-    format: PropTypes.oneOf(Object.values(FieldFormats)),
+    control: PropTypes.oneOf([
```
richardmccartney, line 231: Question is there any way we can still have central PropTypes that are shared using `PropTypes.shape({ })`?? The idea of having common shared Props seems like a good thing to keep, otherwise, we are repeating a lot of the same thing between components?

packages/dynamic-flows/src/formControl/FormControl.js
```diff
@@ -433,13 +433,7 @@ FormControl.propTypes = {
...
   searchValue: PropTypes.string,
   onSearchChange: PropTypes.func,
-  size: PropTypes.oneOf([
-    FormControl.Size.EXTRA_SMALL,
-    FormControl.Size.SMALL,
-    FormControl.Size.MEDIUM,
-    FormControl.Size.LARGE,
-    FormControl.Size.EXTRA_LARGE,
-  ]),
+  size: PropTypes.oneOf(['xs', 'sm', 'md', 'lg', 'xl']),
```
richardmccartney, line 436: Same as before, are we able to use `shape` to enable the shared props?

JulianG-TW (approved):
Thank you so much @dozortsev !

I think we should merge this as a BREAKING CHANGE because the translations imports seem to have changed.

```diff
-import dynamicFlowsTranslations from '@transferwise/dynamic-flows/build/i18n';
+import { translations as dynamicFlowsTranslations } from '@transferwise/dynamic-flows';
```

Also, TypeScript projects would need to remove dynamic-flow entries from their **declarations.d.ts** file.

richardmccartney:
> 

I agree @JulianG-TW if we could make this a breaking change and then update the release notes accordingly that would be great 👍

richardmccartney (approved):

PR #1350, opened 2022-03-02
feat(dynamic-flow): add debouncing when refreshing string schemas

JulianG-TW:
## 🖼 Context

The mobile clients use a debouncing strategy to avoid constantly refreshing as a user types into a textfield.

The `recipient-form` library (used in recipient creation) does a similar thing.

But `DynamicFlow` was not debouncing.

## 🚀 Changes

The `DynamicFlow` component now debounces refresh calls in **some** cases.

This is all handled by a custom hook `useDebouncedRefresh`, so the changes in DynamicFlow.js are only 3 lines.

The decision on which schemas to debounce is based on the schema type:

```js
function shouldDebounceSchema(schema) {
  return (
    getSchemaType(schema) === schemaType.BASIC &&
    schema.type !== 'boolean' &&
    schema.format !== 'base64url'
  );
}
```

We also only debounce refreshes to the same url. When two schemas have different `refreshFormUrl` values, they are independently debounced.

#### Before (no debouncing)
The button is grayed-out while the step is being refreshed (very frequently, with almost every change)
no-debounce

#### After (debouncing)
The button is grayed-out while the step is being refreshed (only 2 seconds after the user stops typing)
debounce

## 🤔 Considerations

De debounce interval/timeout is hardcoded to 1 second, while mobile clients use a 2-second timeout.
The `recipient-form` library currently in use for recipient creation uses 1 second.

## ✅ Checklist


packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -151,6 +153,10 @@ const DynamicFlow = (props) => {
       .finally(() => setRefreshing(false));
   };
 
+  const fetchRefreshWithDebounce = useDebouncedCallback((action, data) => {
```
StevePole, line 156: Why set this up out here rather than inside?

StevePole (commented):
Do we need to debounce on all schemas?  Thinking back to the angular implementation, I think we only did it on string schemas where we had types input.  Perhaps that's not what we want.  We also handled the debounce at a lower level, which meant we could contain that logic within the string schema itself, and not introduce more complexity at the top level.  We would just manage which changes were broadcast up the component stack.

packages/dynamic-flows/src/flow/DynamicFlow.js
JulianG-TW: I'm not sure what do you mean by "inside".

I can't go inside the `onModelChange` callback, because it's a react hook, and needs to be in the component scope, outside of any conditionals, etc.

JulianG-TW:
> Do we need to debounce on all schemas? Thinking back to the angular implementation, I think we only did it on string schemas where we had types input. Perhaps that's not what we want. We also handled the debounce at a lower level, which meant we could contain that logic within the string schema itself, and not introduce more complexity at the top level. We would just manage which changes were broadcast up the component stack.

We're only debouncing "string" schemas that are not `oneOf`:

```js
function shouldDebounceRefresh(triggerSchema) {
  return triggerSchema?.type === 'string' && !triggerSchema.oneOf;
}
```

I believe it's the same in mobile clients. I'll double check.

JulianG-TW:
> I believe it's the same in mobile clients. I'll double check.
 
 Apparently not.
 
 > On iOS we debounce anything while in an "editing" state, and the only component that can be "editing" is a text input - but that could be a string, an integer, or a number schema (Eddie)

So I need to find a condition that covers anything that could result in an `<input type="text" ...`, I think.

StevePole:
Got it, I wonder then, if we should be doing this in the BasicTypeSchema itself, so we only broadcast onChange handlers dependent on the debounce?

JulianG-TW:
> Got it, I wonder then, if we should be doing this in the BasicTypeSchema itself, so we only broadcast onChange handlers dependent on the debounce?

Yes. I think that would work. But I have a few reservations:

- Each schemas/input will debounce itself, and we could end up with a refresh being triggered while there's still typing happening on another input field. Very unlikely in practice, but it could happen if we had two text inputs and the user types on one, then hits tab and continues tabbing on the next field. we could end up missing a few keystrokes entered on the second field, since the refresh response would not contain the latest model.
- The other worry is, if we ever refactor this into stateless components, then debouncing will need to be on the root level.

I'm happy to try moving the debouncing into `BasicTypeSchema` or `SchemaFormControl` if you think those concerns are not worth extracting `getControlType`.

I think it will be fine either way.

yarn.lock
```diff
@@ -22224,6 +22224,11 @@ use-composed-ref@^1.0.0:
   dependencies:
     ts-essentials "^2.0.3"
 
+use-debounce@^7.0.1:
```
dozortsev, line 22227: I would use `lodash.debounce` with 2s interval as we currently have in Typeahead component

https://github.com/transferwise/neptune-web/blob/main/packages/components/src/typeahead/Typeahead.js#L25-L31

JulianG-TW: done!

packages/dynamic-flows/src/common/schemaTypes/getControlType.js
```diff
@@ -0,0 +1,42 @@
+import { FormControlType } from '..';
+
+import { isOneOfSchema } from './schemaTypes';
+
+export function getControlType(schema) {
```
JulianG-TW, line 5: Not new. Extracted from **SchemaFormControl.js**.

packages/dynamic-flows/src/flow/useDebouncedRefresh/useDebouncedRefresh.js
```diff
@@ -0,0 +1,48 @@
+import debounce from 'lodash.debounce';
+import { useRef } from 'react';
+
+import { FormControlType } from '../../common/propsValues/formControlType';
+import { getControlType } from '../../common/schemaTypes/getControlType';
+
+const DEBOUNCE_DELAY = 1000; // milliseconds
+
+const debounceableControlTypes = new Set([
```
StevePole, line 9: Is there a reason we can't use the schemas fro this, rather than control types?  It feels like it adds a layer of indirection.  We need to turn the schema into a control type, then check against a list of control types, could we not just say something like
```
(schema.type === 'string' && schema.format !== 'base64url') || schema.type === 'number'
```
Or is there something I'm missing?

JulianG-TW: We can probably create a good conditional expression. but I'm afraid it would be more complex than `(schema.type === 'string' && schema.format !== 'base64url') || schema.type === 'number'`

If I understand correctly, there are more schema types that would need to be debounced, like:  "integer",
and some more "string" types that should not be debounced, like when it's a `oneOf` or `const`.

Because I could not think of all the cases, and -in a way- we want to debounce when the control is a textfield, I thought we could reuse the `getControlType` function.

On a tangential note, arguably the `function shouldDebounce(schema)` could be moved to `/common/schemaTypes/`.

StevePole: If it's a oneOf or a const it should not have a 'type', their type is determined by their value.  So it's just adding integer as far as I can see?

> On a tangential note, arguably the function shouldDebounce(schema) could be moved to /common/schemaTypes/.

This is a good idea.

JulianG-TW: Done! now we decide to debounce based on schema type:
```js
function shouldDebounceSchema(schema) {
  return getSchemaType(schema) === schemaType.BASIC && schema.type !== 'boolean';
}
```

JulianG-TW: Updated to:
```js
function shouldDebounceSchema(schema) {
  return (
    getSchemaType(schema) === schemaType.BASIC &&
    schema.type !== 'boolean' &&
    schema.format !== 'base64url'
  );
}
```

StevePole (approved):

PR #1353, opened 2022-03-08
fix(dynamic-flows): ignore invalid property names in displayOrder array in ObjectSchema

JulianG-TW:
## 🖼 Context

missing property

When an object schema contains a "displayOrder" property which includes a property name that is missing from the "properties" object, the render throws with:

```
"Cannot read property 'width' of undefined"
```

This is often confusing and makes it difficult to pin-point the problem, especially when source mapping is not working correctly in the consuming codebase.

## 🚀 Changes

We filter-out invalid property names in order to prevent throwing.

The reason behind this is that if the "displayOrder" array is incorrect, we can still render the schema, ignoring the array, or some of its elements.

## 🤔 Considerations

We don't have to throw. We could instead ignore the invalid property and continue rendering.

~~Without changing behaviour, we now throw with a more descriptive message:~~

~~"Invalid object schema. The 'displayOrder' array contains a property name which is not in the 'properties' object. (missing 'some-invalid-property')"~~

## 🤔 Considerations

~~We don't have to throw. We could instead ignore the invalid property and continue rendering.~~

## ✅ Checklist


packages/dynamic-flows/src/jsonSchemaForm/objectSchema/ObjectSchema.js
```diff
@@ -49,7 +49,7 @@ const ObjectSchema = (props) => {
 
   const orderedPropertyNames = [
     ...new Set([...(props.schema.displayOrder || []), ...Object.keys(props.schema.properties)]),
-  ];
+  ].filter((propertyName) => typeof props.schema.properties[propertyName] !== 'undefined');
```
StevePole, line 52: Very hard to read this statement.  It was a problem before, and this makes it worse.  Could break it out?
```
const isPropertyDefined = (propertyName) => {
  typeof props.schema.properties[propertyName] !== 'undefined'
};

const allPropertyNames = new Set([...(props.schema.displayOrder || []), ...Object.keys(props.schema.properties)]);
const orderedPropertyNames = [...allPropertyNames].filter(isPropertyDefined);
```

Does this actually preserve order?  i.e. Is `Set` order preserving?  I suspect it is by implementation, but perhaps not by spec.

JulianG-TW: [According to MDN docs](https://developer.mozilla.org/en-%0AUS/docs/Web/JavaScript/Reference/Global_Objects/Set) "you can iterate through the elements of a set in insertion order"

I'll make your suggested changes. it reads much better. Thanks!

JulianG-TW: done @ e27bf49e256509ece52cc384453c99f762ee0e51

PR #1356, opened 2022-03-10
feat(dynamic-flows): Disable action buttons during persist-async

JulianG-TW:
## 🖼 Context

When a schema uses persist-async we can expect that some data will be uploaded to a different endpoint. The endpoint returns a token or reference which is used in the model. The model is eventually submitted by one of the step actions.

But if a user quickly submits a form before the persist-async schema gets a chance to finish uploading, the model will not contain said token or reference.

persist-async-before

We should wait for persist-async operations to finish before allowing the user to submit the form.

## 🚀 Changes

We disable actions and action buttons while there are more than zero persist-async pending promises.

persist-async-after

## 🤔 Considerations

This can be further improved by queueing actions and run them after persist-async, rather than requiring the user to click the action button a second time.

## ✅ Checklist


packages/dynamic-flows/src/common/contexts/dynamicFlowContext/usePendingPromiseCounter.ts
```diff
@@ -0,0 +1,16 @@
+import { useCallback, useState } from 'react';
+
+export function usePendingPromiseCounter() {
+  const [count, setCount] = useState<number>(0);
+
+  const addPendingPromise = useCallback(
+    (promise: Promise<unknown>) => {
+      setCount((c) => c + 1);
+      promise.finally(() => delayUntilNextCycle(() => setCount((c) => Math.max(0, c - 1))));
```
StevePole, line 9: I can't remember, does a "finally" fire in the event of thrown exception?  If not, we might need to make sure we decrement in that case.

packages/dynamic-flows/src/flow/examples/persist_async_step.json
```diff
@@ -0,0 +1,39 @@
+{
+  "type": "form",
+  "title": "Persist Async",
+  "schemas": [
+    {
+      "$id": "the-schema",
+      "allOf": [
```
StevePole, line 7: Why an allOf?

packages/dynamic-flows/src/step/layoutStep/LayoutStep.js
```diff
@@ -28,6 +33,7 @@ const LayoutStep = (props) => {
       errors={errors}
       onAction={onAction}
       onModelChange={onModelChange}
+      onPersistAsync={registerPersistAsyncPromise}
```
StevePole, line 36: Did this callback already exist?  Was it not used before?

packages/dynamic-flows/src/flow/examples/persist_async_step.json
JulianG-TW: It's not needed. I have now removed it. Thanks!

packages/dynamic-flows/src/step/layoutStep/LayoutStep.js
JulianG-TW: Yes `DynamicLayout` already had a `onPersistAsync` prop declared in prop types.
`LayoutStep` was not using it.

richardbata (approved):

PR #1359, opened 2022-03-14
feat(dynamic-flows): Update the array schema's "required" behaviour

dhananjaykha:
## 🖼 Context

EARLIER BEHAVIOUR:
- For a required array schema input, an empty array is NOT a valid input
- The array must always contain at least one element.

UPDATE:
- For a required array schema input, an empty array is a valid input
- For the at least one restriction the `minItems` construct must be used.

## 🚀 Changes

 
- Removes the check for required array to have at least one element
- Fix: an invocation to the utility function not passing the `required` value to the function.
- removes test expecting the earlier behaviour

## ✅ Checklist


packages/dynamic-flows/src/common/validation/rule-validators/index.js
```diff
@@ -4,9 +4,8 @@ function isValidRequired(value, isRequired) {
   return !isRequired || !isUndefined(value);
 }
 
-function isValidRequiredArray(value, isRequired) {
-  const isNonEmptyArray = isArray(value) && value.length > 0;
-  return !isRequired || isNonEmptyArray;
+function isValidRequiredArray(value, isRequired = false) {
```
JulianG-TW, line 7: is there a reason we need to default to `false` here?

- the `isValidRequired` on line 3 doesn't have a default.
- all the functions in validation-failures/index.js don't have a default either.

packages/dynamic-flows/src/common/validation/validation-failures/index.js
```diff
@@ -141,7 +141,7 @@ function getConstValidationFailures(value, schema, isRequired) {
   return [];
 }
 
-function getArrayValidationFailures(value, schema, isRequired) {
+function getArrayValidationFailures(value, schema, isRequired = false) {
```
JulianG-TW, line 144: is there a reason for having a default value while all the other functions in this module don't?

packages/dynamic-flows/src/common/validation/rule-validators/index.js
dhananjaykha: Sure. I felt the default makes it predictable that all falsy values are defaulted to false. But okay to make it consistent.

packages/dynamic-flows/src/common/validation/validation-failures/index.js
dhananjaykha: Only reason to add it was if some usage does not pass it explicitly, it is mapped to false. Will make it consistent. No hard opinion.

packages/dynamic-flows/src/common/validation/rule-validators/spec.js
```diff
@@ -141,4 +142,24 @@ describe('Given a library for validating json schema rules', () => {
...
     });
   });
+
+  describe('when establishing validations an array', () => {
+    it('should return true all values for a non-required input', () => {
+      expect(isValidRequiredArray([1])).toBe(true);
+      expect(isValidRequiredArray([])).toBe(true);
+
+      expect(isValidRequiredArray(undefined)).toBe(true);
+      expect(isValidRequiredArray(null)).toBe(true);
```
JulianG-TW, line 152: missing a second argument in these calls?

JulianG-TW (approved):

PR #1363, opened 2022-03-16
Fixing type declaration of DynamicFlow

richardbata:
## ✅ Checklist


packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -275,7 +307,15 @@ const DynamicFlow = (props) => {
   );
 };
 
-DynamicFlow.propTypes = {
+const DynamicFlowWrapper = (props) => {
+  return (
+    <ErrorBoundary onError={props.onError}>
+      <DynamicFlow {...props} />
+    </ErrorBoundary>
```
dozortsev, line 314: I would just use `<ErrorBoundary >` in `return (` of `DynamicFlow`

dozortsev: this `DynamicFlowWrapper` looks redundant and rather problematic 

we should have having this
```js
DynamicFlow.propTypes = DynamicFlowWrapper.propTypes = {
```

richardbata: The issue is it doesn't capture JS errors within the DynamicFlow component...

packages/dynamic-flows/src/flow/DynamicFlow.js
```diff
@@ -275,7 +307,15 @@ const DynamicFlow = (props) => {
...
+  );
+};
+
+DynamicFlow.propTypes = DynamicFlowWrapper.propTypes = {
```
dozortsev, line 318: why we need this assignment change?

richardbata: So that the default export (Wrapper) has the same type interface as DynamicFlow component

JulianG-TW (approved):
Thank you!
It would be nice to get rid of the `injectIntl` hook as well. 
Probably in a future PR.

JulianG-TW (approved):

dozortsev (approved):

PR #1364, opened 2022-03-16
refactor(dynamic-flows): Avoid using HoC in ErrorBoundary component

JulianG-TW:
## 🖼 Context

This is not necessary. But wrapping components in HoC usually "ruin" their type declarations, preventing some useful type completion, and type-checking when using the wrapped component.

## 🚀 Changes

Because ErrorBoundary must be a class component, we extract the parts that needed localisation into a new ErrorBoundaryAlert function component that can use the `useIntl` hook.

## ✅ Checklist


packages/dynamic-flows/src/flow/errorBoundary/ErrorBoundaryAlert.tsx
```diff
@@ -0,0 +1,22 @@
+import { Alert, Sentiment } from '@transferwise/components';
+import { useIntl } from 'react-intl';
+
+import messages from './ErrorBoundary.messages';
+
+type Props = { onDismiss: () => void };
+
+export function ErrorBoundaryAlert({ onDismiss }: Props) {
```
dozortsev, line 8: well, I think it's better to refactor `ErrorBoundary` to be functional component or just even better, refactor it to TS

dozortsev (commented):
from PR description:

> ...But wrapping components in HoC usually "ruin" their type declarations, preventing some useful type completion, and type-checking when using the wrapped component.

[those rules](https://github.com/transferwise/neptune-web/blob/main/packages/components/CONTRIBUTING.md#js-component-rules) mostly make sense and applicable to components that are exposed from package, [while `ErrorBoundry` isn't](https://github.com/transferwise/neptune-web/blob/main/packages/dynamic-flows/src/index.ts)

so this change doesn't really add any value, unless you refactor the component to TS

packages/dynamic-flows/src/flow/errorBoundary/ErrorBoundaryAlert.tsx
JulianG-TW: I don't think it's possible to refactor the error boundary to a functional component.
According to the [React docs](https://reactjs.org/docs/error-boundaries.html): "Only class components can be error boundaries."

I think there's little benefit in converting it to TS while the DynamicFlow component is still on JS. I would like to convert both at some point. But this PR is only to remove the HoC. (and perhaps to fix the main branch build that I broke earlier)

JulianG-TW:
@dozortsev 

> so this change doesn't really make sense, unless you refactor the component to TS

oh! I see! you're right! for some reason I assumed that js modules would at least provide some intellisense or code completion when calling a TS React component. But it's not the case, at least not in this setup.

So I've went ahead and converted `ErrorBoundary` to TS. 😅

richardbata (approved):

packages/dynamic-flows/src/flow/errorBoundary/ErrorBoundary.tsx
```diff
@@ -0,0 +1,43 @@
...
+    this.setState({ hasError: false, isFatalError: false });
+  };
+
+  render() {
+    const { children } = this.props;
+    const { hasError, isFatalError } = this.state;
+
+    return (
+      <>
+        {hasError && <ErrorBoundaryAlert onDismiss={this.handleErrorReset} />}
```
dozortsev, line 36: since it's TS file now you could use `injectIntl` and have accurate types, so having `ErrorBoundaryAlert` is redundant

PR #1371, opened 2022-03-22
feat: markdown to optionally add target=_blank to links

JulianG-TW:
## 🖼 Context

Mostly in **dynamic-flows** we want an option to open links in a new tab to avoid taking users away from the flow when clicking on a "read more" link.

## 🚀 Changes

Add an option to `Markdown` component specifically for this.

## 🤔 Considerations

It would be nice to add more generic options but it's not really easy to modify the markup generation.
In fact the proposed implementation in this PR is a bit of a hack. **This is a draft.**

## ✅ Checklist


packages/components/src/markdown/Markdown.js
```diff
@@ -23,18 +24,17 @@ const Markdown = ({ as: Element, children, className, allowList, blockList }) =>
...
       return stripNodes({ parsed, blockList: toExclude });
     }
     return parsed;
   };
 
   const createMarkup = () => {
     const parsed = parser(children);
-    return writer.render(parsed);
+    const markup = writer.render(parsed);
+    return linkTargetBlank ? markup.replace(/<a href="/g, `<a target="_blank" href="`) : markup;
```
dozortsev, line 37: then use the config here

```suggestion
    return linkTargetBlank ? markup.replace(/<a href="/g, `<a target="${config.link.target}" href="`) : markup;
```

packages/components/src/markdown/Markdown.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable testing-library/render-result-naming-convention */
```
dozortsev, line 1: why we need disable rule for testing-library in this file?

packages/components/src/markdown/Markdown.js
```diff
@@ -107,13 +109,23 @@ Markdown.propTypes = {
       'thematic_break',
     ]),
   ),
+  config: PropTypes.shape({
+    link: PropTypes.shape({
+      target: PropTypes.oneOf('_blank' | '_self'),
```
dozortsev, line 114:
```suggestion
      target: PropTypes.oneOf(['_blank', '_self']),
```

dozortsev (approved):

PR #1379, opened 2022-03-28
feat(dynamic-flows): update default validation messages

JulianG-TW:
## 🖼 Context

When a schema doesn't provide validation error messages, the client provides some defaults.

Mobile clients have slightly different validation messages than the web client.

## 🚀 Changes

This PR updates some default validation messages to match those in mobile clients.

Before | After
---- | ----
before | after

## 🤔 Considerations

[According to the spec](https://api-docs.dev-wi.se/dynamic-flow#schemas-string-schema) `date` is a format for a `string` schema. But the `string` schema does not support `minimum` nor `maximum` properties.

## ✅ Checklist


packages/dynamic-flows/src/formControl/FormControl.js
```diff
@@ -264,7 +264,9 @@ export default class FormControl extends PureComponent {
           />
         );
 
-      case FormControlType.DATELOOKUP:
+      case FormControlType.DATELOOKUP: {
+        // const min = minDate || (minimum && new Date(minimum))
+        // const max = maxDate || (maximum && new Date(maximum))
```
JulianG-TW, line 269:
```suggestion
```

packages/dynamic-flows/src/flow/examples/date.json
```diff
@@ -0,0 +1,72 @@
...
+      "$id": "#form",
+      "allOf": [
+        {
+          "type": "object",
+          "properties": {
+            "date-one": {
+              "title": "Enter a date between 2000 and 2010",
+              "type": "string",
+              "placeholder": "YYYY-MM-DD",
+              "pattern": "^\\d{4}\\-\\d{2}\\-\\d{2}$",
```
StevePole, line 16: Do we need this?

packages/dynamic-flows/src/jsonSchemaForm/controlFeedback/ControlFeedback.js
```diff
@@ -72,13 +72,29 @@ ControlFeedback.defaultProps = {
...
+    maximum: formatMessage(controlFeedbackMessages.maximum, { maximum: schema.maximum }),
+    minLength: formatMessage(controlFeedbackMessages.minLength, { minLength: schema.minLength }),
+    maxLength: formatMessage(controlFeedbackMessages.maxLength, { maxLength: schema.maxLength }),
+    pattern: formatMessage(controlFeedbackMessages.pattern),
+    required: formatMessage(controlFeedbackMessages.required),
+  };
+
+  if (schema.format === 'date') {
+    formattedMessages.minimum = formatMessage(controlFeedbackMessages.minimumDate, {
+      minimum: schema.minimum,
```
StevePole, line 90: We could format here

JulianG-TW: Oh, I've pushed a commit before reading this. it now uses `toLocaleDateString()` like this:
```js
minimum: new Date(schema.minimum).toLocaleDateString(),
```

Did you have a different thing in mind?

packages/dynamic-flows/src/flow/examples/date.json
JulianG-TW: I made this storybook fixture to easily see what these 3 types of entering dates would look like.

I left it in the codebase because I think there's a bug where date-lookup allows users to select a date outside of the `minimum/maximum` range. But I didn't want to fix it in this PR because it's about validation **messages**.

JulianG-TW: Here it is:

The user can select a date outside of the range:
image

Then validation shows up, but it could be avoided, using `min/max` props of the `DateLookup` component.

image

packages/dynamic-flows/src/jsonSchemaForm/controlFeedback/ControlFeedback.js
dhananjaykha: We can use the standard neptune [formatting](https://github.com/transferwise/formatting).

JulianG-TW: Thank you! Done at 808d50c693799ecc3cab08367b26939f41736c9c

dhananjaykha (approved):

dhananjaykha:
Do we have these default messages the same on the other platforms too?
I had referred to [this document](https://docs.google.com/spreadsheets/d/18kFkuGJ0J7EXe-oEZsfsjOQR8y9ZCiZ2XpIohJZ2vZ4/edit#gid=0) previously when working on upload component.

Is there a standard collection of messages to refer to?

JulianG-TW:
> I had referred to [this document](https://docs.google.com/spreadsheets/d/18kFkuGJ0J7EXe-oEZsfsjOQR8y9ZCiZ2XpIohJZ2vZ4/edit#gid=0) previously when working on upload component.
that's what I used as reference

PR #1380, opened 2022-03-29
feat(components): update option and navigationOption components

striccaa:
## 🖼 Context

**Jira ticket**: https://transferwise.atlassian.net/browse/CSPROD-2546

In the `help-center` CRAB app, we often use the Next.js-specific `<Link />` component, which has a `passHref` property, and given it's enabled, it passes the href attribute to its child element.

The current implementation of the `NavigationOption` - which we often use wrapped in the `<Link>` component - is not able to receive/forward refs, and produces the following error, which is why I created this PR.
image-20200904-114208

In this PR, I'm wrapping the `NavigationOption` and the `Option` in `forwardRef`, and at the same time converting those components to TS. Also, I added the `Option` component to the package exports, due to the fact that we would like to use this generic component in the `help-center` repo.

## 🚀 Changes
 
- converted `Option` and `NavigationOption` components to TS
- wrapped `Option` and `NavigationOption` components with forwardRef
- added `Option` component to package exports
- updated `Option` test case

## ✅ Checklist


packages/components/src/common/Option/Option.tsx
```diff
@@ -0,0 +1,90 @@
...
+      decision = true,
+      disabled = false,
+      complex,
+      inverseMediaCircle = true,
+      className,
+      showMediaAtAllSizes,
+      showMediaCircle = true,
+      ...rest
+    },
+    reference,
```
richardmccartney, line 47: Should we look too set a type to the `reference` prop? This could look like `export type RefType = React.Ref<HTMLElement | HTMLAnchorElement>;`

richardmccartney (commented):
Great refactor of these components 👍 As we are now bring in `forwardRef` to both of this could we look to adding in some unit tests for both?

Would be ideal to check whether the reference value is `false` first and then `true` after setting the value.

_Example of a Button Ref test_
```
  it(': set `ref` to be true on Button', () => {
    const ref = createRef<HTMLButtonElement>();
    const props = {
      ref,
    };

    expect(ref.current).toBeFalsy();

    render(<Button {...props}>Click me!</Button>);

    expect(ref.current).toBeTruthy();
  });
```

striccaa:
> Great refactor of these components 👍 As we are now bring in `forwardRef` to both of this could we look to adding in some unit tests for both?
> 
> Would be ideal to check whether the reference value is `false` first and then `true` after setting the value.
> 
> _Example of a Button Ref test_
> 
> ```
>   it(': set `ref` to be true on Button', () => {
>     const ref = createRef<HTMLButtonElement>();
>     const props = {
>       ref,
>     };
> 
>     expect(ref.current).toBeFalsy();
> 
>     render(<Button {...props}>Click me!</Button>);
> 
>     expect(ref.current).toBeTruthy();
>   });
> ```

Thanks @richardmccartney, I appreciate it!
You've got a point there with regards to testing, however, I decided not to extend tests with tests that solely check the `ref` of the components due to the fact that we don't have any business logic that could alter or amend the ref that is received in the components, and at that point, we would only be testing whether react and/or `forwardRef` works the intended way.

I also couldn't think of a better test case that would really make sense to include. What do you think?

striccaa:
Update:
I marked this PR as a "breaking change" due to the fact that adding types to the `onClick` event handler as such:
```
onClick?: (event: MouseEvent<HTMLElement>) => void;
```
breaks a number of repos where, for example, the `<NavigationOption />` is used like this:
```
<NavigationOption onClick={onClickFunction}>...</NavigationOption>

# instead of using it like this:
<NavigationOption onClick={() => onClickFunction()}>...</NavigationOption>
```
These onClick properties will have to be updated accordingly, to adhere to the TS types.

cc. @dozortsev @richardmccartney

richard-wiser:
@richardmccartney @dozortsev Just picking this up as Adriano is on a holiday. Is there anything blocking this? 

- Do you agree with @striccaa 's assessment of the tests, or would you like those written?
- What would be a good strategy to release this? It's a breaking change, and a small one as well, so I think this would be a good candidate for the next major version.

richardmccartney:
@richard-wiser if we can still get the `ref` tests written, it's best we cover all tests cases where possible 👍 Once we have made the change, we can ensure we create the appropriate release notes and any upgrade documentation so others are aware in the web guild

packages/components/src/common/Option/Option.tsx
```diff
@@ -0,0 +1,92 @@
...
+  className?: string;
+  showMediaAtAllSizes?: boolean;
+  showMediaCircle?: boolean;
+}
+
+interface AnchorOptionProps extends BaseOptionProps {
+  as: 'a';
+  href: string;
+  target: '_self' | '_blank' | '_parent' | '_top';
+}
```
dozortsev, line 26: we this type defined `LinkProps` (you can import it from `commonProps.ts`)

dozortsev (approved):
Thank for contributing!

Note I would avoid releasing this as breaking change.

packages/components/src/common/Option/Option.tsx
dozortsev: in one of tests of this component on consumer side I noticed this weird types complain 🤔 

I added Link component just to test it's `target` prop (which btw based on type from `commonProps.ts`)

Screen Shot 2022-04-11 at 2 15 55 PM

striccaa: Could you please check if the type complaint is still there if you explicitly define the `as` attribute as so: `as: 'a'`?

striccaa: Could you pleas check if the type complaint still exists if you explicitly set the `as` attribute as so: `as: 'a'`?

dozortsev: yeah the warning disappears 👍 

but please address duplicate types for link https://github.com/transferwise/neptune-web/pull/1380#discussion_r847219922

striccaa: Sure thing, thanks!

richardmccartney (approved):

dozortsev (approved):

PR #1396, opened 2022-04-07
fix: style of list type should be none

dozortsev:
## 🖼 Context

Cards rendered in list (`<ul>`) have marker:

Screenshot 2022-04-01 at 15 30 20

Screenshot 2022-04-01 at 15 30 34

The issue is reproducible on docs https://transferwise.github.io/neptune-web/branch/fix-list-card/components/content/Card#group-of-cards

## ✅ Checklist


packages/css/src/less/list-group.less
```diff
@@ -11,8 +11,7 @@
 // Easily usable on <ul>, <ol>, or <div>.
 
 .list-group {
-  // No need to set list-style: none; since .list-group-item is block level
-
+  list-style-type: none;
```
dozortsev, line 14: I couldn't find how the bug was introduced to just fixing it this way

richardmccartney (approved):

PR #1398, opened 2022-04-07
refactor(dynamic-flows): Revert default parameter value in helper function

JulianG-TW:
## 🖼 Context

In https://github.com/transferwise/neptune-web/pull/1392 a default value was added for the parameter of `isSubmissionMethod`, for no apparent reason.

Considering that the `requestStep` function uses `GET` as a default when the action method is falsy (DynamicFlow.js line 60). I don't see a reason for the parameter to default to `POST`.

## 🚀 Changes

Revert the change, leaving the function parameter without a default value.

## 🤔 Considerations

We could have defaulted to `GET` but it would make no difference, AFAIK.

## ✅ Checklist


OFFLlNE:
This change was done due to Action documentation: https://transferwise.atlassian.net/wiki/spaces/DF/pages/1974569292/Actions

While locally testing the Decision Layout it was also failing when there was no default method given

PR #1424, opened 2022-05-03
[VUX-1582] DF Version Orchestrator

richardbata:
## Context
Following many discussion in the guild about the version orchestrator we are building for verification and Self-Service, here is the implementation of it for web.

More info [here](https://transferwise.atlassian.net/wiki/spaces/DF/pages/2369128718/DFv3+v2+Version+Orchestrator#Transition-between-versions-mid-flow) on how it works.

I put the code in its own package in neptune as suggested by the web guys in the guild.

## Checklist


packages/dynamic-flows/package.json
```diff
@@ -52,10 +52,19 @@
     "@transferwise/less-config": "^3.0.3",
     "@transferwise/neptune-css": "^12.5.0",
     "@transferwise/neptune-tokens": "7.0.0",
+     "@transferwise/ng-browser-info": "^1.0.1",
+    "@transferwise/styleguide-components": "^9.1.0",
     "@transferwise/test-config": "^5.0.0",
+    "@transferwise/verification-components": "^7.2.1",
+    "angular": "~1.5.10",
```
dozortsev, line 59: I don't think we want bring angular dependecy to neptune-web repo, why this angular 2 react mapping can't be done on consumer side of DF components?

dozortsev (changes requested):

packages/dynamic-flows/package.json
pterblgh: If you want to expose a React component that teams can "just import" then the only way forward is to convert the existing AngularJs based component to a React one in a library. Pushing all the logic to consumers about "How to convert AngularJs to React?" would only result in code-duplication all over the places where Dynamic Flows is used in React consumers IMO.

richardbata: But these are only devDependencies for the Storybook 👍

dozortsev: yeah but it still a dependency, and it needs to be maintained, version needs to be occasionally updated, to fix some conflicts in deps or solve security vulnerability

dozortsev: I would place this demo in https://github.com/transferwise/verification-components

packages/dynamic-flows/src/adapter/client/v2.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { DfVersion } from '../DFv2v3Adapter';
+import { isV3Response } from './utils';
+
+const GET_METHOD = 'GET';
+
+export const createCustomV2HttpClient = (
+  Dfv2Client: ReactElement<any>,
+  transitionVersion: (newVersion: DfVersion, transitionStep: any) => void,
+) => {
+  const apiUrl = Dfv2Client.props.apiUrl; // TODO taking from props is risky
```
richardbata, line 11: Is it safe to take values off props here? If not, what alternatives are there??

packages/dynamic-flows/src/adapter/client/v2.tsx
```diff
@@ -0,0 +1,68 @@
...
+          'Accept-Language': lang,
+          'X-Access-Token': 'Tr4n5f3rw153',
+        },
+        method: GET_METHOD,
+        credentials: 'same-origin',
+      });
+
+      return await response.json();
+
+      // return this.$http(request).then((response) => response.data);
```
richardbata, line 35: Safe to not use $http?

packages/dynamic-flows/src/adapter/typings.ts
```diff
@@ -0,0 +1,8 @@
+import { DynamicFlow } from '..';
+import { ReactElement } from 'react';
+
+export interface DFv2v3AdapterProps {
```
richardbata, line 4: Any tips on how to make types better?

packages/dynamic-flows/src/adapter/DFv2v3Adapter.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { createCustomV2HttpClient, createCustomV3Fetcher } from './client';
+
+export enum DfVersion {
+  v2 = 'v2',
+  v3 = 'v3',
+}
+
+const supportedVersions = [DfVersion.v3, DfVersion.v2];
+
+const DFv2v3Adapter: ComponentType<DFv2v3AdapterProps> = ({ flowUrl, Dfv3Client, Dfv2Client }) => {
```
pterblgh, line 14:
```suggestion
const DFv2v3Adapter: VFC<DFv2v3AdapterProps> = ({ flowUrl, Dfv3Client, Dfv2Client }) => {
```

packages/dynamic-flows/src/adapter/DFv2v3Adapter.tsx
```diff
@@ -0,0 +1,92 @@
...
+    });
+  };
+
+  const customV3Fetcher = createCustomV3Fetcher(Dfv3Client, supportedVersions, transitionVersion);
+
+  const customV2HttpClient = createCustomV2HttpClient(
+    Dfv2Client,
+    supportedVersions,
+    transitionVersion,
+  );
```
pterblgh, line 47: Can this be encapsulated with the memoized hook that produces`MyDfv2Client`?

packages/dynamic-flows/src/adapter/client/utils.ts
```diff
@@ -0,0 +1,65 @@
...
+  let headerValue = 'supported-versions=v2,v3';
+
+  if (method?.toUpperCase() === 'POST') {
+    headerValue += `; version=${dfVersion}`;
+  }
+
+  return {
+    [X_DYNAMIC_FLOW]: headerValue,
+  };
+};
```
pterblgh, line 19: Let's investigate whether we can use the "platform": https://developer.mozilla.org/en-US/docs/Web/API/Headers

packages/dynamic-flows/src/adapter/client/utils.ts
```diff
@@ -0,0 +1,65 @@
...
+const isResponseV3Shape = async (originalResponse: Response) => {
+  const response = originalResponse.clone();
+  const body = await response.json();
+
+  return (
+    Object.keys(body).filter((key) =>
+      ['type', 'key', 'schemas', 'model', 'actions', 'layout'].includes(key),
+    ).length >= 2
+  );
+};
```
pterblgh, line 37: Why do we need these methods? According to the proposal the API endpoint will always return what is the version of the currently returned flow.

packages/dynamic-flows/src/adapter/client/utils.ts
```diff
@@ -0,0 +1,65 @@
...
+export const isV2Response = async (response: Response) => {
+  const { headers } = response;
+
+  if (getVersionFromHeader(headers.get(X_DYNAMIC_FLOW)) === 'v2') {
+    return true;
+  }
+
+  if (await isResponseV2Shape(response)) {
+    return true;
+  }
```
pterblgh, line 48: Error handling is probably missing.

packages/dynamic-flows/src/adapter/client/v2.tsx
pterblgh: DfV2Client should not have any props, it's just going to be a function reference to a React component that's capable of rendering V2 flows isn't it? 

I would mark every necessary property form the consumer of the component to be able to create a HTTP client. 

According [to this discussion](https://transferwise.slack.com/archives/C2SGDKP4Y/p1650547825683819) we should use `@transferwise/service-comms`

packages/dynamic-flows/src/adapter/client/v2.tsx
pterblgh: Need to check the error handling in `verification-components` - IIRC the AngularJs component expects a `$http` style error object to be able to show error messages (validation, error messages from the server).

richardbata: What I am trying to achieve is that the DFv_Clients can keep the ability to have their own HttpClient, which the Adapter/Orchestrator would only extend, not completely replace.

packages/dynamic-flows/src/adapter/client/v2.tsx
richardbata: What I am trying to achieve is that the DFv_Clients can keep the ability to have their own HttpClient, which the Adapter/Orchestrator would only extend, not completely override.

richardbata: Actually, DFv3Client will be a React component already created, not a function. This is so the consumer of the Orchestrator can initialize the DF client with options it wants to

packages/dynamic-flows/src/adapter/client/v2.tsx
richardbata: I think all that is shown on the UI is an alert at the top, which takes an Error object to display
error-section: https://github.com/transferwise/verification-components/blob/master/src/components/error-section/error-section.html

packages/dynamic-flows-version-orchestrator/package.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "@transferwise/dynamic-flows-version-orchestrator",
+  "version": "1.0.0",
+  "description": "Wise Dynamic Flows in Multiple Versions",
```
dozortsev, line 4: why we need this package? what it does?

packages/dynamic-flows-version-orchestrator/docs/version_orchestrator.md
```diff
@@ -0,0 +1,73 @@
...
+was too domain specific. From there came the initiative the build a generic DF version
+orchestrator that initially Self-Service and verification can use, but one that potetially
+other teams within the firm could also utilize if a similar use case arises.
+
+## Wikis
+
+The Self-Service to V3 [wiki](https://transferwise.atlassian.net/wiki/spaces/VERXP/pages/2356674799/Self-services+to+DFv3) goes more in depth about the business case for the project.
+
+The DF proposal [wiki](https://transferwise.atlassian.net/wiki/spaces/DF/pages/2369128718/DFv3+v2+Version+Orchestrator) goes into more depth around the technical specs of how its works for all
+platforms.
```
dozortsev, line 22: this repository is open source, you shouldn't mention/expose links to Wise internal docs

packages/dynamic-flows-version-orchestrator/package.json
JulianG-TW: We discussed this in DFv3 Guild meeting. 

Considering that:
- The `dynamic-flows` package only knows about v3.
- Only a few crab apps need transitioning from V2 to V3 flow steps

We decided to have this orchestrator on a separate package so that only those who needed will add the dependency.

JulianG-TW: > by v2 you mean https://github.com/transferwise/verification-components right?

yes

> how @transferwise/dynamic-flows-version-orchestrator know about v2?

There's a `DynamicFlowV2Wrapper` component. There's no dependency AFAIK, but there's awareness of a v2.

dozortsev: I don't see why `DynamicFlowV2Wrapper` can't live `@transferwise/dynamic-flows`

dozortsev: the login in `@transferwise/dynamic-flows-version-orchestrator` looks like an abstraction over V2 and V3 but IMHO it's totally fine to keep it `@transferwise/dynamic-flows`, this way you eventually kill it faster

creating a separate package in this monorepo is a very big maintenance cost

PR #1445, opened 2022-05-11
feat(select): add aria-attributes and expose dropdownProps

jonathan-barnes-wise:
## Context

Follows on from #1443.

Exposes `dropdownProps` to allow for prop injection on internal `ul` DOM element and adds accessibility improvements to the select component. Follows implementation by [this component](https://www.radix-ui.com/docs/primitives/components/select) in the Radix UI component library and [select-only combobox demo on w3.org](https://www.w3.org/TR/wai-aria-practices/examples/combobox/combobox-select-only.html).

Please see [C20-2635](https://transferwise.atlassian.net/browse/C20-2635) for initial a11y finding (and remediation plan).

## Changes
 
- add `dropdownProps` prop to `Select` component
- add `aria-` and `role` attributes to DOM elements to improve a11y
- add consistent behaviour to focus button DOM element when dropdown closes or a selection is made

## Considerations / Results

- No visual changes
- There's possibly still room for improvement in the generated markup (see [w3.org example](https://www.w3.org/TR/wai-aria-practices/examples/combobox/combobox-select-only.html)), but that might require extensive changes so have kept things lean here
- Was unable to write a unit test to test the [focus button behaviour](https://github.com/transferwise/neptune-web/compare/C20-2635-improve-select-dropdown-a11y?expand=1#diff-29d0efb3fed45db7c0cd30b2bd9190372a2db9f970e9ee367c106d230e115665R286). Various googling and GitHub issues has led me to think it could be a quirk with focus behaviour in the JSDOM test environment. Have omitted the test for now.

## Checklist


packages/components/src/select/Select.js
```diff
@@ -337,7 +353,14 @@ export default class Select extends Component {
...
 
     return (
-      <ul className={dropdownClass} aria-labelledby={id}>
+      <ul
+        ref={this.optionsListRef}
+        id={`${id}-listbox`}
+        role="listbox"
+        tabIndex="-1"
+        className={dropdownClass}
+        {...dropdownProps}
```
dozortsev, line 362: what props would you need provide via `...dropdownProps`? do you have an example?

packages/components/src/select/Select.js
```diff
@@ -436,6 +459,10 @@ export default class Select extends Component {
       <li
         key={index}
         className={className}
+        aria-selected={option.header ? 'undefined' : isActive}
```
dozortsev, line 462: `aria-selected` has following types

```ts
'aria-selected'?: boolean | 'false' | 'true' | undefined;
```

so I think it's better pass there `undefined` rather than `'undefined'`

dozortsev: why we check here only for Headers (`option.header`)? we also have Separator options

jonathan-barnes-wise: > `aria-selected` has following types
> 
> ```ts
> 'aria-selected'?: boolean | 'false' | 'true' | undefined;
> ```
> 
> so I think it's better pass there `undefined` rather than `'undefined'`

great catch, thank you 👍  I think using a string here was accidental, will amend to `undefined`.

> why we check here only for Headers (`option.header`)? we also have Separator options, should we handle them here as well

and again, totally agree. If an option isn't directly selectable by the user it should also have `undefined` as the setting here.

jonathan-barnes-wise: Actually I've just realised, I don't think we need this check for `option.header` here, as the markup for headers and separators are returned earlier in the function.

[option.separator](https://github.com/transferwise/neptune-web/pull/1445/files/37a3129fd7f2802a698f1739c1090f66b4a056b8#diff-29d0efb3fed45db7c0cd30b2bd9190372a2db9f970e9ee367c106d230e115665R428)
[option.header](https://github.com/transferwise/neptune-web/pull/1445/files/37a3129fd7f2802a698f1739c1090f66b4a056b8#diff-29d0efb3fed45db7c0cd30b2bd9190372a2db9f970e9ee367c106d230e115665R432)

dozortsev: > ..If an option isn't directly selectable by the user it should also have undefined as the setting here.

try use [`actionableOption`](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/select/Select.js#L24-L26) here to verify this

dozortsev: yeah you right

jonathan-barnes-wise: have addressed the points 👍

dozortsev (approved):

PR #1446, opened 2022-05-11
feat(Button): expose `ref` prop

dozortsev:
## Context

I need `ref` prop in `Button` as part of https://github.com/transferwise/neptune-web/pull/1399.

This change mainly because another today's change to `Select` component https://github.com/transferwise/neptune-web/pull/1445 ([see 👁️  this line](https://github.com/transferwise/neptune-web/pull/1445/files#diff-29d0efb3fed45db7c0cd30b2bd9190372a2db9f970e9ee367c106d230e115665R521)).

## Changes

- expose React `ref` prop in `Button`
- refactor `Button` to TypeScript

## Checklist


packages/components/src/actionButton/ActionButton.tsx
```diff
@@ -1,12 +1,7 @@
...
-import { ButtonHTMLAttributes, MouseEvent, forwardRef } from 'react';
+import { ButtonHTMLAttributes, forwardRef } from 'react';
 
-type ActionButtonProps = {
-  className?: string;
-  disabled?: boolean;
-  text: string;
-  onClick: (event: MouseEvent<HTMLButtonElement>) => void;
-} & ButtonHTMLAttributes<HTMLButtonElement>;
+type ActionButtonProps = { text: string } & ButtonHTMLAttributes<HTMLButtonElement>;
```
dozortsev, line 4: all those props (`className`, `disabled` ...) are coming from `ButtonHTMLAttributes`

packages/components/src/button/Button.tsx
```diff
@@ -0,0 +1,94 @@
...
+      className,
+      disabled,
+      htmlType = 'button',
+      loading = false,
+      priority = Priority.PRIMARY,
+      size = Size.MEDIUM,
+      type = ControlType.ACCENT,
+      ...rest
+    }: Props,
+    reference,
```
richardmccartney, line 52: Can we add a test just to ensure `reference` is working and is `true`

```js
  it(': set `reference` to be true on Button', () => {
    const reference = createRef<HTMLButtonElement>();
    const props = {
      reference,
    };

    expect(reference.current).toBeFalsy();

    render(<Button {...props}>Click me!</Button>);

    expect(reference.current).toBeTruthy();
  });
```

richardmccartney (changes requested):
Let's just add a unit test to this component to test `ref` and we should be good 👍

packages/components/src/button/Button.tsx
dozortsev: added the test https://github.com/transferwise/neptune-web/pull/1446/commits/b4e87a50708d306c415a86fb218b12515ad7a916

richardmccartney (approved):

PR #1452, opened 2022-05-20
Adding types to dynamic-flow

richardbata:
## Context
Adding types to DynamicFlow.

## Changes
Many type fixes were needed in the component code to pass TS and linter checks.

## Considerations / Results
DynamicFlow has a couple of props where either or is required:
- either `baseUrl` or `fetcher`
- either `flowUrl` or `initialStep`

## Checklist


JulianG-TW (changes requested):
This is amazing! I've been wanting to do this for a long time!

Just a few non-practical comments:

We can use `unknown` instead of `any` in all (or most) of these instances.

I think we can also refactor the Action type from:

```ts
export type Action = unknown & {
  url: string;
  method: string;
  data?: any;
  exit?: boolean;
  result?: any;
};
```

to

```ts
export type Action = {
  url: string;
  method: string;
  data?: Record<string, unknown>;
  exit?: boolean;
  result?: Record<string, unknown>;
};
```

The reason I prefer `unknown` to `any` is that, while both options let us get away without defining the type, `unknown` is stricter.

In some cases changing `any` to `unknown` result in other type errors down the line, but in this case (probably because everything else is JS) nothing seems to break. Given the choice, I think I always prefer `unknown`.

Also, is there a reason for `Action` to be `unknown & {...}` ?
For what I understand there's no difference with or without the `unknown`

Thanks for doing this!!

richardbata:
Thanks for the pointers, and no worries on doing the change 👍 

Agree with `unknown` over `any`, and also `unknown & {...}` makes no difference 👍

JulianG-TW (approved):
Thank you!!

PR #1454, opened 2022-05-23
fix(dynamic-flows): use controlType when mapping schemas to options

JulianG-TW:
## Context

When mapping const schemas to be used in oneOfs, the mapping function was only taking Neptune radio buttons into account. Icons and images would not render at all if the control type  was "select".

## Changes

property | Before | After
--- | --- | ---
using "bank" and "emoji" icon names | before | icon-after
using [00b9ff](https://placeholder.pics/svg/96/00b9ff/FFFFFF) and [b900ff](https://placeholder.pics/svg/96/b900ff/FFFFFF) placeholder image urls | before | image-after

## Checklist


packages/dynamic-flows/src/jsonSchemaForm/schemaFormControl/optionMapper.js
```diff
@@ -5,21 +5,53 @@ import DynamicIcon, { isValidIconName } from '../../layout/icon';
...
+    case 'radio':
+    default:
+      return {
+        label: schema.title,
+        value: !isUndefined(schema.const) ? schema.const : schema.enum[0],
+        ...getOptionDescription(schema.title, schema.description),
+        ...mapIconToAvatar(schema.icon),
+        ...mapCurrency(schema.icon),
+        ...mapAvatar(schema.image),
+        ...getDisabled(schema.disabled),
```
jonstieglitz, line 29: any way to reduce the duplicate code here?

JulianG-TW: Something like this?
```ts
  return {
    label: schema.title,
    value: !isUndefined(schema.const) ? schema.const : schema.enum[0],
    ...getOptionDescription(schema.title, schema.description),
    ...(controlType === 'select' ? mapIcon(schema.icon) : mapIconToAvatar(schema.icon)),
    ...mapCurrency(schema.icon),
    ...(controlType === 'select' ? mapImage(schema.image) : mapAvatar(schema.image)),
    ...getDisabled(schema.disabled),
  };
```

jonstieglitz: Looks better to me. Maybe stack the ternary operators at the bottom?

PR #1455, opened 2022-05-26
fix(dynamic-flows): Select the first option by default on oneOf schemas instead of null when children are object schemas

JulianG-TW:
## Context

In order to have parity with mobile clients, we need to always select one option in a oneOf schema of objects.

## Changes

Before | After
--- | ---
before | after
Initially no option is selected; forcing the user to select one | We select the first option by default when the children are object schemas.

## Checklist


StevePole (changes requested):
I don't agree with this change.  This is how it was _originally_ designed, but at some point we changed it to this behaviour because we had a scenario where people needed to make a selection.  If presented with a default a subset of people would just leave it on the default even if incorrect.

The way that it is currently, means that you can supply a model value if you want to default on load, but if you want to force a decision, you just avoid specifying the default value.

JulianG-TW:
Closing in favour of https://github.com/transferwise/neptune-web/pull/1456

PR #1458, opened 2022-05-27
refactor: widen react-intl peer dependency range

kripod-wise:
## Context

According to its changelog, react-intl@6 seems to contain no breaking changes for React:

https://github.com/formatjs/formatjs/blob/main/packages/react-intl/CHANGELOG.md#600-2022-05-19

## Changes

 

Widened `peerDependencies` range for `react-intl` to support a new major version.

## Considerations / Results

N/A

## Checklist


packages/components/package.json
```diff
@@ -88,7 +88,7 @@
     "prop-types": "^15.7.2",
     "react": "^16.14 || ^17",
     "react-dom": "^16.14 || ^17",
-    "react-intl": "^5.10.0"
+    "react-intl": "^5.10.0 || ^6.0.0"
```
dozortsev, line 91: could you please add same in DF package https://github.com/transferwise/neptune-web/blob/main/packages/dynamic-flows/package.json#L70

dozortsev, line 91:
```suggestion
    "react-intl": "^5.10.0 || ^6"
```

maybe this actually would be better?

dozortsev (approved):

packages/components/package.json
kripod-wise: Done, thanks for your feedback 😊

PR #1465, opened 2022-06-06
feat(dynamic-flows): new External Step

JulianG-TW:
## Context

The new External Step opens the specified URL in a new browser tab, and starts polling on another URL to know when the external user activity has completed.

## Changes

image

This PR introduces a new type of step: `external`, which doesn't support layouts. and instead displays `title` and `description` and a `retry/reopen` link button.

## Checklist


packages/dynamic-flows/src/layout/external/usePolling.tsx
```diff
@@ -0,0 +1,63 @@
...
+      ? function poll() {
+          if (isPolling && attemptCount < maxAttempts) {
+            attemptCount++;
+
+            asyncFn()
+              .then((pollingResponse) => {
+                isPolling = onPollingResponse(pollingResponse);
+              })
+              .catch(() => {
+                consecutiveFailures++;
```
richardbata, line 35: Do you want consecutive failure or just failures? If consecutive, you might want to reset counter on `asyncFn ` success

packages/dynamic-flows/src/layout/external/types.ts
```diff
@@ -0,0 +1,34 @@
...
+  result: string;
+  data?: Record<string, unknown>;
+};
+
+export type ResponseHandler = {
+  result: string;
+  action: Action;
+};
+
+export type Action = {
```
richardbata, line 28: Could we use the same action type in /flow/DynamicFlow?

packages/dynamic-flows/src/layout/external/DynamicExternal.tsx
```diff
@@ -0,0 +1,40 @@
...
+import messages from './DynamicExternal.messages';
+import { useIntl } from 'react-intl';
+
+type Props = {
+  component: ExternalConfig;
+  onAction: (action: Action) => void;
+};
+
+export const DynamicExternal = ({ component, onAction }: Props) => {
+  const { requestUrl, responseHandlers, polling, retryTitle } = component;
```
richardbata, line 16: Just a question, but why not reuse the simple `title` key instead of `retryTitle`?

packages/dynamic-flows/src/step/layoutStep/layoutService/convertStepToLayout.js
```diff
@@ -4,7 +4,7 @@ export function convertStepToLayout(step) {
   }
 
   // This condition is meant as a temporary hack until mobile platforms will be refactored to fully support the layout property
-  if (step.layout) {
+  if (step.layout && step.type !== 'external') {
```
richardbata, line 7: :/

packages/dynamic-flows/src/layout/external/DynamicExternal.tsx
JulianG-TW: We need 3 texts: `title`, `description` and `retryTitle`:
image

packages/dynamic-flows/src/step/layoutStep/layoutService/convertStepToLayout.js
JulianG-TW: Good point. I don't like this either.
This was meant to allow for a `layout` property that is subsequently ignored.
But the Decision Step doesn't support a `layout` either, since adding one would result in a return from line 8 instead of the call to `convertDecisionStepToDynamicLayout` in line 15.
So I suppose we don't need this. We can explicitly not support `layout` in External Step, just like in Decision Step.

packages/dynamic-flows/src/layout/external/usePolling.tsx
JulianG-TW: oh! you're right! that's a bug!

JulianG-TW: it should be fixed now. I also added some tests for this.

packages/dynamic-flows/src/step/layoutStep/layoutService/convertStepToLayout.js
JulianG-TW: done!

PR #1482, opened 2022-06-14
feat(dynamic-flows): Display global errors using an alert

JulianG-TW:
## Context

After submitting a form we can get an error response with validation (field errors), but we can also get a global error.

In recipient creation (custom code) we display this error on top of the `JsonSchemaForm` component.

We need to support this in this code base so we can migrate to use `DynamicFlow` directly.

## Changes

Now the `DynamicFlow` component keeps a a `globalError` state which is used to render a `DynamicAlert` component on top of the layout.

This `globalError` state is cleared once the model is changed.

global-error-4

## Considerations / Results

We had to fix the initialisation of `ObjectSchema` to avoid unnecessary calls to `onChange`. This was resulting in extra calls to `onModelChange` which in turn would clear the global error.

## Checklist


packages/dynamic-flows/src/flow/DynamicFlow.tsx
```diff
@@ -322,7 +332,8 @@ const DynamicFlowComponent: VFC<DynamicFlowV3Props> = ({
       stepSpecification,
       submitted,
       model: combineModels(models),
-      errors: validations,
+      fieldErrors: fieldErrors,
```
richardbata, line 335: Can simplify

packages/dynamic-flows/src/flow/DynamicFlow.tsx
```diff
@@ -137,7 +138,8 @@ const DynamicFlowComponent: VFC<DynamicFlowV3Props> = ({
   const [refreshing, setRefreshing] = useState(false);
   const [loading, setLoading] = useState(!initialStep);
   const [submitted, setSubmitted] = useState(false);
-  const [validations, setValidations] = useState<Errors>();
+  const [fieldErrors, setFieldErrors] = useState<FieldErrors>();
```
richardbata, line 141: fieldErrors vs formErrors

packages/dynamic-flows/src/step/layoutStep/layoutService/convertStepToLayout.js
```diff
@@ -229,10 +241,11 @@ function isWideForm() {
   return false;
 }
 
-function addMissingTitleAndDescriptionToStep(step) {
+function addMissingTitleAndDescriptionToStep(step, globalError) {
   return [
     ...(step.title ? [convertStepTitleToDynamicHeading(step.title)] : []),
     ...(step.description ? [convertStepDescriptionToDynamicParagraph(step.description)] : []),
+    ...(globalError ? [convertGlobalErrorToAlert(globalError)] : []),
```
richardbata, line 248: Does the globalError alert need to be bellow the title and description? In v2 I believe it was at the very top of the screen. I ask because if it can be at the top, we can simplify a lot of this logic, and put the <Alert> in DynamicFlows JSX even.

packages/dynamic-flows/src/jsonSchemaForm/objectSchema/ObjectSchema.js
```diff
@@ -16,7 +16,9 @@ const getSchemaColumnClasses = (width) => {
 };
 
 const ObjectSchema = (props) => {
-  const [model, setModel] = useState({ ...(props.model || {}) });
```
richardbata, line 19: Probably a stupid question, but does the error changes affect ObjectSchema?

packages/dynamic-flows/src/flow/DynamicFlow.tsx
```diff
@@ -230,8 +233,13 @@ const DynamicFlowComponent: VFC<DynamicFlowV3Props> = ({
   const handleFetchError = async (response: Response) => {
     try {
       const errorBody: ErrorStep = await response.json();
-      if (errorBody?.validation) {
-        setValidations(errorBody?.validation);
+      if (errorBody?.validation || errorBody?.error) {
+        if (errorBody?.validation) {
+          setFieldErrors(errorBody?.validation);
+        }
+        if (errorBody?.error) {
```
richardbata, line 240: Does the DF spec support such a global error yet?

JulianG-TW: Yes. It does. See https://api-docs.dev-wi.se/dynamic-flow#features-error-handling

packages/dynamic-flows/src/jsonSchemaForm/objectSchema/ObjectSchema.js
JulianG-TW: Fair question. I made this change in a [separate commit](https://github.com/transferwise/neptune-web/pull/1482/commits/f3ad1236ac4fcab853ce3fc5f2d2e8eac10a52dd) with a message. But now I see it's not super clear:

> fix(dynamic-flows): Fix ObjectSchema model initialisation
> This is to avoid an initial call to props.onChange when the useEffect is triggered.

The problem was that ObjectSchema would always trigger `onChange` after initialisation. 
So for steps with more than one schema, after a a global error response was received, DF would get its `onModelChange` called and therefore the global error alert would be immediately dismissed.

packages/dynamic-flows/src/flow/DynamicFlow.tsx
JulianG-TW: I'm not sure I understand. Are you suggesting we rename `fieldErrors` to `formErrors`?

packages/dynamic-flows/src/flow/DynamicFlow.tsx
JulianG-TW: fixed!

packages/dynamic-flows/src/flow/DynamicFlow.tsx
richardbata: Yeah, thats what I meant. Up to you ofc

packages/dynamic-flows/src/step/layoutStep/layoutService/convertStepToLayout.js
richardbata: Can check with Design Systems if we want to be sure. In terms of code, having it above all would simplify things a little

packages/dynamic-flows/src/jsonSchemaForm/objectSchema/ObjectSchema.js
richardbata: Are we sure we want to hide global error when editing fields? Isn't that more the behaviour of fieldErrors?

packages/dynamic-flows/src/flow/DynamicFlow.tsx
JulianG-TW: Makes sense. Thank you!

packages/dynamic-flows/src/step/layoutStep/layoutService/convertStepToLayout.js
JulianG-TW: With your `Unexpected error` in the passport screen, I have now seen 4 screens with global error:
- 2 with the error on top of the title
- 1 with the error on top of the form (and there's no title)
- 1 with the error in between the title and the rest of the step

I think it makes sense to have it on top of everything for now. it makes the code simpler. I'll change it as soon as I can.

JulianG-TW:
> A really minor thing, but I feel like the alert need a little bottom margin between it and the title

I have pushed 2baeaee5c9ff44d33dc200d074155d579bec8a18 adding a margin.

richardbata (approved):

PR #1485, opened 2022-06-15
fix(dynamic-flows): fix input value when model is an empty object

JulianG-TW:
## Context

The following schema renders just fine:

```json
  "schemas": [
    {
      "$id": "#form",
      "allOf": [
        {
          "type": "object",
          "properties": {
            "some-string": {
              "type": "string",
              "title": "Please, type any string"
            }
          }
        }
      ]
    }
  ]
```
ok

But if the string schema is not part of an “object”, like this:

```json
  "schemas": [
    {
      "$id": "#form",
      "allOf": [
        {
          "type": "string",
          "title": "Please, type any string"
        }
      ]
    }
  ]
```
Then the text field is initialised with “[object Object]” as a value.

image

This is not a placeholder. The user can freely edit it.

## Changes

When rendering `<input>`, `<InputWithDisplayFormat>` or `<textarea>`, if the `value` property is an object, we change it to `null`.

## Checklist


AllanHasegawaTw (approved):
LGTM! 🚀

PR #1493, opened 2022-06-24
Add support for DF icons with `text` (initials) without breaking `Select` `Option`

JulianG-TW:
## Context

The DFv3 specification for **Icons** states that the icon object may contain a `name` or a `text` property. Until now we only supported `name`, which rendered an icon.

## Changes

We now support `text` and render an `Avatar` component of type `initials`.

In addition we also render icons wrapped in an `Avatar` component, this time, of type `icon`. This is to align with the [Neptune guidelines](https://neptune.wise.com/components/navigation-option/). See also [Decision](https://transferwise.github.io/neptune-web/components/navigation/Decision#Decision) and [Avatar](https://transferwise.github.io/neptune-web/components/content/Avatar#Avatar).

### Before

before

### After

after

**Text icons were not supported in "radio" buttons nor in dropdowns**

Screenshot 2022-06-24 at 11 10 43

## Checklist


packages/dynamic-flows/src/jsonSchemaForm/schemaFormControl/optionMapper.js
```diff
@@ -61,6 +62,11 @@ export const mapIconToAvatar = (icon) => {
       ),
     };
   }
+  if (icon?.text) {
+    return {
+      avatar: <Avatar type={AvatarType.INITIALS}>{icon.text}</Avatar>,
```
fergusjordantw, line 67: Will the text always fit in an Avatar as one or two characters? Or should we be splitting the string into initials for safety?

fergusjordantw: I guess it's fine to leave up to whoever is passing down the schema

packages/dynamic-flows/src/layout/decision/index.js
```diff
@@ -1,14 +1,27 @@
...
+        <DynamicIcon type={option.icon.name} />
+      </Avatar>
+    );
+  }
+  if (option.icon?.text) {
+    return <Avatar type={AvatarType.INITIALS}>{option.icon.text}</Avatar>;
+  }
+  if (option.image?.url) {
     // eslint-disable-next-line jsx-a11y/img-redundant-alt
     return <img src={option.image?.url} alt={option.image?.text} />;
```
fergusjordantw, line 26: Out of scope of this change but should this also be thumbnail avatar? Any reason it isn't?

fergusjordantw (approved):
Didn't look at the schema changes in detail

packages/dynamic-flows/src/jsonSchemaForm/schemaFormControl/optionMapper.js
JulianG-TW: Exactly. Neptune Web Avatar doesn't care if the text fits either.
It's up to the service generating the schema.
image

packages/dynamic-flows/src/layout/decision/index.js
JulianG-TW: Definitely out of scope. We could use Avatar here as well:
image

But I need to check what mobile clients are doing. I've seen this used for brands/logos, and I'm not sure we can assume the can be masked into a circle.

JulianG-TW:
Thanks for reviewing @fergusjordantw !

AllanHasegawaTw (approved):

PR #1494, opened 2022-06-24
WIP

JulianG-TW:
## Context

Forms rendered using `DynamicFlow` do not contain an HTML `<form>` element, and therefore cannot be submitted by using the `ENTER` key. Users **MUST** click one of the action buttons in order to submit theform.

## Changes

In the `DynamicForm` component we now wrap `JsonSchemaForm` in a `<form>` element. When the user attempts to submit the form by hitting `ENTER` the `onSubmit` callback is handled by `DynamicLayout` which looks through all the components, finds the first primary action, and triggers `onAction` with it.

https://user-images.githubusercontent.com/67635620/175551823-20800bd7-e3e4-4070-a5de-ff5afb06f5e8.mov

## Considerations / Results

We're traversing the layout "component" tree in order to find the the first `primary` button, if none are present, we try to find the first `secondary` button, if that fails, then the first button component.

## Checklist


packages/dynamic-flows/src/layout/form/index.js
```diff
@@ -13,14 +13,23 @@ const DynamicForm = (props) => {
...
+      >
+        <JsonSchemaForm
+          schema={form.schema}
+          model={props.model}
+          submitted={props.submitted}
+          errors={props.errors}
+          onChange={onChange}
+          onPersistAsync={props.onPersistAsync}
+        />
+        <input type="submit" hidden />
```
StevePole, line 31: I wonder if we would be better adding the form at page level, and then making the action button a submit?

JulianG-TW: @StevePole my concern with rendering some action buttons as `<input type="submit" ...` is that AFAIK the input element would need to be styled properly so that it looks like a Neptune button:

```jsx
<form method="POST">
  <Button size="md" priority="primary">
    Button
  </Button>
  vs.
  <input
    type="submit"
    value="Input Type Submit"
    className="btn btn-md np-btn np-btn-md btn-accent btn-priority-1"
  />
</form>
```

We would need to maintain the value of `className` whenever `@transferwise/components` is updated.

PR #1513, opened 2022-07-05
Remove enum schema functionality

JulianG-TW:
## Context

We are dropping support for enum schema, in favour of using a oneOf const schemas.

## Changes

Surgically extract `enum` schema and all of its tests.
Also rename some test schemas and validation error to avoid using the "enum" keyword.

## Considerations / Results

What could possibly go wron?

## Checklist


packages/dynamic-flows/src/common/requirements.js
```diff
@@ -1,40 +1,6 @@
 /* eslint-disable no-unused-vars */
 /* eslint-disable no-param-reassign */
 
-const prepRequirements = (alternatives) => {
```
phellipealexandre, line 4: Does this code has something to do with the enums?

phellipealexandre (approved):

packages/dynamic-flows/src/common/requirements.js
JulianG-TW: No. except that it contains the string "enum" in it.
I found it while searching for `enum` and I discovered that it was an unused function.
[The first commit in this branch](https://github.com/transferwise/neptune-web/pull/1513/commits/a23abfa4293f9843f91aec3fca05ff9cfba70c72) is just removing this function.

PR #1514, opened 2022-07-05
fix(dynamic-flows): Fix timezone issue when using DateLookup

JulianG-TW:
## Context

**There's a problem when using the DateLookup component in DF when the client system is in a negative timezone offset because we store its value as a string.**

1. Change your computer timezone anywhere in America.
2. Open [this codesandbox](https://codesandbox.io/s/df-datelookup-issue-r5bix7) explaining the problem: 
3. Select a date using the example on the left. It should be correct.
4. Select a date using the example on the right. It should display the wrong date.

British Summer Time | Mountain Daylight Time
--- | ---
date-lookup-before-bst | date-lookup-before-mdt
✅ any timezone on UTC or with a positive offset | ❌ any timezone with a negative offset

### Why this happens

When a user makes a selection, the `DateLookup` component creates a `Date` object with the specified year, month and day at 00:00:00 in UTC timezone. This represents a specific point in time. Not a whole day, in the sense of a date of birth, for example.

For example, when a user selects the 6th of July, 2022 in a system set to Pacific Time (which is 7 hours behind UTC), the resulting Date object is `2022-07-06T07:00:00.000Z` (note it's 7am, not midnight). Because the object represents 2022-07-06 at 00:00 in UTC.

When using this component in a controlled way, we keep a React state and pass a value prop and a onChange callback prop. If those are Date objects everything works fine.

The problem is that in Dynamic Flow, we keep the React state as a string in the YYYY-MM-DD format, and we can't change that. So, depending on how we convert the `Date` object to `string` and back to `Date`, we get different results.

## Changes

This PR changes the way we convert the date to a string, by omitting the hours, minutes, seconds in order to work around this problem.

The existing code was doing this:

```js
const dateStringToDate = (dateString) => (dateString ? new Date(dateString) : null);
```

The new code omits only uses the year, month and date, ignoring hours, minutes and seconds.

```js
const dateStringToDate = (dateString) => {
  if (!dateString) {
    return null;
  }
  const [year, month, date] = dateString.split('-');
  return new Date(year, month - 1, date);
};
```

Now the correct date is displayed after selecting.

Before | After
--- | ---
date-lookup-before-mdt | date-lookup-after-mdt
❌ Mountain Daylight Time | ✅ Mountain Daylight Time

## Checklist


sbrandwoo:
When you finish with the component, what value do you want to have in the rest of your app? Do you want `2022-07-07` or
`Thu Jul 07 2022 00:00:00 GMT-0300`?

For regular statements we actually want the timezone, so we want the latter, we forward the timezone to the server and it uses it to display everything in the user's timezone.

For tax statements we just want the date, so we want the former. But JavaScript dates don't actually store the timezone on them, they are just points in time - the timezone only appears when you print them out. A European date of `Jul 07 2022 00:00:00 GMT+0300` is actually `Jul 06 2022 21:00:00 UTC` and if we send that to the server then it will be the wrong date. So in theory we need to "fix" the day by adding 3 hours onto it to get the correct UTC time before sending to the server.

Is the correct fix to keep storing `2022-07-07` and fix how this is rendered to the user?

JulianG-TW:
> When you finish with the component, what value do you want to have in the rest of your app? Do you want 2022-07-07 or
Thu Jul 07 2022 00:00:00 GMT-0300?

Good question. The code clearly keeps only `YYYY-MM-DD`. In DF date and times are string schemas with an additionally "format" property which can be `'date' | 'date-time' | 'date-lookup'`. 

The spec says:

Format | Example Value | Description
-- | -- | --
date | `2018-11-13` | An ISO-8601 format date containing only day, month, and year.
date-time | `2018-11-13T20:20:39+00:00` | An ISO-8601 format date containing day, month, year, and time.

It doesn't mention `date-lookup` but I assume it would be `YYYY-MM-DD`.

JulianG-TW:
**UPDATE**:

We're removing `date-time` format from the spec.

This means we can assume the `date-lookup` control is being used to specify a date without time component. This is a string with the `YYYY-MM-DD` format.

This means the current PR is correct.

When/if we add `date-time` back to the spec, we'll may need to consider this again.

sbrandwoo (approved):

JulianG-TW:
> As noted, the component only deals with dates, and not times, so involving timezones at all seems an unnecessary complication.

I agree. I think it makes sense to use string for dates such as date of birth.

I was about to suggest #neptune-web to consider using string, but then @sbrandwoo commented with a use case (outside of DF) where the time (or at least the timezone) is important.

Because DF currently only supports `date` and not `date-time` format, I think it’s ok to fix the problem in this PR without affecting the Neptune component.

Perhaps in the future the Neptune component could be improved to explicitly support only dates (as a string) or Date objects.

phellipealexandre (approved):
Approving (with auto-merge disabled) to unblock Julian

PR #1520, opened 2022-07-11
Refactor OneOfSchema (and related files)

JulianG-TW:
## Context

Before changing behaviour I'd like to refactor OneOfSchema.js

## Changes

It's definitely easier to review each individual commit: 

**[Refactor functions to be about values rather than model or default](https://github.com/transferwise/neptune-web/pull/1520/commits/1eb4147a96c54a5d1ae80340eb2771580cc97bce)**
There were very similar functions that would validate a schema against a model or a schema against its default value. Those can  be generalised without adding complexity, so we can later used them to check a schema against both a model and their own default value. For instance: `getValidIndexFromModel` becomes `getValidIndexFromValue`, and then we can delete `getValidIndexFromDefault`.
  
**[Move function to module scope](https://github.com/transferwise/neptune-web/pull/1520/commits/8b8b9a2af4d7321039601028d8037a06434d0afa)**
The diff in this commit doesn't reflect how trivial this change was. The `getActiveSchemaIndex` function defined inside the `OneOfSchema` function component can be moved outside of the component, to the module scope. This is analog to making a method static when it doesn't access any of the instance members. In JavaScript, this means the function is now only defined one, and not on every re-render. In addition, the function can now be moved below the component, as we do in the following commit.
  
**[Move module-scoped functions to below the React component function](https://github.com/transferwise/neptune-web/pull/1520/commits/d94226f5e3a3077ab842fcec1ae0b7a7e770f9a9)**
The main purpose of the OneOfSchema.js file is to define the `OneOfSchema` React component. In this file were a lot of module-scoped functions defined **above** the component definition. These functions are helper or ancillary functions which I prefer having below the component definition. (In the same way a class may define its public methods above its private ones). I think this improves readability of the module.
  
**[Extract some functions to their own files. Renamed a file](https://github.com/transferwise/neptune-web/pull/1520/commits/9676be074f06856ef234b455228cf2ad438b543b)** 
In this commit I extracted `getActiveSchemaIndex` and `isConstSchema` to their own files. This reduces the size of the main React component file and makes dependencies between functions more clear.
I also renamed OneOfSchemaModelMatcher.js to getBestMatchingSchemaIndexForValue.js since that is the name of the main exported function in that file. There was no entity called `OneOfSchemaModelMatcher` in the codebase.
  
**[Convert arrow functions to function declarations. Rename function. Reorder functions.](https://github.com/transferwise/neptune-web/pull/1520/commits/9201eb282445e6f9557ac52313cb66db35c52e45)** 
Finally I converted those module-scoped arrow functions into function declarations. This is because some linters may complain that the arrow functions are defined after they're used. By using function declarations, we can take advantage of [function hoisting](https://developer.mozilla.org/en-US/docs/Glossary/Hoisting).

**All these changes required no changes to the tests.**

## Checklist


AllanHasegawaTw (approved):
Looks good 🚀

PR #1522, opened 2022-07-12
test(dynamic-flows): Refactor OneOfSchema test suite

JulianG-TW:
## Context

I was having trouble reading this test suite, mostly because all the properties used to shallow render or mount the component under test were being mutated in `beforeEach` functions in different describe blocks.

## Changes

We avoid mutating props by generating them every time. Also we render the component in each test rather than in `beforeEach` functions which makes it easier to understand each test without having to scroll up and down all the time.

## Considerations / Results

Refactoring tests is always risky. We don't have tests for the tests. The fact that they're all green, doesn't mean a mistake wasn't made. I recommend reviewing this PR by using the split view and hiding whitespace, which makes it easier to review each test to ensure their logic remains intact.
image

## Checklist


AllanHasegawaTw (approved):
Noice 🚀 Looks clean 💅

PR #1535, opened 2022-07-22
Mc 2356  button margin

JulianG-TW:
## Context

The `margin` property in a layout button component was not being used.

The code was trying to use the `action.margin` instead.

## Changes

We now use the `component.margin`.

## Considerations / Results

We need TypeScript types and interfaces for all of the DF entities. This wouldn't have happened in TS, since `action.margin` would have been caught as a type error at build time, or even highlighted as an error in the IDE.

## Checklist


mSobhy90:
Does this need to be rebased since the icons one was merged?

packages/dynamic-flows/src/layout/button/index.js
```diff
@@ -46,13 +46,16 @@ const DynamicButton = (props) => {
 
   const { refreshing } = useDynamicFlow();
 
+  const className =
+    getMarginBottom(component.margin || 'md') + ' ' + getActionClasses(component.action);
```
mSobhy90, line 50: 🤔 This changes the behaviour from the original logic by adding the space to all results of the `getActionClasses` instead of just the default ` btn-primary` or am I missing something?

mSobhy90 (approved):
Just a small comment 🚀

packages/dynamic-flows/src/layout/button/index.js
JulianG-TW: it does, but it makes no difference. This string ends up in an HTML's `class` attribute. Different CSS classes are separated by a space. An extra space at the end makes no difference.

PR #1540, opened 2022-07-27
fix: adding correct accent tertiary button styles when disabled

richardmccartney:
## Context

This resolves a visual bug where our tertiary button, when disabled, was presenting the incorrect designs.

## Changes

- Made CSS changes to `buttons.less` that includes the override styles for `accent` buttons when `disabled`

## Considerations / Results

- Looking further into this, our CSS logic for buttons isn't overriding the correct styles, and we only use `greyscale` to affect the component. We will need to look at how to refactor this component in the future.
- Resolves bug 🐛 [#2473](https://transferwise.atlassian.net/browse/DS-2473?atlOrigin=eyJpIjoiZTRlMjljMmY0ZjYzNGViNjlkNzYyNzljMzBhNDdjYzIiLCJwIjoiaiJ9)

## Checklist


helendurranttw (approved):
Agreed that this needs refactoring - the colour isn't correct. But it's an improvement so approved!

PR #1541, opened 2022-07-28
fix(dynamic-flows): revert numeric-string control type

JulianG-TW:
## Context

The original intention was to make mobile browsers display a numeric keyboard, but marking an `input` as `type="number"` has a few undesirable side-effects:
- scrolling over the field changes its value
- display patterns don't work (e.g. 11-11-11)

## Changes

The scroll issue on number and integer schemas is fixed by adding a onWheel prop to the input.
At the moment we don't have a way to keep the display patterns so we're reverting the change.

## Checklist


packages/dynamic-flows/src/formControl/FormControl.js
```diff
@@ -226,6 +225,7 @@ export default class FormControl extends PureComponent {
             onBlur={this.handleOnBlur}
             onChange={this.handleOnChange}
             onFocus={this.handleOnFocus}
+            onWheel={(event) => event.target.blur()}
```
fergusjordantw, line 228: Can we add a test that this does what we want it to?

JulianG-TW: I have tried focusing the element and then firing a "wheel" event like this:
```js
    fireEvent.focus(input);
    fireEvent.wheel(input, { deltaY: -2 });
```
The `onWheel` callback is triggered, but the value of the input field does not change in jest-dom.

I also tried adding more properties to the event, but nothing seems to make a difference:
```js
    fireEvent.focus(input);
    fireEvent.wheel(input, {
      deltaY: -2,
      eventPhase: 3,
      offsetY: 1,
      isTrusted: true,
      currentTarget: global.window,
      target: input,
      srcElement: input,
    });
```

fergusjordantw: Oh well, it was worth a shot. Thanks for trying 🙇

fergusjordantw (approved):

eddiewoodleytw (approved):

PR #1542, opened 2022-07-29
feat!(dynamic-flows): Add loaderConfig prop to DynamicFlow component

JulianG-TW:
## Context

Currently, the only configuration available is to specify the `loaderSize` prop to be either `xs | sm | md | lg | xl`.
The loader is displayed when DF is loading the initial step or submitting / executing an action that results in a new step or exiting the flow.

A common issue is when the user submits a valid form, that then returns server-side validation errors. The form disappears for a few seconds during the request, only to come back with some validation errors.

This could be avoided by specifying that we don't want a loader animation in between steps.

#### Before:
```ts
loaderSize?: Size;
```

#### After:
```ts
loaderConfig?: {
  size?: Size;
  initial?: boolean;
  submission?: boolean;
};
```

## BREAKING CHANGES:

By default we do not render a loading animation during form submissions.

If you REALLY want a loading animation during submission, you now need to specify a `loaderConfig`:
```tsx
loaderConfig={ { submission: true, size: "xl" } }
```

The `loaderSize` prop is deprecated. If the `loaderSize` prop was being set to anything other than "xl", consumers will need to instead pass a `loaderConfig` object with at least the `size` property set:

#### Before:
```tsx
loaderSize="sm"
```
#### After
```tsx
loaderConfig={ { size: "sm" } }
```

## Checklist


packages/dynamic-flows/src/flow/typings.ts
```diff
@@ -5,7 +5,11 @@ export type DynamicFlowV3PropsBasic = {
   onStepChange?: (newStep: Step, previousStep: Step) => void;
   onError: (error: unknown, status?: number) => void;
   onTrackableEvent?: (eventName: string, properties: Record<string, unknown>) => void;
-  loaderSize?: Size;
+  loaderConfig?: {
+    size?: Size;
+    initial?: boolean;
+    submission?: boolean;
```
mikknarusk, line 11: where should the behavior this controlls be documented. I can kind of get what `initial` _might_ do based on the PR description, but there should be a more explict documentation somewhere, right?

packages/dynamic-flows/src/flow/DynamicFlow.LoaderConfig.spec.js
```diff
@@ -0,0 +1,191 @@
...
+  ...jest.requireActual('@transferwise/components'),
+  Loader: ({ size }) => <div>Loader Component of size: {size}</div>,
+}));
+
+const i18n = { locale: 'en-GB', messages: { ...componentTranslations['en'] } };
+
+describe('DynamicFlow LoaderConfig', () => {
+  beforeAll(() => jest.useFakeTimers());
+  beforeEach(() => mockFetcher.mockClear());
+  afterAll(() => jest.useFakeTimers());
```
mikknarusk, line 70: should this be `jest.useRealTimers()` instead?

packages/dynamic-flows/src/flow/typings.ts
JulianG-TW: I think the time has come to update that README.md file

JulianG-TW: I've pushed a commit with a lot of changes to README.md

packages/dynamic-flows/src/flow/DynamicFlow.LoaderConfig.spec.js
JulianG-TW: Thank you! it turns out this is not needed! DELETED!

mikknarusk (approved):

AllanHasegawaTw (approved):

PR #1543, opened 2022-07-31
fix(dynamic-flows): Upload component not deleting the files when landed with existing uploaded file ids

dhananjaykha:
## Context

Upload component from DF schema when is rendered with a schema, where the model already contains file Ids, the files can not be deleted from the UI also the form actions do not work.

This is because there exists a bug which does not pick the correct file ids from model.

https://user-images.githubusercontent.com/91905919/182197776-0daad6c3-c653-464c-9811-338a8e546862.mov

## Changes

 
- Changes to the places where the model is consumed from schema and cleaned up.
- Also added a condition to prevent the infinite re-rendering of the UploadInput component.

## Checklist


packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/arrayListSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js
```diff
@@ -25,6 +25,9 @@ const MultipleFileUploadSchema = (props) => {
   const fileSchemaDescriptor = props.schema.items.persistAsync.schema;
 
   function handleFilesChange(newFiles) {
+    if (newFiles.length === files.length) {
+      return;
+    }
```
JulianG-TW, line 30: I'm a bit worried about this.

Could there be a case where `UploadInput` triggers `onFilesChange` with the same number of files, but one of the has changed its `status` or `error` properties?

I wonder if we may need to compare the entire arrays and each element. The y have `id`, `filename`, `status`, `url`, `error`.

dhananjaykha: I agree. It is the status that is changing. Others are always static. I'll address this.

dhananjaykha: @JulianG-TW can you take a look at the logic. I could not find a way to write tests about it though.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/arrayListSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js
```diff
@@ -13,6 +13,25 @@ import ControlFeedback from '../../../controlFeedback';
...
+  for (const newFile of newFiles) {
+    const existingFileObject = fileIdToExistingFileMap[newFile.id];
+    const fileDoesNotExistAlready = !existingFileObject;
+
+    if (fileDoesNotExistAlready) {
+      return false;
+    }
+    if (existingFileObject.status !== newFile.status) {
+      return false;
+    }
```
JulianG-TW, line 30: Consider comparing other properties as well: `filename`, `url`, `error` ?
I'm not sure. Could an existing file in the list be updated with an error?

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/arrayListSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js
JulianG-TW: I would export the function, or maybe extract it to its own file, and then write some tests for it.
The cases would cover identical lists, a list where an item is added, a list where an item is removed, a list where an item changes status, or a list where an item changes to contain an error.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/arrayListSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js
dhananjaykha: All these fields are updated at once. id, url, error, or status. So not each of them seemed fine.

packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/arrayListSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js
dhananjaykha: Done

packages/dynamic-flows/src/common/validation/valid-model/index.js
```diff
@@ -57,9 +60,14 @@ function cleanModelWithObjectSchema(model, schema) {
...
-  if (isArray(model)) {
-    return model.map((childModel) => getValidModelParts(childModel, schema));
+  if (!isArray(model)) {
+    return null;
+  }
+
+  if (!isTupleArraySchema(schema)) {
+    return model.map((childModel) => getValidModelParts(childModel, schema.items));
   }
+  // handle the tuple here when implementing it
```
richardbata, line 70: Do you need to handle when tupleArray?

richardbata: Does schema type `array` always have `items` property for individual entries of array model?

packages/dynamic-flows/src/common/validation/valid-model/spec.js
```diff
@@ -390,4 +390,36 @@ describe('Given a library for returning the valid parts of a model based on a sc
       expect(result).toStrictEqual(model);
     });
   });
+
+  describe('when cleaning file array schema', () => {
```
richardbata, line 394: Maybe some more tests here for is items is array, etc.

packages/dynamic-flows/src/common/validation/valid-model/index.js
dhananjaykha: The array schema can be `list` (same item repeating) or `tuple` (different items in array).

List: `items` is a schema object defining the repeating items
Tuple: `items` is an array of different schemas

We have List so far. That too is a special case, Upload component. That's only handled here. 
Tuple is not implemented on any DF client yet, I believe. So when we implement is, we handle it here.

Added comment specifically for folks to not miss.

packages/dynamic-flows/src/common/validation/valid-model/spec.js
dhananjaykha: Like mentioned, `items` as array is not implemented yet.

richardbata (approved):
LGTM

PR #1553, opened 2022-08-04
feat: Rewriting the camera component

richardbata:
## Context
Rewriting the camera component and fixing many existing issues with is.

So we did an experiment in the KYC Experience team testing different versions of the DF camera component:
- Variant 0: Control
- Variant 1: Increased space allocated to the document frame
- Variant 2: Always choose the main camera when more are available (particularly an issue on Samsung devices where default is the macro camera)
- Variant 3: Take larger resolution images, 1440p

Read more into it on our [wiki](https://transferwise.atlassian.net/l/cp/btABiJFV).

The final component is variant 2, the best performing one.

## Changes

Complete rewrite of the component in TS using react-webcam lib. The overlay now uses SVG rather than canvas.

There is a new error page for when camera isn't supported by the browser.

I've added an [image review page](https://www.figma.com/file/kiSuvU2xf2wwPqgDygIdZd/Camera-capture-experience) which is now disabled, but there for the future.

Many exceptions should also be gone now, reducing rollbar noise.

## Considerations / Results

https://user-images.githubusercontent.com/51439827/190632102-95be7874-89f2-465c-9032-d5a3a561a61e.mov

Screenshot 2022-09-15 at 17 36 42

Screenshot 2022-09-15 at 17 37 05

image

Screenshot 2022-09-16 at 13 44 13

## Checklist


packages/dynamic-flows/src/step/cameraStep/cameraCapture/CameraCapture.tsx
```diff
@@ -0,0 +1,216 @@
...
+  trackCameraFeedStarted,
+  trackCameraNotSupported,
+  trackCameraPermissionDenied,
+} from './tracking';
+import { generateCanvasFromVideo, isSelfieCamera } from './utils';
+
+export enum CAMERA_DIRECTION {
+  FRONT = 'FRONT',
+  BACK = 'BACK',
+}
```
JulianG-TW, line 22: What's the difference between this `enum CAMERA_DIRECTION` and the `direction` under `cameraConfig` in `FileUploadSchema` type in `/src/flow/typings.ts`, line 88?

I can see these are uppercase and the type is lowercase. So perhaps they're unrelated. But I'm confused.

packages/dynamic-flows/src/step/cameraStep/cameraCapture/CameraCapture.tsx
```diff
@@ -0,0 +1,216 @@
...
+      </div>
+    </div>
+  );
+};
+
+const CaptureButton = ({ onClick }: { onClick: () => void }) => (
+  // eslint-disable-next-line jsx-a11y/click-events-have-key-events,jsx-a11y/no-static-element-interactions
+  <span className="camera-capture-btn m-b-2" data-testid="camera-capture-button" onClick={onClick}>
+    <span className="camera-capture-btn-inner" />
+  </span>
```
JulianG-TW, line 70: I see there's a few jsx-a11y rules disabled. Does the button have to be a `span`?
can it not be a `button` element?

I'm sure you tried this already.

One idea could be to extract this component to its own file where it's more quiet, and we can add a comment paragraph explaining the limitations and reasons for doing it this way.
That way it's easier to review in the future, when the context may have changed and we can make more accessible.

packages/dynamic-flows/src/step/cameraStep/cameraCapture/screens/CameraNotSupported/CameraNotSupported.tsx
```diff
@@ -0,0 +1,28 @@
...
+  const intl = useIntl();
+
+  return (
+    <div id="camera-not-supported">
+      <div className="container p-t-5">
+        <div className="row">
+          <div className="col-md-6 col-md-offset-3">
+            <img
+              src="https://wise.com/public-resources/assets/illustrations/general_error_bob_and_naina.svg"
+              alt=""
```
JulianG-TW, line 16: We shouldn't be referencing Wise specific media or URLs.

The most we can do is import components from `@transferwise/components`. Ideally, in the future we will decouple DF from neptune components. At that point we would need to initialise `DynamicFlow` with a prop that injects all the UI components required for the flow.

But for now... to be honest I'm not sure what we can do.

inject this component specifically? but it sounds horrible to do it for just this.

packages/dynamic-flows/src/step/cameraStep/cameraCapture/screens/NoCameraAccess/NoCameraAccess.tsx
```diff
@@ -1,19 +1,23 @@
...
 
   return (
-    <div id="no-camera-access" className={hidden ? 'hidden' : ''}>
+    <div id="no-camera-access">
       <div className="container p-t-5">
         <div className="row">
           <div className="col-md-6 col-md-offset-3">
             <img
-              src="/public-resources/assets/illustrations/general_error_bob_and_naina.svg"
+              src="https://wise.com/public-resources/assets/illustrations/general_error_bob_and_naina.svg"
```
JulianG-TW, line 20: why did this change from relative to absolute?

JulianG-TW (commented):
Thank you for tackling this, and for converting so many files to TS along the way!
I left some comments/questions.

packages/dynamic-flows/src/step/cameraStep/cameraCapture/screens/CameraNotSupported/CameraNotSupported.tsx
JulianG-TW: or we could make it part of the `cameraConfig`. I can se we have `assets.overlay` and `assets.outline`.

BTW is this `cameraConfig` in the spec? I don't see it anywhere.

packages/dynamic-flows/src/step/cameraStep/cameraCapture/CameraCapture.tsx
```diff
@@ -0,0 +1,216 @@
...
+  FRONT = 'FRONT',
+  BACK = 'BACK',
+}
+
+enum MODE {
+  CAPTURE,
+  REVIEW,
+  NO_CAMERA_ACCESS,
+  CAMERA_NOT_SUPPORTED,
+}
```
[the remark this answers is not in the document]
richardbata: Good point! I thought of enum because then you can't mistype a string type union lets say, but actually TS will catch such mistyped string union types! Changing it 👍

packages/dynamic-flows/src/step/cameraStep/cameraCapture/CameraCapture.tsx
richardbata: Changed it to a button and could remove the eslint issues. Only thing is I had to add another style for :focus, some weird orange outline appeared on the button.

packages/dynamic-flows/src/step/cameraStep/cameraCapture/screens/CameraNotSupported/CameraNotSupported.tsx
richardbata: There is a proposal which we accepted for [cameraConfig](https://transferwise.atlassian.net/wiki/spaces/DF/pages/2113142902/DFV3+Camera+Config).

Good question whats the best approach here 🤔 Passing an initialization prop of the page or just image URL to DynamicFlows seems too much for this small thing. 

For adding it to the JSON schema, then we should add the title and paragraph as well probably. But then Mobile does not have the NotSupported page, only permisssions denied, and also recovery of permissions is slightly different for web and mobile...

Lets see what @StevePole has to say about this one

packages/dynamic-flows/src/step/cameraStep/cameraCapture/CameraCapture.tsx
```diff
@@ -0,0 +1,216 @@
...
+          type={ControlType.ACCENT}
+          priority={Priority.SECONDARY}
+          onClick={onRetry}
+        >
+          {intl.formatMessage(messages.reviewRetry)}
+        </Button>
+      </div>
+    </div>
+  );
+};
```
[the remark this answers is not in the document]
richardbata: Moved them to components/index.tsx

packages/dynamic-flows/src/step/cameraStep/cameraCapture/screens/CameraNotSupported/CameraNotSupported.tsx
richardbata: I think I prefer the simple approach for now, and not over engineer it before its needed. I can change the img paths back to relative, and keep the copies on the UI for now. Wdyt?

JulianG-TW: Yes. I discussed with the team this morning and we agreed that using relative paths is fine for now.

JulianG-TW (approved):

packages/dynamic-flows/src/flow/DynamicFlow.tsx
```diff
@@ -21,17 +21,19 @@ import ErrorBoundary from './errorBoundary/ErrorBoundary';
...
   ETag,
   Fetcher,
+  FormErrors,
+  GlobalError,
+  isSchema,
+  LayoutComponent,
   Model,
   ModelsMap,
+  onActionType,
+  onModelChangeType,
```
StevePole, line 34: Why 'Type'?

packages/dynamic-flows/src/step/cameraStep/cameraCapture/components/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+  onSubmit,
+  onRetry,
+}: {
+  onSubmit: () => void;
+  onRetry: () => void;
+}) => {
+  const intl = useIntl();
+  return (
+    <div className="bottom-bar p-x-2">
+      <div className="col-xs-12 col-md-6 p-a-0">
```
StevePole, line 22: Should this be in a `.row` not a `.p-x-2`?  The 'col' provides negative margin which must be compensated by padding, but that might be responsive, so better to use the corresponding class.

richardbata: The issue is bottom-bar is positioned: absolute. I've added a new row div in between, and adjusted the styles as needed. Was able to remove `p-a-0` as well

StevePole (approved):

PR #1557, opened 2022-08-05
fix: shouldn't use filled/circle icon for alert close button and default icons size should be 16px

jonstieglitz:
## Context

Filled Cross doesn't look right on the Alert, Modal, and Drawer - all close icons

We meant to keep icon default size to 16px during upgrade to v3 but it got switched to 24px. this puts it back. also updated the snapshots

Fixed [bug in docs](https://wise.slack.com/archives/CJNBX9CP6/p1660057343147859) where ActionButton page was broken

## Changes
consumed v3.0.6 of transferwise/icons with default icon size as 16px

imported settings icon instead of Cog to ActionButton doc

 
Screenshot 2022-08-05 at 14 38 15

Switched back to un-filled Cross

image

## Checklist


packages/components/src/common/closeButton/CloseButton.js
```diff
@@ -1,4 +1,4 @@
-import { CrossCircleFill as CrossCircleIcon } from '@transferwise/icons';
+import { CrossCircle as CrossCircleIcon } from '@transferwise/icons';
```
dozortsev, line 1: same needs to be for Drawer, Modal, Flow Navigation and maybe in some other places

jonstieglitz: it's the `Close Button` file so it is consumed in all of those places - i checked

dozortsev (approved):

richardmccartney (approved):

phellipealexandre (approved):

PR #1558, opened 2022-08-07
feat(dynamic-flows): Implement list layout component

JulianG-TW:
## Context

MC-2606

The DF List Layout component was not implemented on web.

Example:

```json
  {
    "type": "list",
    "title": "This is a List",
    "items": [
      {
        "title": "Item 1",
        "description": "The first item is positive",
        "icon": "bank",
        "status": "positive"
      },
      {
        "title": "Item 2",
        "description": "The second item is neutral",
        "icon": "emoji",
        "status": "neutral"
      },
      {
        "title": "Item 3",
        "description": "The third item is warning",
        "icon": "gift-box",
        "status": "warning"
      }
    ]
  }
```

## Changes

Before | After
--- | ---
 | image

## Checklist


packages/dynamic-flows/src/layout/list/spec.js
```diff
@@ -0,0 +1,73 @@
...
+
+const i18n = { locale: 'en-GB', messages: { ...componentTranslations['en'] } };
+
+function renderComponent(step) {
+  render(
+    <Provider i18n={i18n}>
+      <DynamicFlow initialStep={step} fetcher={mockFetcher} onClose={onClose} onError={onError} />
+    </Provider>,
+  );
+}
```
dozortsev, line 18: I would suggest to keep such function in [test utils](https://github.com/transferwise/neptune-web/blob/main/packages/dynamic-flows/src/test-utils/index.js)

packages/dynamic-flows/src/layout/layouts/list.json
```diff
@@ -0,0 +1,26 @@
+[
+  {
+    "type": "list",
```
dozortsev, line 3: please just keep this JSON inside story

dhananjaykha (approved):
I have not explored all of components but we don't have margin as a property implemented in the Web Df implementations, right?
https://api-docs.dev-wi.se/dynamic-flow#layouts-list-layout

packages/dynamic-flows/src/layout/layouts/list.json
JulianG-TW: Do you mean inside the `.storybook` directory?
I added this JSON file here because all other layout JSON examples are here.
Is this getting bundled?

packages/dynamic-flows/src/layout/list/spec.js
JulianG-TW: Yes. We now have a few of these. We don't always initialise DF with the same props, but the provider and `i18n` can definitely be in test utils. I'll do that in a separate PR.

Thanks!

JulianG-TW:
> I have not explored all of the components but we don't have margin as a property implemented in the Web Df implementations, right? https://api-docs.dev-wi.se/dynamic-flow#layouts-list-layout

Well spotted! I had forgotten to add the `margin` prop. It is indeed in the specification. Thank you!

PR #1568, opened 2022-08-15
fix(android-web):Fix Dimmers hidden behind address bar on Android web Chrome

jonstieglitz:
## Context

https://wise.slack.com/archives/CJNBX9CP6/p1657795371891599
Drawers and modals are not closable on Android web - anything using the Dimmer

**notes for future devs** 

- Thank you Richard for suggesting [ngrok](https://ngrok.com/). I'm not sure how I could have tested this without it. It didn't work for me while on Wise Wifi (I guess because of firewalls). I used my phone's hotspot.
- It's really useful to test on your desktop so turn off the `isMobile` check in places [like this](https://github.com/transferwise/neptune-web/blob/a435f61391934d232ce1038c74fe52afd06853b3/packages/components/src/modal/Modal.tsx#L57) to be able to use inspector on the code

## Changes
Switch from using `%` to `vh/w` for Dimmer height (and width to be consistent)
 

## Considerations / Results
Before:

After:

## Checklist


packages/components/src/dimmer/Dimmer.less
```diff
@@ -33,8 +33,8 @@
   bottom: 0;
   .left(0);
 
-  width: 100%;
-  height: 100%;
+  width: 100vh;
```
jonstieglitz, line 36: i'll switch this back to `%` if anyone makes a fuss. i changed it for consistency, not for the bug fix

mikknarusk: shouldn't this be 100vw?

jonstieglitz: sigh. yes.

packages/components/src/dimmer/Dimmer.less
```diff
@@ -33,8 +33,8 @@
   bottom: 0;
   .left(0);
 
-  width: 100%;
-  height: 100%;
+  width: 100vw;
+  height: 100vh;
```
richardmccartney, line 37: Could we just have a little test on Mobile Safari, I know there is always a bug with `100vh` and just want to make sure we're not introducing any more 👍 

https://allthingssmitty.com/2020/05/11/css-fix-for-100vh-in-mobile-webkit/

richardmccartney (approved):

PR #1576, opened 2022-08-22
fix(dynamic-flows): Fix schema.default in BasicTypeSchema

JulianG-TW:
## Context

In BasicTypeSchema, the condition that determines whether or not to apply the default value was using type coercion when evaluating `model` and `schema.default`, so a nullish value would be considered false.

As a result when we have a `null` model in a schema like `{ "type": "boolean",  "default": false }`, the default value of `false` is not used.

Even worse, when the model is `false` and the `schema.default` is `true`, the default value would override the model.

In the video below we can see the model us `null` even though the `schema.default` is set to `false`.
https://user-images.githubusercontent.com/67635620/186028335-c7dccf26-20b3-41e7-b2ec-b0b5b92ac9c0.mov

## Changes

We now check for null(ish) values rather than evaluating `!model` and `schema.default` as booleans.

image

## Considerations / Results

If the boolean schema does not provide a default value and the model does not have a value for this schema, then the checkbox will look like the value is false while it is null. Attempting to submit the form will still result in the validation error.

## Checklist


artem-storozhuk (approved):

packages/dynamic-flows/src/jsonSchemaForm/basicTypeSchema/BasicTypeSchema.js
```diff
@@ -57,7 +57,7 @@ const BasicTypeSchema = (props) => {
 
   const onSchemaChange = () => {
     // if no model, change to the default, only run this when the schema changes
-    if (!model && props.schema.default) {
+    if (model == null && props.schema.default != null) {
```
mikknarusk, line 60: shouldn't it be `===` and `!==`?

richardbata: I guess we need to check for model === null || model === undefined, which is the same as above.

I'm thinking it could be easier to read to explicitly check for both, possibly with isNull etc. utility methods.

Both approaches work, I'm happy to approve either way

JulianG-TW: I'm intentionally using `==` because it covers both `null` and `undefined`.

When the main `model` for the entire form does not include a value for this particular schema, it would be `undefined` by default, but at some point in the code, we're passing `null` instead. So in practice the model will usually be either `null` or some value, but since it's JS, I cannot be sure it won't be undefined some times.

In the case of `props.schema.default` when we don't provide a default then it's `undefined`, but again, some JSON schema may contain `"default": null"` so I rather cover both cases using `!=`.

A more explicit version if this line would be
```js
 if ((model === null || model === undefined) && (props.schema.default !== null || props.schema.default !== undefined) {
```
But I think it's too verbose.

Maybe this one declares intent more accurately ?
```js
if (model == null && props.schema.default != undefined) {
```

JulianG-TW: or, using the utility function:
```js
if (model == null && !isUndefined(props.schema.default) {
```

mikknarusk: Ah okay I understand now. I'm worried that this is fragile and may be refactored in the future, not realising the importance of the missing `=`. Maybe it would be clearer with a helper function that does an explicit check for null and undefined+

JulianG-TW: using utility functions...
```js
 if ((isNull(model) || isUndefined(model)) && (!isNull(props.schema.default) && !isUndefined(props.schema.default)) {
 ```
 it still reads very long to me.
 
 what about making a `isNullish` function?
 
 ```js
 if (isNullish(model) && !isNullish(props.schema.default) {
 ```

mikknarusk: yeah something like isNullish is what I was thinking of

JulianG-TW: Added `isNullish` in https://github.com/transferwise/neptune-web/pull/1576/commits/4c720a74bdff0e93307a55c834f3badbaae01caf

mikknarusk (approved):
nice

richardbata (approved):
Could move isNullish to a common utils file, but looks good

JulianG-TW:
> Could move isNullish to a common utils file, but looks good
yes. we could move it to neptune-validation, next to `isNull` and `isUndefined`.

PR #1582, opened 2022-08-25
feat(dynamic-flows): Boolean Schema now has an implicit default value of false

JulianG-TW:
## Context

When a boolean schema is used with a `null` model, its initial value is `null`, (unless specified in with a `default` property).

https://user-images.githubusercontent.com/67635620/186693749-40aaf2b7-79bf-4c54-a656-a74933fbab89.mov

However, to the user this looks like they're submitting `false`.

Mobile clients default to `false` in this cases. This PR is to align the web client to mobile clients.

## Changes

This PR makes boolean schema have an implicit default value of `false`.

Now the initial value of the boolean is `false`:

image
(I could record a video, but nothing changes)

## Considerations / Results

When a boolean model is null, it is immediately updated to `false` (or `true` if it is explicitly specified) and that results in an `onChange` event. Some tests had to be fixed to account fir this extra callback call.

## Checklist


packages/dynamic-flows/src/jsonSchemaForm/basicTypeSchema/BasicTypeSchema.js
```diff
@@ -16,6 +16,10 @@ const generateId = () => {
   return String(Math.floor(100000000 * Math.random()));
 };
 
+const getDefaultValue = (schema) => {
+  return schema.type === 'boolean' && isNullish(schema.default) ? false : schema.default;
```
dozortsev, line 20: note most likely you wouldn't end up need for function like `isNullish` if you were using TypeScript so please consider to migrate component as you make changes in them

JulianG-TW: It's true.

But I think, in this case, converting `BasicTypeSchema` to TS would not give us any confidence that the props passed to it are of the types we specify, because other JS files will render `<BasicTypeSchema`.

JulianG-TW: The other friction point to converting to TS is that I don't want a PR that both makes functionality changes AND converts to TS. Since the changes could be difficult to see in the middle of it.

Ideally a PR would only convert some modules to TS without changing behaviour.

JulianG-TW: I'm just listing my excuses for not having converted almost anything to TS in the past few months.

packages/dynamic-flows/src/jsonSchemaForm/basicTypeSchema/BasicTypeSchema.errors.spec.js
```diff
@@ -174,27 +174,29 @@ describe('Given a component for rendering a number schema', () => {
...
           mountOptions,
         );
 
-        expect(onChange).not.toHaveBeenCalled();
+        expect(onChange).toHaveBeenCalledWith(
+          false, //  model
+          schema, // schema
+          false, //  trigger model
+          null, //   previous model
+        );
```
JulianG-TW, line 199: this test is covering the changes

richardbata (approved):

PR #1583, opened 2022-08-26
Keep form in loading state after submitting

mikknarusk:
## Context
https://transferwise.atlassian.net/browse/SEND-3692
Currently, when the form is successfully submitted and we get the x-df-exit header, the spec is cleared which replaces the form with a blank page. This assumes clients will listen to onClose and will immediately be able to render new content. 

In the send money flow we're trying to keep random floating loaders to a minimum, and since the next step in the flow (verification) requires some data to be loaded before we can render anything, users are seeing a blank white page while that's going on. 

We decided that it would make sense to keep showing the form until the consumer reacts to onClose, unmounts the form and renders something else. We're releasing this as a breaking change as some behavior changes although most consumers will not have to change anything to adopt the latest version.

## Changes
Don't set the specifications to an empty object after submission.
Explicitly set the loader state to idle if we got a new step or if we got an error. Keep loader state if we reached the exit header.
 

## Checklist


JulianG-TW (approved):
Thank you. I'll make sure this is merged with BREAKING CHANGE in the description.

JulianG-TW (approved):
Thank you. I'll make sure this is merged with BREAKING CHANGE in the description.

PR #1585, opened 2022-08-30
[WIP] feat!: updating theme provider and docs for personal theme – Kristóf’s branch

kripod-wise:
## Context

https://transferwise.atlassian.net/browse/DS-2677

As the part of our rebranding process, Neptune Web components’ colors have to be updated to reflect our new look and feel.

I’ve tried my best documenting my progress on a per-Figma-component basis here: https://transferwise.atlassian.net/wiki/spaces/DS/pages/2503181525/Web+redesign+progress

Additionally, I’ve proposed major improvements for the Neptune Web package ecosystem, detailed on a wiki page: https://transferwise.atlassian.net/wiki/spaces/WEB/pages/2493943945/Neptune+design+system+packages+Issues+Suggestions

## Changes

 

- Added personal theme styles for many components
- Extended `useTheme` hook to return whether the selected theme is modern or not
- Fixed minor issues that I came across while working on my primary task

## Considerations / Results

I’ve tried not to break existing markup and kept all the legacy styles intact.

## Checklist


packages/components/src/actionButton/ActionButton.less
```diff
@@ -27,4 +27,37 @@
       box-shadow: 0 0 0 4px var(--color-background-accent);
     }
   }
+
+  .np-theme-personal & {
+    color: var(--color-interactive-primary);
```
dozortsev, line 32: is it okey to use `-interactive-` token for text color? should we use only `-content-` for text?

packages/components/src/chips/Chip.less
```diff
@@ -70,4 +70,42 @@
       color: white;
     }
   }
+
+  .np-theme-personal & {
+    font-weight: var(--font-weight-semi-bold);
```
dozortsev, line 75: typography styles for all components are done in another PR please don't fix typography as part of colors PR

packages/components/src/chips/Chip.less
```diff
@@ -70,4 +70,42 @@
...
+    color: var(--color-interactive-primary);
+    padding: 0 var(--size-16);
```
dozortsev, line 77: padding/margin and other spacing styles needs to be done separately, as part of shape work

packages/components/src/common/bottomSheet/BottomSheet.less
```diff
@@ -31,4 +31,8 @@
     // https://webkit.org/blog/7929/designing-websites-for-iphone-x/
     padding-bottom: env(safe-area-inset-bottom);
   }
+
+  .np-theme-personal & {
+    border-radius: var(--radius-large) var(--radius-large) 0 0;
```
dozortsev, line 36: please leave shapes change for separate PR

packages/components/src/common/bottomSheet/BottomSheet.tsx
```diff
@@ -166,8 +166,17 @@ const BottomSheet = (props: Props): ReactElement => {
       maxHeight: `calc(${screenHeight} - ${safeZoneHeight} - ${topbarHeight})`,
     };
   }
+
+  const theme = useTheme();
```
dozortsev, line 170: BottomSheet started slide from right to left when opens up, instead of from bottom to top

not sure what causing this bug

https://user-images.githubusercontent.com/4015628/188632345-5704ab58-597e-422a-af19-846258e0e44f.mov

packages/components/src/accordion/AccordionItem/AccordionItem.js
```diff
@@ -30,7 +30,7 @@ const AccordionItem = ({
         as="button"
         media={iconElement}
         title={title}
-        button={<Chevron orientation={open ? Position.TOP : Position.DOWN} />}
+        button={<Chevron orientation={open ? Position.TOP : Position.DOWN} size={Size.MEDIUM} />}
```
dozortsev, line 33: in light theme size should be small, so I guess you need define size here based on theme (via `useTheme` hook)

packages/components/src/chips/Chip.tsx
```diff
@@ -33,6 +34,10 @@ const Chip = ({
   const isActionable = onClick || onKeyPress;
   const defaultRole = isActionable ? 'button' : undefined;
   const tabIndex = isActionable ? 0 : -1;
+
+  const theme = useTheme();
+  const isThemeModern = useMemo(() => modernThemes.has(theme), [theme]);
```
dozortsev, line 39: I would instead make `useTheme()` more useful

```js
const { isPersonalThem, isLightTheme } = useTheme()
```

this way you will safe a tone of `if` and equality (`===`) statements inside Neptune web and for consumers of Neptune

dozortsev: we have similar thing do page direction data

https://github.com/transferwise/neptune-web/blob/main/packages/components/src/common/hooks/useDirection/useDirection.ts#L6-L10

stylelint.config.js
```diff
@@ -24,7 +24,10 @@ module.exports = {
     'csstools/value-no-unknown-custom-properties': [
       true,
       {
-        importFrom: ['./node_modules/@transferwise/neptune-tokens/tokens.css'],
+        importFrom: [
+          './node_modules/@transferwise/neptune-tokens/tokens.css',
+          './node_modules/@transferwise/neptune-tokens/themes/personal/tokens.css',
```
dozortsev, line 29: you should not need define `personal/tokens.css` here as tokens names are the same (only values are different) that is what it matters for this Stylelint plugin

packages/components/src/actionButton/ActionButton.less
kripod-wise: I’ve used [the component remapping specs](https://www.figma.com/file/GO6vjcaEIukCbRdZ2yZ1Ms/VE---Components-remapping-specs?node-id=65%3A21950) as a reference, as instructed within the [Jira ticket](https://transferwise.atlassian.net/browse/DS-2677) my work is based upon.

packages/components/src/common/bottomSheet/BottomSheet.tsx
kripod-wise: This change shouldn’t affect that. If I remember correctly, the bottom sheet came in from the left even before this change, but I may be mistaken. and we should see whether this behavior is the same on the original `DS-2378-personal-theme` branch.

packages/components/src/chips/Chip.tsx
kripod-wise: I’ve tried not to introduce any breaking changes within this PR, so that app owners who rely upon this package could upgrade to the latest version of Neptune Web without hassle, introducing no breaking changes throughout the flow.

However, if we really want return an object from `useTheme()`, I could refactor the code to do so.

stylelint.config.js
kripod-wise: We did introduce some new colors, as my IDE’s complaints led me to make this change. E.g. [`--color-interactive-accent` has been replaced by `--color-interactive-primary`](https://transferwise.atlassian.net/wiki/spaces/VR/pages/2437844264/Colour+tokens+old+vs+new#Interactive-colours).

packages/components/src/common/bottomSheet/BottomSheet.tsx
dozortsev: you can always double check on our prod 🙂

https://transferwise.github.io/neptune-web/components/overlays/BottomSheet

there it opens from bottom to top

packages/components/src/chips/Chip.tsx
dozortsev: what breaking change will be introduced the thing that I suggested https://github.com/transferwise/neptune-web/pull/1585#discussion_r963647766?

I don't see any 🤔

stylelint.config.js
dozortsev: do you mean that `--color-interactive-primary` isn't defined in `neptune-tokens/tokens.css`? if so than this is a bug in neptune-tokens package

packages/components/src/common/bottomSheet/BottomSheet.tsx
kripod-wise: Yep, but I’d assume the production branch isn’t based upon `DS-2378-personal-theme`, on which my branch is based.

kripod-wise: It turns out that the base `DS-2378-personal-theme` branch has the same issue. Therefore, further investigation should be done over there, as this isn’t the result of this PR.

richardmccartney (approved):

PR #1586, opened 2022-08-31
refactor(dynamic-flow): Refactor DynamicParagraph and its test

JulianG-TW:
## Context

Before making changes to `DynamicParagraph` to support the "copyable" control, I would like to take a moment to convert it to TypeScript, and refactor its tests to use React Testing Library instead of Enzyme.

## Changes

- Rename index.js and spec.js to DynamicParagraph.ts and DynamicParagraph.spec.js
- Removed prop-types declarations in favour of TS types.
- Refactor the test to use RTL instead of Enzyme

## Considerations / Results

It may be easier to review each commit separately. 

## Checklist


packages/dynamic-flows/src/layout/paragraph/DynamicParagraph.tsx
```diff
@@ -0,0 +1,14 @@
+import { getTextAlignmentAndMargin } from '../utils';
+
+type Props = {
+  component: {
+    text: string;
+    margin?: 'xs' | 'sm' | 'md' | 'lg' | 'xl';
```
andreapiras, line 6: I imagine these classes are defined somewhere. In that case why not reuse `m-x-1, m-x-2` so we won't produce any more css? Same for alignment `text-xs-center` etc

packages/dynamic-flows/src/layout/paragraph/DynamicParagraph.tsx
```diff
@@ -0,0 +1,14 @@
...
+    align?: 'left' | 'right' | 'center';
+  };
+};
+
+export const DynamicParagraph = ({ component }: Props) => {
+  const classes = getTextAlignmentAndMargin(component);
+  return <p className={classes}> {component.text} </p>;
+};
```
andreapiras, line 14:
```suggestion
export const DynamicParagraph = ({ component }: Props) => (
  <p className={getTextAlignmentAndMargin(component)}> {component.text} </p>
);
```

packages/dynamic-flows/src/layout/paragraph/DynamicParagraph.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+type Props = {
+  component: {
+    text: string;
+    margin?: 'xs' | 'sm' | 'md' | 'lg' | 'xl';
+    align?: 'left' | 'right' | 'center';
+  };
+};
+
+export const DynamicParagraph = ({ component }: Props) => {
```
andreapiras, line 11: Should component be renamed to something more intuitive like styles?

packages/dynamic-flows/src/layout/utils.ts
```diff
@@ -27,7 +30,7 @@ const getTextAlignment = (align) => {
   }
 };
 
-const getTextAlignmentAndMargin = (component) => {
+const getTextAlignmentAndMargin = (component: { align?: Align; margin?: Size }) => {
   return `${getTextAlignment(component.align)} ${getMarginBottom(component.margin)}`;
 };
```
andreapiras, line 35:
```suggestion
const getTextAlignmentAndMargin = (component: { align?: Align, margin?: Size }) =>
  `${getTextAlignment(component.align)} ${getMarginBottom(component.margin)}`;
```

packages/dynamic-flows/src/layout/paragraph/DynamicParagraph.tsx
```diff
@@ -0,0 +1,14 @@
...
+  const classes = getTextAlignmentAndMargin(component);
```
andreapiras, line 12: How about getTextAlignmentAndMargin -> getClassNames

packages/dynamic-flows/src/layout/paragraph/DynamicParagraph.spec.js
```diff
@@ -0,0 +1,16 @@
+import { render, screen } from '@testing-library/react';
+
+import { DynamicParagraph } from './DynamicParagraph';
+
+describe('Given a component for dynamically rendering paragraphs', () => {
+  const component = {
+    type: 'paragraph',
```
andreapiras, line 7: Is type part of components definition?
component: {
    text: string;
    margin?: 'xs' | 'sm' | 'md' | 'lg' | 'xl';
    align?: 'left' | 'right' | 'center';
  };

packages/dynamic-flows/src/layout/paragraph/DynamicParagraph.tsx
JulianG-TW: Perhaps, but if you look at [DynamicLayout.js#L32-L101](https://github.com/transferwise/neptune-web/blob/main/packages/dynamic-flows/src/layout/DynamicLayout.js#L32-L101) all layout components take a `component` prop.
I think it would be out of scope to rename it for `DynamicParagraph` only.

packages/dynamic-flows/src/layout/paragraph/DynamicParagraph.tsx
JulianG-TW: This function is used in other layout components. I think it's out of scope for this PR to rename it.

packages/dynamic-flows/src/layout/paragraph/DynamicParagraph.spec.js
JulianG-TW: Yes. it's not really used in DynamicParagraph because DynamicLayout will look at the `type` property and decide which layout component to render ([here](https://github.com/transferwise/neptune-web/blob/main/packages/dynamic-flows/src/layout/DynamicLayout.js#L32-L101)).

So, technically, DynamicParagraph doesn't care what the `type` property says.

In the near future I plan to create some TypeScript types based on the DF specification. I'm hoping that will help consolidating the types in one place rather than being defined all over the place. I anticipate the `type` property in this case will be used as the discriminator in a big discriminated union type called `LayoutComponent` or something like that.

JulianG-TW: I just think it will take some time to write all those Types. I could hold off converting anything to TS until I have those types, but I don't see a problem with converting some components now.

JulianG-TW: It's already part of the DF spec, and being used. I haven't introduced it in this PR, and it's not exclusive to this component.
I wish I could point to the DF spec but it's not public yet.

andreapiras (approved):

phellipealexandre (approved):

PR #1587, opened 2022-09-01
feat(dynamic-flows): Add support for copyable paragraph

JulianG-TW:
## Context

A new feature was added to the DFv3 spec. It's a new control to the exiting `paragraph` layout component.
When `control: "copyable"` a giant "Copy" button appears and allows users to copy the text to their clipboad.

## Changes

We implemented the feature.

copyable

### Snackbar availability

Currently DF can be used without a SnackbarProvider. To avoid adding a dependency and a breaking change, this new feature works with and without snackbar.
When the snackbar provider is in the render tree it is used, when it's missing, there's no visual feedback, but the copy functionality continues to work.

### Browser compatibility

Tested on Chrome, Safari and Firefox:

Screenshot 2022-09-01 at 09 49 30

Screenshot 2022-09-01 at 09 47 40

## Checklist

- [x] Changes work in all supported browsers (nevermind IE11)


richardbata:
I could also see [InlineAlert](https://transferwise.github.io/neptune-web/components/alerts/InlineAlert) being used for onCopy, but I also like Snackbar

packages/dynamic-flows/src/layout/paragraph/DynamicParagraph.tsx
```diff
@@ -11,3 +17,33 @@ type Props = {
...
+      ?.writeText(text)
+      .then(() => createSnackbar({ text: formatMessage(messages.copied) }))
+      .catch(noop);
+  };
+
+  const classNames =
+    getTextAlignmentAndMargin({ align: component.align, margin: 'sm' }) + ' form-control';
+  return (
+    <div className={getTextAlignmentAndMargin(component)}>
+      <input type="text" className={classNames} value={text} readOnly />
```
richardbata, line 39: Does input support longer texts with line breaks needed?

packages/dynamic-flows/src/i18n/en.json
```diff
@@ -15,6 +15,8 @@
   "dynamicFlows.ControlFeedback.required": "Please fill out this field.",
   "dynamicFlows.ControlFeedback.type": "Incorrect type",
   "dynamicFlows.DynamicExternal.retryTitle": "Reopen window",
+  "dynamicFlows.DynamicParagraph.copied": "Copied to clipboard",
```
richardbata, line 18: Sidenote: do these actually get translated through Crowdin?

richardbata (approved):

packages/dynamic-flows/src/layout/paragraph/DynamicParagraph.tsx
JulianG-TW: No. and it's unfortunate that this component is called "paragraph". or even that we added "copyable" to this component.
Web was the last platform to implement this, and only when I implemented it I realised how far from a paragraph this is. It's basically two different components.

packages/dynamic-flows/src/i18n/en.json
JulianG-TW: Good point. The english file seems to have a lot more entries than the other languages. Maybe at some point the integration got broken. I'll investigate.

packages/dynamic-flows/src/layout/DynamicLayout.js
```diff
@@ -18,7 +18,7 @@ import DynamicImage from './image';
 import DynamicInfo from './info';
 import { DynamicList } from './list';
 import { componentModel } from './models';
-import { DynamicParagraph } from './paragraph';
+import { DynamicParagraph, DynamicCopyableParagraph } from './paragraph';
```
StevePole, line 21: Shouldn't the switch be internal?  Feels like we're polluting this global component.  In both cases this is a DynamicParagraph, so the switch should really be inside.

JulianG-TW: Good point. I'll do that.

JulianG-TW: Done. Now there are no changes to DynamicLayout in this PR.

PR #1588, opened 2022-09-02
test(dynamic-flows): Refactor integration tests for DynamicFlow component

JulianG-TW:
## Context

There is one big DynamicFlow.end-to-end.spec.js file that is testing two different ways of initialising the DF component: `initialStep` and `flowUrl`. 

This test suite has a `mockFetcher` that handles requests for multiple tests. 
Tests that cause refresh-on-change are difficult to read and reason about because we need to constantly scroll up to see what the schema looks like and then also find what the mockFetcher returns for each request.

**This will help writing some upcoming tests while fixing a bug.**

## Changes

We separate the test suite in two use cases: `initialStep` and `flowUrl`. It turns out that for the `initialStep` case the `mockFetcher` is now very simple.

In the `flowUrl` suite, the `mockFetcher` is also very simple for most tests, and only requires some complexity in one of the tests. Because of this, the `renderComponent` function accepts a `fetcher` property and we only pass a complex fetcher in the test that needs it.

## Considerations / Results

Unfortunately, the changes to the `flowUrl` tests are not easy to spot in this PR because they're in a new file.

## Checklist


packages/dynamic-flows/src/flow/DynamicFlow.flowUrl.spec.js
```diff
@@ -0,0 +1,195 @@
...
+        url: '/persist-async-upload',
+        param: 'cardNumber',
+        idProperty: 'cardToken',
+        schema: {
+          title: 'Enter your card number',
+          type: 'string',
+        },
+      },
+    },
+    ...extraProperties,
```
JulianG-TW, line 53: We can now easily add new properties to the main object schema using `extraProperties`.

packages/dynamic-flows/src/flow/DynamicFlow.flowUrl.spec.js
```diff
@@ -0,0 +1,195 @@
...
+    };
+    render(
+      <Provider i18n={i18n}>
+        <DynamicFlow {...props} />
+      </Provider>,
+    );
+  }
+
+  it('renders the form', async () => {
+    renderComponent({ fetcher: () => delayedJsonResponse(getFormStep()) });
```
JulianG-TW, line 101: Most tests use a very simple `fetcher` that returns the default form step.

packages/dynamic-flows/src/flow/DynamicFlow.flowUrl.spec.js
```diff
@@ -0,0 +1,195 @@
...
+                  email: 'some.other@email.com',
+                },
+              });
+              return delayedJsonResponse(step);
+            }
+            default:
+              throw new Error('unexpected request');
+          }
+        });
+        renderComponent({ flowUrl: '/example', fetcher: mockFetcher });
```
JulianG-TW, line 180: Only this test uses a special fetcher, which returns a step with slightly different schema and model.

phellipealexandre (approved):
I don't have much knowledge on JS but approving as these changes have no impact in production code

PR #1589, opened 2022-09-05
fix(dynamic-flows): Fix refresh-on-change edge case

JulianG-TW:
## Context

When a model change triggers a refresh-on-change and the refreshed step contains the exact same schema, but different model, the form is not updated with the new model.

Note that this problem doesn't happen if the schema changes, which is what happens in most real-world use cases.

## Changes

- In the first commit we add a test that fails because of the bug.
- in the second commit we fix the issue and the new test passes.

## Considerations / Results

I additionally tried this locally on recipient creation and I could not see any adverse side-effects or regressions.

The reason to fix this edge-case is that we need to continue fixing issues with refresh-on-change, and it's better to do that on top of a more stable implementation.

## Checklist


packages/dynamic-flows/src/flow/DynamicFlow.flowUrl.spec.js
```diff
@@ -34,7 +34,7 @@ const getSchema = ({ extraProperties } = {}) => ({
         { title: 'Blue', const: 'blue' },
       ],
       refreshFormOnChange: true,
-      refreshFormUrl: '/example-account-refresh',
+      refreshFormUrl: '/example-color-refresh',
```
JulianG-TW, line 37: irrelevant

dhananjaykha (approved):

PR #1593, opened 2022-09-08
fix(dynamic-flows): retain the entered value of persist async fields across refreshes

JulianG-TW:
## Context

Values entered into persist async fields (such as credit card numbers which are stored in compliant servers) must be persisted across refreshes.

In the case where such fields are part of a oneOf, we also need to make sure they're kept when switching schemas (e.g. when switching tabs back and forth). 

## Changes

Before | After
--- | ---
persist-async-fix-before | persist-async-fix-after

## Considerations / Results

In order to keep the entered value I had to resort to rendering all schemas for each oneOf, keeping only the selected schema visible, and the rest hidden. Some tests had to be fixed because **jsdom** was "finding" components that were actually hidden.

## Checklist


packages/dynamic-flows/src/layout/DynamicLayout.js
```diff
@@ -32,23 +32,22 @@ const DynamicLayout = (props) => {
   const renderComponent = (component) => {
     switch (component.type) {
       case 'heading':
-        return <DynamicHeading key={getKey(component)} component={component} />;
```
StevePole, line 35: Are unique keys no longer necessary?  React always used to complain.

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.js
```diff
@@ -84,49 +84,61 @@ const OneOfSchema = (props) => {
...
-          hideTitle
-          disabled={props.disabled}
-          onChange={(model, triggerSchema, triggerModel, lastTriggerModel) =>
-            onChildModelChange(schemaIndex, model, triggerSchema, triggerModel, lastTriggerModel)
-          }
-          onPersistAsync={props.onPersistAsync}
-        />
-      )}
+      {controlAndFeedback}
+      {isNoNConstSchema(props.schema.oneOf[schemaIndex]) ? genericSchemas : null}
```
StevePole, line 141: Can we fix this typo `NoN`, ideally !isConstSchema.

packages/dynamic-flows/src/jsonSchemaForm/genericSchema/GenericSchema.js
```diff
@@ -61,15 +65,16 @@ GenericSchemaForm.propTypes = {
   onChange: PropTypes.func.isRequired,
   submitted: PropTypes.bool.isRequired,
   hideTitle: PropTypes.bool,
+  hideSchema: PropTypes.bool,
```
StevePole, line 68: I don't understand this prop.  Why would I tell a child to hide itself, and not just hide it myself?

packages/dynamic-flows/src/layout/DynamicLayout.js
JulianG-TW: Good point. The linter doesn't complain, even though these are in a map function predicate.
We can no longer use the schemas as keys, but maybe we need to use something else.
Using the stringified schemas as keys result in a new instance of each component as soon as the schema changes. This makes it impossible for us to keep (otherwise unaltered) component instances.
We could replace the schemas with a representation of their position in the tree or something like that, which should not change in between refreshes. however this is similar to using indices for keys which is also a discouraged practice, although, I think in this case it may be what we want.

JulianG-TW: UPDATE: React DOES complain. It was the linter not complaining. My bad. I can add `key={index}` and that satisfies React without upsetting the linter.

packages/dynamic-flows/src/jsonSchemaForm/oneOfSchema/OneOfSchema.js
JulianG-TW: Yes. I fixed the typo. I would not replace it with `!isConstSchema` because is not exactly the same.

```js
export function isNonConstSchema(schema) {
  return !!schema && !isConstSchema(schema);
}
```

packages/dynamic-flows/src/jsonSchemaForm/genericSchema/GenericSchema.js
JulianG-TW: The reason for this is testing. 

In **OneOfSchema.spec.js**:
```js
expect(component.find('GenericSchema')).toHaveLength(3);
expect(component.find('GenericSchema[hideSchema=false]')).toHaveLength(1); // 1 visible
expect(component.find('GenericSchema[hideSchema=true]')).toHaveLength(2); // 2 hidden
```
If in OneOfSchema.js I wrap each `GenericSchema` in a `div`, it wouldn't be so easy to make these assertions, I believe.Ï

PR #1594, opened 2022-09-12
feat: Add optional Action Button to Definition List

NewmanJWise:
## Context

We want to be able to include an action button with a definition list so that users can do things like "Copy" text.

image

## Changes

 

- Refactored value content to have a possible action button on the right side

## Considerations / Results

- I decided to put the button inside the content section for the `dd` because adding a third column with this structure causes a lot of issues with alignment for the button especially complicated with the vertical/horizontal differences

- Should we consider a refactor of this component in the future to use CSS grid to make it more expandable with new columns / rows? This would mean I wouldn't have to do so many workarounds to get a button on the right hand side.

## Checklist


packages/components/src/definitionList/DefinitionList.less
```diff
@@ -6,6 +6,15 @@
     word-break: break-word;
   }
 
+  &__action {
+    align-self: center;
+    flex-shrink: 0;
+  }
+
+  &__button {
+    margin-top: -50%;
```
NewmanJWise, line 15: This feels nasty but if I used align items centre then the text on the left would be further down from the heading than the rest of the definitions.

packages/components/src/definitionList/DefinitionList.js
```diff
@@ -26,10 +27,34 @@ const DefinitionList = ({ definitions, layout, muted }) => (
...
     {definitions
       .filter((definition) => definition)
-      .map(({ title, value, key }) => (
+      .map(({ action, title, value, key }) => (
         <div key={key} className="tw-definition-list__item">
           <dt>{title}</dt>
-          <dd className={alignmentClassMap[layout] || ''}>{value}</dd>
+          <dd
+            className={classNames(
+              classNames(alignmentClassMap[layout] || '', {
```
dozortsev, line 35: please remove one `classNames(` function here

packages/components/src/definitionList/DefinitionList.js
```diff
@@ -26,10 +27,34 @@ const DefinitionList = ({ definitions, layout, muted }) => (
...
+                'd-flex': action,
+                [isLayoutHorizontal(layout) ? 'align-items-center' : 'align-items-start']: action,
+                [layout === Layout.HORIZONTAL_RIGHT_ALIGNED
+                  ? 'justify-content-end'
+                  : 'justify-content-between']: action,
+              }),
+            )}
+          >
+            <div>{value}</div>
```
dozortsev, line 44: I think we should put Action Button beneath value on small screens

Screen Shot 2022-09-13 at 11 36 53 AM

dozortsev: just a suggestion, `useLayout` hook might be useful for you in this case

example: https://github.com/transferwise/neptune-web/blob/main/packages/components/src/drawer/Drawer.js#L17

packages/components/src/definitionList/DefinitionList.js
```diff
@@ -38,6 +63,10 @@ const DefinitionList = ({ definitions, layout, muted }) => (
 DefinitionList.propTypes = {
   definitions: PropTypes.arrayOf(
     PropTypes.shape({
+      action: PropTypes.shape({
+        label: PropTypes.string.isRequired,
```
dozortsev, line 67: this should be `node` type, because [Action Button accepts node via `children` prop](https://github.com/transferwise/neptune-web/blob/main/packages/components/src/actionButton/ActionButton.tsx#L18)

consumer might want put icon the button

helendurranttw: But we don't officially support icons in buttons right? It's not something we want to encourage

packages/components/src/definitionList/DefinitionList.js
```diff
@@ -26,10 +27,34 @@ const DefinitionList = ({ definitions, layout, muted }) => (
...
-          <dd className={alignmentClassMap[layout] || ''}>{value}</dd>
+          <dd
+            className={classNames(
+              classNames(alignmentClassMap[layout] || '', {
+                'd-flex': action,
+                [isLayoutHorizontal(layout) ? 'align-items-center' : 'align-items-start']: action,
+                [layout === Layout.HORIZONTAL_RIGHT_ALIGNED
+                  ? 'justify-content-end'
+                  : 'justify-content-between']: action,
+              }),
```
helendurranttw, line 41: The `alignmentClassMap[layout]` worked when it was just applying a few classes, but I think you should refactor this code to find a cleaner way of constructing all these classes based on layout and action. Maybe in its own function, even - but at least, not mixing `alignmentClassMap[layout]`, `isLayoutHorizontal(layout)` and `layout === Layout.HORIZONTAL_RIGHT_ALIGNED` on subsequent lines - it's just too difficult to reason on.

packages/components/src/definitionList/DefinitionList.js
NewmanJWise: I couldn't seem to replicate this, is it still happening?
image

Do you think we should put it down the line below? On mobile it's always on the right side
image

packages/components/src/definitionList/DefinitionList.js
NewmanJWise: We could leave it as text for now and always change it & discuss later on if a request comes in?

richardmccartney: Yes, let's leave it for now 👍

packages/components/src/definitionList/DefinitionList.js
richardmccartney: Happy for us to leave this as it is currently, we can pick up any design changes after 👍 @dozortsev can you let us know what browser this issue was showing on?

packages/docs/liveEditorCode/definitionList.code.js
```diff
@@ -17,5 +17,13 @@
...
       key: 'third',
     },
+    {
+      title: 'Action buttons are useful',
+      value: 'You can do all kinds of neat things with actions buttons!',
+      key: 'fourth',
+      action: {
+        label: 'Hey',
+        onClick: () => console.log('Hello World!!!'),
+      },
```
richardmccartney, line 27: Looks like we're getting a syntax error on this page: 

```
SyntaxError: Unexpected token (28:2)
24 :       action: {
25 :         label: 'Hey',
26 :         onClick: () => console.log('Hello World!!!'),
27 :       },
28 :   ]}
       ^
```

If we can resolve this issue then we can look to merging your changes.

richardmccartney (changes requested):

packages/docs/liveEditorCode/definitionList.code.js
richardmccartney: You can see it happening live here https://transferwise.github.io/neptune-web/branch/GPAID-1145/components/content/DefinitionList

NewmanJWise: Fixed it, sorry about that!

PR #1595, opened 2022-09-13
feat(dynamic-flows)!: Remove X-Access-Token from step requests

JulianG-TW:
## Context

Originally, all requests to fetch a DF step would add a Wise-specific `X-Access-Token` with a hardcoded value.

This request header is only necessary for SOME Wise endpoints, and therefore should not be added by default.

If you need to add this header, there is now an example example in the README file.

## Considerations / Results

BREAKING CHANGE: If your flow endpoints expect a hard-coded X-Access-Token request header you need to initialise DF using a fetcher function that provides that header.

## Checklist


packages/dynamic-flows/src/common/hooks/usePersistAsync.js
```diff
@@ -52,7 +52,6 @@ export const usePersistAsync = (persistAsyncSchema, onPersistAsync) => {
       method: persistAsyncSchema.method,
       headers: {
         ...(!isBlobSchema && { 'Content-Type': 'application/json' }),
-        'X-Access-Token': 'Tr4n5f3rw153',
```
dhananjaykha, line 55: Hardcoded value is bad. 
However, in the PR description, when we say - 
> This request header is only necessary for SOME Wise endpoints, and therefore should not be added by default.

Would this update be a breaking change?

dhananjaykha (approved):
The token seems just a random value. However, if it's must, should we do a major release?

packages/dynamic-flows/src/common/hooks/usePersistAsync.js
dhananjaykha: Oh you mentioned it up.

PR #1601, opened 2022-09-16
refactor(component): refactor CheckboxOption component

gmukul01:
## Context

Refactor CheckboxOption

## Changes

1. Changed files of `CheckboxOption` & `CheckboxButton` component files to `typescript`.
2. Add `as` prop to `CheckboxOption` component.

## Checklist


packages/components/src/checkboxButton/types.ts
```diff
@@ -0,0 +1,3 @@
+import { HTMLProps } from 'react';
+
+export type CheckboxButtonProps = HTMLProps<HTMLInputElement>;
```
dozortsev, line 3: we tend to define types of component properties inside component file above the component code, could you please move them?

packages/components/src/checkboxOption/types.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export type CheckboxOptionProps = Omit<BaseOptionProps, 'onChange'> & {
+  /**
+   * The checked state of the checkbox.
+   */
+  checked?: boolean;
+  /**
+   * Function to call when the checkbox option is clicked
+   */
+  onChange: (value: boolean) => void;
```
dozortsev, line 11: why we need omit and then redefine `onChange` prop?

packages/components/src/checkboxOption/CheckboxOption.tsx
```diff
@@ -0,0 +1,18 @@
+import CheckboxButton from '../checkboxButton';
+import Option from '../common/Option';
+
+import { CheckboxOptionProps } from './types';
+
+const CheckboxOption = ({ checked, disabled, onChange, ...rest }: CheckboxOptionProps) => {
```
dozortsev, line 6: could you please also add `ref` support for this component

packages/components/src/common/Option/Option.tsx
```diff
@@ -3,9 +3,10 @@ import { MouseEvent, forwardRef, HTMLAttributes, ReactNode, ElementType } from '
 
 import { LinkProps } from '../commonProps';
 
-interface BaseOptionProps extends Omit<HTMLAttributes<HTMLElement>, 'title'> {
+export interface BaseOptionProps extends Omit<HTMLAttributes<HTMLElement>, 'title'> {
   'aria-label'?: string;
   media?: ReactNode;
+  name?: string;
```
dozortsev, line 9: just out of curiosity why you need `name` prop?

packages/components/scripts/generate-type-declarations.js
```diff
@@ -24,7 +23,6 @@ const jsFiles = [
   { filename: 'Popover', folder: 'popover' },
   { filename: 'Money', folder: 'money' },
   { filename: 'RadioOption', folder: 'radioOption' },
-  { filename: 'CheckboxOption', folder: 'checkboxOption' },
```
dozortsev, line 27: 👏

dozortsev (commented):
look great, left a few minor comments

packages/components/src/checkboxOption/types.ts
gmukul01: Actually the `onChange` prop in `BaseOptionProps ` is of `FormEventHandler<HTMLElement>`, so I had to omit that one to change it to `(value: boolean) => void`

packages/components/src/common/Option/Option.tsx
gmukul01: We are passing the name prop also in `CheckboxOption.story.tsx`
image

packages/components/src/checkboxButton/types.ts
gmukul01: Should we not keep the types in separate files for better separation of concern and better readability ?

packages/components/src/checkboxOption/CheckboxOption.tsx
gmukul01: Done

packages/components/src/checkboxButton/types.ts
dozortsev: IMHO it's more convenient to keep Props types in component file, it's easier to scan 👀 code and understand what this component about, I think this way you get better readability

re. separation of concern, we can review this idea later, for please just follow approach that we have in all other components

packages/components/src/checkboxOption/types.ts
dozortsev: why can't we use native one (`FormEventHandler<HTMLElement>`)?

dozortsev: why can't we use native one (`FormEventHandler<HTMLElement>`)?

packages/components/src/checkboxButton/types.ts
gmukul01: Perfect. Done.

packages/components/src/checkboxOption/types.ts
dozortsev: wait, sorry, using native interface for `onChange` in this case will lead to breaking change, which we don't want now, so please ignore my above https://github.com/transferwise/neptune-web/pull/1601#discussion_r976298284.

dozortsev (approved):

PR #1602, opened 2022-09-21
feat(dynamic-flows): Replace Upload component with UploadInput in PersistAsyncBlobSchema

JulianG-TW:
## Context

Reasons for this change:

1. The Verification Experience team prefers this smaller component.
2. As opposed to the `Upload` component the `UploadInput` component can be "managed", so we can initialise it with a file id, for instance after a refresh-on-change. 

## Changes

Before | After
--- | ---
image | image
image | image

## Considerations / Results

I wrote an adaptor component for this in order to minimise the change to PersistAsyncBlobSchema, in fact, one can easily revert to using `Upload` by changing the imports. This means it is trivial to import both and provide a mechanism to select which control to use at runtime.

The downside of this is that it prevents us from simplifying some things. For instance the `onSuccess` callback expects an argument of type `Response`, but the `onFailure` callback expects an object like `{ response: Response }` for some reason. We could clean this up, but then the adaptor would not be compatible with `Upload` component any more.

I think we can clean up in a follow-up PR; once we're sure this is what we want. And that we're not going to offer the old upload component any more.

**Pre-populating the value**

As mentioned above, this new component can be managed, and initialised with a file id. However we don't have a way to specify the file name. I think this is a limitation in the blob schema, but I may be wrong.

## Checklist


artem-storozhuk:
1. How does the component look when the file is uploading/uploaded?
2. Can the user remove the file when it is uploaded?
3. Can I specify the value of the file? We sometimes need to pre-populate the file input component with previously uploaded file.

JulianG-TW:
> 1. How does the component look when the file is uploading/uploaded?
> 2. Can the user remove the file when it is uploaded?
> 3. Can I specify the value of the file? We sometimes need to pre-populate the file input component with previously uploaded file.

1. It looks like this https://transferwise.github.io/neptune-web/components/inputs/UploadInput#UploadInput
2. Yes
3. Yes, but only the fileId. we don't have a way to specify the filename. Thanks! this is one thing I should add to "Considerations".

packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/blob/UploadInputAdapter.tsx
```diff
@@ -0,0 +1,81 @@
...
+      }
+    });
+  };
+
+  return (
+    <UploadInput
+      fileInputName="fileInputName"
+      files={files}
+      uploadButtonTitle={props.usLabel}
+      sizeLimit={maxSize ? maxSize / 1024 : undefined}
```
dhananjaykha, line 71: we can reuse some logic from [file-utils.js](https://github.com/transferwise/neptune-web/blob/main/packages/dynamic-flows/src/common/utils/file-utils.js)

packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/blob/UploadInputAdapter.tsx
```diff
@@ -0,0 +1,81 @@
+import { Status, UploadedFile, UploadInput } from '@transferwise/components';
```
dhananjaykha, line 1: this file and [MultipleFileUploadSchema.js](https://github.com/transferwise/neptune-web/blob/main/packages/dynamic-flows/src/jsonSchemaForm/arrayTypeSchema/arrayListSchema/multipleFileUploadSchema/MultipleFileUploadSchema.js) do have multiple common things (they are anyway using the `UploadInput` under the hood).

dhananjaykha: Also. Asking for myself, 

- Is it fair to that the upload component would always exists with persist async? (fair to assume? what do you think?) In Wise's context it is. Other clients, not sure if it would be a good design.
- To avoid duplication, should we reuse the same adapter component both places maybe (in a separate PR)?

packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/blob/UploadInputAdapter.tsx
```diff
@@ -0,0 +1,81 @@
...
+    onFailure,
+    onCancel,
+  } = props;
+
+  const { formatMessage } = useIntl();
+
+  const files = useMemo((): UploadedFile[] => {
+    if (fileId) {
+      const id = fileId;
+      const filename = formatMessage(MESSAGES.UploadedFile);
```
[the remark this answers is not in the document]
JulianG-TW: Thank you! Done @a466764316ed584e167ab3f6d86822255c53688f

packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/blob/UploadInputAdapter.tsx
JulianG-TW: Thank you! Done @ e336cfe2ad43f907adf2f9f24421d364b2b257c2

packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/blob/UploadInputAdapter.tsx
JulianG-TW: Yes.I'm sure a refactoring could be done. Like I said in the "considerations section":

> The downside of this is that it prevents us from simplifying some things. For instance the onSuccess callback expects an argument of type Response, but the onFailure callback expects an object like { response: Response } for some reason. We could clean this up, but then the adaptor would not be compatible with Upload component any more.

I can see a future PR cleaning up and refactoring what's common in single-file and multiple-file uploads.

packages/dynamic-flows/src/i18n/en.json
```diff
@@ -28,6 +28,8 @@
   "dynamicFlows.MultipleFileUploadSchema.maxFileSizeError": "Sorry, that file is too big. Please upload a smaller file.",
   "dynamicFlows.MultipleFileUploadSchema.maxItemsError": "Please upload {maxItems} or fewer files.",
   "dynamicFlows.MultipleFileUploadSchema.minItemsError": "Please upload at least {minItems} file(s).",
+  "dynamicFlows.PersistAsyncSchema.UploadFailed": "Upload failed. Please, try again.",
+  "dynamicFlows.PersistAsyncSchema.UploadedFile": "Uploaded file",
```
dhananjaykha, line 32: We can get this removed. LGTM, otherwise.

JulianG-TW: ah! of course! I need to run `yarn build`. Thank you!!!

dhananjaykha (approved):

richardmccartney (approved):

PR #1616, opened 2022-09-26
Fix uploadinputadapter

JulianG-TW:
## Context

After [#1602 Replacing Upload component with UploadInput in PersistAsyncBlobSchema](https://github.com/transferwise/neptune-web/pull/1602) a potential bug was identified: if more than one `UploadInput` component is present in the same page, they need to different `id` properties, to avoid the problem illustrated below:

Clicking on the second upload component and selecting a file results in that file being "sent" to the first upload component.

### Before
before

## Changes

We now pass an `id` property taken from either the schema `$id` or the schema `idProperty`.

### After
after

## Checklist


packages/dynamic-flows/src/jsonSchemaForm/persistAsyncSchema/blob/PersistAsyncBlobSchema.js
```diff
@@ -58,6 +58,11 @@ const PersistAsyncBlobSchema = (props) => {
   return (
     <div className={classNames(formGroupClasses)}>
       <UploadInputAdapter
+        id={
+          props.schema.$id ||
```
richardbata, line 62: Safe access? `?.`

richardbata (approved):
Nice!

PR #1619, opened 2022-09-26
feat(components): update Button component

striccaa:
## Context

**Jira ticket**: https://transferwise.atlassian.net/browse/CSPROD-4455

In the `help-center` CRAB app, we often use the Next.js-specific `<Link />` component, which has a `passHref` property, and given it's enabled, it passes the href attribute to its child element.

The current implementation of the `Button` component - which we often use wrapped in the `<Link>` component - does not support the `as` prop which would alter the rendered HTML tag, which causes two things:
1. It produces semantically incorrect HTML (see screenshot 1)
2. It breaks the styles of the button (`text-decoration: underline`) due to the button being rendered in an anchor tag (see screenshot 2)

| Issue 1 | Issue 2 |
| --- | --- |
| Screenshot 2022-09-26 at 23 36 05 | Screenshot 2022-09-26 at 23 35 08 |

In this PR I'm adding support to anchor tags in the `Button` component.

## Changes

 
- extended `Button` with `as` prop to support anchor tags
- added new anchor variant to `Button` story
- updated `Button` test case

## Considerations / Results

I marked this PR as a breaking change due to the types of the `Button` changing.

## Checklist


packages/components/src/button/Button.tsx
```diff
@@ -26,23 +27,41 @@ type DeprecatedTypes = 'primary' | 'pay' | 'secondary' | 'danger' | 'link';
 /** @deprecated */
 type DeprecatedSizes = SizeExtraSmall;
 
-export type Props = {
+export type CommonProps = {
+  as?: ElementType;
```
richard-better, line 31: Hmm, is this a breaking change? Since the new behaviour depends on the presence of a new prop, the existing behaviour is intact, right? 

If that is correct, this classifies as a minor:

> MINOR version when you add functionality in a backwards compatible manner
(from https://semver.org/ )

richardmccartney: @richard-better you are correct here, the prop looks to be optional so we're not breaking the component API here at all. I'd just rebase the commit message when possible 👍

richardmccartney (changes requested):

packages/components/src/button/Button.story.js
```diff
@@ -24,13 +25,14 @@ export const Basic = () => {
   return (
     <div className={background === 'blue' && 'bg--dark p-a-3'}>
       <Button
+        as={element}
```
tiborsaas-tw, line 28: It would be nice to see an example of assigning a `href` parameter as well so we could preview how it behaves as a link.

packages/components/src/button/Button.tsx
```diff
@@ -26,23 +27,41 @@ type DeprecatedTypes = 'primary' | 'pay' | 'secondary' | 'danger' | 'link';
...
+  Omit<ButtonHTMLAttributes<HTMLButtonElement>, 'type'> & {
+    as?: 'button';
+    htmlType?: 'submit' | 'reset' | 'button';
+  };
+
+export type AnchorProps = CommonProps &
+  AnchorHTMLAttributes<HTMLAnchorElement> &
+  Omit<LinkProps, 'href'> & {
+    as?: 'a';
+    href?: string; // Optional due to usage of component with next/link's passHref behavior
```
tiborsaas-tw, line 50: Have you considered adding a `target` property? In some cases it might make sense to open a link in a new window.

richard-better: I believe you won't need it  here explicitly in this case. If you add `as="a"`, typescript will automatically narrow the props of the component to be `AnchorProps`, and that means you can add any props supported by `a`.  The `{...props}` spread on Line 115 means it will pass those on automatically.

packages/components/src/button/Button.story.js
striccaa: Added ✅

packages/components/src/button/Button.tsx
striccaa: As @richard-better mentioned, the `AnchorHTMLAttributes<HTMLAnchorElement>` and `LinkProps` types should cover all properties that an anchor tag can have, thus, we don't need to add properties such as `target` manually.

packages/components/src/button/Button.story.js
tiborsaas-tw: [🙇](https://emojipedia.org/person-bowing/)

richardmccartney (approved):

PR #1622, opened 2022-09-28
fix(dynamic-flows): Fix step validation errors when using initialStep

JulianG-TW:
## Context

Steps can contain validation errors. These are displayed immediately when the step loads:

```json
{
  "key": "any",
  "type": "form",
  "errors": {
    "validation": {
      "name": "This validation error comes with the step"
    }
  },
  "schemas": [
...
```
expected-bahaviour

This works fine when a step is fetched, for instance as a result of initialising the `DynamicFlow` component using `flowUrl`, after a refresh-on-change, or when loading a new step as a result of a submission.

But it doesn't work when initialising DF using `initialStep`.

## Changes

This PR fixes that.

Before | After 
---- | ----
before | after

## Considerations / Results

**Check the first two commits, they're a test and a fix.**
The third commit adds storybook fixtures and makes the PR 10x more complex than it should be.

## Checklist


packages/dynamic-flows/src/flow/DynamicFlow.tsx
```diff
@@ -139,7 +139,9 @@ const DynamicFlowComponent: VFC<DynamicFlowV3Props> = ({
 }) => {
   const { locale } = useIntl();
   const [submitted, setSubmitted] = useState(false);
-  const [formErrors, setFormErrors] = useState<FormErrors>();
+  const [formErrors, setFormErrors] = useState<FormErrors>(
+    initialStep ? initialStep?.errors?.validation : undefined,
+  );
```
JulianG-TW, line 144: This is the bugfix!

phellipealexandre (approved):
Did my best to review it and unblock you Julian 🙇 It looks good to my non-web developer eyes!

PR #1647, opened 2022-10-11
feat(dynamic-flows): Add a displayStepTitle prop to DynamicFlow

JulianG-TW:
## Context

We need a way to tell `DynamicFlow` NOT TO inject the step title into the layout, in some cases. Otherwise, when we render a DF flow in a modal body, we get the title duplicated once in the modal header and once in the modal body:

image

## Changes

We add a new `displayStepTitle` prop to `DynamicFlow`.
This prop is optional and when omitted it defaults to `true`, to maintain backwards compatibility.

## Checklist


packages/dynamic-flows/src/step/layoutStep/LayoutStep.js
```diff
@@ -44,6 +45,7 @@ const LayoutStep = (props) => {
 
 LayoutStep.propTypes = {
   stepSpecification: PropTypes.object.isRequired,
+  stepConversionOptions: PropTypes.object.isRequired,
```
richardbata, line 48: One of those examples where we can rewrite file to TS, if not too much effort 👍

packages/dynamic-flows/src/step/layoutStep/LayoutStep.js
```diff
@@ -6,12 +6,12 @@ import DynamicLayout from '../../layout';
 
 import { convertStepToLayout, inlineReferences } from './layoutService';
 
-const getComponents = (step) => {
+const getComponents = (step, stepConversionOptions) => {
```
richardbata, line 9: Name param just: options?

richardbata (approved):

packages/dynamic-flows/src/step/layoutStep/LayoutStep.js
JulianG-TW: I was about to do it. I even renamed the file. But then the changes to the file get lost in the diff.
I'm more inclined to make a PR exclusively about JS->TS.

I'm planning to bring some TS types that I wrote based on the DFv3 spec into this repo and change a bunch of files to TS without changing functionality.

packages/dynamic-flows/src/step/layoutStep/LayoutStep.js
JulianG-TW: I wanted to convey that they're the options for step conversion (something that I hope to get rid of some day) and not the options for "getting components".

JulianG-TW: oh, I see. in thiscontext.... yes. it could be called just options, I suppose.

JulianG-TW: renamed!

PR #1648, opened 2022-10-11
feat(Typography): display number styles

dozortsev:
## Context

https://www.figma.com/file/T9heSKeXBLXrR4NDIKw1V5/Typography-spec-doc?node-id=248%3A23

## Changes

Add CSS class that implements Display Number header.

## Considerations / Results

Screen Shot 2022-10-11 at 2 53 53 PM

## Checklist


packages/css/src/less/core/_typography.less
```diff
@@ -189,6 +189,13 @@ a,
 
 // DISPLAY
 
+.np-text-display-number {
+  font-weight: var(--font-weight-semi-bold);
+  font-size: 40px;
+  line-height: 44px;
+  letter-spacing: -0.03em;
```
dozortsev, line 196: note not using CSS custom props here because we need fix some of them (see https://transferwise.atlassian.net/browse/DS-2998)

